PR #4, opened 2015-10-08
Initial Peninsula test setup

mlange05:
Initial setup of the "Peninsula" test case from North et al., 2009, section 2.2.2. This includes all the bare necessities to start exploring further tests and functionalities, such as API details and performance optimisation. The merge includes:
* Initial travis setup with flake8 checks and requirement installs, including `numpy`/`scipy`/`netCDF4`
* Basic encoding of NEMO grid file conventions (read and write via `netCF4`)
* Simple 2D spatial interpolation via `scipy` with user-defined interpolation degree
* Simple (slow!) RK4 implementation in a pure Python loop

Note that the initial Travis setup only runs the test but does not yet check for a correct solution.


erikvansebille:
Thanks for getting this code together, Michael. Great job!

PR #35, opened 2016-02-15
Generic netcdf grids

Jacketless:
Wrote an example script to test with defaults that should work with the
simple OFAM data now pushed to the parcels-examples repository.
NetCDF_grid class now moved to grid.py
Slight edit to plotParticles.py to allow loading of single OFAM grid data (currently hardcoded, so will need generalising at some point).

Still an issue with running under jit mode, not sure why. Default is scipy.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -164,7 +164,7 @@ class Field(object):
         if varname is None:
             varname = self.name
         # Derive name of 'depth' variable for NEMO convention
-        vname_depth = 'depth%s' % self.name.lower()
+        vname_depth = 'depth'
 
         # Create DataArray objects for file I/O
         t, d, x, y = (self.time.size, self.depth.size,
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -57,7 +57,7 @@ class Field(object):
         self.time_index_cache = LRUCache(maxsize=2)
 
     @classmethod
-    def from_netcdf(cls, name, varname, datasets, **kwargs):
+    def from_netcdf(cls, name, varname, datasets, dimensions={}, **kwargs):
         """Create field from netCDF file using NEMO conventions
 
         :param name: Name of the field to create
@@ -68,13 +68,14 @@ class Field(object):
         """
         if not isinstance(datasets, Iterable):
             datasets = [datasets]
-        lon = datasets[0]['nav_lon'][0, :]
-        lat = datasets[0]['nav_lat'][:, 0]
+        lon = datasets[0].variables[dimensions["long"]]
+        lat = datasets[0].variables[dimensions["lat"]]
+        #lat = datasets[0][dimensions["lat"]][:, 0]
         # Default depth to zeros until we implement 3D grids properly
         depth = np.zeros(1, dtype=np.float32)
         # Concatenate time variable to determine overall dimension
         # across multiple files
-        timeslices = [dset['time_counter'][:] for dset in datasets]
+        timeslices = [dset[dimensions["time"]][:] for dset in datasets]
         time = np.concatenate(timeslices)
 
         # Pre-allocate grid data before reading files into buffer
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -68,9 +68,9 @@ class Field(object):
         """
         if not isinstance(datasets, Iterable):
             datasets = [datasets]
-        lon = datasets[0].variables[dimensions["long"]]
+        lon = datasets[0].variables[dimensions["lon"]]
         lat = datasets[0].variables[dimensions["lat"]]
-        #lat = datasets[0][dimensions["lat"]][:, 0]
+        # lat = datasets[0][dimensions["lat"]][:, 0]
         # Default depth to zeros until we implement 3D grids properly
         depth = np.zeros(1, dtype=np.float32)
         # Concatenate time variable to determine overall dimension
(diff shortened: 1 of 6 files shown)


mlange05:
Ok, there are a few things here:
* You accidentally committed loads of files that should not be in the repo, including generated kernel code files and Python bytecode files (`.pyc`). We should probably add the latter to our `.gitignore` file.
* I would prefer if test/example files live in the same repo as the according data files. However, in order to keep these tests as part of our automated test suite we need to first turn the examples into a git-submodule that travis can checkout and run regression tests from.
* Introducing a new `NetCDFGrid` class is definitely the right approach to keep things generic, but ideally the `NEMOGrid` would then be a specialised sub-class of the more generic `NetCFGrid` class that derives the relevant default parameters for NEMO conventions.
* There are quite a few PEP8 violations in the new test. I know the PEP8 guidelines can be a bit annoying at times, but they do result in cleaner code in the long run.

parcels/field.py
```diff
@@ -68,13 +68,14 @@ def from_netcdf(cls, name, varname, datasets, **kwargs):
         """
         if not isinstance(datasets, Iterable):
             datasets = [datasets]
-        lon = datasets[0]['nav_lon'][0, :]
-        lat = datasets[0]['nav_lat'][:, 0]
+        lon = datasets[0].variables[dimensions["long"]]
```
mlange05: Is there any particular reason why we switch from `lon` to `long` here?

scripts/plotParticles.py
```diff
@@ -18,10 +18,10 @@ def particleplotting(filename,tracerfile, mode):
     fig, ax = plt.subplots()
     if tracerfile != 'none':
       tfile = Dataset(tracerfile,'r')
-      X = tfile.variables['x']
-      Y = tfile.variables['y']
-      P = tfile.variables['P']
-      plt.contourf(np.squeeze(X),np.squeeze(Y),np.squeeze(P))
+      X = tfile.variables['xu_ocean']
+      Y = tfile.variables['yu_ocean']
```
mlange05: This will most likely break plotting for other grid files that follow a different convention. We need to make this more generic and configurable, possibly even using our own `Grid` or `Field` classes.

tests/test_generic_netcdf_grid.py
```diff
@@ -0,0 +1,68 @@
...
+    pset = grid.ParticleSet(size=args.particles, pclass=ParticleClass,
+                            start=((lon_min+lon_max)/2, lat_min), finish=((lon_min+lon_max)/2, lat_max))
+
+    hours = 25*24
+    substeps = 60
+
+    pset.execute(AdvectionRK4, timesteps=hours*substeps, dt=300.,
+                       output_file=pset.ParticleFile(name=args.output_file),
+                       output_steps=substeps)
+
```
mlange05: it would be great to have some kind of acceptance criteria here to better check that the example behaves as expected.

parcels/field.py
Jacketless: - Ah yes, apologies, I completely forgot about the tests, I should have
   run them first. They check for PEP8 right? Getting used to the workflow.
   - If you want to create a directory structure for tests and datafiles
   that fits those requirements, I can change the locations of files and
   default settings in the test file easily enough.
   - Yes we discussed the superclass framework- I thought I would make my
   first pull request based on this standalone class for the moment. In a few
   weeks I should *hopefully* have yet another type of field data from our
   colleagues in Toulouse (habitat-maps), so perhaps it is worth waiting to
   see what these look like before we build a higher level class?
   - 'Long' was my instinctive choice for longitude dimension naming- it
   was only afterwards that I saw that the field class uses 'lon', sorry.
   - The previous plotParticles was already hardcoded to NEMO so it was
   already breaking for files outside that convention. Agree on making it
   generic.
   - Acceptance criteria... Perhaps that the resulting grid actually
   contains some data? Animating the flowfields as background images in
   plotParticles would be really great as well.

parcels/field.py
```diff
@@ -68,13 +68,14 @@ def from_netcdf(cls, name, varname, datasets, **kwargs):
         """
         if not isinstance(datasets, Iterable):
             datasets = [datasets]
-        lon = datasets[0]['nav_lon'][0, :]
-        lat = datasets[0]['nav_lat'][:, 0]
```
Jacketless: These removed lines in field.py are now breaking test_peninsula each time, in which a grid is created, saved, then re-loaded using NEMOgrid.from_file. I will try and fix this tomorrow. You're right @mlange05, this would not break if NEMOgrid was a proper subclass of NetCDFGrid.

mlange05:
@Jacketless , so I think I figured out the necessary fix and how to do the constructor inheritance in the grid class in a somewhat nice way. Apologies for re-doing some of your work and submitting a separate PR for this, but this way we can build this in conjunction with the new `examples` submodule and without some of the commit noise. Please have a look at PR #38, especially the way we define the OFAM convention in the example test, and see if this is sufficient for your models. If you are happy with this, please feel free to implement an additional `Grid.from_ofam()` utility constructor to put the convention in the central repo.

Also, there is still a lot more good stuff in here, like the plotting script change and your full test setup for the OFAM grid. Can you please resubmit these changes in the relevant repositories to get them merged as well?

Jacketless:
Deleting this branch, as useful commits were put into PR #38 and merged with the master.

PR #36, opened 2016-02-17
Updates

Abobie:
Added EE method, made RK4 correctly time dependent and adjusting for lon/lat, and added the option to give numerical method as an argument

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -33,6 +33,7 @@ class Kernel(object):
             stack = inspect.stack()
             try:
                 user_ctx = stack[-1][0].f_globals
+                user_ctx['math'] = globals()['math']
             except:
                 print("Warning: Could not access user context when merging kernels")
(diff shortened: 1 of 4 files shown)


Abobie:
I don't understand why the test says that 'math' is not defined, I've imported it in test_peninsula.py, particle.py and kernel.py, do I need to import it somewhere else too?

mlange05:
@Abobie, sorry this took so long to review. This is an excellent contribution, a very nice set of clean commits. The bug stems from a subtlety in the way the JIT engine fuses kernels, and I have fixed this now.

Also, due to previous changes earlier today (PR #38) I will rebase this on the current master before we merge. I have also removed the accidentally committed test to keep the history clean.

mlange05:
Ok, one more amendment: the `locals()['Advection' + methodname]` trick is cute but slightly confusing to non-experts, so instead I changed it to spell the method out explicitly.

PR #38, opened 2016-02-22
Generic netCF grids and examples submodule

mlange05:
This merge brings two features at the same time:
* And `examples` sub-module that contains data-intensive example scripts, as well as a few extra tests. It is hosted under `OceanParcels/parcels-examples` and still small, but it allows us to separate demonstrative examples with real data from more low-level feature tests. More re-shuffling of the test infrastructure is needed though.
* Adds generic grid initialisation from netCDF files, as demonstrated by the OFAM example grid. This also includes:
  * Rename `NEMOGrid` to `Grid`
  * Add `Grid.from_netcdf()` that allows specific initialisation from a set of netCDF files with specified variable and dimension names.
  * Replace `NEMOGrid.from_file()` with `Grid.from_nemo()`, which encapsulates some explicit assumptions about NEMO naming conventions.

The two features are landed together in order to activate testing the `examples` subdirectory straight away. Many thanks go to @Jacketless for providing the test and doing all the groundwork for this merge. The merge supersedes some of the core parts of PR #35, so @Jacketless can you please have a thorough look at this before we merge?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "examples"]
+	path = examples
+	url = https://github.com/OceanPARCELS/parcels-examples.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
     - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy netcdf4 pandas
     - pip install --upgrade pip
     - pip install -r requirements.txt
-    - python setup.py build_ext --inplace
+    - git submodule update --init --recursive
 
 script:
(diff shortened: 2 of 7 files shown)


Jacketless:
Thanks @mlange05 - yes actually what you have done makes more sense. This new grid class with your edits is now essentially the base class I see. Until we need to load more complex files, we might as well just create from_xxx() functions with hard coded naming conventions for any ocean data we use a lot (NEMO, OFAM etc.).
I'll take the just plotParticles and generic grid tests and resubmit them when I have a moment.

PR #43, opened 2016-03-07
Particle sampling

Jacketless:
Added in the functionality for custom user particle subclasses of
Particle or JITParticle to define “user_vars” and sampling kernels
which will record an arbitrary field to sample from other than lat and
lon.
Not sure if I have done this best way, i’ve tried to edit the current
structure as little as possible. 

Something not working properly with
JITParticle subclasses- suspect it has something to do with me not
allocating memory for the new user_vars OrderedDict in the Particle
Class. Maybe this could blow up our memory allowance as it’s in the superclass?

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -39,14 +39,18 @@ class Particle(object):
     :param lon: Initial longitude of particle
     :param lat: Initial latitude of particle
     :param grid: :Class Grid: object to track this particle on
+    :param user_vars: Dictionary of any user variables that might be defined in subclasses
     """
+    user_vars = OrderedDict()
 
     def __init__(self, lon, lat, grid, cptr=None):
         self.lon = lon
         self.lat = lat
-
         self.xi = np.where(self.lon >= grid.U.lon)[0][-1]
         self.yi = np.where(self.lat >= grid.U.lat)[0][-1]
+
+        for var in self.user_vars:
+            setattr(self, var, 0)
 
     def __repr__(self):
         return "P(%f, %f)[%d, %d]" % (self.lon, self.lat, self.xi, self.yi)
@@ -86,7 +90,7 @@ class JITParticle(Particle):
 class ParticleType(object):
     """Class encapsulating the type information for custom particles
 
-    :param user: Optional list of (name, dtype) tuples for custom variables
+    :param user_vars: Optional list of (name, dtype) tuples for custom variables
     """
 
     def __init__(self, pclass):
@@ -100,8 +104,8 @@ class ParticleType(object):
         if self.uses_jit:
             self.var_types = pclass.base_vars
             self.var_types.update(pclass.user_vars)
-        else:
-            self.var_types = None
+
+        self.user_vars = pclass.user_vars
 
     def __repr__(self):
         return self.name
@@ -246,6 +250,7 @@ class ParticleFile(object):
         :param name: Basename of the output file
         :param particlset: ParticleSet to output
         :param initial_dump: Perform initial output at time 0.
+        :param user_vars: A list of additional user defined particle variables to write
         """
         self.dataset = netCDF4.Dataset("%s.nc" % name, "w", format="NETCDF4")
         self.dataset.createDimension("obs", None)
@@ -286,6 +291,16 @@ class ParticleFile(object):
         self.z.units = "m"
         self.z.positive = "down"
 
+        if particleset.ptype.user_vars is not None:
+            self.user_vars = particleset.ptype.user_vars.keys()
+            for var in self.user_vars:
+                setattr(self, var, self.dataset.createVariable(var, "f4", ("trajectory", "obs"), fill_value=0.))
+                getattr(self, var).long_name = ""
+                getattr(self, var).standard_name = var
+                getattr(self, var).units = "unknown"
+        else:
+            self.user_vars = {}
+
         self.idx = 0
 
         if initial_dump:
@@ -302,6 +317,8 @@ class ParticleFile(object):
             self.lat[:, self.idx] = np.array([p.lat for p in pset])
             self.lon[:, self.idx] = np.array([p.lon for p in pset])
             self.z[:, self.idx] = np.zeros(pset.size, dtype=np.float32)
+            for var in self.user_vars:
+                getattr(self, var)[:, self.idx] = np.array([getattr(p, var) for p in pset])
         else:
             raise TypeError("NetCDF output is only enabled for ParticleSet obects")
 
--- a/scripts/plotParticles.py
+++ b/scripts/plotParticles.py
@@ -1,64 +1,68 @@
 #!/usr/bin/env python
 from netCDF4 import Dataset
 import numpy as np
-import matplotlib as mpl
 import matplotlib.pyplot as plt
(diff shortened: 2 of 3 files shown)


parcels/particle.py
```diff
@@ -39,15 +39,24 @@ class Particle(object):
...
         self.xi = np.where(self.lon >= grid.U.lon)[0][-1]
         self.yi = np.where(self.lat >= grid.U.lat)[0][-1]
 
+        # If particle is a superclass with user variables,
+        # initiate these attributes from the grid here
+        for var in self.user_vars:
+            if hasattr(grid, var):
+                setattr(self, var, getattr(grid, var)[0, lon, lat])
+            else:
+                raise AttributeError('The current grid does contain the "%s" field defined in particle.user_vars.' % var)
```
mlange05: The way this is currently written every `user_var` must exist on the grid to not cause the error. I don't think this is the intended behaviour and it's what causes the current Travis failures, I believe. A better way would probably be to just initialise all `user_vars` to 0 and let the user kernel do the sampling. 

My reasoning here is that automatically inferring a relationship between grid and particle variables solely based on name might cause unexpected behaviour from a users point of view and I don't think adding a single line to make the sampling process explicit is too much to ask. I'm open to discussion on this point though. Also, please note that I do not object to adding a helper kernel `UpdateUserVars` that makes this assumption, but this should be clearly documented.

parcels/particle.py
```diff
@@ -100,8 +109,9 @@ def __init__(self, pclass):
         if self.uses_jit:
             self.var_types = pclass.base_vars
             self.var_types.update(pclass.user_vars)
+            self.user_vars = pclass.user_vars
         else:
-            self.var_types = None
+            self.user_vars = pclass.user_vars
```
mlange05: Why not just move the line out of the conditional?

parcels/particle.py
```diff
@@ -224,12 +234,12 @@ def Kernel(self, pyfunc):
...
 
     def ParticleFile(self, *args, **kwargs):
-        return ParticleFile(*args, particleset=self, **kwargs)
+        return ParticleFile(*args, particleset=self, user_vars=self.ptype.user_vars, **kwargs)
 
 
 class ParticleFile(object):
 
-    def __init__(self, name, particleset, initial_dump=True):
+    def __init__(self, name, particleset, initial_dump=True, user_vars=None):
```
mlange05: The `particleset` has an associated `ParticleType ptype` attribute, which contains the `user_vars`. We could just infer it from there instead of adding to the constructor, no?

SEAPODYM_Data_Test.py
```diff
@@ -0,0 +1,94 @@
...
+                                              self.habitat)
+
+    pset = grid.ParticleSet(size=args.particles, pclass=Particle,
+                            start=((lon_min + lon_max) / 2, lat_min), finish=((lon_min + lon_max) / 2, lat_max))
+
+    hours = 3 * 24
+    substeps = 30
+
+    k_adv = pset.Kernel(AdvectionRK4)
+    sense_habitat = pset.Kernel(updateUserVars)
```
mlange05: I don't think `sense_habitat` is run anywhere?

SEAPODYM_Data_Test.py
```diff
@@ -0,0 +1,94 @@
+from parcels import Particle, JITParticle, AdvectionRK4
+from parcels.grid import Grid
+from argparse import ArgumentParser
+from py import path
+import numpy as np
+
+
+def updateUserVars(particle, grid, time, dt):
+    for var in particle.user_vars.keys():
+        setattr(particle, var, getattr(grid, var)[time, particle.lon, particle.lat])
```
mlange05: If you capitalise the name and add documentation we could move this over to `particle.py` as a helper kernel. It clearly needs to state that this makes an assumption about particle and grid variables names though. Once we have parameterisable kernels we could even let the user define the name via a name map, but that is best left for later.

SEAPODYM_Data_Test.py
```diff
@@ -0,0 +1,94 @@
...
+                   help='List of field variables to extract, using PARCELS naming convention')
+    p.add_argument('-n', '--netcdf_vars', default=['u', 'v', 'habitat', 'skipjack_cohort_20021015_density'],
+                   help='List of field variable names, as given in the NetCDF file. Order must match --variables args')
+    p.add_argument('-d', '--dimensions', default=['lat', 'lon', 'time'],
+                   help='List of PARCELS convention named dimensions across which field variables occur')
+    p.add_argument('-m', '--map_dimensions', default=['lat', 'lon', 'time'],
+                   help='List of dimensions across which field variables occur, as given in the NetCDF files, to map to the --dimensions args')
+    p.add_argument('-o', '--output_file', default='SEAPODYM_Particles',
+                   help='Name of output NetCDF file for particle data')
+    args = p.parse_args()
```
mlange05: This setup assumes a whole lot of default files that are not (and should not be) part of the main repo. If you are free to upload down-sampled versions of the these, I think we should move the entire example to `parcels-examples`.

As far as I can see it also doesn't add any automated tests to our regressions test suite, and the newly added features should easily be testable with one of the existing tests. This demonstrates that we need to start caring about what is a test and should be in the main repo and what goes into the examples sub-repo. I'll create an issue to that effect.

SEAPODYM_Data_Test.py
Jacketless: Actually, it would have been better to have not commit the test script on
my part, as it was more a working piece of test code and as you note now-
some of the functions I'm not really running. Similarly with the
updateUserVars kernel function, I kept this in the test script until we
were happy with the way I'd put the functionality in particle.py.

Perhaps I should instead write a more generic test based on peninsula or
moving eddies where we create an arbitrary grid to be sampled by the
particules rather than relying on the SEAPODYM data which does not really
need to be used as a test specifically.

parcels/particle.py
Jacketless: No, you're totally right- the kernel can do sampling for the initial
values. As you say, users can put their own error handling in a sampling
kernel they have written, and for now I have moved the error code into the
UpdateUserVars kernel to keep the Particle class clean.

parcels/particle.py
Jacketless: Ah yes, that's much cleaner- I'd forgotten i'd added that attribute to
ParticleType. Cool, thanks Michael.

erikvansebille:
Is this PR ready for merge now? @mlange05 what do you think?

mlange05:
So the the `SEAPODYM_Data_Test.py` files seems obsolete now that we have a much nicer and cleaner test for this. Otherwise this looks good and should be ready to go.

Jacketless:
Yes, I hoped it was ready to merge. I'm now working on the behavioural
stuff from this branch, and extending the test_sampler case for that. Let's
just delete the SEAPODYM test commit, I should never have included that.

Removing that is a case of rebasing, right @mlange05 ?

mlange05:
@Jacketless I hope you don't mind but I did the rebase for you and we now have two nice and clean commits. I'm happy for those to go in once travis is green.

Jacketless:
Au contraire, I'm super happy you did it for me!

PR #48, opened 2016-03-14
RK45, analytical test and Stommel test

Abobie:
The option to use an RK45 scheme to solve the equations using variable step sizes was added.

The analytical test case describes an eddy that flows eastward and decays with time. This test case is only time dependent.

The Stommel test case uses the steady-state solution of the Stommel equation, which describes periodic trajectories with large velocities along the western boundary. This test case is only space dependent.

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -9,7 +9,7 @@ import math
 import datetime
 
 __all__ = ['Particle', 'ParticleSet', 'JITParticle',
-           'ParticleFile', 'AdvectionRK4', 'AdvectionEE']
+           'ParticleFile', 'AdvectionRK4', 'AdvectionEE', 'AdvectionRK45']
 
 
 def AdvectionRK4(particle, grid, time, dt):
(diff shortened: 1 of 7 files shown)


mlange05:
Ok, so the branch is currently in an state that prevents auto-merging, but it should rebase cleanly on the current master. @Abobie can you please update your master branch and simply run `git rebase master` on this branch and then force-push to get it into a clean state again on github?

Another small note is that it would be great to have a rough description of the features and tests added in the description field of the PR. This makes the review process much easier, especially on large changes like this.

Abobie:
@mlange05 so sorry about the mess! I followed the process you described above now, and this seems to have cleaned things up, right?

mlange05:
Yes, this looks much better. We're still getting test failures on travis though, so could you please look at the assert values for the failing tests and see if they make sense?

Abobie:
I'm getting a segmentation fault in travis when running the peninsula test between two other tests. I don't know what causes it, but it seems to stem from the fact that all tests except the peninsula one use the same names for the C-files. Just changing the file name of the Stommel test so that it is run before the peninsula one solves the problem, but I haven't found a nicer clean solution yet... Any idea what's causing it?

Abobie:
@mlange05 in this last commit I just explicitly changed the file name of the output in the Stommel test to avoid any conflicts, so if this solution is acceptable this branch should be ready to merge now.

mlange05:
@Abobie I actually figured out the bug that caused this and fixed it in another branch (e4f476ed in dynamic-particle-sets). If you don't mind I'll do a bit of rebasing to get a clean commit history before we merge this.

Abobie:
@mlange05 ok, great! Sure, go ahead.

mlange05:
@Abobie  Ok, I just pushed a rebased version of this branch (`numerics_clean-rebased`). The rebase is not quite done yet and I'll continue with this soon, so this is just a temporary copy. 

All in all your stuff looks quite good already, but the biggest problem I see at the moment is that none of your new tests actually execute RK45 as part of the automated test suite and so all `execute_adaptive()` loops are effectively untested. Can you please pick up the rebased branch and run some of your validation tests by hand? Once you confirm that all functionality is still working after the rebase, we need to add some more acceptance tests that actually compares the RK45/RK4/EE runs to `ground_truth()` results.

Abobie:
@mlange05 I tried all the tests with RK45 now and everything worked fine except for a small typo in moving eddies. Should I push the correction to `numerics_clean` or to `numerics_clean_rebased`?

mlange05:
@Abobie Thanks for checking this. Please push to `numerics_clean_rebased` for now. Once I have completed the rebase and increased test coverage I will replace `numerics_clean` with it to update the PR.

Abobie:
@mlange05 Pushed it now. I think I did something weird when I tried to get a local copy of the `numerics_clean-rebased` branch. I ended up pulling it to my failed attempt at a copy in the end, resulting in some merge conflicts which I had to solve. What would have been the correct way to get a local copy of a remote branch? I tried `fetch` and `checkout --track -b origin/numerics_clean-rebased`.

mlange05:
Ok, so on a clean repo, simply doing `git fetch` or `git pull`, followed by `git checkout numerics_clean-rebased` should have worked. Once something is slightly out of sync, a hard reset always ensure you have exactly the state that is on the server: `git fetch origin && git reset --hard origin/numerics_clean-rebased`. Careful with the latter though, it will delete all unsaved changes from your current checkout.

Abobie:
Ok, thanks!

Abobie:
@mlange05 Hey, this should be ready to pull right? Would you like me to make a new PR of the `numerics_clean-rebased` branch or something?

mlange05:
@Abobie Unfortunately I don't think this is quite ready to be merged yet. I have just finished the initial rebase so that we now have a more logical commit history, but there are still quite a few problems in this branch:
* The filename hack (commit 190745d) should be removed and the segfault needs to be fixed properly. I can have a look at that, since it seems to be a low-level technical issue
* The tests that you added only run RK4, so essentially all the new functionality is untested. We should really parameter sweep across the 3 three methods, as well as `jit`/`scipy` and compare against ground truth values.
* There are still quite a few hacks in your timestep loop changes that assume single-particle lists or check for hard-coded kernel names. We need to generalise this before we can land the new functionality. I can have a look at re-structuring the loops a little once I find some more time.

Abobie:
@mlange05 I started adding tests that also use EE and RK45. However, since EE just isn't as accurate as the others it fails the current tests. I'm leaning towards writing separate asserts for EE, or do you think there is a better way to handle it?

Regarding the hacks/assumptions I think you are referring partly to the code I've used to write all steps taken by RK45 to the output file. This is something I have used for development purposes and should probably not have included in the PR. If you want I can have a look at it and remove those parts. There are still places where I check e.g. the name of the kernel though, and I'm not sure how that should best be generalized.

Abobie:
As embarrassing as it is, I noticed I had actually forgotten to add code in the kernel to make RK45 actually use JIT, so I added it now.

parcels/particle.py
```diff
@@ -277,16 +355,33 @@ def execute(self, pyfunc=AdvectionRK4, time=None, dt=1., timesteps=1,
...
-        timeleaps = int(timesteps / output_steps)
+        timeleaps = int(timesteps / output_steps)   # Number of output points
         # Execute kernel in sub-stepping intervals (leaps)
         current = time or self.grid.time[0]
-        for _ in range(timeleaps):
-            self.kernel.execute(self, output_steps, current, dt)
-            current += output_steps * dt
-            if output_file:
-                output_file.write(self, current)
+        if self.kernel.funcname == 'AdvectionRK45UpdateP' or\
```
erikvansebille: Wouldn't it make more sense to use `self.kernel.funcname.find('RK45')` here? Because what happens if a user adds yet another kernel? This seems like a slight hack

mlange05: Or even the slightly more pythonic version `if 'RK45' in self.kernel.funcname:`. :snake: :wink: I'll fix this during the rebase and further integration of this branch following PR #65.

mlange05:
Superseded by PR #78.

Issue #47, opened 2016-03-11
Check that particles remain within grid domain


erikvansebille:
Currently, there is no check on whether particles stay within the grid domain. For example, setting the advection time in test_peninsula to `time = 24 * 3600. * 100` will cause the particles to be moved well beyond the domain boundary

Now, it is not clear what intended behaviour is when a particle reaches the domain boundary. Should it stop? Should it 'die'? 
And what about global domains that are periodic (i.e. where the western edge directly connects to the eastern edge, as on a globe)? CMS uses a option here to flag whether the domain is periodic or not

Raising the issue now so that we can think about this

erikvansebille:
I had another think about this, after discussions with @Jacketless, and it is easy to create a function `check_withinbounds`, as I did in the commit above. However, this simple function now stops execution of the entire set as soon as at least one particle is out of bounds.

Better would be to have a set of optional behaviour if a particle gets out of bounds, and then let users choose. These option can be

1. Stop execution altogether
2. Stop execution only for particles that are out of bound (this would be expected behaviour for physical oceanographers); requires the particles to be able to be 'popped', as in #14.
3. Do not execute a particle if it means getting out-of-bound. This requires the old position to be saved, so we can revert back to it if a particle get out-of-bound
4. More sophisticated behaviour like 'bouncing back from a wall'. This can be implemented at later stages

Jacketless:
Since the new spatial interpolation routines implemented in PR #81, boundary conditions have become more apparent as PARCELS currently grinds to a halt if attempts at sampling field values outside the boundary are made.

The animal behavioural work requires diffusion/random jumps/walks etc. This often results in particles ending up past boundaries or at least much closer to them than would occur for passive particles. 
At present, PARCELS fails whenever sampled fields are out of bounds. This occurs not just when a particle ends up on land during the previous time-step, when we might wish the particle to "die", but also when the RK4 algorithm samples points as fractions of dt (e.g. calculating u1, u2, u3 & u4).

I suggest to keep things simple, we allow an option for particles to "die" if any sampling takes place outside a field boundary (even if it's just part of the RK4). I think we also need to allow an option for some kind of error or other object to be returned in place of the interpolated value from `interpolator1d()` in `field.py`. If this error object can propagate back to a user-defined kernel, the user can then choose some alternative action himself/herself (re-drawing a random jump, for example).

At the moment, putting a try-except statement within my user-kernels does not catch the error: they occur in the interpolation functions of `field.py`.

[...]

PR #49, opened 2016-03-15
Initial from density field and test

Jacketless:
added static function to initial a given number of particles
stochastically from a grid.field
small utility method within grid to add fields
short test script to show things working. Envision that it will be
added to for future additions to particle release code

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -35,17 +35,16 @@ def AdvectionEE(particle, grid, time, dt):
 
 def positions_from_density_field(pnum, startfield, mode='monte_carlo'):
     # initialise particles from a field
-    print("Starting from " + startfield.name + " field")
-
-    # Gradient ends up with overall negative sum of cells, not sure if this makes sense
-    # so working with lat long points still for the moment
-    # startcells = np.gradient(startfield.data[0,:,:])
-    # total = np.sum(startcells)
-
+    print("Initialising particles from " + startfield.name + " field")
     total = np.sum(startfield.data[0, :, :])
     startfield.data[0, :, :] = startfield.data[0, :, :]/total
     lonwidth = (startfield.lon[1] - startfield.lon[0])/2
     latwidth = (startfield.lat[1] - startfield.lat[0])/2
+
+    def jitter_pos(pos, width, list=[]):
+        list[-1] = pos + np.random.uniform(-width, width)
+        return list
+
     if(mode is 'monte_carlo'):
         probs = np.random.uniform(size=pnum)
         lon = []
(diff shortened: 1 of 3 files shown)


parcels/particle.py
```diff
@@ -148,6 +176,9 @@ def cptr(i):
             lon = np.linspace(start[0], finish[0], size, dtype=np.float32)
             lat = np.linspace(start[1], finish[1], size, dtype=np.float32)
 
+        if start_field is not None:
+            lon, lat = positions_from_density_field(size, getattr(self.grid, start_field))
```
mlange05: Could we not simply pass in the `Field` object itself instead of the name? We could even do some type checking at this point.

parcels/particle.py
```diff
@@ -33,6 +33,34 @@ def AdvectionEE(particle, grid, time, dt):
...
 
 
+def positions_from_density_field(pnum, startfield, mode='monte_carlo'):
+    # initialise particles from a field
+    print("Starting from " + startfield.name + " field")
+
+    # Gradient ends up with overall negative sum of cells, not sure if this makes sense
+    # so working with lat long points still for the moment
+    # startcells = np.gradient(startfield.data[0,:,:])
+    # total = np.sum(startcells)
```
mlange05: Are these just leftover debug statements?

mlange05:
Ok, so I've tried to run this with one of my own test fields and I believe the added jitter took quite a few particles outside the domain, which then causes errors during particle initialisation. Is there some way to avoid this?

parcels/particle.py
Jacketless: No- these are a commented out attempts to move from assigning particle positions at lat-lon points to assigning them to cells (you remember we discussed using np.gradient to do this last week?). But it doesn't work out as I expected because of course we end up with some negative gradient values assigned to some cells, which is not what we wanted I think.

The print statement can be removed if you want, I thought it was just a useful line out for the user.

parcels/particle.py
Jacketless: We could... My feeling was that by initialising from a field already in the grid, any type checking or QA that we include in `Grid` object initialisation later on would have already been carried out. Additionally, if we have new particle release events later during the simulation, they can be initialised from a field within the grid that was defined at initialisation, rather than passing around a 'Field' object in the main loop. 
.....But passing the Field object as you suggest doesn't really stop us from doing either of these, so if you think it's more Pythonic to do it this way, then cool! 🐍👍

Jacketless:
I guess it would be fairly trivial to just ensure this doesn't happen in the initialisation method here, by either re-drawing the random jitter or capping the particle position at the boundary, but this seems to be another example of issue #47 - perhaps it's time we started writing some simple functions that get called when particles end up beyond the boundaries of the grid. Even if they're just pushed back on for the moment...

parcels/particle.py
mlange05: You are right about the initialisation checks, but most of them happen in the `Field` constructor, so ensuring the object is of type `parcels.Field` should be enough. And yes, passing the object rather than its name is definitely more pythonic. :snake: :wink:

mlange05:
For now I would probably re-draw the jitter, but capping would be fine too. But I agree, once we have one or several solutions to #47 we should revisit this.

Jacketless:
@mlange05 Take a quick look at this changes in light of our comments. I still think there's some more convenience functions that I'd like to add under the same heading, but it would be nice to merge this before I'm away for a week or so with travel to Aus.
Cheers!

mlange05:
Yes, this looks good now. We'll probably want to add some more automated testing once we beef up the test suite, but this is good to go for now.

PR #51, opened 2016-03-18
Fix moving eddies test grid resolution

erikvansebille:
Fixed how eddy shapes are calculated in the test_moving_eddies script. Longitude and latitude are now treated separately; and all variables are now independent of xdim and ydim


erikvansebille:
Travis tests seem to have failed because the particle trajectories are now slightly different, as the velocity fields themselves have changed. Should we change the assert values, @mlange05 ?

mlange05:
The trajectories still look sensible to me, so feel free to change the assert values. All they do is make sure that the particles end up in roughly the right region. The rest looks good too, so feel free to merge once travis is green.

Issue #12, opened 2016-01-13
Add capacity to run in time-backward mode


erikvansebille:
For the GMD manuscript, we will need to have functionality to run particles in time-backward mode. As for the integration, this can be done by simply negating velocity fields. The trick is to also handle input fields in reverse order

erikvansebille:
OK, so I'm started looking into this. Initially, I thought I would just write a simple kernel 
```
def reverseUV(particle, grid, time, dt):
    grid.U[time, particle.lon, particle.lat] = -1 * grid.U[time, particle.lon, particle.lat]
    grid.V[time, particle.lon, particle.lat] = -1 * grid.V[time, particle.lon, particle.lat]
```
And then prepend that kernel to the `pset.execute` set

However, this will not work for three reasons:

1. While the right hand sides of these equations can be executes, they can't be assigned to the left hand side because `[time, particle.lon, particle.lat]` are floats, rather than integers
2. Even if we would be able to make this work, it will only negate velocities on the location where the particles are now. However, RK4 will also require the velocities at the intermediate points to be negated
3. There is no way to check that we haven't doubly-negated a velocity field if we call reverseUV for the same location again

So really, it seems that we have a few options: 

1. Put the negation inside the Advection* kernels (with the disadvantage that is a user defines its own kernel it will not necessarily use the same negated velocities)
2. Negate all the grid cells around the particles (although this will not fix problem 3. above)
3. Negate the entire velocity field when we read it in

Option 3. is the easiest to implement, but it can also be quite slow. In scipy mode python will probably do some under the hood vector magic; but not sure how this works in JIT mode. @mlange05 any thoughts?

mlange05:
Ok, I hadn't really thought about this problem in detail yet. Option 3 should work for both modes, but will probably be costly, as you point out. So how about option 4: Optionally perform negation in-place as part of the sampling routine. What I had in mind is this:
* Add an attribute `negate` to `Field` objects and ensure that it is `True` for U and V if we call `execute(..., reverse=True)`. We don't want to negate temperature, right?
* For Python, we then adjust `Field.eval()` to do:
```
val = ...
if self.negate:
    val *= -1.
return val
```
[...]

PR #52, opened 2016-03-24
Time backward mode

erikvansebille:
Time backward mode (#12) should now be working with this PR

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -164,6 +164,8 @@ class KernelGenerator(ast.NodeVisitor):
         self.grid = grid
         self.ptype = ptype
         self.field_args = OrderedDict()
+        # Hack alert: JIT requires U field to update grid indexes
+        self.field_args['U'] = grid.U
 
     def generate(self, py_ast, funcvars):
         # Untangle Pythonic tuple-assignment statements
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -279,7 +279,11 @@ class ParticleSet(object):
             output_steps = timesteps
         timeleaps = int(timesteps / output_steps)
         # Execute kernel in sub-stepping intervals (leaps)
-        current = time or self.grid.time[0]
+        if dt < 0:
+            startind = -1
+        else:
+            startind = 0
+        current = time or self.grid.time[startind]
         for _ in range(timeleaps):
             self.kernel.execute(self, output_steps, current, dt)
             current += output_steps * dt
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -249,16 +249,16 @@ class ParticleSet(object):
         self.particles = np.delete(self.particles, indices)
         return particles
 
-    def execute(self, pyfunc=AdvectionRK4, time=None, dt=1., timesteps=1,
+    def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
                 output_file=None, show_movie=False, output_steps=-1):
         """Execute a given kernel function over the particle set for
         multiple timesteps. Optionally also provide sub-timestepping
         for particle output.
 
         :param pyfunc: Kernel funtion to execute
-        :param time: Starting time for the timestepping loop
+        :param starttime: Starting time for the timestepping loop
+        :param endtime: End time for the timestepping loop
         :param dt: Timestep interval to be passed to the kernel
-        :param timesteps: Number of individual timesteps to execute
         :param output_file: ParticleFile object for particle output
         :param output_steps: Size of output intervals in timesteps
         :param show_movie: True shows particles; name of field plots that field as background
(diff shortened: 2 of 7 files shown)


tests/test_moving_eddies.py
```diff
@@ -96,6 +96,39 @@ def moving_eddies_example(grid, npart=2, mode='jit', verbose=False,
 
 
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
+def moving_eddies_fwdbwd(mode, npart=2):
```
mlange05: The new test isn't being picked up by pytest. According to conventions it should be called `test_moving_eddies_fwdbwd` to add it to the automated test suite.

parcels/grid.py
```diff
@@ -104,6 +108,12 @@ def eval(self, x, y):
         v = self.V.eval(x, y)
         return u, v
 
+    def reversetime(self):
+        self.U.dtsign = -1.
+        self.V.dtsign = -1.
+#        self.W.dtsign = -1.  # uncomment line once vertical-vels are implemented
+        return self
+
```
mlange05: For convenience, how about adding a `reverse` or even `reversetime` argument to the signature of `grid.execute()` that calls this if `reverse=True`? Or we could even move the setting of `dtsign` into `exeute()` altogether, since we currently have no way of cleanly deactivating the "reverse" mode?

parcels/particle.py
```diff
@@ -206,10 +206,14 @@ def execute(self, pyfunc=AdvectionRK4, time=None, dt=1., timesteps=1,
             output_steps = timesteps
         timeleaps = int(timesteps / output_steps)
         # Execute kernel in sub-stepping intervals (leaps)
-        current = time or self.grid.time[0]
+        if self.grid.U.dtsign < 0 and self.grid.V.dtsign < 0:
```
mlange05: Again, adding the `reverse=False` default argument to execute would make this line a little cleaner.

mlange05:
This looks really great now. Apart from the small API change and the test suite comment I'm happy for this to land.

@erikvansebille Do you mind if I squash a few commits before merging this to keep the history clean?

mlange05:
I had another thought on this regarding the API changes: Why are we still passing a positive `dt` in `pset.execute(..., dt=dt)` in the moving eddies test? Wouldn't setting `dt=-dt` be the most intuitive way to trigger time-backward mode?

That said, if the `dt` value passed into a kernel is negative, do we still need to invert U and V sampling? Or should we instead just make sure that we have more intuitive time-stepping parameters in `ParticleSet.execute()` and let negative `dt` values propagate to the kernel. Something like `execute(starttime=start, endtime=end, dt=dt` and equivalently `execute(starttime=end, endtime=start, dt=-dt)`?

erikvansebille:
This idea of `execute(starttime=end, endtime=start, dt=-dt)` is an interesting idea yes. Perhaps we could even infer that `dt` < 0 if `starttime` > `endtime`. 

And you may be right that if `dt` < 0, we don't need to negate the velocity fields. 

I'll have a look at it over the next few days

erikvansebille:
I've been reworking the whole PR today, and have revamped the way time is handled by `.execute`. Rather than the old `time` and `timesteps` variables, `.execute` now requires `starttime` and `endtime`, which hopefully are more intuitive. If the sign of `dt` is not equal to the sign of `endtime - starttime`, PARCELS will now negate `dt`.

Something more to do is to change output_steps to the more intuitive output_freq (every how many _seconds_ should output be saved, rather than every how many `dt`)

mlange05:
@erikvansebille Ok, so I found out what was causing this bug and added a patch that fixes it for this particular case. However, the problem is really a deeper one that highlights that numpy arrays are really the wrong container type for our JIT approach once you add particle removal, since `np.delete` invalidates the Python memory views for our `JITParticles`.  Anyway, it works for now and I also added an xfailing test that highlights the case where it doesn't (see b703af9). 

I also rebased the branch to get a clean history for the time backward changes and I believe it is now ready to be merged. Please have another look first though.

parcels/codegenerator.py
```diff
@@ -164,6 +164,8 @@ def __init__(self, grid, ptype):
         self.grid = grid
         self.ptype = ptype
         self.field_args = OrderedDict()
+        # Hack alert: JIT requires U field to update grid indexes
+        self.field_args['U'] = grid.U
```
erikvansebille: Thanks @mlange05, the PR looks all good. Only thing I don't complete understand is why we only do this line 168 in `codegenerator.py` for `grid.U` and not for the other `grid` fields. Is this because it's a hack?

Anyway, I'm happy with you pulling this one into master

mlange05: Yes, it's totally hacky, but the underlying JIT code to update `p->xi` and `p->yi` always uses `U`, which is already a quite hacky. I'll clean that up soon hopefully and do a more generalised field -> index mapping.

PR #53, opened 2016-03-24
Fixed a small type-o in moving eddies test

erikvansebille:
I suddenly spotted a very small typo in moving eddies example (a V where there should be a U). Quickly fixed it here. 

Am doing it through a PR rather than directly on master as I understand this is the way to fix even simple bugs like this?


mlange05:
Looks good, merging.

PR #55, opened 2016-03-30
Dynamic particle sets

mlange05:
This merge brings additional set manipulation capabilities for the `ParticleSet` class,  including the addition and removal of one or more particles from any given set. The merge comes with a battery of unit tests for all manipulation routines, as well as two important bug-fixes, in particular:
* Single particle instantiation
* Addition of single particles and merging of `ParticleSets`
* Removal of particles from a set via indexes
* Self-removal of particles from within kernels via `particle.delete()`
* Bug-fix for auto-unindentation for in-lined kernel definitions
* Bug-fix in kernel naming to avoid clashes in the test suite

Note that certain set operations are not yet supported, such as duplicate-and-merge (`pset3 = pset1 + pset2`) and removal of real particle objects rather than indices; these are currently xfailing tests. Comments on the desired details of the API itself are very welcome. 

Also note that the addition and removal operations we are using are currently slow, especially for large sets, since we rely on numpy arrays under the hood. Optimisation of these data structures is left for another day though since it's non-trivial.

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -217,6 +217,10 @@ class ParticleSet(object):
     def __setitem__(self, key, value):
         self.particles[key] = value
 
+    def __iadd__(self, particles):
+        self.add(particles)
+        return self
+
     def add(self, particles):
         if isinstance(particles, ParticleSet):
             particles = particles.particles
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -77,6 +77,7 @@ class Particle(object):
         self.lat = lat
         self.xi = np.where(self.lon >= grid.U.lon)[0][-1]
         self.yi = np.where(self.lat >= grid.U.lat)[0][-1]
+        self.active = 1
 
         for var in self.user_vars:
             setattr(self, var, 0)
@@ -87,6 +88,9 @@ class Particle(object):
     @classmethod
     def getPType(cls):
         return ParticleType(cls)
+
+    def delete(self):
+        self.active = 0
 
 
 class JITParticle(Particle):
@@ -100,7 +104,8 @@ class JITParticle(Particle):
     """
 
     base_vars = OrderedDict([('lon', np.float32), ('lat', np.float32),
-                             ('xi', np.int32), ('yi', np.int32)])
+                             ('xi', np.int32), ('yi', np.int32),
+                             ('active', np.int32)])
     user_vars = OrderedDict()
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


parcels/particle.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, lon, lat, grid, cptr=None):
         self.lat = lat
         self.xi = np.where(self.lon >= grid.U.lon)[0][-1]
         self.yi = np.where(self.lat >= grid.U.lat)[0][-1]
+        self.active = 1
```
erikvansebille: Is there a reason we use numeral `1` here, rather than `True`? And same of curse for inactive particles?

parcels/particle.py
```diff
@@ -96,11 +104,16 @@ class JITParticle(Particle):
     """
 
     base_vars = OrderedDict([('lon', np.float32), ('lat', np.float32),
-                             ('xi', np.int32), ('yi', np.int32)])
+                             ('xi', np.int32), ('yi', np.int32),
+                             ('active', np.int32)])
```
erikvansebille: Again, couldn't we reduce memory footprint by using booleans here? Or is this not implemented?

tests/test_particle_sets.py
```diff
@@ -0,0 +1,156 @@
...
+def test_pset_remove_particle(grid, mode, npart=100):
+    lon = np.linspace(0, 1, npart, dtype=np.float32)
+    lat = np.linspace(1, 0, npart, dtype=np.float32)
+    pset = grid.ParticleSet(npart, lon=lon, lat=lat, pclass=ptype[mode])
+    for ilon, ilat in zip(lon[::-1], lat[::-1]):
+        p = pset.remove(pset[-1])
+        assert(p.lon == ilon)
+        assert(p.lat == ilat)
+    assert(pset.size == 0)
+
```
erikvansebille: I must say I don't completely understand what the difference is between `p = pset.remove(-1)` (which does work) and `p = pset.remove(pset[-1])` (which doesn't). Does the former mean we can only remove the last particle? Why would one want to use the latter indexed version rather than the former? What is the advantage?

parcels/particle.py
mlange05: C traditionally represents booleans as integers and this way the interfacing with the JIT backend seemed easiest. That said, I haven't tried interfacing through native booleans yet.

parcels/particle.py
mlange05: As above, the C compiler will use an integer under the hood, I believe. Also, a native single-byte boolean might skew memory alignment.

tests/test_particle_sets.py
mlange05: The difference is that the latter passes the particle object itself, which would allow more user friendly (and more Pythonic) constructs, such as:
```
for p in pset:
    if p.lon > 0.2:
        pset.remove(p)
```
The problem with this is that `np.delete` only works with array indices, not objects, and finding the index for a given object in a set adds another costly  search to an already expensive operation.

erikvansebille:
merging dynamic-particle sets. This should also be useful for #47, where we can now stop particles when they reach a grid/domain boundary

Issue #16, opened 2016-01-15
Time-varying tracer fields in plotting script


mlange05:
The visualisation script can currently not handle time-varying tracer fields. Fixing this will require interpolation between the individual frames of tracer data. Since this can be quite a computationally intense process, depending on your tracer field, we might also want to start thinking about creating permanent output, ie. plot and video files to avoid having to recompute this.

PR #56, opened 2016-04-02
On the fly movies with fields

erikvansebille:
Improvements to the way on-the-fly-movies (i.e. movies while the code is run, particularly useful for testing/debugging) are made. Added a variable `show_movie` to .execute. If `show_movie = True`, then only the particle locations are plotted. If `show_movie = U` or any other grid field, that field will be shown in the background. See also #16.

Note that travis builds fail because of `Invalid DISPLAY variable`. Simplest solution to fix this travis error is to set `show_movie = False` in `moving_eddies_example` but then the code won't be tested, of course. Any suggestions?

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -146,7 +146,13 @@ class Field(object):
     def show(self, **kwargs):
         import matplotlib.pyplot as plt
         t = kwargs.get('t', 0)
-        data = np.squeeze(self.data[t, :])
+        idx = self.time_index(t)
+        t0 = self.time[idx-1]
+        t1 = self.time[idx]
+        f0 = self.data[idx-1, :]
+        f1 = self.data[idx, :]
+        val = f0 + (f1 - f0) * ((t - t0) / (t1 - t0))
+        data = np.squeeze(val)
         vmin = kwargs.get('vmin', data.min())
         vmax = kwargs.get('vmax', data.max())
         cs = plt.contourf(self.lon, self.lat, data,
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -168,10 +168,6 @@ class Field(object):
         cs.cmap.set_under('w')
         cs.set_clim(vmin, vmax)
         plt.colorbar(cs)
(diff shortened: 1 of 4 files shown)


parcels/field.py
```diff
@@ -91,10 +92,24 @@ def __getitem__(self, key):
...
     @cachedmethod(operator.attrgetter('interpolator_cache'))
-    def interpolator(self, t_idx):
+    def interpolator2D(self, t_idx):
         return RectBivariateSpline(self.lat, self.lon,
                                    self.data[t_idx, :])
 
+    def interpolator1D(self, idx, time, y, x):
+        # Return linearly interpolated field value:
+        if x is None and y is None:
+            t0 = self.time[idx-1]
```
mlange05: Unfortunately this breaks for `idx=0`, ie. with static fields like we use in the peninsula test.

parcels/particle.py
```diff
@@ -244,17 +246,25 @@ def execute(self, pyfunc=AdvectionRK4, time=None, dt=1., timesteps=1,
...
+        plt.clf()
         plt.plot(np.transpose(lon), np.transpose(lat), 'ko')
-        if field is None:
+        if field is True:
+            axes = plt.gca()
+            axes.set_xlim([self.grid.U.lon[0], self.grid.U.lon[-1]])
+            axes.set_ylim([self.grid.U.lat[0], self.grid.U.lat[-1]])
             plt.show()
         else:
+            field = getattr(self.grid, field)
```
mlange05: I find passing the field itself, rather than it's name more intuitive and pythonic :snake:, so I would suggest something like:
```
if not isinstance(field, Field):
    field = getattr(self.grid, field)
```

mlange05:
I don't think we need to test interactive features explicitly through travis, since we can't auto-verify correctness anyway. Other than that, and a few in-lined comments above, this looks really great!

erikvansebille:
I've implemented all your suggestions now, @mlange05. Thanks for these! Good to pull, then?

PR #57, opened 2016-04-07
Animated field plotting

Jacketless:
Made some general improvements to plotParticles, including simpler
colour plotting of a recorded variable in particles
command line arg tracerfile now replaced with more generic
backgroundfield taking NetCDF and plots first timestep for static
plots, and animates through time for movie2d


scripts/plotParticles.py
```diff
@@ -1,64 +1,86 @@
...
-      plt.contourf(np.squeeze(X),np.squeeze(Y),np.squeeze(P))
+    if(recordedvar is not 'none'):
+        rMax = np.max(pfile.variables[recordedvar])
+        record = pfile.variables[recordedvar]
+
+    if backgroundfield is not 'none':
+        bfile = Dataset(backgroundfield, 'r')
+        bX = bfile.variables['x']
+        bY = bfile.variables['y']
+        bT = bfile.variables['time_counter']
```
erikvansebille: Why not use the same functions here as we use in the main parcels code? Can we use the `Grid.from_netcdf `function? That would also allow users to use the `variables = {}` and `dimensions={}` dictionaries in a similar way as in the main PARCELS code and make updating codebases much easier.

scripts/plotParticles.py
```diff
@@ -1,64 +1,86 @@
...
+            ax.cla()
+            if recordedvar is not 'none':
+                scat = ax.scatter(lon[:, i], lat[:, i], s=60, c=record[:, i]/rMax)
+            else:
+                scat = ax.scatter(lon[:, i], lat[:, i], s=60, c='white', edgecolors='black')
+            if backgroundfield is not 'none':
+                field_time = np.argmax(bT > time[0, i]) - 1
+                plt.contourf(bX[:], bY[:], bVar[field_time, 0, :, :], zorder=-1, vmin=0, vmax=bMax,
+                             levels=np.linspace(0, bMax, 10))
+            return scat,
```
erikvansebille: A lot of new code has landed with PR #56 and PR #59  doing prettyprinting of titles and axes etc; Again, is there some way to hook into that code in `parcels.py`?

Jacketless:
I have a couple of new commits to finalise, and then I suggest we merge this branch (as it's only changes to plotParticles.py), then look at taking the good parts from this PR and PR #56 and adding them to each other.

Jacketless:
Are we good to have this PR merged? I don't understand why github is telling me that this branch has conflicts that must be resolved...

erikvansebille:
PR looks good to me; although I think that in the longer term we want to have more overlap between the online and offline plotting functions. But that can be done later.
Still, I think it would be good if @mlange05 has one more glance over it
And as for why automatic merging is not possible, it seems there's a bit of a mess-up in the plotParticles.py itself. There's some conflicts between the HEAD and master there. Could you try fixing this, @Jacketless?

Jacketless:
I agree, it is a hefty dependency. I'm not adverse to losing it, but I find it really useful to have some kind of landmask for reference, and this was the easiest way I could get this. Perhaps someone knows and alternative method with less dependency on large packages?

erikvansebille:
I agree that the Basemap dependency is a bit too hefty for how core it is to PARCELS. 
Unfortunately, Basemap seems to have cornered the market so I couldn't find any simpler packages to do this. Shall we land this PR once the Basemap is removed?

Jacketless:
I've made a few adjustments to this branch since. Let me clean it up and remove the Basemap and I'll re-pull request it.

Jacketless:
I think this should be ready to land, minus the "branch conflicts". I'm sorry, I don't really know what that means/refers to... 🤔

erikvansebille:
Hi @Jacketless , what's the status of this PR? Except for a few minor conflicts in scripts/plotParticle.py, this one should be good to land. Can you use the command line (see below) to resolve these conflicts? Then I will merge

Jacketless:
Ok, I think I've resolved the conflicts, but now all tests are failing because the travis build doesn't have matplotlib... I'm sure this wasn't an issue before?

Jacketless:
Closing pull request as this is long out of date. I have a local version of plotParticles which contains many of this new functionality. I will endeavour to make a new branch.

PR #59, opened 2016-04-07
Include netcdf timeorigin

erikvansebille:
Added option to pass `time_origin` to `Grid.from_netcdf`, which is useful for grids that use calendar time.

By passing a time_origin (often given in netcdf files like those from NEMO, http://gws-access.ceda.ac.uk/public/nemo/runs/ORCA0083-N06/means) to `.execute`, the time on the movies is pretty-printed, and the output netcdf files will show the correct time

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -1,4 +1,4 @@
-from netCDF4 import Dataset
+from netCDF4 import Dataset, num2date
 from parcels.field import Field
 from parcels.particle import ParticleSet
 from py import path
@@ -17,11 +17,12 @@ class Grid(object):
     :param time: Time coordinates of the grid
     :param fields: Dictionary of additional fields
     """
-    def __init__(self, U, V, depth, time, fields={}):
+    def __init__(self, U, V, depth, time, time_origin=0, fields={}):
         self.U = U
         self.V = V
         self.depth = depth
         self.time = time
+        self.time_origin = time_origin
         self.fields = fields
 
         # Add additional fields as attributes
@@ -78,7 +79,16 @@ class Grid(object):
             fields[var] = Field.from_netcdf(var, dimensions, dsets, **kwargs)
         u = fields.pop('U')
         v = fields.pop('V')
-        return cls(u, v, u.depth, u.time, fields=fields)
+
+        # assign time_origin if the time dimension has units and calendar
+        try:
+            time_units = dsets[0][dimensions['time']].units
+            calendar = dsets[0][dimensions['time']].calendar
+            time_origin = num2date(0, time_units, calendar)
+        except:
+            time_origin = 0
+
+        return cls(u, v, u.depth, u.time, fields=fields, time_origin=time_origin)
 
     @classmethod
     def from_nemo(cls, basename, uvar='vozocrtx', vvar='vomecrty',
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -78,7 +78,6 @@ class Grid(object):
             fields[var] = Field.from_netcdf(var, dimensions, dsets, **kwargs)
         u = fields.pop('U')
         v = fields.pop('V')
-
         return cls(u, v, u.depth, u.time, fields=fields)
 
     @classmethod
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -1,4 +1,4 @@
-from netCDF4 import Dataset, num2date
+from netCDF4 import Dataset
 from parcels.field import Field
 from parcels.particle import ParticleSet
 from py import path
@@ -17,12 +17,11 @@ class Grid(object):
     :param time: Time coordinates of the grid
     :param fields: Dictionary of additional fields
     """
-    def __init__(self, U, V, depth, time, time_origin=0, fields={}):
+    def __init__(self, U, V, depth, time, fields={}):
         self.U = U
         self.V = V
         self.depth = depth
         self.time = time
-        self.time_origin = time_origin
         self.fields = fields
 
         # Add additional fields as attributes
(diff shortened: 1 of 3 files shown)


parcels/grid.py
```diff
@@ -78,7 +79,16 @@ def from_netcdf(cls, filenames, variables, dimensions, **kwargs):
             fields[var] = Field.from_netcdf(var, dimensions, dsets, **kwargs)
         u = fields.pop('U')
         v = fields.pop('V')
-        return cls(u, v, u.depth, u.time, fields=fields)
+
+        # assign time_origin if the time dimension has units and calendar
+        try:
+            time_units = dsets[0][dimensions['time']].units
+            calendar = dsets[0][dimensions['time']].calendar
+            time_origin = num2date(0, time_units, calendar)
```
Jacketless: We are now assuming in a couple of places (another being in the `particle.py` constructor) that the time dimension from one field is the same for all. This is usually the U field, but here it's whatever is the first `dset`.

Is it time that we put some validation code in to ensure that `time` is consistent across fields, or if that does not need to be forced, that the user can specify the "time-origin" field? I had some odd results recently that took me a couple of days to track down because the time dimension from one field was at a different scale to another...

erikvansebille: This is a good point @Jacketless. I guess the solution is to move the `time` vector from the `grid` class to the `field` class altogether. In that way, each `field` in a `grid` can have its own `time` (and `time_origin`)

Is that how you solved the problem when you encountered it? Or did you do something else?

Jacketless: Actually I just manually rescaled the time data in the original netCDF. The time was been days since an origin for a particular, whereas other fields were in seconds since the same origin. The interpolator appeared to just be using the final time-step of the incorrectly scaled field for the entire execution loop, and so initially I did not pick up that there was a problem.

I suggest that fields have their own `time`, but that a warning is displayed when values are interpolated outside of that time range for any field? This check could be added in `eval` function of the `Field` class?

erikvansebille: OK, so  as I understand it, the whole idea of the `num2date` function of netCDF4 is to convert everything in the same units (seconds in this case). 

I just pushed a commit where `time_origin` is in the `Field` class. So your situation, where one field is in seconds and the other in days, shouldn't happen anymore with this new PR. Could you check, @Jacketless ?

Jacketless: Nice! Looks good, it appears that it makes the calculation for me now when `time_origin` is being assigned in `Field`. Cool, one less thing to worry about! 😎

erikvansebille: Great, glad to hear it works now! Squashing bugs one at a time ;-)

On Thu, 14 Apr 2016 at 04:37 Joe Scutt Phillips <notifications@github.com>
wrote:

> In parcels/grid.py
> <https://github.com/OceanPARCELS/parcels/pull/59#discussion_r59659677>:
>
> > @@ -78,7 +79,16 @@ def from_netcdf(cls, filenames, variables, dimensions, **kwargs):
> >              fields[var] = Field.from_netcdf(var, dimensions, dsets, **kwargs)
> >          u = fields.pop('U')
> >          v = fields.pop('V')
> > -        return cls(u, v, u.depth, u.time, fields=fields)
> > +
> > +        # assign time_origin if the time dimension has units and calendar
> > +        try:
> > +            time_units = dsets[0][dimensions['time']].units
> > +            calendar = dsets[0][dimensions['time']].calendar
> > +            time_origin = num2date(0, time_units, calendar)
>
> Nice! Looks good, it appears that it makes the calculation for me now when
> time_origin is being assigned in Field. Cool, one less thing to worry
> about! 😎
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/OceanPARCELS/parcels/pull/59/files/cd4818b0ce498a704314881e1389881df26f71a8..581a6324f6a5a2601eb90f1850b62d88c0bb794d#r59659677>
>
-- 
Dr. Erik van Sebille
Lecturer in Oceanography and Climate Change
Grantham Institute & Department of Physics
Imperial College London
London, United Kingdom
---------------------------------
e: E.van-Sebille@imperial.ac.uk
t: +44 7 481 294 686
w: www.erik.vansebille.com

erikvansebille: Great, glad to hear it works now! Squashing bugs one at a time ;-)

parcels/particle.py
```diff
@@ -180,6 +181,7 @@ def __init__(self, size, grid, pclass=JITParticle,
         self.particles = np.empty(size, dtype=pclass)
         self.ptype = ParticleType(pclass)
         self.kernel = None
+        self.time_origin = grid.U.time_origin
```
mlange05: Do we need to store this on the particle set if we always pull `time_origin` from the field?

parcels/particle.py
```diff
@@ -356,26 +362,29 @@ def __init__(self, name, particleset, initial_dump=True):
         self.trajectory[:] = np.arange(particleset.size, dtype=np.int32)
 
         # Create time, lat, lon and z variables according to CF conventions:
-        self.time = self.dataset.createVariable("time", "f8", ("trajectory", "obs"), fill_value=0.)
+        self.time = self.dataset.createVariable("time", "f8", ("trajectory", "obs"), fill_value=-9999.)
```
mlange05: Arbitrary numerical fill values are rarely a good idea. Would `np.nan` not work?

mlange05:
Apart from a few minor nitpciks this looks good.

parcels/particle.py
erikvansebille: Since we need `time_origin` in the `__init__` of `ParticleFile` (see below), and that `__init__` function doesn't have access to the `field` class, I copy the value here to the `Particle` Class.
It was either that, or passing a `grid` or `field` argument to  `ParticleFile`

parcels/particle.py
erikvansebille: Yes, works! Thanks for the suggestion!

PR #62, opened 2016-04-18
Gradient calculation and test

Jacketless:
Functionality for field-level gradient calculation, including variable sample distances longitudinally by latitude, was required. This PR now allows users to call a `Field` object function returning new gradient fields in X and Y for the whole field, or a defined subset in space and/or time.

Added `.gradient` function to `Field` objects, using central
differencing and forward/backward differencing at edges.

Created test that creates small grid and compares internal `Field`
object gradient calculation to Numpy’s gradient function that cannot
use unequal sample distances.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -8,7 +8,36 @@ import operator
 from ctypes import Structure, c_int, c_float, c_double, POINTER
 
 
-__all__ = ['Field']
+__all__ = ['CentralDifferences', 'Field']
+
+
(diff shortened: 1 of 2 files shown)


mlange05:
Ok, as a first step towards advanced field numerics this is really great. In the future we might want to extend this further to do on-the-fly gradient calculations within kernels, and we might also want to start thinking about how to generalise this for "all grid types", but this provides a really neat first-level test for further development. Thanks a lot @Jacketless ! IMHO this can go in as is.

PR #63, opened 2016-04-26
Improved testing and time handling

mlange05:
This merge brings a few small bug fixes, improvements to the test base and automated handling of `datetime.timedelta` objects in the execution loop. In detail:
* Handle and interpret `datetime.timedelta` objects in the execution loop
* Use timedelta objects in example tests
* Add baseline tests for grid creation and field sampling functionalities
* Fix standalone `field` visualisation
* Allow setting `output_interval` without specifying an output file

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -296,7 +296,7 @@ class ParticleSet(object):
         timesteps = int((endtime - starttime) / dt)
 
         # Check if output is required and compute outer leaps
-        if output_file is None or output_interval <= 0:
+        if output_interval <= 0:
             output_steps = timesteps
         else:
             output_steps = abs(int(output_interval / dt))
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -6,7 +6,7 @@ import netCDF4
 from collections import OrderedDict, Iterable
 import matplotlib.pyplot as plt
 import math
-import datetime
+from datetime import timedelta as delta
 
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -35,5 +35,5 @@ script:
     - flake8 tests
     - flake8 examples
     - python tests/test_peninsula.py --grid 20 10
-    - py.test -v -s tests/
+    - py.test tests/
```
erikvansebille: Why would we want less verbosivity here? I liked being able to see which tests worked and which didn't. Is there a reason for this?

parcels/particle.py
```diff
@@ -6,7 +6,7 @@
 from collections import OrderedDict, Iterable
 import matplotlib.pyplot as plt
 import math
-import datetime
+from datetime import timedelta
```
erikvansebille: Since we import `timedelta` in the tests as `delta`, should we do the same here, for consistency?

erikvansebille:
Apart from two minor issues (level of verbosivity in test.py and whether to import `timedelta` as `delta` or not), I think this is a very nice housekeeping PR! Cleans up some of the code!

.travis.yml
mlange05: Hehe, I was getting annoyed with reading so many compilation messages, but sure we can drop this change until our test base has grown a bit further.

mlange05:
Cool, addressed comments, rebased and green travis. This should be good to go.

PR #64, opened 2016-04-30
Vertical velocities

erikvansebille:
Adding vertical velocities to the Advection schemes


erikvansebille:
Only scipy version of vertical velocity has been implemented yet. jit doesn't work at all. Needs to be fixed, hoping @mlange05 can lead the way

parcels/field.py
```diff
@@ -39,7 +39,11 @@ def __init__(self, name, data, lon, lat, depth=None, time=None,
             # Make a copy of the transposed array to enforce
             # C-contiguous memory layout for JIT mode.
             self.data = np.transpose(self.data).copy()
-        self.data = self.data.reshape((self.time.size, self.lat.size, self.lon.size))
+        if self.depth.size > 1:
```
mlange05: Having two potential shapes for `self.data` creates a whole lot of special casing that can be avoided by defaulting to a `np.zeros(1, dtype=np.float32)` in the constructor above. This should be treated the same way we treat time and it will also make things easier when dealing with the interpolators below, I think.

parcels/particle.py
```diff
@@ -15,25 +15,55 @@
...
-    lon2, lat2 = (particle.lon + u2*.5*f_lon, particle.lat + v2*.5*f_lat)
-    u3, v3 = (grid.U[time + .5 * dt, lon2, lat2], grid.V[time + .5 * dt, lon2, lat2])
-    lon3, lat3 = (particle.lon + u3*f_lon, particle.lat + v3*f_lat)
-    u4, v4 = (grid.U[time + dt, lon3, lat3], grid.V[time + dt, lon3, lat3])
+    posvertdir = -1.  # TODO: Decide on direction of positive w
+    u1 = grid.U[time, particle.lon, particle.lat, particle.dep]
+    lon1 = particle.lon + u1*.5*f_lon
+    v1 = grid.V[time, particle.lon, particle.lat, particle.dep]
+    lat1 = particle.lat + v1*.5*f_lat
+    if hasattr(grid, 'W'):
```
mlange05: I really don't like the extreme special casing in this routine; it goes against the entire philosophy of writing simple, easy-to-understand kernels. And also, `hasattr` is a pure Python construct that has no real equivalent in C code. 

However, since we don't have a dimension-independent "position" abstraction yet (ie. a vector variable, instead of several scalar variables), I would argue we simply always compute `w` and let it compute a whole lot of zeros for now. We then really need to create a "vector variable" abstraction though before we can don any sincere benchmarking, since we want to get rid of this blatant performance overhead.

erikvansebille: I agree indeed that this is ugly code, and take full responsibility. We should indeed strive to make it much cleaner before it lands, yes!

So do you suggest to always assume a grid has W? Or how would we deal with grids that don't have W (of which there are many) if we can't use `hasattr`? Any nice pythonic suggestions would be great here!

mlange05: Ok, as far as I can see we have 3 potential solutions here:
1. Bite the bullet and implement vector fields / vector variables. This would mean we sample the field with something like `vel = grid.Velocity[time, particle.position]`, where `grid.Velocity` is `{U, V}` in 2D and `{U, V, W}` in 3D. Note that `vel` and `particle.position` are now vector variables with 2 or 3 components according to grid dimension.
2. Simply replicate the kernels and write out `AdvectionRK4_2D` and `AdvectionRK4_3D`. Note, that might require us to do some sanity checking about whether a kernel is applicable, given a particular grid. This, in turn, will require us to store some meta-information with kernels, such "local variables used" or "fields sampled", which we can get out of the JIT analysis engine. 
3. Assume every grid has `W` and create a kind of "dummy field" that always returns zeros for 2D runs. As mentioned before this will be wasteful in terms of performance, and will prevent us running any meaningful benchmarks for a while.

Now, 1 is clearly "the right thing to do" in the long run, since we are trying abstract grid assumptions, such as the U-V-W split of velocity, away from the user. However, it is also by far the most complicated to implement and might take a significant portion of someone's time, so I'm ok with putting in one of the other quick fixes first, as long as we are all aware this needs to happen eventually. I'm also open to alternatives, of course...

Jacketless: I think for the moment, separate RK4_2D and RK4_3D methods is the way to
go. The user will have to choose the appropriate kernel, and we can put a
single check before the main loop of `ParticleSet.execute` to ensure that
the grid has the appropriate dimensions.

erikvansebille:
I feel that we're almost there with the vertical velocities. Python 2.7 works completely (including vertical velocities in JIT, it seems!), there is just a strange python3 bug. @mlange05 could you have a quick look at some point?

PR #65, opened 2016-05-04
Automated in-place unit conversion (m to degree)

mlange05:
This merge adds automatic unit conversion via `UnitConverter` objects and implements converters for meter-to-degree conversion (`Geographic`) and geographic with pole correction (`GeographicPolar`). The new API allows the units of a `Field` to be set as `field.units = Geographic()` or via the constructor. NEMO grids (`Grid.from_nemo()`) default to `U->GeographicPolar` and `V->Geographic` units, while all other grid constructors assume meters.

Note, we also introduce hashed cache keys to the JIT pipeline, so that we can add the converter types to the cache key for the generated files. This fixes errors due to several different kernels trying read/write/execute from the same set of files.

### DIFF
--- a/tests/test_grid_sampling.py
+++ b/tests/test_grid_sampling.py
@@ -172,10 +172,10 @@ def test_meridionalflow_sperical(mode, xdim=100, ydim=200):
     pset = grid.ParticleSet(2, pclass=pclass(mode), lon=lonstart, lat=latstart)
     pset.execute(pset.Kernel(AdvectionRK4), endtime=endtime, dt=delta(hours=1))
 
-    assert(pset[0].lat - (latstart[0] + endtime.total_seconds() * maxvel / 1852 / 60) < 1e-7)
-    assert(pset[0].lon - lonstart[0] < 1e-7)
-    assert(pset[1].lat - (latstart[1] + endtime.total_seconds() * maxvel / 1852 / 60) < 1e-7)
-    assert(pset[1].lon - lonstart[1] < 1e-7)
+    assert(pset[0].lat - (latstart[0] + endtime.total_seconds() * maxvel / 1852 / 60) < 1e-4)
+    assert(pset[0].lon - lonstart[0] < 1e-4)
+    assert(pset[1].lat - (latstart[1] + endtime.total_seconds() * maxvel / 1852 / 60) < 1e-4)
+    assert(pset[1].lon - lonstart[1] < 1e-4)
 
 
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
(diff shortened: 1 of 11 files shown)


tests/test_grid_sampling.py
```diff
@@ -21,6 +22,40 @@ def grid(xdim=200, ydim=100):
...
+
+
+@pytest.fixture
+def grid_geographic_polar(grid_geographic):
+    grid = grid_geographic
+    for i, y in enumerate(grid.U.lat):
+        # Apply inverse conversion for each latitude value
+        grid.U.data[0, i, :] *= cos(y * pi / 180)
+    grid.U.units = GeographicPolar()
+    return grid
```
erikvansebille: Note that in virtually all cases, U and V will have the same grid layout. It is only the lon and lat dimensions that are different in geographic_polar

tests/test_moving_eddies.py
```diff
@@ -61,8 +61,8 @@ def cosd(x):
         U[:, :-1, t] = np.diff(P[:, :, t], axis=1) / dy / corio_0 * g
         U[:, -1, t] = U[:, -2, t]  # Fill in the last row
 
-    return Grid.from_data(U, lon, lat, V, lon, lat,
-                          depth, time, field_data={'P': P})
+    return Grid.from_data(U, lon, lat, V, lon, lat, depth, time, field_data={'P': P},
+                          u_units=GeographicPolar(), v_units=Geographic())
```
erikvansebille: I think we should have simply one `units=GeographicPolar()` here, because we can assume that U and V have the same grid. Also, I think that `units=GeographicPolar()` should be the default (i.e. if you don't provide units, it should assume GeographicPolar()

mlange05:
Ok, I've added some changes to the JIT pipeline to this PR, so that we can enable the appropriate sampling tests. Unfortunately one still seems bugged :frowning:.

Following discussion with @erikvansebille we should probably also add some explicit tests for meridional and zonal advection before we merge to verify the effect of the pole correction.

mlange05:
Ok, I've added the discussed advection tests and the convenience API for for defining unit types from a keyword in the grid constructor. 

Following this, a nasty bug that was causing issues in the RK4/5 branch re-appeared and it took me quite a while to track that down. During the bug hunt I have added some small convenience changes for debugging, mostly to do with the lower-level details of JIT mode, so it would be great if @erikvansebille could have a second look and ok these before they get merged.

erikvansebille:
I've just added two unit tests that I think are more powerful than the ones that were pushed before. They work on a very simple zonal or meridional flow on a sphere, advect two particles in that from different start locations, and `assert` whether AdvectionRK4 moves them according to the analytical solution which is
- for the meridional case `lat = latstart + vel * Dt` and `lon = lonstart`
- for the zonal case `lon = lonstart + vel * Dt * cos(latstart)` and `lat = latstart`

By the way, the precision with which `AdvectionRK4` works in this case is` <1e-7` for `scipy`, but only `<1e-4` for `JIT`; quite shocking that there is such a big difference (and that `JIT` is less precise than `scipy`)

mlange05:
Thanks @erikvansebille the additional tests look great. I take it this branch is ready to land then?

Regarding the precision problem, unless there is a bug somewhere, I believe the reason is in the difference between the interpolation schemes used; our JIT mode uses a very simple bilinear intra-cell interpolation, whereas we rely on `scipy.interpolate` routines that use a spline evaluation backend for the pure-python mode. I would hope that switching to bicubic interpolation in JIT mode might alleviate the problem for now, but we should raise a separate issue for that.

erikvansebille:
Upon browsing through the code one more time, I suddenly realise there is one key potential issue we still need to test for: that, even in spherical mode, any 'other' fields (i.e. not U or V) are not converted. I've now added that to the last test case, hope is ok. 
@mlange05 if you approve of this addition, then I'll pull the code into master. Not sure why Travis failed; going through the log there doesn't seem to be any problem...

mlange05:
@erikvansebille Yes, the new test looks great and even uncovered a small bug in the JIT kernel caching. I just pushed a fix, so feel free to merge once travis is green.

PR #67, opened 2016-05-18
Particle Set Density Calculation and Test

Jacketless:
Additions to `particle.py` for a `ParticleSet` density function, which
returns a the binned density per area for each vertex in `grid`

Small test script

Still to do:
Add density timestep option to `ParticleSet.execute’ for execution loop calculation of particle densities

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -344,15 +344,10 @@ class ParticleSet(object):
             mainleap = abs(int(inner_interval / dt))
             # Calculate density field areas outside main loop
             density_field.area = np.zeros(np.shape(density_field.data[0,:,:]), dtype=np.float32)
-            dx = (density_field.lon[1] - density_field.lon[0]) * 1852 * 60 * np.cos(density_field.lat*math.pi/180)
+            dx = (density_field.lon[1] - density_field.lon[0]) * 1852 * 60 * np.cos(density_field.lat*np.pi/180)
             dy = (density_field.lat[1] - density_field.lat[0]) * 1852 * 60
             for y in range(len(density_field.lat)):
                 density_field.area[y, :] = dy * dx[y]
-=======
-        # Check if output is required and compute outer leaps
-        if output_interval <= 0 or output_interval > abs(endtime-starttime):
-            output_steps = timesteps
->>>>>>> master
         else:
             outerleap = 1
             outerfunction = empty
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -253,12 +253,11 @@ class ParticleSet(object):
 
     def density(self, field):
         Density = np.zeros((field.lon.size, field.lat.size), dtype=np.float32)
-
         # For each particle, find closest vertex in x and y
         for p in self.particles:
             Density[np.argmin(np.abs(p.lon - field.lon)), np.argmin(np.abs(p.lat - field.lat))] += 1
         # Scale by cell area (assumes field has previously had area data calculated)
-        Density /= field.area
+        Density /= np.transpose(field.area)
         return Density
 
     def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
@@ -316,16 +315,17 @@ class ParticleSet(object):
         timesteps = int((endtime - starttime) / dt)
 
         # Some simple wrapper functions used to abstract the main execution loop functions
-        def density_wrapper(field, current, output_file, movie_field):
-            field.data[np.where(field.time == current), :, :] = self.density(field)
+        def density_wrapper(density_field, current, output_file, movie_field):
+            dtime = np.argmin(np.abs(current - density_field.time))
+            density_field.data[dtime, :, :] = np.transpose(self.density(density_field))
 
-        def output_wrapper(field, current, output_file, movie_field):
+        def output_wrapper(density_field, current, output_file, movie_field):
             if output_file:
                 output_file.write(self, current)
             if movie_field:
                 show(field=movie_field, t=current)
 
-        def empty(field, current, output_file, movie_field):
+        def empty(density_field, current, output_file, movie_field):
             pass
 
         # Check if output is required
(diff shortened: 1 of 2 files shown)


parcels/particle.py
```diff
@@ -251,6 +254,26 @@ def remove(self, indices):
...
 
+    def density(self, lon=None, lat=None):
+        if lon is None:
+            lon = self.grid.U.lon
+        if lat is None:
+            lat = self.grid.U.lat
+        Density = np.zeros((lon.size, lat.size), dtype=np.float32)
+        Area = np.zeros(np.shape(Density), dtype=np.float32)
+
+        dx = (lon[1] - lon[0]) * 1852 * 60 * np.cos(lat*math.pi/180)
```
erikvansebille: Should we use the new geographic_conversion routines here? (PR #65)

parcels/particle.py
```diff
@@ -85,6 +85,9 @@ def __init__(self, lon, lat, grid, cptr=None):
         for var in self.user_vars:
             setattr(self, var, 0)
 
+        for var in self.user_vars:
+            setattr(self, var, 0)
+
     def __repr__(self):
```
erikvansebille: Why are these three new lines duplicated now? Or aren't they the same as the three lines above?

parcels/particle.py
```diff
@@ -251,6 +254,26 @@ def remove(self, indices):
...
+
+        dx = (lon[1] - lon[0]) * 1852 * 60 * np.cos(lat*math.pi/180)
+        dy = (lat[1] - lat[0]) * 1852 * 60
+        for y in range(len(lat)):
+            Area[0:len(lon), y] = dy * dx[y]
+
+        # For each particle, find closest vertex in x and y
+        for p in self.particles:
+            Density[np.argmin(np.abs(p.lon - lon)), np.argmin(np.abs(p.lat - lat))] += 1
+        # Scale by cell area
```
erikvansebille: This is quite a computing-intensive operation. For one, don't we know `xi` and `yi` for each particle already? So I think we can replace `np.argmin(np.abs(p.lon - lon))` with `np.xi`, right?

And also, we don't need to recalculate `Area` on every call of `density()`; it is not going to change. Perhaps calculate only on first call of `density() `and then store it in the `grid` class?

tests/test_density.py
```diff
@@ -0,0 +1,168 @@
...
+    lon1, lat1 = (lon + u1*.5*f_lon, lat + v1*.5*f_lat)
+    u2, v2 = (fieldx[time + .5 * dt, lon1, lat1], fieldy[time + .5 * dt, lon1, lat1])
+    lon2, lat2 = (lon + u2*.5*f_lon, lat + v2*.5*f_lat)
+    u3, v3 = (fieldx[time + .5 * dt, lon2, lat2], fieldy[time + .5 * dt, lon2, lat2])
+    lon3, lat3 = (lon + u3*f_lon, lat + v3*f_lat)
+    u4, v4 = (fieldx[time + dt, lon3, lat3], fieldy[time + dt, lon3, lat3])
+    Vx = (u1 + 2*u2 + 2*u3 + u4) / 6.
+    Vy = (v1 + 2*v2 + 2*v3 + v4) / 6.
+    return [Vx, Vy]
+
```
erikvansebille: Why are you using your own RK4 here? Why not the built-in AdvectionRK4?

erikvansebille:
Nice PR, Joe! I've made a few comments on the files, have a look at these. 

As for adding density timestep to `ParticleSet.execute’, I'm assuming you don't want to be doing this every single timestep, right? So an option similar to the `output_interval`?

parcels/particle.py
Jacketless: Yes! This was written before that PR was finalised. I will familiarise myself with the conversion routines and incorporate.

parcels/particle.py
Jacketless: Ah yes, I forgot `particle.xi` already exists- we can use that.

Regarding area, `density()` can actually be called with arbitrary lat/lon grid spacings, so in this case, yes we would want to recalculate area. However, you're right that in most cases it is unlikely to change. I suggest that at the creation of a `Grid` object, area is calculated and stored from the U field spacing, and then this is used if the default lon/lat arguments are passed to `density()`

Jacketless:
An `output_interval` type argument is exactly what I was working on! Will finish up and re PR 👍

tests/test_density.py
Jacketless: So as I have been developing alternative kernels for the diffusion/random walk behaviours referenced in #68, I find myself having to interpolate through time using RK4 for a lot of different fields. It seems needless to continually re-write the routine substituting `U` and `V` for my field variables of interest, and so I've abstracted the RK4 slightly in this version- you just pass the x and y fields you wish to get the forcing from.

@mlange05, not sure if this is mucking up things on the JIT side, but it would be really useful to have a generic RK4 method that can be called for fields other than `U` and `V`.

Jacketless:
In making additions to the branch, it struck me as much more pythonic to pass the actual `Field` object that defined and stored particle densities to the `execute` function. Particle density will be calculated at the timesteps of this field, and stored on the grid for writing later.

However, as we can now have three different operations occurring in the main particle advection execution (density calculation, writing output to file, and kernel execution), i have completely re-written the loop to account for different possibilities. 

I've run a few of the test scripts to ensure that things still work, but would **really** appreciate a couple of people taking a good look at this- a couple of the travis tests are failing and this may be due to the number of iterations in "main loop" (kernel execution) calculated as zero... I will try and take another look at these fails.

I'm not sure of the new geographic conversion routines- I need to spend some more time looking at how you have implemented them in other examples.

parcels/particle.py
Jacketless: It turns out `particle.xi` and `yi` are not updated each time-step, so we need to do this binning somewhere every time we want to calculate densities anyway.

Jacketless:
Closing this PR, as complex innerloop stepping is something we agreed to come back to. A simplified version of this density calculation now exists in PR #111

Issue #70, opened 2016-05-20
RK4/Jit problem in GlobCurrent data


erikvansebille:
Simple advection now works for scipy in the GlibCurrent data, but somewhat surprising the jit doesn't work at all. The velocities are way off, and particles simply get swept outside of the domain very quickly.

I have made a test script in the [parcels-examples/tree/globcurrent_support](https://github.com/OceanPARCELS/parcels-examples/tree/globcurrent_support) branch, but because we don't have travis enabled on the parcels-examples repository I had to do it in the `__main__`

So you'll see that python `examples/test_globcurrent.py` works for scipy, but not for jit

mlange05:
Ok, I think there are a few things going on here:
* Examples are enabled with travis in `parcels`, but you need to pull in the changes from `parcels-examples` through a separate branch in the main repo (eg.  commit 4624e89).
* The `starttime`, which comes from the grid itself, is much greater than the `endtime`, and thus we go into an assumed backward-in-time mode. I'm not sure that this is intentional, but it highlights that our time arguments convention in `execute`, including the use of `deltas`, is not entirely clear yet (eg. should `delta` indicate `starttime + delta`) and needs more rigorous unit testing to capture all possible uses and misuses.
* None of the above explains why the JIT mode gives a different field sampling result than the Python interpolators on the same time frame, so this is definitely a bug and I'll keep digging...

erikvansebille:
Thank you @mlange05 for taking a look
Yes, I saw the point about the starttime too, and agree it will need to be fixed. Another issue is that `time_origin` is not properly parsed because the time units are spelled `Unit` rather than `units`.

However, the bigger issue is the different velocities in JIT and scipy. printing e.g. `u4` at every step in the AdvectionRK4 shows that they are very different. I expect something with the unit conversion?

mlange05:
Ok, I think I have this triaged: The `lat` and `lon` arrays from the file are 64-bit doubles, but our JIT code assumes 32-bit floats. This causes all the grid indexing to go out of the window and hence erroneous field sampling. Since re-casting might be kind a bit tedious (and potentially slow) I would argue we should be able to adjust our JIT code to the data type in the file, but that might require a careful re-think of the engine. 

I'll try to come up with a clean fix later, but in the mean-time feel free to cast `lon` and `lat` to the right data type in the field constructor to bypass this for now.

erikvansebille:
Thanks @mlange05, I can indeed confirm that adding the four lines
```
[...]

PR #71, opened 2016-05-28
64bit lon/lat and better time support

erikvansebille:
This PR:
- fixes bug #70 by casting lon/lat to 32 bits
- adds a new .execute keyword `runtime` that sets `endtime = starttime + runtime`
- fixes some small bugs in calendar support

While Travis is green for python2.7, there is something wrong with python3. However, I don't have that installed, so can't easily test what goes wrong. @mlange05, any idea?

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -353,15 +353,17 @@ class ParticleSet(object):
             axes.set_xlim([self.grid.U.lon[0], self.grid.U.lon[-1]])
             axes.set_ylim([self.grid.U.lat[0], self.grid.U.lat[-1]])
             namestr = ''
+            time_origin = self.grid.U.time_origin
         else:
             if not isinstance(field, Field):
                 field = getattr(self.grid, field)
             field.show(animation=True, **kwargs)
             namestr = ' on ' + field.name
-        if field.time_origin == 0:
+            time_origin = field.time_origin
+        if time_origin is 0:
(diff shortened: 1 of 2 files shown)


parcels/particle.py
```diff
@@ -280,22 +283,36 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
-
-        # Check if starttime, endtime and dt are consistent and compute timesteps
+        if isinstance(starttime, datetime):
+            starttime = (starttime - self.time_origin).total_seconds()
+        if isinstance(endtime, datetime):
+            endtime = (endtime - self.time_origin).total_seconds()
+
+        # Check if starttime, endtime, runtime and dt are consistent and compute timesteps
+        if runtime is not None and endtime is not None:
+            sys.exit("ERROR: Only one of (endtime, runtime) can be specified")
```
mlange05: Why not just `raise RuntimeError('Only one ....)`? I think that would be more :snake:.

mlange05:
Ok, Python3 error is fixed, and other than the error handling this looks good to go. Sorry for the delay in reviewing/fixing.

parcels/particle.py
erikvansebille: Fixed now to suggested pythonic version of error handling (learning something new every day!)

PR #72, opened 2016-05-31
Lagrangian diffusion

Jacketless:
This test script provides kernels for Lagrangian diffusion in
non-uniform fields of diffusivity

Uses implementation from the particle density calculation PR #67 to show how a simple random walk does not result in true diffusion


Jacketless:
This functionality is now addressed in the updated PR #130

PR #73, opened 2016-06-01
Install requirement fixes

mlange05:
Due to an unexpected hardware fault (:cry:), I recently had to re-install parcels from scratch and tripped over a few dependency issues for non-anaconda installs. These changes should fix them :smile:. We now also deal with a missing `matplotlib` more gracefully, since this is quite often missing on larger machines.


erikvansebille:
Looks like a good cleanup, and removing the matplotlib requirement makes sense. 
Another issue I often have is with cachetools. I couldn't get it to work on cx1, and also not very easily on my mac in python3 (it's not part of macports, for example). How important is cachetools?

No worries if it is key to performance, then let's keep it. But if it's not super-critical perhaps we can also do something like with matplotlib?

mlange05:
Ok, so we use `cachetools.LRU` for caching the interpolation splines for field evaluation, and without it the SciPy mode runs crippling slow. I had no problems with it for Python3 though, since it just came from pip via `pip install -r requirements.txt`. Did that not work for you?

erikvansebille:
OK, if it really serves a purpose, then let's keep it in. I didn't try pip install on my mac, wanted to do everything through macports (long story...)

I'll pull this one in then

Issue #74, opened 2016-06-02
Odd field sampling


Jacketless:
I've been getting some strange results when sampling from none U/V forcing fields, and I'm not sure where they were coming from.

I have a `Field` object, with data that I have checked is > 0 in all cases. Occasionally however, when a kernel samples this field for continuous [time, lon, lat] values I'm given a negative number by the interpolator. This can still occur when I remove the RK4 temporal interpolation and use discrete time.

Is there a way that the `RectBivariateSpline' function can be somehow returning me negative numbers? Perhaps it is interpolating with null cell values present in some points on the field...

mlange05:
This type of problem is virtually impossible to track down without a [minimal working example](https://en.wikipedia.org/wiki/Minimal_Working_Example). Can you please provide a (down-sampled) grid file and a gist that demonstrates this potential bug? Also, have you run this with JIT by any chance?

Jacketless:
Yes sorry I'm aware that bug tracking is difficult without this. I had started writing minimal example, but ran out of time and thought it better to at least send an email off. I will try and put something together over the weekend with a downscaled field.

Jacketless:
There is now a minimal working example of this as part of PR #77 . It's possible that there is a casting issue going on, although some of the returned values from grid sampling are well below the grid minimum.

I'm aware this example is not something we necessarily want to merge, but I did it this way in case it later becomes part of an internal test.

Jacketless:
Having spent a little more time looking into this, I've narrowed the strange field sampling down to be a result of using Scipy's `RectBivariateSpline`. When there is a strong gradient and vertex values are close to zero, the spline can return negative values when interpolating between all positive values.

I've got a temporary fix going using linear 2d interpolation `interp2d` from Scipy, but this raises the issue of us perhaps wanting to use linear interpolation when a spline evaluation returns values outside a range specified by the user (although this would be slow). Alternatively, users could just specify interpolation on each `Field` object?

erikvansebille:
Did PR #81 resolve this question now? Can it be closed?

mlange05:
Yes, this is now fixed in PR #81.

PR #77, opened 2016-06-08
Test Script to Show potential interpolation bugs

Jacketless:
As discussed in #74 this test should hopefully show this potential interpolation bug.

An arbitrary field of random numbers is generated, and shown to contain only values 0-1.
When particles sample the field from within a kernel, occasionally values of < 1 are returned. Due to the stochasticity of particle positions and field values, you may have to run the script a couple of times.

### DIFF
--- a/FieldSampleBug.py
+++ b/FieldSampleBug.py
@@ -1,0 +1,99 @@
+from parcels import *
+import numpy as np
+from argparse import ArgumentParser
+
+
+def CreateArbitraryField(grid):
+    """Generate a grid that contains the physical eddy dynamics
+    from the original moving eddies test example, overlaying a
+    habitat map containing a single maxima and gaussian gradient
+    """
+    depth = grid.U.depth
+    time = grid.U.time
+
+    # Coordinates of the test grid (on A-grid in deg)
+    lon = grid.U.lon
+    lat = grid.U.lat
+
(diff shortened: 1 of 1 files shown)


mlange05:
Ok, so I had a look at this and indeed this is a very neat test that exposes a subtle bug. I have repackaged the test to be part of our regular test runs and found a potential solution in branch `regular-grid-sampling`. This switches to using `RegularGridInterpolator` instead of `RectBivariateSpline`, which I suspect to cause the overshoot, and this sees to fix the original test. 

However, the branch currently fails due to the `ParticleSet(start_field)` initialiser creating lat/lon values just outside the domain for the repackages test. I'm not sure this is an unrelated bug or whether this positional error was the original cause of the spline interpolator creating overshoots. I'll continue investigations...

mlange05:
Uhh also, setting `numpy.random.seed(somevalue)` ensures that the stochasticity remains recreatable and avoids unnecessary re-running.

mlange05:
Ok, it turns out the stochastic initialisation bug was not to blame for the overshoots and we do need to change to a more conservative interpolator in SciPy mode. PR #81 should address both bugs and also adds a miniature version of this test to the regular test base.

Jacketless:
Great stuff, good job @mlange05 - thanks for looking into this.

PR #78, opened 2016-06-13
RK45 and fully adaptive execution loops

mlange05:
This merge is an adaptation of @Abobie's initial work (PR #48) that primarily uses "adaptive" While-loops for particles throughout and implements RK45 advection. It also adds a cleaner versions of the advection tests according to Fabbroni, 2009, and re-structures the internal main loop to allow a kernel to "fail", ie. causing the engine to not update time while keeping the time loop going. Further work is clearly needed to make the internal execution structure more robust and handle different outcomes of a kernel execution (and remove the forward/backward special-casing).  

@Abobie and @erikvansebille, please have a detailed look and test the new loop infrastructure thoroughly before merging, since some of these changes are quite invasive.

Please also note that I have disabled the Stommel test for now (although the timestepping has been updated to reflect the new API), since the acceptance criteria was not actually using the computed truth values, and I currently don't understand the details of this test enough. This should be fixed in a follow-up PR.


parcels/particle.py
```diff
@@ -257,7 +314,8 @@ def remove(self, indices):
         return particles
 
     def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
-                runtime=None, output_file=None, output_interval=-1, show_movie=False):
+                runtime=None, interval=None, output_file=None, tol=None,
+                show_movie=False):
```
erikvansebille: I don't think that the keyword `interval` is specific enough here. Users who don't go into the code won't understand the difference with the `dt` keyword.
I think `output_interval` worked better. Yes, the keyword does more than control output, but to a user, that's what it's about

parcels/particle.py
```diff
@@ -290,55 +350,58 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
+            if dt > 0.:
+                dt *= -1.
+                print("negating dt because running in time-backward mode")
+            if interval > 0.:
+                interval *= -1.
+                print("negating interval because running in time-backward mode")
+
+        # Initialise particle timestepping
+        for p in self:
+            p.time = starttime
```
erikvansebille: So we can't have particles in the same set start at a different time? What about an experiment where a particle is released every day or so? Should we allow users to set the `p.time` before calling the `.execute`? Or is this a separate keyword (`p.starttime`?)

parcels/particle.py
```diff
@@ -290,55 +350,58 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
+            p.dt = dt
+        # Execute time loop in sub-steps (timeleaps)
+        timeleaps = int((endtime - starttime) / interval)
+        assert(timeleaps >= 0)
+        leaptime = starttime
         for _ in range(timeleaps):
```
Jacketless: I also think it would be worth thinking about the particle density calculation here and how it goes in the loop. You'll see I had a go at this in PR #67. Perhaps you could take a look and include the relevant code from there and we can scrap that PR?

tests/test_stommel.py
```diff
@@ -0,0 +1,141 @@
...
+def stommel_grid(xdim=200, ydim=200):
+    """Generate a grid encapsulating the flow field consisting of two
+    moving eddies, one moving westward and the other moving northwestward.
+
+    The original test description can be found in: K. Doos,
+    J. Kjellsson and B. F. Jonsson. 2013 TRACMASS - A Lagrangian
+    Trajectory Model, in Preventive Methods for Coastal Protection,
+    T. Soomere and E. Quak (Eds.),
+    http://www.springer.com/gb/book/9783319004396
+    """
```
erikvansebille: This reference here is incorrect. @Abobie could you suggest the correct reference?

parcels/particle.py
mlange05: This has not been possible before either, so this changes does not affect any current features. In fact it is a first step towards a more complex time-stepping run-time that allows different start times. However, such changes are beyond the scope of this PR, and require some further careful design.

parcels/particle.py
mlange05: Agreed, I didn't want yet another level of sub-timestepping before we have a better concept of how to deal with the growing complexity of the timestepping loop. Once this PR lands we definitely need to look at multi-levelled sub-timestepping again.

parcels/particle.py
mlange05: Since complex multi-level sub-timestepping will require `interval` to be derived from multiple sources, I currently favour the more general term. I understand that this is not the most user friendly, but this variable should not actually be set by users once we have a more detailed API for our increasingly complex time loop. In the meantime I believe a detailed docstring should do the trick, while we should really focus on designing a clean and concise time stepping API.

parcels/particle.py
```diff
@@ -36,6 +37,57 @@ def AdvectionEE(particle, grid, time, dt):
     particle.lat += v1 * dt
 
 
+def AdvectionRK45(particle, grid, time, dt):
+    tol = [1e-9]
```
erikvansebille: Shouldn't this `tol` be set by the user? It is a keyword in the `.execute` now, but doesn't seem to be passed on? (Thanks @Abobie for pointing out!)

Abobie:
I'm getting an error when trying to run RK45 in scipy:
`NameError: global name 'KernelOp' is not defined`
I think I got the same error earlier with scipy not finding `math` even though it is imported as it should, do you remember the fix for this @mlange05 ?

parcels/particle.py
mlange05: Technically yes, but it breaks all the current abstractions we have. By that I mean it shouldn't really be in the `execute` interface in the first place, since it has no meaning for any other algorithm. It also doesn't seem to have any effect on the current tests so I would leave this until someone re-enables the Stommel test and implements appropriate acceptance criteria that highlight the sensitivity to `tol`.

In the long term, one really cool, but complicated way to do this would be to have smart kernel-factories that generate parameterisable versions of known kernels on-the-fly, for example `AdvectionRKN(N=8, tol=1.e-6)`. A somewhat easier approach would be to have a global "parameter dictionary" to expose these algorithm-specific options and thus allow users to override them. However, none of these suggestions are feasible with the current infrastructure, so I just hard-coded it to get this branch merged.

mlange05:
@Abobie are you sure you are picking up the right version of parcels? Have you set your PYTHONPATH to a checkout of the new branch?

Abobie:
@mlange05 oh you're right, I realized had my PYTHONPATH set to `/home/joel/Documents/parcels`, switched it to `/home/joel/Documents/parcels/parcels` which seemed to fix it

erikvansebille:
OK I think we're all good to land this one then. @Abobie, do you want to do the honours ;-)

PR #81, opened 2016-06-14
Regular grid sampling

mlange05:
This merge changes the SciPy interpolator in our Python/SciPy implementation from `RectBivariateSpline` to `RegularGridInterpolator`. This fixes sampling overshoots that can arise from a random-value field with sharp gradients, as reported by @Jacketless in issue #74. Note that the new interpolator is closer to what we do in JIT, but it has two distinct side effects:
* Lower accuracy of the `field.eval()` function, which requires lowering tolerances in sampling tests
* Particles moving out of the domain bounds now cause interpolator errors. In Python mode this was previously handled by extrapolation through the spline interpolator and we are still extrapolating in JIT mode.

Please note also that another bug in the stochastic particle set initialiser was found during this bug fix, which was fixed through a small re-write of the initialiser function. An appropriate test has of course also been added. @Jacketless, please review this carefully.

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -96,10 +96,10 @@ def positions_from_density_field(pnum, field, mode='monte_carlo'):
     lonwidth = (field.lon[1] - field.lon[0]) / 2
     latwidth = (field.lat[1] - field.lat[0]) / 2
 
-    def add_jitter(value, width, min, max):
-        value += np.random.uniform(-lonwidth, lonwidth)
+    def add_jitter(pos, width, min, max):
+        value = pos + np.random.uniform(-width, width)
         while not (min <= value <= max):
-            value += np.random.uniform(-width, width)
+            value = pos + np.random.uniform(-width, width)
         return value
 
     if mode == 'monte_carlo':
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -88,31 +88,33 @@ def AdvectionRK45(particle, grid, time, dt):
         return KernelOp.FAILURE
 
 
-def positions_from_density_field(pnum, startfield, mode='monte_carlo'):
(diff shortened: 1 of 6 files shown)


parcels/particle.py
```diff
@@ -88,31 +88,33 @@ def AdvectionRK45(particle, grid, time, dt):
...
+def positions_from_density_field(pnum, field, mode='monte_carlo'):
+    """Initialise particles from a given density field"""
+    print("Initialising particles from " + field.name + " field")
+    total = np.sum(field.data[0, :, :])
+    field.data[0, :, :] = field.data[0, :, :] / total
+    lonwidth = (field.lon[1] - field.lon[0]) / 2
+    latwidth = (field.lat[1] - field.lat[0]) / 2
+
+    def add_jitter(value, width, min, max):
+        value += np.random.uniform(-lonwidth, lonwidth)
```
Jacketless: shouldn't this line get a random.uniform between -width and width, rather than lonwidth? We use this same function for jittering values of latitude

parcels/particle.py
```diff
@@ -88,31 +88,33 @@ def AdvectionRK45(particle, grid, time, dt):
...
+        while not (min <= value <= max):
+            value += np.random.uniform(-width, width)
+        return value
+
```
Jacketless: In the case of `value` being outside min or max, should we not re-jitter `value` from it's original position, rather than add yet another random jitter? `value` could conceivably move further and further away from our defined min or max this way.

Jacketless:
Apart from two minor issues in the monte carlo particle initialiser function `positions_from_density_field`, this looks good to merge. Thanks for cleaning up @mlange05 , small and neat tests too 👍

PR #82, opened 2016-06-15
Custom `random` module for use in JIT mode

mlange05:
This merge adds a custom Parcels-specific `random` module to the kernel API that works similar to the `math` module and is available in JIT kernels or through pure python. It provides the functions:
* `random()`: Random float between 0 and 1
* `uniform(low, high)`: Random float between `low` and `high`
* `randint(low, high)`: Random integer between `low` and `high`

The merge also adds a set of basic arithmetic tests that covers the core arithmetic and logical functionalities of the JIT engine. It is also worth noting that the `random` module uses the basic C `rand()`/`srand()` API, so I make no guarantees about the quality of the RNG sequence. Once this becomes an issue we can incorporate other RNGs, like [PCG](http://www.pcg-random.org/download.html), but for now plain old C avoids licensing issues.


parcels/kernel.py
```diff
@@ -35,30 +36,34 @@ def fix_indentation(string):
 class Kernel(object):
     """Kernel object that encapsulates auto-generated code.
 
-    :arg filename: Basename for kernel files to generate"""
+    :arg grid: Grid object providing the field information
+    :arg ptype: PType object for the kernel particle
+    :arg grid: Grid object providing the field information for the kernel
```
erikvansebille: Why twice the `grid` argument here?

erikvansebille:
Looks really good! Happy for this to land, if the minor issue with the double 'grid' in kernel.py is sorted

mlange05:
Cool, thanks, I just pushed the fix. Let's wait for travis to finish before we merge.

PR #83, opened 2016-06-16
JITParticle float32/float64 declaration bug

erikvansebille:
There was a bug in JIT mode where the `time` in the JITParticle was declared as a `float32`, even though time of the `Field` class is `float64`, causing issues in the globcurrent test. 
When I fixed this by changing `time` in JITParticle to `float64` (see 8f2662f), other errors popped up where `py.test` hangs on `test_kernel_language`.

Inspection of the compiled code showed that the compilation adds a line `unsigned char _cgen_pad[4];` to the end of the `typedef struct` of the particles. 

For `test_kernel_language` that can be fixed by also making `dt` a `float64`, but then the same issue appears with `test_advection`. This probably needs someone with deeper understanding of the JIT compilation to have a look at, @mlange05 ?

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -163,7 +163,7 @@ class JITParticle(Particle):
     """
 
     base_vars = OrderedDict([('lon', np.float32), ('lat', np.float32),
-                             ('time', np.float32), ('dt', np.float32),
+                             ('time', np.float64), ('dt', np.float32),
                              ('xi', np.int32), ('yi', np.int32),
                              ('active', np.int32)])
     user_vars = OrderedDict()


mlange05:
OK, so the main problem here is that we still use use `numpy` to allocate our particle arrays, which does not honour the padding `cgen` generates for the raw C datatypes we create for JIT, and thus the two data type definitions start deviating once we start introducing alignment. Unfortunately, `numpy` does not provide great control over the allocation under its arrays, which means our data access assumptions are no longer true for the JIT loops. 

After having played with this for a little while, I don't see a safe solution to this problem other than implementing our own particle list API in C/C++, which as we know is quite tedious and troublesome (see branches `cpp-dynamic-lists` and `cpp-dynamic-lists-boost`). So, before we continue burning time on this, do we really need to switch `particle.time` to `double`, ie. are the "globcurrrent" files hitting the 2^32 limit?

erikvansebille:
Thanks for investigating further, @mlange05. What's happening is that the .execute hangs in an infinite loop.
To see, add the line
`    print('"%d\\n", time')`
somewhere in AdvectionRK4 in particle.py, and then run the examples/test_globcurrent.py (the one in the  globcurrent_support branch)
You'll see that time stays 1 infinitely. So the new while loop in .execute never seems to be finished, in JIT at least (all's good in scipy). I think that's because of an integer overflow, because if I make `particle.time` a double, it does work

mlange05:
Yes, I encountered the infinite time loop, and I believe that this is due to the alignment issues I mentioned earlier. The problem is that we currently rely on the `np.zeros(dtype=JITParticle)` to allocate memory for our particle array, where `JITParticle` defines the amount of bytes that make up a single particle. Since so far all variables are 32-bit (`np.float32` and `np.int32`) there is no extra padding and the amount of bytes per particle is the same as the C compiler derives from our `struct` definition in the C code. Our JIT engine currently relies entirely on the fact that these two definitions agree and that they create the exact same offset for a single particle in Python and C mode.

Now, once we start introducing 64 variables, `cgen` attempts to auto-align the entire C struct to 64bits, although this can be suppressed. For `numpy` there might be something similar happening under the hood, but we have no control over it. My current hypothesis now is that the offset calculations for individual particles into the array do not agree any more once we introduce `np.float64` between Python and C, so that the `particle.time` corresponds to a different set of bytes in memory between the Python and the C engines, thus causing the endless loops.

Now, I have played with a few tricks but so far I haven't found any neat way to get around this problem. These included suppressing the auto-padding and making `particle.dt` a `np.flaot64`, which should naturally align at least the basic `JITParticle`, but neither of these seemed to work. Probably the safest thing to do, which I haven't attempted yet, would be to utilise [Cython's typed memoryviews](http://docs.cython.org/src/userguide/memoryviews.html) feature, which is kind of designed for this type of thing. However, that is again a quite invasive change that requires a lot of work and time, and it will re-introduce a Cython compile step during installation on top of the git checkout.

erikvansebille:
Fixed with PR #94. Closing this PR

PR #84, opened 2016-07-19
Ipython notebook tutorial

erikvansebille:
First draft of an ipython notebook. Most works, so keen to get some feedback

A few things still open
- @Thomas-95 is going to write an additional custom module example to show how to track distance travelled by the particle (showing off use of `extra_vars`)
- The movie doesn't work. While I found an example of embedding movies directly in juypter notebooks [here](http://jakevdp.github.io/blog/2013/05/12/embedding-matplotlib-animations/), that requires significant rewriting of the plotting function. Perhaps easier to just take a screencast and put that in separately?
- Slightly worryingly, I had to resort to scipy mode in both the WestVel kernel example and the GlobCurrent one, as JIT didn't work(?)
- I've added a `Dockerfile` for possible use at [mybinder.org](mybinder.org) for live interactive running of this tutorial. However, I can't get parcels installed there because of trouble with netcdf. Will investigate further. If we don't get it to work, we might want to delete the `Dockerfile` again


PARCELStutorial.ipynb
```diff
@@ -0,0 +1,508 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "grid = Grid.from_nemo(\"examples/MovingEddies_data/moving_eddies\")"
```
mlange05: It might be good to visualise the loaded fields here, so maybe adding `grid.U.show()` would make things more obvious? We could also show `P`, but that requires adding `extra_vars={'P': 'P'}` to the load line and we would need to explain what that does.

PARCELStutorial.ipynb
```diff
@@ -0,0 +1,508 @@
...
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "pset = grid.ParticleSet(size=2,             # the number of particles\n",
+    "                        pclass=JITParticle, # the type of particles (JITParticle or Particle)\n",
+    "                        lon=[ 3.3,  3.3],   # a vector of release longitudes \n",
+    "                        lat=[46.0, 47.8])   # a vector of release latitudes"
+   ]
+  },
```
mlange05: The same here; visually following the process makes things much easier to comprehend. So maybe add `pset.show(field=grid.P)`?

mlange05:
Ok, box 12 and 13 (the `WestVel` kernel) just worked fine for me with JIT mode.

mlange05:
Ok, the GlobCurrent example really is hanging in JIT mode. I suspect the problem highlighted in PR #83 so, I would leave this in SciPy mode for now.

mlange05:
Ok, I've had another run through and this looks great as a first example.

erikvansebille:
I've made one more change: calling plotParticle as a function, rather than a script. See my more recent commit. If good, then I'm happy for this to be pulled. We can open other PRs for the remaining issues (the 'distance' custom kernel, the inline movies, the Agulhas in JIT etc)

mlange05:
Ok, this all looks good to me. The only thing I found was that I get a different set of paths for the final Aghulas result than what is in the committed notebook. I suspect that this is due to an update in the examples branch that hasn't been pulled into this PR yet? Or do we just need to update the notebook itself? In any case, if you are happy with the result it produces this is good to go.

Issue #76, opened 2016-06-03
Allow circumpolar/periodic movement for particles on global grids


erikvansebille:
On a global grid, a particle that exits on the eastern boundary should reappear on the western boundary; i.e. the grid should allow periodic motion.

Easiest to do this with the `%360` operator; but not sure if this should be implemented throughout the entire code by default, or only if the grid is predefined to be global and/or circumpolar

And while this is relatively straightforward for the zonal dimension (east-west), it is much harder in the north-south direction. They will need to be transferred across the Northpole (no ocean on the Southpole, so no need to worry there ;-)

erikvansebille:
Following up, it would be good to generalise the periodic boundary also for non-global grids. E.g. zonal channels with periodic boundary conditions

erikvansebille:
Fixed in #133

Issue #61, opened 2016-04-14
What to do if particles are on land?


erikvansebille:
We need to decide how to treat particles that are on land. @Abobie has worked hard on the `boundaries` branch where particles are 'pushed back' using ghost points, when they come very close to land. However, there can still be instances where particles end up on land. For example when the timestep is so large that particles step over multiple grid cells in one timestep. Or if random-walk diffusion is added

My thinking is that at the end of the kernel, we need a function that, for each timestep, assesses whether a particle is on land (and whether it has gone out of the domain, see #47). If that function return `True`, the user should have a few options:
1. Remove the particle (to simulate beaching)
2. Try computing the particle position again (essentially redrawing the random number in the diffusion; this should probably be defualt behaviour)
3. Move the particle to the closest ocean point (might be difficult/expensive to compute, though)

@Jacketless, you have been thinking about this too for your diffusion steps, right? Any comments? 

And @mlange05 what do you think about a general test at the end of each particle timestep to accept/reject the move in that timestep? Would that be feasible?

Jacketless:
I have been returning to this problem (and related out-of-domain issues) as it is becoming increasingly important for simulations involving diffusion or other, non-advection, forcing.

I'll attempt to summarise my understanding of our current functionality, to open discussion. Please correct me where if I am misunderstanding the code, particularly JIT mode where it's harder to debug the C-code converted kernels.
At present, the master branch:
- Interpolates sampled field values using `interpolator2d`
- If this interpolation involves reading values from outside the domain:
    - In Scipy mode this is caught, an error message printed, and a value of `None` returned for the interpolated value
    - In JIT mode, NaNs are returned, but no error or warning message is generated
- NaN values *within* the boundary demain are set to zero during initiation of individual `Field` objects

Is this an accurate summary?

This allows the user to check if sampled field values are None (for scipy) or nan (using `math.isnan` for JIT mode), and act accordingly for out-of-domain instances (e.g. calling `particle.die()` or perhaps moving the particle back to a previously stored position).

However, it seems that there is still no way of checking for presence on land masses within the model domain, aside from checking for zero field values within kernels (which in some instances may be a valid field value).
From the point of view of writing custom kernels, it would more useful to have a consistent NaN value returned, rather than a built in check-rejection test. It would allow users to, say, redraw the diffusion component of a time-step but keep the advection, for example. 

I know NaNs cause problems with Scipy, but what options do we have here?

mlange05:
[...]

PR #85, opened 2016-07-29
Out of bounds fix

Jacketless:
Not suggesting that this be merged with the master branch, but wanted to have a short-term fix branch that can be checked out to the HPC cluster so I can keep running simulations there.

However, if we want, this branch could form the baseline for our eventual solutions discussed in #47 #61 and #76


erikvansebille:
I totally understand that you need a quick fix to the out-of-bounds error, so thanks for creating this (slightly dirty) hack, @jacketless

While I'd be happy to temporarily add this code to master, we need a better solution in the long term of course. 

The other option, if we don't want to 'spoil' master with these quick fixes, is that you temporarily fork parcels and then add it in your own forked version. @mlange05, what do you think?

mlange05:
Ok, IMHO I would prefer a well-documented hack over maintaining a separate fork, since a hacky solution is better than no solution at all. I would like a clear in-place developer comment though that points to the relevant github issue, in case interested people stumble across this.

erikvansebille:
I agree with @mlange05. So if you can add some more comments, pointing to the relevant issues on the github page, @Jacketless, then I'll pull this into master

Issue #17, opened 2016-01-15
Parallel execution


mlange05:
Now that we are homing in on what Parcels will be and how one uses it we should start considering parallel execution. I propose that we should aim to add automatically MPI parallelism the key ideas outlined in the workshop. Ultimately the user should only be required to execute `mpiexec -n num_proc python my_parcels_script.py`. The key ingredients for this would be:
- Grids need to compute a parallel partitioning in parallel.
- Particle serialisation is required to send small particle groups to neighbouring partitions via MPI.
- We need to detect once a particle enters a "halo" region to trigger sending. This avoids global communications, which otherwise would impede scalability.

Ideally we would implement this for both modes equally, but depending on feasibility I would prioritise C/JIT. A pre-requisite for the particle serialisation would be more dynamic data structures for the particle sets, since we would need to add and remove particles frequently on local processes.

jianzhou722:
Hi all - any timeline for the parallel version of PARCELS?

erikvansebille:
Discussions on Parallelisation in this very old Issue is now being continued in #553, so closing here

PR #86, opened 2016-08-02
masking update for AdvectionRK4 method

nathanieltarshish:
Before interpolating to find the velocity at a given (lat, lon), we
check to see if (lat, lon) are in the grid domain. If they are in the
domain , we proceed to the velocity interpolation step. If they are
outside the domain, the particle has exited and the coordinates get
masked to np.nan. Note that np.nan masking is compatible with the show
movie feature of the execute method and the plotParticles 2d mode
script.

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -18,16 +18,30 @@ __all__ = ['Particle', 'ParticleSet', 'JITParticle',
 
 
 def AdvectionRK4(particle, grid, time, dt):
-    u1 = grid.U[time, particle.lon, particle.lat]
-    v1 = grid.V[time, particle.lon, particle.lat]
-    lon1, lat1 = (particle.lon + u1*.5*dt, particle.lat + v1*.5*dt)
-    u2, v2 = (grid.U[time + .5 * dt, lon1, lat1], grid.V[time + .5 * dt, lon1, lat1])
-    lon2, lat2 = (particle.lon + u2*.5*dt, particle.lat + v2*.5*dt)
-    u3, v3 = (grid.U[time + .5 * dt, lon2, lat2], grid.V[time + .5 * dt, lon2, lat2])
-    lon3, lat3 = (particle.lon + u3*dt, particle.lat + v3*dt)
-    u4, v4 = (grid.U[time + dt, lon3, lat3], grid.V[time + dt, lon3, lat3])
-    particle.lon += (u1 + 2*u2 + 2*u3 + u4) / 6. * dt
-    particle.lat += (v1 + 2*v2 + 2*v3 + v4) / 6. * dt
+    if not np.isnan(particle.lon):
+        u1 = grid.U[time, particle.lon, particle.lat]
+        v1 = grid.V[time, particle.lon, particle.lat]
+        lon1, lat1 = (particle.lon + u1*.5*dt, particle.lat + v1*.5*dt)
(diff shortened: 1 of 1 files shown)


nathanieltarshish:
The scipy version works, but the jit version fails. I am looking into it.

mlange05:
Ok, so unfortunately this approach looks somewhat incompatible with our current JIT engine, since we currently don't support generic numpy functionality in the kernels. The main point goes deeper though, since the plan for this type of feature is to perform the containment check as part of the underlying time loop engine itself, and then invoke a separate, possibly user-defined, "failure kernel" that specifies the recovery behaviour. This planned approach is essentially part of the planned parallelisation, since one of the "recovery mechanisms" here is to send the particle to a different partition of the grid.

That being said though, I don't mind intermediate short-term workarounds, but they do need to consider the JIT engine, I'm afraid.

nathanieltarshish:
Thanks for the explanation. I did not understand that you wanted a generic time loop engine with a distinct recovery kernel. I could only think of a few different behaviors that a user would desire for particles reaching the boundary: mask the particle, remove the particle data, or enforce circumpolar boundary conditions. Given the limited number of cases, I thought there would be a different time loop engine corresponding to each case (and by passing a flag to ParticleSet.execute, the user would specify the desired engine/behavior at the boundary). However, I understand that ideally the code would be as modular as possible to facilitate parallelization.  

As you mentioned, it would be great to have an intermediate workaround. I assume that the generic numpy functionality that was problematic was the usage of the np.nan datatype and the np.isnan function. Do you know of any JIT-compatible alternatives? Are there other JIT incompatibilities to the proposed code that I have overlooked?

erikvansebille:
We have just landed PR #85, which tries to solve the issue in the field interpolator, rather than the RK4 function. I had a chat with @mlange05 and since that solution was so much smaller and less impactful (it doesn't interfere with the 3D velocities PR), we decided to first land the field interpolator hack. 

We are aware that the 'true' solution to this out-of-bounds problem requires proper error handling and passing. This is also tightly linked in to the parallelisation (Issue #17), since another case where particles reach domain boundaries is  when they need to be moved from one tile to another, in parallel execution.

As for JIT functionality of np.nan, that is not supported yet. Will need some work to implement, unfortunately. Hopefully we can land a proper solution in a week or so!

nathanieltarshish:
Thanks for the update! I could not tell if the "proper solution" that you mention at the end of your comment refers to building the ideal solution (error handling and passing ) or a short-term solution (like building in np.nan functionality). In case it referred to a short-term solution, I just wanted to mention that on the scipy side, the code works equally well with the math package's nan implementation (math.isnan() function and float('nan') datatype ). I am new to the JIT world, but thought that the math implementation may be easier to make JIT-compatible than the numpy implementation. Either way, I appreciate your help and explanations.

nathanieltarshish:
Thanks for the update! I could not tell if the "proper solution" that you
mention at the end of your comment refers to building the ideal solution
(error handling and passing ) or a short-term solution (like building
in np.nan functionality). In case it referred to a short-term solution, I
just wanted to mention that on the scipy side, the code works equally well
with the math package's nan implementation (math.isnan() function and
float('nan') datatype ). I am new to the JIT world, but thought that the
math implementation may be easier to make JIT-compatible than the numpy
implementation. Either way, I appreciate your help and explanations.

Issue #44, opened 2016-03-08
Re-organise tests and examples


mlange05:
Now that we have a separate `examples` sub-repo we should start caring about the way we test internal functionality and how we demonstrate it by separating low-level feature and unit tests from full-scale example setups. The current test setups are all small examples and should thus be migrated to the examples repo eventually. In their place we require a whole range of additional low-level feature tests that cover the basic functionalities individually, for example:
* Grid and field initialisation from file using different formats
* Internally provided kernels, such as `AdvectionRK4/EE` should be tested against known solutions
* Various constructs of our internal JIT kernel language
* Particle trajectory and data output

As we add more and more features and cover a wider range of formats, this list should grow, while the `examples` repo will be used to maintain a useful set of setups that demonstrate complete scenarios.

erikvansebille:
So the idea now is to host all the netcdf data on a separate server (ideally oceanparcels.org, http://plasticadrift.org/parcelsdata/ for now). There will then be a script to download this data if users want it.

Following from some initial investigation by @Thomas-95, I've had a further look into pydap, and it seems that it doesn't play nice with hdf5. It gives strange error messages (also detailed in the old email thread [here](https://groups.google.com/forum/#!topic/pydap/Drswq52UveM))

However, I'm not sure we actually need pydap; might it not be overkill? Since we simply want to download the entire set of files anyway, can't we use something much simpler like `wget`, or a python version of that (see [here](https://pypi.python.org/pypi/wget))?

mlange05:
Ok, I don't think we want a script to "bulk download" all test on install. Instead I would thing we integrate `pydap` into our `Grid` objects to allow users to access and cache their own remote files via OpenDAP and put the according addresses in the `test_xxx.py` files, ie.: `grid=Grid.from_nemo("http://plasticadrift.org/parcelsdata/GlobCurrent_xxx.nc", ...)`). In this scenario we need to make sure `pydap` buffers the files, so the download is only triggered the first time the test is run.

However, if `pydap` really doesn't work, the bulk download can be used as a temporary fix. Have you tried it yourself though (the bug report you link looks old)?

erikvansebille:
Sorry for not being clear that the `pydap` indeed fails.

The code
`from pydap.client import open_url`
`dataset = open_url('http://test.opendap.org/dap/data/nc/coads_climatology.nc')`
`print 'read coads'`
[...]

PR #92, opened 2016-08-06
Examples test update clean

erikvansebille:
Clean version of the branch to reorganise tests and examples; see also issue #44. Thanks @Thomas-95 for doing most of the legwork!


scripts/pull_data.py
```diff
@@ -0,0 +1,80 @@
...
+                                temp_file.write(f.read())
+                        print("%s written to %s" % (filename, path))
+                else:
+                        print("%s already exists within %s" % (filename, path))
+
+
+if __name__ == "__main__":
+        import_MovingEddies_example()
+        import_globcurrent_example()
+        import_OFAM_example()
```
mlange05: These functions essentially do the same thing with minor tweaks right? Would it make sense to create a kind of `Example` class that holds filenames, paths, urls, etc? We could then write only one download function and it would be much easier to add new examples.

Thomas-95: I was thinking that myself when I wrote them. It would certainly make the code cleaner and shorter - I was simply prioritising having code that works.

mlange05: Sure, that is how all code development works: First we get things to work, and then we generalise to clean up the code. The trick is to not forget part two :wink:.

examples/test_case_globcurrent.py
```diff
@@ -0,0 +1,40 @@
...
+    grid = set_globcurrent_grid()
+    assert(grid.U.lon.size == 81)
+    assert(grid.U.lat.size == 41)
+    assert(grid.U.data.shape == (365, 41, 81))
+    assert(grid.V.lon.size == 81)
+    assert(grid.V.lat.size == 41)
+    assert(grid.V.data.shape == (365, 41, 81))
+
+
+@pytest.mark.xfail(run=False, reason="32/64 bit bug in time (github PR #83)")
```
mlange05: We should be able to remove the `xfail`, now that PR #94 has landed.

erikvansebille: Just did that, and also changed file names to example_* and tox.ini to setup.cfg. See 188c611

mlange05:
Ok, other than a generic downloader function in `pull_data.py` this looks great and is good to go.

Thomas-95:
Rewritten `pull_data.py` - functions replaced by single class with three instances, code generally cleaner. @erikvansebille has given the all clear.

mlange05:
Ok, this looks great now. Thanks a lot. Merging....

PR #93, opened 2016-08-08
Cleaning, testing and fixing time interpolation

mlange05:
This merge cleans up the current SciPy interpolation cascade and adds tests and bug fixes for the time interpolation in JIT:
* We now only do temporal interpoaltion by hand, and defer all spatial interpolation to a wrapper routine that invokes the SciPy wrappers, allowing us to intercept NaN values or errors.
* The SciPy interpolators now return NaN and we detect this to insert 0 as before. We can hook into this to build a generic "recovery kernel" dispatch system later.
* A new test explicitly tests temporal interpolation inside, on the edges and outside the given time interval.
* Check and force `field.time` to always be `np.float64`.
* Re-writes the JIT time loop to fix subsequent temporal sampling bugs in JIT mode. This now hardcodes a "tolerance" value of `1.e-6` for the time loop condition.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -227,50 +227,62 @@ class Field(object):
 
     @cachedmethod(operator.attrgetter('interpolator_cache'))
     def interpolator2D(self, t_idx):
-        return RegularGridInterpolator((self.lat, self.lon),
-                                       self.data[t_idx, :])
+        """Provide a cached SciPy interpolator for spatial interpolation
 
-    def interpolator1D(self, idx, time, y, x):
-        # Return linearly interpolated field value:
-        if x is None and y is None:
-            t0 = self.time[idx-1]
-            t1 = self.time[idx]
-            f0 = self.data[idx-1, :]
-            f1 = self.data[idx, :]
-            val = f0 + (f1 - f0) * ((time - t0) / (t1 - t0))
-        else:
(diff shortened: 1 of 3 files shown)


parcels/field.py
```diff
@@ -227,50 +230,60 @@ def gradient(self, timerange=None, lonrange=None, latrange=None, name=None):
...
+        """
+        return RegularGridInterpolator((self.lat, self.lon), self.data[t_idx, :],
+                                       bounds_error=False, fill_value=np.nan)
+
+    def spatial_interpolation(self, tidx, y, x):
+        """Interpolate horizontal field values using a SciPy interpolator"""
+        val = self.interpolator2D(tidx)((y, x))
+        if np.isnan(val):
+            # Temporary hotfix for out-of-bounds sampling
+            # until we have a graceful kernel recovery system.
```
erikvansebille: Do we want to explicitly refer to #47, #61 and #85, as in the old comment?

erikvansebille:
Looks good to me! Only minor thing is to add back reference to the issues on github, as we also asked @Jacketless to do. After that, I think this is good to go into master

mlange05:
Yes, sorry, I meant to put this back in. It's done now, so good to go once travis is happy.

PR #95, opened 2016-08-09
Fixing field.show bug

erikvansebille:
Fixing a bug caused by PR #93, where field.show didn’t have access to a
function that interpolates a full field. Created a new function to do
this now


mlange05:
Yes, looks good to me.

PR #97, opened 2016-08-10
Ipython updates clean

erikvansebille:
Updating the notebook:
- animations and an extra example kernel to the PARCELStutorial notebook
- automated testing of the notebook via the ipynb plugin of py.test (which annoyingly doesn't seem to allow testing of files that don't start with test_*, so created a symlink)


scripts/plotParticles.py
```diff
@@ -4,10 +4,11 @@
 import matplotlib.pyplot as plt
 from argparse import ArgumentParser
 import matplotlib.animation as animation
+from matplotlib import rc
 
 
-def particleplotting(filename, tracerfile='none', tracerlon='x', tracerlat='y',
-                     tracerfield='P', recordedvar='none', mode='2d'):
+def plotTrajectoriesFile(filename, tracerfile='none', tracerlon='x', tracerlat='y',
+                         tracerfield='P', recordedvar='none', mode='2d'):
```
mlange05: We should really replace `'none'` with `None`, here and below.

mlange05:
Except for `'none'`-> `None` this looks good to me. Good to go.

mlange05:
Looks great now, merging.

Issue #96, opened 2016-08-09
new feature: plot velocites in full vector form  


nathanieltarshish:
I wrote a new method that utilizes the `Basemap.quiver` module to produce vector plots of the velocity data. The usage of `Basemap` necessitates that `mpl_toolkits` be added to the requirements list.  The method, called `show_velocity()`, belongs to the `ParticleSet` class and relies on the ` temporal_interpolate_fullfield` method to find the velocities over the domain at any arbitrary time. It produces plots of the normalized velocity vector field colored by speed.

 `show_velocity()` has the following arguments: 
` t`: datetime or timedelta object that specifies time at which to plot velocity
`land`: boolean that is true iff land should be drawn on plot
 `latN, latS, lonW, lonE`: floats that delineate the geographic domain to be plotted. 

The user-specified domain is projected onto the velocity grid (using a new `nearest_index` function). This feature allows the user to be ignorant of the precise values of the velocity grid's lat/lon coordinates. If no domain is specified by the user, the method defaults to plotting the full domain. The method also plots the current locations of particles in the ParticleSet. Using the GlobCurrent example data set, after instantiating a grid and pset, calling `pset.show_velocity(t=datetime(2002, 1, 2), land=True)` produces the following sample output: 
![globcur_sample](https://cloud.githubusercontent.com/assets/20796154/17530408/62b955f8-5e45-11e6-86e5-9610a477fd98.png). 
By alternating between advecting the particles and saving velocity plots, one can create very useful animations: 
![animation_fast](https://cloud.githubusercontent.com/assets/20796154/17530534/f0f55e7a-5e45-11e6-9228-eef3d9de4c77.gif)

The new code is over on my fork along with a file called example_vector_plots.py that is located in the example folder. This file demonstrates the features of the new method with the MovingEddies and the GlobCurrent example datasets. If you are interested in including this feature, the master branch on my fork is ready to be merged with the main fork. I am looking forward to hearing any suggestions or feedback!

StephenGriffies:
Very nice diagnostic.  It will be quite useful! Thanks, stephen

mlange05:
This looks awesome! Can you please create a branch on the main repo for this and submit a Pull Request?

erikvansebille:
Indeed, very cool!! It would be fantastic to land this into master, yes. Really cool eye-candy and a great showcase for why python trumps Fortran :-)

A few small comments:
[...]

PR #98, opened 2016-08-10
new vector plotting features 

nathanieltarshish:
ParticleSet now has a new method `show_velocity` that plots vector fields as described in #96


parcels/particle.py
```diff
@@ -457,6 +473,94 @@ def show(self, **kwargs):
         plt.show()
         plt.pause(0.0001)
 
+    def show_velocity(self, **kwargs):
```
erikvansebille: Would it make sense to make this actually part of the original `pset.show()` function? There's quite a bit of overlap with that function, so for example handling of the time and title doesn't need to be replicated.

parcels/particle.py
```diff
@@ -457,6 +473,94 @@ def show(self, **kwargs):
...
+        t = kwargs.get('t', None)
+        if t is None:
+            t = self.particles[0].time
+        # flag to drawing land on plot
+        land = kwargs.get('land', False)
+        # plot domain latitude longitude parameters
+        latN = kwargs.get('latN', np.nan)
+        latS = kwargs.get('latS', np.nan)
+        lonE = kwargs.get('lonE', np.nan)
+        lonW = kwargs.get('lonW', np.nan)
```
erikvansebille: I can see that is some cases it would be good for a user to specify the plotting domain, but perhaps cleaner to do that with one keyword? E.g. `domain =[lonW, lonE, latS, latN]`? 

Also, this is again something that would be nice to have in `pset.show` too. So for example you could make quiver plot when a user calls `pset.show(field=vectors)` or something like that.

parcels/particle.py
```diff
@@ -457,6 +473,94 @@ def show(self, **kwargs):
...
+            V = V[latS:latN, lonW:lonE]
+
+        # configuring plot
+        lat_median = np.median(lat)
+        lon_median = np.median(lon)
+        plt.figure()
+        m = Basemap(projection='merc', lat_0=lat_median, lon_0=lon_median,
+                    resolution='h', area_thresh=100,
+                    llcrnrlon=lon[0], llcrnrlat=lat[0],
+                    urcrnrlon=lon[-1], urcrnrlat=lat[-1])
```
erikvansebille: Have you checked that this doesn't crash when `Basemap` is not available? I.e. what happens when there is an exception up in line 16 of this file?

parcels/particle.py
```diff
@@ -457,6 +473,94 @@ def show(self, **kwargs):
...
+        if(self.grid.U.time_origin == 0):
+            plt.title(delta(seconds=t))
+        else:
+            plt.title(netCDF4.num2date(t, 'seconds since '+str(self.grid.U.time_origin)))
+
+        if savefile is None:
+            plt.show()
+        else:
+            plt.savefig(savefile)
+            plt.close()
```
erikvansebille: Again, saving is a nice feature that would be useful more generally than just in `pset.show_velocity()` :-)

examples/example_vector_plots.py
```diff
@@ -0,0 +1,25 @@
...
+
+pset = grid.ParticleSet(size=10, pclass=Particle, start=(31, -31), finish=(34, -31))
+pset.show_velocity(t=datetime(2002, 1, 2), land=True, vmax=2)
+pset.execute(AdvectionRK4, starttime=datetime(2002, 1, 2), runtime=timedelta(days=1),
+             dt=timedelta(minutes=5), interval=timedelta(hours=6))
+pset.show_velocity(land=True, latN=-30, latS=-36, lonE=33, lonW=18, vmax=2)
+
+grid = Grid.from_nemo("examples/MovingEddies_data/moving_eddies*")
+pset = grid.ParticleSet(size=2, pclass=Particle, lon=[3.3,  3.3], lat=[46.0, 47.8])
+pset.show_velocity(lonW=2.8, lonE=3.5, latS=47.5, latN=48.3, vmax=3)
```
erikvansebille: This fails the travis automatic testing because travis doesn't like explicit calls to matplotlib.

In this case, I suggest turning it into a jupyter notebook. This allows you to better explain all the different options, and embed the figures directly. perhaps rename to `tutorial_vectorplots.ipynb`?

erikvansebille:
Implemented in other PR

PR #99, opened 2016-08-11
Particle housekeeping

mlange05:
A few housekeeping changes that distribute the various things in `particle.py` to separate submodules:
* Separate `particleset.py` submodule
* Separate `particlefile.py` submodule
* New `kernels` submodule with `kernels.advection` to hold advection kernels
* This leaves `particle.py` to only define the Particle types and the relevant utility methods for JIT (`ParticleType`)

### DIFF
--- a/parcels/kernels/__init__.py
+++ b/parcels/kernels/__init__.py
@@ -1,0 +1,1 @@
+from .advection import *  # noqa
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -1,0 +1,98 @@
+import numpy as np
+import netCDF4
+
+
+__all__ = ['ParticleFile']
+
+
+class ParticleFile(object):
+
+    def __init__(self, name, particleset, initial_dump=True):
+        """Initialise netCDF4.Dataset for trajectory output.
+
+        The output follows the format outlined in the Discrete
+        Sampling Geometries section of the CF-conventions:
+        http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#discrete-sampling-geometries
+
+        The current implementation is based on the NCEI template:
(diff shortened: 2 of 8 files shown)


parcels/particlefile.py
```diff
@@ -0,0 +1,98 @@
+import numpy as np
```
erikvansebille: Would it make sense to rename this file to fileIO.py or something, and also add the reading and writing of netcdf files from field.py here?

parcels/kernels/__init__.py
```diff
@@ -0,0 +1 @@
+from .advection import *  # noqa
```
erikvansebille: Since we want users to add kernels to this  `kernels` directory, would it make more sense to have it directly in the root of the repo? I.e. not a subdirectory of `parcels`?

erikvansebille:
PR looks good. Just had two questions about reordering some more: 
1) Would it make sense to bundle all fileIO into one file? Also so that plotParticles.py can access it?
2) Would it be more intuitive for users and easier to add new kernels when we put the new `kernels` directory straight at the root of the repo, rather than within the `parcels` directory?

parcels/particlefile.py
mlange05: Yes and no. Conceptually yes, but since the internal format for grid and particle data is completely different, I'd say no for now. At the moment, all they have in common is the `import netcDF` statement, until we use a slightly higher-level abstraction that encapsulates various netCDF conventions , for example.

parcels/kernels/__init__.py
mlange05: No, since `advection.py` needs to on the PYHTONPATH, so if you move it outside the `parcels` directory/module it becomes a separate python package on it's own. Creating a separate sub-module is actually the whole point of separating kernels from the core engine, but the y should still be part of `parcels`.

PR #102, opened 2016-08-12
Created file for 2D diffusion.

Thomas-95:
Test function written in `/examples`for 2D diffusion kernel on flat grid with lenient tolerances. Limited to `ScipyParticle` mode for the time being.


examples/example_brownian.py
```diff
@@ -0,0 +1,62 @@
...
+import random
+
+
+def two_dim_brownian_flat(particle, grid, time, dt):
+
+    # random.seed() - should a seed be included for reproducibility/testing purposes?
+    # Use equation for particle diffusion.
+    particle.lat += random.normalvariate(0, 1)*math.sqrt(2*dt*grid.K_lat)
+    particle.lon += random.normalvariate(0, 1)*math.sqrt(2*dt*grid.K_lon)
+
```
erikvansebille: If we add a normal distribution to rng.py, this should also work in JIT?

examples/example_brownian.py
```diff
@@ -0,0 +1,62 @@
...
+
+def brownian_grid(xdim=200, ydim=200):     # Define a flat grid of zeros, for simplicity.
+    depth = np.zeros(1, dtype=np.float32)
+    time = np.linspace(0., 100000. * 86400., 2, dtype=np.float64)
+
```
erikvansebille: Both `time` and `depth` are superfluous in this grid. I would simply remove them, and have the function `return Grid.from_data(U, lon, lat, V, lon, lat, mesh='flat')` only, to keep the code as concise as possible

examples/example_brownian.py
```diff
@@ -0,0 +1,62 @@
...
+    V = np.zeros((lon.size, lat.size, time.size), dtype=np.float32)
+
+    return Grid.from_data(U, lon, lat, V, lon, lat, depth, time, mesh='flat')
+
+
+def test_brownian_example(npart=3000):
+
+    grid = brownian_grid()
+    grid.K_lat = 100                  # Set diffusion constants.
+    grid.K_lon = 100
```
erikvansebille: @mlange05 what do you think of this idea to add extra variables to the `grid` via `grid.K = 100`? In this case, we need to pass a diffusion coefficient for use in the `two_dim_brownian_flat` kernel, and since arguments for kernels are fixed, this is the way around that @Thomas-95 and I came up with. Good? Or do you have a better idea?

examples/example_brownian.py
erikvansebille: Note that I have now created a new branch (`brownian_rng_attempt`) that has an attempt to implement a Normal distribution using the Box-Muller transform (https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform). Currently has an assertion error, but will look into that

examples/example_brownian.py
mlange05: I think this is generally a good idea, but incompatible with JIT in it's current form. But how about something like `grid.K_lat = Constant(100., dtype=np.float32)`? This could be translated to JIT, since the code-generator can query the grid and insert values directly.

Jacketless:
I thought I would chime in here too, as I have similar brownian motion/diffusion/random walk kernels that I use for the tuna work. Apologies, I am on a few days annual leave at the moment so have been a bit slow on a lot of these PRs. What suite of these kinds of kernels do we think should be included in PARCELS "out-of-the-box"? My feeling is maybe a diffusion kernel, a brownian motion kernel, and a true random-walk kernel would offer a nice choice of stochastic kernels for users initially.

In addition, may I suggest I slight modification to the kernel you have written here @Thomas-95 ? When spatial diffusivity is non-uniform, these lagrangian brownian motion algorithms result in erroneous accumulation of particles in areas of low K, which is not the case for true Eulerian diffusion. You need to add a deterministic "boosting" term which pushes particles out of these regions of low diffusivity. I have some test scripts that show this in the [lagrangian-diffusion branch](https://github.com/OceanPARCELS/parcels/tree/lagrangian-diffusion), which I will try and clean up next week and we can combine any useful code into this PR.
[Ross & Sharples 2004](http://mobile.tube.aslo.net/lomethods/free/2004/0289.pdf) give an example, which is what I have built my kernels from.

Assuming a single, spatially varying K field, and that you have already calculated the gradient of the K field using `Field.gradient()` to give dK_dx and dK_dy, the kernel looks something like this (and uses @mlange05 rng function for use in JIT mode. Using a random.uniform stochastic term prevents particles from having potentially infinite diffusion, if that is what you require. Perhaps we should also offer a normal distribution term as well?):

`r = 1/3.
  Rx = rng.uniform(-1., 1.)
  Ry = rng.uniform(-1., 1.)
  dK = [grid.dK_dx[time, particle.lon, particle.lat], grid.dK_dy[time, particle.lon, particle.lat]]
  half_dx = 0.5 * dK[0] * dt
  half_dy = 0.5 * dK[1] * dt
  K = [grid.K[time, particle.lon + half_dx, particle.lat + half_dy]] * 2
  Rx_component = Rx * np.sqrt(2 * K[0] * dt / r)
  Ry_component = Ry * np.sqrt(2 * K[1] * dt / r)
  CorrectionX = dK[0] * dt
  CorrectionY = dK[1] * dt
  particle.lon += Rx_compontent + CorrectionX
  particle.lat += Rx_component + CorrectionY`

erikvansebille:
Thanks @Jacketless, for these comments. Totally agree that we will want a whole suit of diffusion kernels 'out-of-the-box'. In fact, now that we have a `parcels/kernels` directory, we can make a new file `diffusion.py` which features all of these. It would be great to have some of the kernels you mentioned in there, too

This PR is only about Brownian motion. While as you say it has limited use in biology, it is widely used in physical oceanography, so I still think it's useful to have in the standard set.

Finally, I did create a `random.normal` function, see the `brownian_rng_attempt` branch. @Thomas-95 will pull some of that into this PR.

Jacketless:
I'm looking to consolidate some of the incomplete PRs I opened a while back, and would like to put some of the diffusion tests and kernels I wrote into the new PARCELS build. @Thomas-95 should I commit them to this branch, or would rather I opened a fresh branch? Don't mind either way 🙂

erikvansebille:
Hi @jacketless, probably best to start a separate branch for now. We can always merge later.

parcels/codegenerator.py
```diff
@@ -38,10 +38,11 @@ def __getattr__(self, attr):
 class RandomNode(IntrinsicNode):
     symbol_map = {'random': 'parcels_random',
                   'uniform': 'parcels_uniform',
-                  'randint': 'parcels_randint'}
+                  'randint': 'parcels_randint',
+                  'normal': 'parcels_normal'}
 
     def __getattr__(self, attr):
-        if hasattr(random, attr):
+        if hasattr(random, attr) or attr is 'normal':
```
erikvansebille: Without this statement `or attr is 'normal'`, JIT mode throws an error
`AttributeError: Unknown random function encountered: normal`

Not sure how to fix this...

examples/example_brownian.py
```diff
@@ -0,0 +1,65 @@
...
+import math
+from parcels import random
+import pytest
+
+ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
+
+
+def two_dim_brownian_flat(particle, grid, time, dt):
+
+    # random.seed() - should a seed be included for reproducibility/testing purposes?
```
mlange05: Yes it should. How about adding it as a `pytest` parameter or even fixture?

include/parcels.h
```diff
@@ -142,3 +143,20 @@ static inline int parcels_randint(int low, int high)
 {
   return (rand() % (high-low)) + low;
 }
+
+static inline float parcels_normalvariate(float loc, float scale)
```
mlange05: Some type of docstring would be nice here, since the details of this are not immediately obvious.

erikvansebille: Done; see below

PR #103, opened 2016-08-12
vector plotting added to pset.show() and accompanying tutorial 

nathanieltarshish:
updated from #98 to incorporate the changes discussed with @erikvansebille. Also, includes a new ipython tutorial that describes the different features of pset.show()


erikvansebille:
This is great @nathanieltarshish! Really nice!

I'll comment inline on some of the changes in `parcels.py`, but here are some minor comments on the tutorial
- Could you rename the tutorial to e.g. `tutorial_plotting.ipynb` to make it clearer to users what this is about? Ideally, we'll get a whole set of `tutorial_*` notebooks :-)

- Could you create a symbolic link to the tutorial `test_tutorial_plotting.ipynb`. In that way, travis will be able to test it too? (There seems to be no way to override default test-naming-conventions in the ipynb plugin for `py.test`, so files need to start with `test_*`. Hence also the `test_PARCELStutorial.ipynb` symlink)

- The title in cell [4] of the tutorial states 1 Jan 1900. Is this a bug?

- As you also acknowledge, in cell [10] the `t` argument only changes the field, but not the location of the particles itself. This may be confusing behaviour to a user, when they think that by changing `t` they can see the evolution of _both_ the field and the particles (i.e. what the `particlePlotting` function does). What is the use-case for showing particles on a time that is not the internal time of the particles? Do we need this option?

- Would it be worth to add the 'trick' on how to create a movie from saved files in the tutorial, perhaps only as text?

- A very, very minor point: currently we can't test for PEP8 compatibility in notebooks, but I noticed that sometimes you don't put spaces within arguments (e.g. `savefile='particles'`) and sometimes you do (e.g. `vector = True`). Best to stay consistent with PEP8, or at least the PARCELStutorial

- Similarly, in cell [1], there are a lot of imported modules that are not needed. I think the only ones needed are 
```
%matplotlib inline
from parcels import *
from datetime import timedelta, datetime
```

Again, really nice work! Don't be put off by this long(ish) list of suggestions ;-)

parcels/particleset.py
```diff
@@ -238,38 +254,113 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
 
     def show(self, **kwargs):
-        if plt is None:
-            raise RuntimeError("Visualisation not possible: matplotlib not found!")
-
-        field = kwargs.get('field', True)
-        lon = [p.lon for p in self]
-        lat = [p.lat for p in self]
+        savefile = kwargs.get('savefile', None)
+        vector = kwargs.get('vector', False)
```
erikvansebille: My original idea was for vectorplotting to be initiated by calling `field='vector'`. In this implementation, `vector` and `field` are separate arguments. 
Not sure what's best. The advantage of the `field='vector'` notation is that there can never be confusion about what to plot if both `field` _and_ `vector` are provided as arguments (currently, only only vectors are plotted in that case).

parcels/particleset.py
```diff
@@ -238,38 +254,113 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
+        t = kwargs.get('t', time[0])
+        if isinstance(t, datetime):
+            t = (t - self.grid.U.time_origin).total_seconds()
+        if isinstance(t, delta):
+            t = t.total_seconds()
+        if domain is not None:
+            latN = nearest_index(self.grid.U.lat, domain[0])
+            latS = nearest_index(self.grid.U.lat, domain[1])
+            lonE = nearest_index(self.grid.U.lon, domain[2])
+            lonW = nearest_index(self.grid.U.lon, domain[3])
```
erikvansebille: Does this ordering of `domain` make most sense? I know that I suggested providing domain as a vector, but now I think perhaps better as separate arguments for `latS`, `lonE` etc. Especially if these default to the grid edges..

parcels/particleset.py
```diff
@@ -48,6 +53,17 @@ def add_jitter(pos, width, min, max):
...
 
+def nearest_index(array, value):
+    """returns index of the nearest value in array using O(log n) bisection method"""
+    y = bisect.bisect(array, value)
+    if y == len(array):
+        return y-1
+    elif(abs(array[y-1] - value) < abs(array[y] - value)):
+        return y-1
+    else:
+        return y
```
erikvansebille: A more fundamental issue: I have seen netcdf output of (global) ocean models where longitude is not monotic. For example, longitude would be [180:360, 0:179] or something.

However, I suspect that other parts of `parcels` would also break on these grids. SO perhaps we should write a wrapper function that checks on read-in of longitude whether it is monotonic, and otherwise adjusts it so that it is (by adding 360 degrees).
So no need to do anything here, but just wanted to flag it now that I'm thinking about it

parcels/particleset.py
```diff
@@ -238,38 +254,113 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
+            idx = self.grid.U.time_index(t)
+            U = np.array(self.grid.U.temporal_interpolate_fullfield(idx, t))
+            V = np.array(self.grid.V.temporal_interpolate_fullfield(idx, t))
+            lon = self.grid.U.lon
+            lat = self.grid.U.lat
+            if domain is not None:
+                lon = lon[lonW:lonE]
+                lat = lat[latS:latN]
+                U = U[latS:latN, lonW:lonE]
+                V = V[latS:latN, lonW:lonE]
```
erikvansebille: Given that we set `latN, latS, lonE, lonW = (-1, 0, 0, -1)` on line 274 above, can't this always be executed (i.e. is there need for the `if domain is not None:` statement here)?

parcels/particleset.py
```diff
@@ -238,38 +254,113 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
+                plt.title(num2date(t, 'seconds since '+str(self.grid.U.time_origin)))
+
+        if savefile is None:
+            plt.show()
+            plt.pause(0.0001)
         else:
-            timestr = ' on ' + str(time_origin + delta(seconds=t))
-        plt.xlabel('Longitude')
-        plt.ylabel('Latitude')
-        plt.title('Particles' + namestr + timestr)
```
erikvansebille: Why not use this old way of creating the title?

parcels/particleset.py
```diff
@@ -238,38 +254,113 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
-        plt.show()
-        plt.pause(0.0001)
+            plt.savefig(savefile)
+            print 'Plot saved to '+savefile+'.png'
```
erikvansebille: This `print 'Plot saved to '+savefile+'.png'` will raise a python3 error because printing is different there. And we aim to be python3 compatible. Hence the red cross in travis..

nathanieltarshish:
Hi @erikvansebille , 

Thanks for the comments. Aside from the points I address below, I incorporated all of the changes you suggested and the branch now passes the Travis tests and is ready to be merged. In reply to your comments, 

>    the title in cell [4] of the tutorial states 1 Jan 1900. Is this a bug?

 No, I don't think so. Prior to cell [4],  no start time has been specified for either the particles themselves (they have not yet been advected) or the `show()` function. Therefore, the plot shows the data at the beginning of the calendar which is 1 Jan 1900. 

> As you also acknowledge, in cell [10] the t argument only changes the field, but not the location of the particles itself. This may be confusing behaviour to a user, when they think that by changing t they can see the evolution of both the field and the particles (i.e. what the particlePlotting function does). What is the use-case for showing particles on a time that is not the internal time of the particles? Do we need this option?

I agree that it is somewhat confusing. Seems to me that there are uses for plotting the velocity vector field at arbitrary times (without plotting the particles themselves). Therefore, I have added a new boolean argument `particles`  which makes showing the particles optional. 

> Would it be worth it to add the 'trick' on how to create a movie from saved files in the tutorial, perhaps only as text?

I'm leaning towards no. The 'trick' is just calling an external program that turns a folder of images into a gif. There are countless ways and programs that do this. I think we should assume a user knows/can easily figure out how to best accomplish this given their OS.

> Does this ordering of domain make most sense? I know that I suggested providing domain as a vector, but now I think perhaps better as separate arguments for latS, lonE etc. Especially if these default to the grid edges..

I am happy to change the ordering, but I think it might make sense to stick with domain as a vector. I also don't understand your point about why defaulting to grid edges supports the idea of switching back to   the `latN`, ` latS`, etc variables.  The `latN`, ` latS` variables are the indices closest to the domain specified latitudes in grid.U.lat not the actual latitudes themselves. Therefore, we would have to create another set of four variables for the user to specify true latitudes and longitudes and then convert that info to the `latN`, ` latS` variables  . This seems a little clunkier that just using a domain vector. Also, after the user becomes familiar with the domain vector format, it is obviously faster and fewer lines of code.

erikvansebille:
Sorry for the long delay on getting back on this PR; we're almost there!

Just two small things

>> the title in cell [4] of the tutorial states 1 Jan 1900. Is this a bug?

> No, I don't think so. Prior to cell [4], no start time has been specified for either the particles themselves (they have not yet been advected) or the show() function. Therefore, the plot shows the data at the beginning of the calendar which is 1 Jan 1900.

But the `pset` lives on a `grid` which has a calendar already, so the particles should be shown on the first day of that grid. It is true that `starttime` has not been set yet, but the default is the first day on the `grid`; this is also the behaviour of cell [6] in the PARCELStutorial notebook

All good for the rest; let me know what you think of this and then ready to pull in!

erikvansebille:
Hi @nathanieltarshish, I had a look today at the issue of default time set to 0 when `pset` is created but not executed yet. In fact, it doesn't really make sense to have the default time set to 0. Rather, it should be the first time on the grid. Updated that with the last commit. Can you update the notebook? Then we should be good to pull in

nathanieltarshish:
Hi @erikvansebille, I think it is now ready to pull in.

Issue #88, opened 2016-08-04
KernelOp doen't work in scipy with RK45


erikvansebille:
While working on Issue #80, I found a strange bug where Kernel additions (`AdvectionRK45 + UpdateP`) don't work in scipy mode when using the AdvectionRK45 kernel.

The AdvectionRK4 and AdvectionEE kernels work, and all works in JIT; out of the 6 possible permutations of modes and advection kernels, it's only RK45/scipy that doesn't work

I've created a branch fix_scipyRK45KernelOps (commit ae88f19) to highlight this error. @mlange05, could you have a look at it when you have time?

erikvansebille:
In fact, looking a little deeper in it, AdvectionRK45 also doesn't seem to work in jit mode. While running with jit doesn't throw an error, the trajectories are actually wrong.

See for example the following terminal output
```
$ python tests/test_peninsula.py jit -m RK4 -v -p 1
Initial particle positions:
P(0.0270, 0.0315)[p=0.00000, p_start=-3.096488]
Peninsula: Advecting 1 particles for 1 day, 0:00:00
Compiled MyParticleAdvectionRK4UpdateP ==> /var/folders/r2/8593q8z93kd7t4j9kbb_f7p00000gn/T/parcels-501/efc436f109ce8e1080a7178258ed2154.so
Final particle positions:
P(0.6971, 0.0416)[p=-3.09723, p_start=-3.096488]

$ python tests/test_peninsula.py jit -m RK45 -v -p 1
Initial particle positions:
P(0.0270, 0.0315)[p=0.00000, p_start=-3.096488]
Peninsula: Advecting 1 particles for 1 day, 0:00:00
Compiled MyParticleAdvectionRK45UpdateP ==> /var/folders/r2/8593q8z93kd7t4j9kbb_f7p00000gn/T/parcels-501/cc976f8cdf8b094181058feb95443e1d.so
Final particle positions:
P(4.7692, 0.3769)[p=0.00000, p_start=-3.096488]
```
The first call (`-m RK4`) yields correct Final particle positions (with longitude around 0.7)
The second call (`-m RK45`), however, leads to a particle actually exiting the domain (with longitude more than 4.7!)
The particle goes _much_ too fast in the RK45 case...

When reducing `interval` to `delta(minutes=5)`, so that it is equal to `dt`, there is no change to the `-m RK4` run, but the longitude in the `-m RK45` run gets even larger, clearly showing that the AdvectionRK45 implementation is broken

[...]

Issue #80, opened 2016-06-14
make the Stommel test case work


erikvansebille:
The Stommel test case (`test_stommel.py`) is currently `xfail` and should be fixed.

erikvansebille:
Fixed in PR #105

PR #105, opened 2016-08-15
Updating stommel test case

erikvansebille:
A rewrite of the Stommel test case, now with working RK4. See also Issue #80

A few points
- EE doesn’t work because assert tolerance is too low (EE is just not good enough in this case), so is not tested
- RK45 doesn’t work at all (see Issue #88). This is a much bigger issue, though


mlange05:
I think this looks great and is good to go. Many thanks for the bug fixes @erikvansebille

PR #107, opened 2016-08-16
Delay particle start example kernel & script

erikvansebille:
Created an example script to show how to implement delayed start of particles (i.e. have particles only be advected some time after the start of the `.execute`). This follows discussions with @Jacketless 

Delayed start is now implemented by setting `particle.active` to a value other than `0` (particle will be deleted) or `1` (particle will be advected)

- Changed the three advection kernels to only work on particles where `active == 1`
- Fixed a bug in `ParticleSet` where deletion of particles with `active == 0` was only done at the end of `.execute`, rather than every `timeleap`
- Added the Peninsula data to oceanparcels.org/examples-data/ and `pull_data`, for a cleaner `example_delay_start` script

### DIFF
--- a/examples/example_delay_start.py
+++ b/examples/example_delay_start.py
@@ -1,0 +1,41 @@
+from parcels import Grid, JITParticle, ScipyParticle, Variable
+from parcels import AdvectionRK4
+import numpy as np
+from datetime import timedelta as delta
+
+
+ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
+
+
+def DelayStart(particle, grid, time, dt):
+    """An example kernel showing how particle starts can be delayed, using a starttime variable
+    """
+    if time > particle.starttime:
+        particle.active = 1
+
+
+def pensinsula_example(mode, npart=10):
+
+    grid = Grid.from_nemo('examples/Peninsula_data/peninsula', extra_vars={'P': 'P'})
+
+    class DelayStartParticle(ptype[mode]):
+        starttime = Variable('starttime', dtype=np.float32, default=0.)
+
+    # Initialise particles
+    x = 3. * (1. / 1.852 / 60)  # 3 km offset from boundary
+    y = (grid.U.lat[0] + x, grid.U.lat[-1] - x)  # latitude range, including offsets
+
+    pset = grid.ParticleSet(npart, pclass=DelayStartParticle, start=(x, y[0]), finish=(x, y[1]))
+    p = 0
+    for particle in pset:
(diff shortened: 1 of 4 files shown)


examples/example_delay_start.py
```diff
@@ -0,0 +1,41 @@
...
+    x = 3. * (1. / 1.852 / 60)  # 3 km offset from boundary
+    y = (grid.U.lat[0] + x, grid.U.lat[-1] - x)  # latitude range, including offsets
+
+    pset = grid.ParticleSet(npart, pclass=DelayStartParticle, start=(x, y[0]), finish=(x, y[1]))
+    p = 0
+    for particle in pset:
+        particle.active = -1  # Using -1 value to signal that particle has not yet started
+        particle.starttime = 3600 * p  # set starttime for each particle
+        p += 1
+
```
Jacketless: The `start` and `finish` arguments in the declaration of `pset` place particles equidistantly between these two points. I think it would be useful to allow users to pass a list of x and y positions, which must be the same length as `npart`, which explicitly defines the start position of each particle.

Of course, this can always be done manually in the `for particle in pset:` loop before `pset.execute` is called :)

Jacketless:
I think this looks great Erik, and is pretty much exactly what I had in mind (except for the comment on particle start positions).

I would merge, but I don't seem to have permission to merge PRs any more... Have I offended the team in some way? 😢

examples/example_delay_start.py
erikvansebille: It has been possible since the beginning to define a set of particles using the `lon =[]` and `lat = []` options. Start and finish are just higher-level keywords. So this has always been possible. The only small inconvenience is that `npart` always needs to be provided too, even if in theory that could be taken from the length of Lon and lat..

erikvansebille:
Closing this pull request after discussions with @mlange05. The 'proper' way to set a delayed start time is by adding particles to the `pset` during runtime using the `pset.add(particle)` syntax. This can easily be achieved with an extra loop around `.execute`

One advantage of this method is that it is much less demanding on memory (only particles that are run at current time are stored in memory; particles that are 'dormant' don't take up memory as in the implementation in this branch)

So I opened another branch `example_delay_start_pset_add` that holds an example of how this `pset add` implementation would work. 

However, at the moment there is a strange JIT bug, where the particles are not actually advected. Scipy works, though. Since @Jacketless still requires delayed start to work in JIT, we will keep this branch open for the moment. Nevertheless, I consider the `particle.active = -1` a hack and do not support this being merged into `master`.

PR #108, opened 2016-08-16
Created test case for simple rotating fluid.

Thomas-95:
Test case 1 (of 8) in the review paper - "Radial rotation with known period". 

Two partcles in grid with rotating fluid of known period. One remains static, the other moves.


erikvansebille:
Looks great to me! Thanks @Thomas-95 for coding up this first testcase from the review paper

PR #111, opened 2016-08-24
Simple particle density

Jacketless:
Simple particleset density calculation 

From a given field attributes, the current density of particles in a particleset object is calculated. 

Density of number of particles per unit area is calculated as default, or a user-defined value held by each particle can be used.

Short test script included.


Jacketless:
Please take a look at the code for this density calculation function, but hopefully this is ready to merge now I have fixed some flake8 fails. Would be good to get this included as some other behaviour kernel branches I am working on require density calculation as test criteria.

parcels/particleset.py
```diff
@@ -382,6 +382,69 @@ def show(self, **kwargs):
             print('Plot saved to '+savefile+'.png')
             plt.close()
 
+    # def density(self, field, particle_val=None):
```
erikvansebille: Is there a reason for leaving all of this commented out in the code? Seems an old version of the newer density() function below

parcels/particleset.py
```diff
@@ -382,6 +382,69 @@ def show(self, **kwargs):
...
+        if area_scale:
+            area = np.zeros(np.shape(field.data[0, :, :]), dtype=np.float32)
+            dx = (field.lon[1] - field.lon[0]) * 1852 * 60 * np.cos(field.lat*np.pi/180)
+            dy = (field.lat[1] - field.lat[0]) * 1852 * 60
+            for y in range(len(field.lat)):
+                area[y, :] = dy * dx[y]
+            # Scale by cell area
+            Density /= np.transpose(area)
+
+        # print(Density[np.where(Density > 0)])
```
erikvansebille: debug statement?

tests/test_density.py
```diff
@@ -0,0 +1,170 @@
...
+    Num = np.zeros((lon.size, lat.size, time.size), dtype=np.float32)
+
+    for i, x in enumerate(lon):
+        for j, y in enumerate(lat):
+            Num[i, j, :] = 1
+
+    return Field('Start', Num, lon, lat, depth, time, transpose=True)
+
+
+if __name__ == "__main__":
```
erikvansebille: The test_denisty.py function doesn't seem to be py-tested. Have a look at the other functions in this test-directory, and make sure it gets tested on both 'jit' and 'scipy' mode

tests/test_density.py
```diff
@@ -0,0 +1,170 @@
...
+    # Simple evenly distribution starting density field
+    depth = grid.U.depth
+    time = grid.U.time
+    lon = grid.U.lon
+    lat = grid.U.lat
+    Num = np.zeros((lon.size, lat.size, time.size), dtype=np.float32)
+
+    for i, x in enumerate(lon):
+        for j, y in enumerate(lat):
+            Num[i, j, :] = 1
```
erikvansebille: Why not initiate `Num` with ones directly, using `np.ones`? Or am I missing something?

tests/test_density.py
```diff
@@ -0,0 +1,170 @@
+from argparse import ArgumentParser
+import numpy as np
+import math
+from parcels import Field, Grid, JITParticle, ScipyParticle, Variable
+
+
+def GradientClimber(particle, grid, time, dt):
+    f_lat = dt / 1000. / 1.852 / 60.
+    f_lon = f_lat / math.cos(particle.lat*math.pi/180)
```
erikvansebille: Can't we use the `mesh=flat` option here, to have field.py do these conversions automatically? In that way, users can use the same GradientClimber on both spherical and flat meshes

tests/test_density.py
```diff
@@ -0,0 +1,170 @@
...
+    lon1, lat1 = (particle.lon + u1*.5*f_lon, particle.lat + v1*.5*f_lat)
+    u2, v2 = (grid.dK_dx[time + .5 * dt, lon1, lat1], grid.dK_dy[time + .5 * dt, lon1, lat1])
+    lon2, lat2 = (particle.lon + u2*.5*f_lon, particle.lat + v2*.5*f_lat)
+    u3, v3 = (grid.dK_dx[time + .5 * dt, lon2, lat2], grid.dK_dy[time + .5 * dt, lon2, lat2])
+    lon3, lat3 = (particle.lon + u3*f_lon, particle.lat + v3*f_lat)
+    u4, v4 = (grid.dK_dx[time + dt, lon3, lat3], grid.dK_dy[time + dt, lon3, lat3])
+    Vx = (u1 + 2*u2 + 2*u3 + u4) / 6.
+    Vy = (v1 + 2*v2 + 2*v3 + v4) / 6.
+    particle.lon += Vx * f_lon * 50000
+    particle.lat += Vy * f_lat * 50000
```
erikvansebille: What's this 50000? Is that a variable?

parcels/particleset.py
```diff
@@ -382,6 +382,69 @@ def show(self, **kwargs):
...
+        half_lat = (field.lat[1] - field.lat[0])/2
+        lons = []
+        lats = []
+        for p in self.particles:
+            lons.append(p.lon)
+            lats.append(p.lat)
+        # Kick out particles that are not within the limits of our density field
+        dparticles = (lons > (np.min(field.lon)-half_lon)) * (lons < (np.max(field.lon)+half_lon)) * \
+                     (lats > (np.min(field.lat)-half_lat)) * (lats < (np.max(field.lat)+half_lat))
+        dparticles = np.where(dparticles)[0]
```
erikvansebille: Can we use the inbuilt `pset.delete()` functionality here?

parcels/particleset.py
```diff
@@ -382,6 +382,69 @@ def show(self, **kwargs):
...
+
+    def density(self, field, particle_val=None, relative=False, area_scale=True):
+        Density = np.zeros((field.lon.size, field.lat.size), dtype=np.float32)
+        half_lon = (field.lon[1] - field.lon[0])/2
+        half_lat = (field.lat[1] - field.lat[0])/2
+        lons = []
+        lats = []
+        for p in self.particles:
+            lons.append(p.lon)
+            lats.append(p.lat)
```
erikvansebille: These lines to build `lons` and `lats` seem fairly slow, by appending. Isn't there a faster way to do this?

parcels/particleset.py
Jacketless: Ah, apologies- given it's been a while I made and PARCELS commits, I've gotten a bit sloppy. Spent too much time throwing together IKAMOANA code in a rush whilst in Nouméa! You're right, there are old versions and debug statements which I will clean up!

parcels/particleset.py
Jacketless: The purpose is to pull all particle lons and lats out so we can subsequently make a subset of those which are outside the user-defined density region/resolution specified by `field`.
Using an enumerator would probably more efficient, I'll look into it

parcels/particleset.py
Jacketless: `ParticleSet.density()` is supposed to be a generic function that could eventually be called from within the inner loop. These statements just find a subset of the particles that are present within the boundaries of the user defined `field` passed to `density()`. 
We don't wish to delete these particles, because at a later time we may want to calculate particle density again or further advect the particleset etc.

erikvansebille: OK, I see your point. But wouldn't the typical usercase be to have particle density calculated over the entire domain? Now, for the vast majority of cases, we're looping over all particles. Would it make sense to do this subsetting only if the domain is smaller than the original model domain?

tests/test_density.py
Jacketless: I can only see a way of declaring `mesh = flat` during construction of a new grid object from netcdf or data. Here, we wish to say that the abstract `K` field, which is used by `GradientClimber` and added to the grid after construction, is on a mesh. I can't see a way to do this?

erikvansebille:
Code is much better now @Jacketless! However, I'm not sure the test_density really is a unit-test of the density calculation itself. Instead, it looks much more like a test of the GradientClimber. This is fine in principle, of course; but it might be better suited in the 'examples' directory?
A unit test of the density calculation would simply only test the `particleset.density` application, see the commit 5b32df2 I've just pushed. Does this make sense?

Jacketless:
Yes @erikvansebille, you're right- I suppose there does not need to be any actual advection or movement of particles to test the density function. I used this because I am building a non-uniform K diffusion test from this branch. Your test of simply distributing particles to known locations is much simpler and more elegant. 
Looks good- shall we kick out the `test_density.py` from this PR?

tests/test_density.py
mlange05: OK, so the `mesh=flat` argument to the grid is just a short-hand notation for setting the appropriate `UnitConverter` classes on the `U` and `V` field. For an individual field you want to set the converter directly by passing `unit=GeographicPolar()`, which will make the conversion implicitly happen every time you sample the field `K` either from Python or within the kernels.

parcels/particleset.py
```diff
@@ -382,6 +382,45 @@ def show(self, **kwargs):
...
+                nearest_lon = np.argmin(np.abs(lons[p] - field.lon))
+                nearest_lat = np.argmin(np.abs(lats[p] - field.lat))
+                Density[nearest_lon, nearest_lat] += 1
+            if relative:
+                Density /= len(dparticles)
+
+        if area_scale:
+            area = np.zeros(np.shape(field.data[0, :, :]), dtype=np.float32)
+            dx = (field.lon[1] - field.lon[0]) * 1852 * 60 * np.cos(field.lat*np.pi/180)
+            dy = (field.lat[1] - field.lat[0]) * 1852 * 60
```
mlange05: Ok, while we may not be able to internalise this conversion directly, it would be great if it were done by one of our internal `Converter` classes. For that I would probably create a `geoconv = Geographic()` or a `polarconv=GeographicPolar()` at the top of this routine, or better yet pull the converter from the field, and call `dx = polarconv.to_source(field.lon[1] - field.lon[0])`. However, as I'm writing this I'm realising that I haven't implemented the inverse call `to_source()` yet, so I'm happy for this to go in as is and be fixed later.

parcels/particleset.py
```diff
@@ -382,6 +382,45 @@ def show(self, **kwargs):
...
+
+        # For each particle, find closest vertex in x and y and add 1 or val to the count
+        if particle_val is not None:
+            for p in dparticles:
+                Density[np.argmin(np.abs(lons[p] - field.lon)), np.argmin(np.abs(lats[p] - field.lat))] \
+                    += getattr(self.particles[p], particle_val)
+        else:
+            for p in dparticles:
+                nearest_lon = np.argmin(np.abs(lons[p] - field.lon))
+                nearest_lat = np.argmin(np.abs(lats[p] - field.lat))
```
mlange05: Ok, ideally we want to add the same cell-tracking to SciPy particle as we do in JIT, so that the grid cell is always known and doesn't need to be searched here. This will probably be much faster, since we can do this on a local search, whereas `np.argmin` traverses the entire lon/lat arrays _for each particle_, which is **very inefficient**. However, until we have cell tracking in Python mode, this is probably ok.

tests/test_density.py
```diff
@@ -0,0 +1,167 @@
...
+                forcing_fields['U'].time, fields=forcing_fields)
+    K = CreateGradientField(grid, mu=[np.mean(grid.U.lon), np.mean(grid.U.lat)])
+    grid.add_field(K)
+    K_gradients = grid.K.gradient()
+    for field in K_gradients:
+        grid.add_field(field)
+
+    grid.add_field(CreateInitialPositionField(grid))
+
+    # grid.add_field(CreateInitialPositionField(grid))
```
mlange05: Debug leftovver

erikvansebille:
OK, so if we don't commit test_density.py in this commit, then the only two remaining issues are the ones that @mlange05 flagged in particleset.py. Both need more under-the-hood work to be done 'properly', but we can probably land this version right now with perhaps some comments that these two code snippets (the `np.argmin` search and the `polarconverter`) need to be improved. Then also file two issues here on github to that extent, and we should be good All agree?

mlange05:
I think we can land the whole thing with just a few small changes to `test_density.py`:
* Rename to `examples/example_density.py`
* Add `unit=Geographic()` and `unit=GeographicPolar` to the `dK_dx`/`dK_dy` fields.
* Multiply by `dt` instead of `f_lon`/`f_lat` in the `GradientClimber` kernel

Jacketless:
As these changes relate only to `test_density` and the gradient climber kernel/field, can I suggest that we drop this file from the branch and simply use Erik's more elegant test function as part of `test_particle_sets.py`?

erikvansebille:
Yes, I agree. You may want to keep test_density in your own IKAMOANA repo, of course!

If you can remove test_density for now, add short comments about the two small 'hacks' that are still in `particleset.density()`, and update with master, then we should be good to pull in.

The only things left is then perhaps to git rebase/squash the branch a bit to keep `master` clean. Is that necessary @mlange05 ?

Jacketless:
Are we able to land this PR? The changes requested by @mlange05 are now defunct as `test_density.py` has been replaced by a simpler test.

erikvansebille:
I've just pushes a small update to this PR where I reworked the `area_scale` bit so that it now uses built-in `units_to_target`
- Note that assumption still is that grid spacing (in degrees) is uniform
- Also, it would probably be faster to calculate this first-time only and then store on `grid`?

Regardless, I think this should now be merged, so that @Jacketless can continue. 

@mlange05, let us know this before Sunday if you still think this needs more work now. Otherwise, I'll merge

PR #112, opened 2016-09-10
Write flag added to particle variables

Jacketless:
Added a simple `to_write` flag to the custom variable definition
(default is true). `ParticleFile` objects now only write particle
variables to netcdf if this flag is on.

This saves on file size for internal particle variables that are used
at runtime, but have no need to be written to results.

### DIFF
--- a/examples/PARCELStutorial.ipynb
+++ b/examples/PARCELStutorial.ipynb
@@ -1474,7 +1474,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


erikvansebille:
This is great @Jacketless! Just added two commits, one that uses the `to_write` keyword to suppress output of the variables `active` and `dt` (and `xi` and `yi` in JIT mode), and the other to show use of the `to_write` keyword in the tutorial
Happy for this to land!

Jacketless:
Looks good @erikvansebille, I would merge but I'm not authorised.

PR #113, opened 2016-09-13
Options for alternative spatial interpolation methods

Jacketless:
Added a few lines fields alternative interpolation method arguments to
be passed to SciPy’s `RegularGridInterpolator` interpolator method.

At present, fields’ `self.interp_method` defaults to linear, and must
be set manually by the user for each field.

@mlange05 or @erikvansebille can you take a quick look at this?
`Field.interpolator2D` is cached method, and so I’m concerned that my
additions may not actually do anything if the field.interp_method is
changed after initialisation (which it always is at present).


mlange05:
Yes, this looks nice and it should not interfere with caching at all - the `RegularGridInterpolator` object is cached and you're only changing a constructor argument. Would be nice to add the new parameter to the dosctring for `Field`.

erikvansebille:
I like it too, @Jacketless! Also, we need to make very clear that this `interp_method` only applies to Scipy-mode. This can be done 

1. in the docstring, or
2. as a warning message when interp_method is changed but execute is in JIT

In JIT, we've written our own linear interpolation method, and while eventually we want more options there too, that's not really a priority for the v0.9 we're working towards at the moment

PR #124, opened 2016-11-20
Small fix in field loading

Jacketless:
Small change to `Field.from_netcdf()` to fix bug in loading fields from
multiple files, each of which contain time slices > 1


Jacketless:
Ah, I'm not sure why travis is failing on the flake8 tests for the examples, it passes fine on my local machine...

Issue #125, opened 2016-11-29
Implement JIT version of alternate spatial interpolation methods


erikvansebille:
PR #113 has added support for alternative interpolation methods to be set on a per-field basis. Currently this only works in SciPy. JIT needs to be implemented too.

As a first example, `nearest` would be a great alternate interpolation method to be implemented in JIT

erikvansebille:
Fixed in #126

PR #126, opened 2016-11-30
Alternative spatial interpolation in JIT

erikvansebille:
This PR fixes issue #125 where alternative spatial interpolations (via `Field.interp_method`) will now also work in JIT

It also implements a simple linear nearest neighbour interpolation

**Note that this PR was branched from PR #122, so best to land that one first.**

### DIFF
--- a/include/parcels.h
+++ b/include/parcels.h
@@ -6,6 +6,11 @@ typedef enum
     SUCCESS=0, REPEAT=1, DELETE=2, ERROR=3, ERROR_OUT_OF_BOUNDS=4
   } ErrorCode;
 
+typedef enum
+  {
+    LINEAR=0, NEAREST=1
+  } InterpCode;
+
 #define CHECKERROR(res) do {if (res != SUCCESS) return res;} while (0)
 
 typedef struct
@@ -81,13 +86,13 @@ static inline ErrorCode temporal_interpolation_linear(float x, float y, int xi, 
   err = search_linear_double(time, f->tdim, f->time, &(f->tidx));
   if (f->tidx < f->tdim-1 && time > f->time[f->tidx]) {
     t0 = f->time[f->tidx]; t1 = f->time[f->tidx+1];
-    if (interp_method==0){
+    if (interp_method == LINEAR){
       err = spatial_interpolation_bilinear(x, y, i, j, f->xdim, f->lon, f->lat,
                                           (float**)(data[f->tidx]), &f0);
       err = spatial_interpolation_bilinear(x, y, i, j, f->xdim, f->lon, f->lat,
                                           (float**)(data[f->tidx+1]), &f1);
     }
-    else if  (interp_method==1){
+    else if  (interp_method == NEAREST){
       err = spatial_interpolation_nearest2D(x, y, i, j, f->xdim, f->lon, f->lat,
                                            (float**)(data[f->tidx]), &f0);
       err = spatial_interpolation_nearest2D(x, y, i, j, f->xdim, f->lon, f->lat,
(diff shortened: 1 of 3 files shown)


include/parcels.h
```diff
@@ -66,15 +81,35 @@ static inline ErrorCode temporal_interpolation_linear(float x, float y, int xi,
   err = search_linear_double(time, f->tdim, f->time, &(f->tidx));
   if (f->tidx < f->tdim-1 && time > f->time[f->tidx]) {
     t0 = f->time[f->tidx]; t1 = f->time[f->tidx+1];
-    err = spatial_interpolation_bilinear(x, y, i, j, f->xdim, f->lon, f->lat,
-                                        (float**)(data[f->tidx]), &f0);
-    err = spatial_interpolation_bilinear(x, y, i, j, f->xdim, f->lon, f->lat,
-                                        (float**)(data[f->tidx+1]), &f1);
+    if (interp_method==0){
```
mlange05: Comparing to integer values here is confusing and makes it hard to guess what value encapsulates which method. How about adding an `enum` similar to `ErrorCode` so that you can write `interp_method==NEAREST` here?

tests/test_grid_sampling.py
```diff
@@ -99,6 +99,25 @@ def test_grid_sample_eval(grid, xdim=60, ydim=60):
...
+    P = np.zeros((xdim, ydim), dtype=np.float32)
+    P[0, 0] = 1.
+    grid = Grid.from_data(U, lon, lat, V, lon, lat, mesh='flat',
+                          field_data={'P': np.asarray(P, dtype=np.float32)})
+    grid.P.interp_method = 'nearest'
+    xv, yv = np.meshgrid(np.linspace(0.1, 0.9, np.sqrt(npart)), np.linspace(0.1, 0.9, np.sqrt(npart)))
+    pset = grid.ParticleSet(size=np.size(xv), pclass=pclass(mode),
+                            lon=xv.flatten(), lat=yv.flatten())
+    pset.execute(k_sample_p, endtime=1, dt=1)
+    assert np.allclose(np.array([p.p for p in pset if p.lon < 0.5 and p.lat < 0.5]), 1.0, rtol=1e-5)
```
mlange05: Does this still work with a differently sized grid, eg. 3x3? If not, we might want to hard-code the 2x2 grid and remove `xdim=2` and `ydim=2` from the function args.

parcels/field.py
```diff
@@ -310,8 +310,13 @@ def eval(self, time, x, y):
         return self.units.to_target(value, x, y)
 
     def ccode_eval(self, var, t, x, y):
-        return "temporal_interpolation_linear(%s, %s, %s, %s, %s, %s, &%s)" \
-            % (x, y, "particle->xi", "particle->yi", t, self.name, var)
+        # Casting interp_method to int as easier to pass on in C-code
+        if self.interp_method is 'linear':
+            interp_method = 0
```
mlange05: If we use enum value for the method selection, this here also needs changing to `interp_method = 'LINEAR/NEAREST'`, I think.

tests/test_grid_sampling.py
erikvansebille: Changed so that xdim and ydim are set within the function

parcels/field.py
erikvansebille: Left it as is now. I tried creating a class `InterpCode` just like in kernel.error
```
class InterpCode(IntEnum):
    Linear = 0
    Nearest = 1
```
 but not sure how to implement then in `ccode_eval`. When I try
```
    def ccode_eval(self, var, t, x, y):
        # Casting interp_method to int as easier to pass on in C-code
        if self.interp_method is 'linear':
            interp_method = InterpCode.Linear
        elif self.interp_method is 'nearest':
            interp_method = InterpCode.Nearest
        return "temporal_interpolation_linear(%s, %s, %s, %s, %s, %s, &%s, %s)" \
            % (x, y, "particle->xi", "particle->yi", t, self.name, var, interp_method)
```
I get a compilation error
```
error: unexpected type name 'InterpCode': expected expression
    err = temporal_interpolation_linear(particle->lon, particle->lat, particle->xi, particle->yi, time, P, &tmp0, InterpCode.Nearest);
                                                                                                                  ^
```

Am I missing something?

mlange05: Hmm, ok. All I was after was that we pass in the string `NEAREST` into the C kernel rather than its equivalent `1` for readability of the kernel code. I've just pushed a small adjustment for this purpose.

That being said, encapsulating this into a custom utility class to avoid string comparison on the Python side is probably a good idea but since SciPy expects the string it seems a little pointless at this point.

mlange05:
Ok, good to go once Travis is green.

PR #129, opened 2016-12-02
Updating initialisation of netcdf output file

erikvansebille:
This PR does two things to the ParticleFile in netcdf
1) Makes sure the first time written out is the `starttime` of the `.execute`, rather than the default 0
2) Does not by default allocate `z` anymore. Depth is now treated as any other variable, that does not need to be written into output


erikvansebille:
Closing pull request, as changes make more sense in the bigger 'indexed ParticleSet' PR that I am preparing

PR #130, opened 2016-12-08
Non uniform K diffusion

Jacketless:
This commit adds a 2d diffusion kernel for particles in spatial varying
diffusivity fields.

A test file is included demonstrating how brownian motion results in
erroneous concentrations of particles in areas of low diffusivity.

Not sure if we wish to add a new kernel file such as `diffusion.py`,
but please let me know your comments.


Jacketless:
Travis failures seem to be coming from the previously implemented density calculation branch- can you check and approve PR #111 @mlange05, if you feel it's good to go? That might fix these issues.

erikvansebille:
Actually, the travis failure is because you did not specify `pnum` in `test_density_calculation`. Even though you pass it as an argument when you call the function in `__main__`, the py.test does not go through `main` so you need to either give it a default value in the function declaration, or add `pnum` as a `pytest-fixture`

erikvansebille:
This looks good, @Jacketless. See above for a bunch of comments on the code; most should be near-trivial to implement. Let me know if you have any questions!

Jacketless:
Great, thanks Erik. Looking through the pull request on github, I can't
find the "above comments on the code" you mention. In fact, none of the
commits seem to have comments associated with them. Am i missing something?
Sorry

tests/test_diffusion.py
```diff
@@ -0,0 +1,199 @@
...
+    # Evenly spread starting distribution
+    Start_Field = Field('Start', CreateStartField(grid.U.lon, grid.U.lat),
+                        grid.U.lon, grid.U.lat, depth=grid.U.depth, time=grid.U.time, transpose=True)
+
+    # Calculate appropriate timestep
+    min_timestep = np.floor(np.min([np.min(1/np.abs(grid.d2K_dx.data)), np.min(1/np.abs(grid.d2K_dy.data))]))
+    print('Timestep should be < %s' % min_timestep)
+    timestep = 500
+
+    steps = timesteps
```
Jacketless: This is actually a method to find the minimum timestep required to accurately replicate Eulerian diffusion- but you're right, it serves no purpose in this example, so I will remove it.

tests/test_diffusion.py
```diff
@@ -0,0 +1,199 @@
...
+    Ry_component = Ry * math.sqrt(2 * Kfield * dt / r_var) * to_lat
+    particle.lon += Rx_component
+    particle.lat += Ry_component
+
+
+def UpdatePosition(particle, grid, time, dt):
+    particle.prev_lon = particle.new_lon
+    particle.new_lon = particle.lon
+    particle.prev_lat = particle.new_lat
+    particle.new_lat = particle.lat
```
Jacketless: This always ends up in an infinite loop on my machine. The reason is when the diffusion kernel moves a particle to outside the grid bounds, no error is triggered. Update position then saves this new position to `prev_lon`. During the next timestep, field sampling out of bounds triggers and error, but it's too late to as `particle.prev_lon` is already saving the out of bounds position... I'm surprised this worked for you. 

Each particle needs to remember it's position _before_ the previous time-step that caused it to move out of bounds. Hence the requirement for a "memory" of two timesteps, if we don't store positions _within_ the diffusion kernels (which I'm trying to keep generic).

tests/test_diffusion.py
```diff
@@ -0,0 +1,199 @@
...
+    densities1 = diffusion_test(args.mode, 'true_diffusion', args.particles, args.timesteps, args.output)
+
+    tol = 0.0001
+    assert np.abs(np.var(densities1[0]) - np.var(densities1[1])) < tol, \
+        'Variance in diffuser particle density across cells is significantly different from start to end of simulation!'
+
+    densities2 = diffusion_test(args.mode, 'brownian_motion', args.particles, args.timesteps, args.output)
+    print(np.abs(np.var(densities2[0]) - np.var(densities2[1])))
+    assert np.abs(np.var(densities2[0]) - np.var(densities2[1])) > tol, \
+        'Variance in brownian motion particle density across cells is not significantly different from start to end of simulation!'
```
Jacketless: As you pointed out, this does not really belong in the group of unit tests, I'll address the rest of the comments first. Once we're happy with these, I'll change the file name so that this sits in the examples directory, and remove these pytest markers

mlange05:
Ok, so the `prev_lon` problem sounds like a bug, since updating a particle position to something out-of-bounds is conceptually wrong and should thus trigger the kernel error. I'll have a look at this and see if it's easy to fix.

mlange05:
Ahhh, updating `p.lon` to an out-of-bounds value actually triggers an error in JIT mode, but does not in SciPy mode, which probably explains why behaviour is different on different machines.

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
+from parcels import Field, Grid, JITParticle, ScipyParticle, Variable, SaptiallyVaryingDiffusion2D, random, ErrorCode
```
erikvansebille: I still think this function is more an 'example' than a unit-test, so I think it would be better placed in the `examples` directory (and be renamed `example_SpatiallyVaryingDiffusion2D.py`

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
...
+                         np.linspace(np.min(grid.U.lat), np.max(grid.U.lat), 5, dtype=np.float32))
+    StartDensity = diffusers.density(DensityField, area_scale=False, relative=True)
+    grid.add_field(Field(type + 'StartDensity', StartDensity,
+                         DensityField.lon,
+                         DensityField.lat))
+
+    diffuse = diffusers.Kernel(SaptiallyVaryingDiffusion2D) if type == 'true_diffusion' else diffusers.Kernel(LagrangianDiffusionNoCorrection)
+
+    diffusers.execute(diffusers.Kernel(UpdatePosition) + diffuse, endtime=grid.U.time[0]+timestep*steps, dt=timestep,
+                      output_file=diffusers.ParticleFile(name=args.output+type),
```
erikvansebille: `args` is unknown, so this should be `output_file=diffusers.ParticleFile(name=output_file+type)`?

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
...
+                         np.linspace(np.min(grid.U.lon), np.max(grid.U.lon), 5, dtype=np.float32),
+                         np.linspace(np.min(grid.U.lat), np.max(grid.U.lat), 5, dtype=np.float32))
+    StartDensity = diffusers.density(DensityField, area_scale=False, relative=True)
+    grid.add_field(Field(type + 'StartDensity', StartDensity,
+                         DensityField.lon,
+                         DensityField.lat))
+
+    diffuse = diffusers.Kernel(SaptiallyVaryingDiffusion2D) if type == 'true_diffusion' else diffusers.Kernel(LagrangianDiffusionNoCorrection)
+
+    diffusers.execute(diffusers.Kernel(UpdatePosition) + diffuse, endtime=grid.U.time[0]+timestep*steps, dt=timestep,
```
erikvansebille: Since `steps = timesteps` above, why not simply use `timesteps` here, and get rid of the `steps = timesteps` above?

Also, it is probably easier here to use the `runtime` keyword (i.e. `runtime=timesteps*steps`), which sets `endtime` as `starttime + runtime`

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
...
+    # Diffusers should spread out evenly regardless of this same low-diffusivity
+    densities1 = diffusion_test(args.mode, 'true_diffusion', args.particles, args.timesteps, args.output)
+
+    tol = 0.0001
+    assert np.abs(np.var(densities1[0]) - np.var(densities1[1])) < tol, \
+        'Variance in diffuser particle density across cells is significantly different from start to end of simulation!'
+
+    densities2 = diffusion_test(args.mode, 'brownian_motion', args.particles, args.timesteps, args.output)
+    assert np.abs(np.var(densities2[0]) - np.var(densities2[1])) > tol, \
+        'Variance in brownian motion particle density across cells is not significantly different from start to end of simulation!'
```
erikvansebille: These `asserts` still won't be executed by Travis, as the py.test doesn't run the `main`

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
...
+# Recovery Kernal for particles that diffuse outside boundary
+def Send2PreviousPoint(particle):
+    # print("Recovery triggered at %s | %s!" % (particle.lon, particle.lat))
+    # print("Moving particle back to %s | %s" % (particle.prev_lon, particle.prev_lat))
+    particle.lon = particle.prev_lon
+    particle.lat = particle.prev_lat
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+def diffusion_test(mode, type='true_diffusion', particles=1000, timesteps=1000, output_file='diffusion_test'):
```
erikvansebille: Since there are no `asserts` in the function, py.test can't check anything. Also, for py.test to work a function name has to start with `test_`. Run `py.test -v -s tests/test_diffusion.py` to check

parcels/kernels/diffusion.py
```diff
@@ -0,0 +1,25 @@
+from parcels.rng import random
+import math
+
+
+__all__ = ['SaptiallyVaryingDiffusion2D']
+
+
+def SaptiallyVaryingDiffusion2D(particle, grid, time, dt):
```
erikvansebille: Type-o? `Saptially` -> `Spatially`??

parcels/kernels/diffusion.py
```diff
@@ -0,0 +1,25 @@
...
+    # Diffusion equations for particles in non-uniform diffusivity fields
+    # from Ross &  Sharples 2004 and Spagnol et al. 2002
+    to_lat = 1 / 1000. / 1.852 / 60.
+    to_lon = to_lat / math.cos(particle.lat*math.pi/180)
```
erikvansebille: I am very worried about these kind of explicit geographic statements in default kernels! This will not work for users who run with flat meshes. It's probably acceptable to have these explicit `to_lon` statements in custom kernels in example scripts, but the kernels in `parcels/kernels/*` should be generic w.r.t. meshes

So we need to get this to work through the unit-converter. I'll have a look in the coming days

tests/test_diffusion.py
```diff
@@ -0,0 +1,188 @@
...
+
+# Recovery Kernal for particles that diffuse outside boundary
+def Send2PreviousPoint(particle):
+    # print("Recovery triggered at %s | %s!" % (particle.lon, particle.lat))
+    # print("Moving particle back to %s | %s" % (particle.prev_lon, particle.prev_lat))
+    particle.lon = particle.prev_lon
+    particle.lat = particle.prev_lat
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
```
erikvansebille: Execution is extremely slow on my machine in scipy mode (taking > minutes). Since Travis testing already takes quite long (10+ minutes), can't we make the test a bit smaller/shorter? Ideally it would run in a few seconds...

Issue #75, opened 2016-06-03
Ability to specify subset of netcdf data to read in when setting grid


erikvansebille:
It would be nice if users could set a subset of the data they want to read in only (e.g. only one depth level) when they specify the grid. NetCDF has ability to read subsets, so we only need way for users to set the limits in `Grid.from_netcdf`

mlange05:
That should be reasonably straightforward to implement. We already pass the naming convention for each field into the `FileBuffer` when reading files, which then simply does `[:, :, ]` for each known dimensions. One way to do this would be to extend the `dimensions` dict to hold nested tuples `(name, shape)`, where `shape = (d0, d1, d2)` for the relevant number of data dimensions. This way we can expose the the shape argument to the user via optional arguments to `Grid.from_netcdf` and `Grid.from_nemo`, for example `dimensions['data'] = ('myvariable', (:, :, 3:5))` for depth layers 3 and 4 only. One thing we then need to do is sanity checking that the passed shape has the right number of dimensions, ie. something like `len(dataset[dimensions[0]]) == len(dimensions['data'][1])`. Does that sound feasible?

erikvansebille:
Yes, this sounds like a very good way forward.

Eventually, in some future version, we might want to add the capacity for users to select the subset based on dimension values, rather than just indices (i.e. selecting depths 0-100m, rather than index 0:5; or latitudes 30N-50N, rather than index 400:500), but that is just sugarcoating that can always be added later

So let's go for this indexing solution first. No high priority, but certainly a nice feature to have

erikvansebille:
Fixed in #136

Issue #106, opened 2016-08-16
grid.from_netcdf does not efficiently allocate memory for models with many time snapshots


erikvansebille:
The filebuffer currently used does not appear to be very efficient when a model has a large number of time snapshots. 

Take for example the Globcurrent data. Running `memory_profiler` on it shows that if all files are used (i.e. `'U': "examples/GlobCurrent_example_data/20*.nc"`) then the total memory footprint of the `grid` is 12.5 MiB.

However, if only the January data is read in (i.e. `'U': "examples/GlobCurrent_example_data/200201*.nc"`), then the total memory footprint of the `grid` is 2.5 MiB.

This becomes more problematic with much larger grids. For example the 0.1 degree global OFES grid (3600 \* 1800 grid points in each snapshot) takes 70 MiB for one snapshot, but if the filebuffer is given 10 snapshots it require 512 MiB. And if it given 73 snapshots (a year's worth, at 3-day interval), the memory footprint of `grid` is 5,075 MiB.

For Parcels to be useful for long experiments on large grids, it needs to be smarter with memory allocation. For example the CMS always has only three consecutive snapshots of the velocity fields in memory, and shifts them through as `time` runs forward.

Jacketless:
A somewhat related issue which I am coming up against increasingly is that of loading multiple netCDF files that contain the same variable across different time periods.

Many times long time-series of high resolution variable data are split up into separate files, e.g. by month, day etc. When these are on a computing cluster, it's not really appropriate to stitch them together using NCO tools into new (and enormous) netcdfs.

It would be neat if multiple files could be specified to PARCELS during grid creating/additions, and multiple instances of the same variable were stitched together in the time dimension. Not too complicated an addition, but possibly fiddly to do. What's the best approach for this?

erikvansebille:
Hi @Jacketless, maybe I misunderstand, but isn't the wildcard operator `*` what you are after? See e.g. the `examples/example_globcurrent.py` script. The code that handles this is in the `from_netcdf` classmethod in `grid.py` (line 106).

Let me know if this fixes your issue with multiple netCDF files, or if you want something else

Jacketless:
Apologies for my slow reply, this week has been very busy with PI meetings and prep for the Nouméa visit.

@erikvansebille You're absolutely right, PARCELS does already have this functionality. Sorry about that. I have not yet had a chance to try it out, but it should work fine.
[...]

Issue #39, opened 2016-02-23
Implement vertical velocities 


erikvansebille:
Currently, parcels can only advect in 2D (horizontal) flow. However, for many applications particles will need to be tracked in 3D. 

This is actually quite a critical functionality, which once we have it will allow much more uptake in the physical oceanographic community

This will require an extension of the advection functions, but should for the rest be relatively straightforward?

erikvansebille:
I started working on the vertical velocities today, but quickly hit a (what I think is a) pretty major roadblock.

I've committed changes to test_particle that allow creation of multiple depth layers in the netcdf files. That was easy. However, making parcels read these fields is much more tricky

The problem, I think, is that we are using `scipy.interpolate.RectBivariateSpline` in field.py (line 96) and that this function as it names suggests can only handle 2D fields. There does not appear to be a TrivariateSpline.

I've looked around, and scipy also has the `scipy.interpolate.RegularGridInterpolator` function, but that will require significant rewriting to work. Plus, this function only works on regular grids, so we will again hit a roadblock once we go to irregular grids.

Another option is to assume that grids are always staggered in the vertical (this is a good assumption, I've never seen an ocean model which where vertical grid cells are not column-wise on top of each other). If we assume this, we could treat the interpolation in the vertical separately: interpolate the grid level above and below the particle in 2D, and then vertically interpolate between these two points. Does this make sense?

Any ideas how to progress from here very welcome! 3D tracking of particles is a pretty major enhancement which we will absolutely need for the GMD manuscript

erikvansebille:
Following up on my comment earlier today, what I suggested with the separate treatment of vertical direction from the horizontal direction is equivalent to how interpolation in time is treated. So we could build on that code

mlange05:
Ok, I propose that we eventually want to do both for the GMD manuscript, since one of parcel's major abstractions is the generic interaction with different types of grids and different interpolation methods. I like the idea of using `scipy.interpolate.RegularGridInterpolator`, since it uses trilinear interpolation and should thus be almost perfectly equivalent to what we do in JIT mode. Similarly though, I also would like the opportunity to treat vertical interpolation separately, since I can see various reasons and use cases for this too. 

[...]

PR #136, opened 2016-12-20
Subsetting from NetCDF

erikvansebille:
Adding option to only read in subset of the data of netCDF files, by providing `indices` dictionary to `grid.from_file` and/or `grid.from_nemo`. This fixes issue #75 and addresses issue #106 (more to come on that in a PR in a few days)

This PR currently only works for `lon`, `lat` and `time`. `depth` will be added once we have support for vertical velocities (issue #39).

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -181,12 +181,8 @@ class Field(object):
         if not isinstance(filenames, Iterable):
             filenames = [filenames]
         with FileBuffer(filenames[0], dimensions) as filebuffer:
-            lon = filebuffer.lon
-            indslon = indices['lon'] if 'lon' in indices else range(lon.size)
-            lon = lon[indslon]
-            lat = filebuffer.lat
-            indslat = indices['lat'] if 'lat' in indices else range(lat.size)
-            lat = lat[indslat]
+            lon, indslon = filebuffer.read_dimension('lon', indices)
+            lat, indslat = filebuffer.read_dimension('lat', indices)
             # Assign time_units if the time dimension has units and calendar
             time_units = filebuffer.time_units
             calendar = filebuffer.calendar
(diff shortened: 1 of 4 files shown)


parcels/field.py
```diff
@@ -168,20 +168,21 @@ def __init__(self, name, data, lon, lat, depth=None, time=None,
         self.time_index_cache = LRUCache(maxsize=2)
 
     @classmethod
-    def from_netcdf(cls, name, dimensions, filenames, **kwargs):
+    def from_netcdf(cls, name, dimensions, filenames, indices, **kwargs):
```
Jacketless: I think `indices` needs to have a default value here of an empty dictionary, which should then default to all indices in `read_dimension`, as is the case in `grid`. Some users may load fields after creation of the grid, then add them to the grid themselves.

erikvansebille:
OK thanks @mlange. I guess you are right about the filebuffering for time yes. But to me that doesn't seem trivial to implement. Happy to have a go, but would need some pointers as to where to start

Also, the netcdf subsetting was intended mainly is for `lon`, `lat` and in future `depth`. If dor example you want to release particles only in the Atlantic but your netcdf is global, then you can obtain a significant speed increase by only reading in a fraction of each grid. Again, this would ideally be done 'on the fly', but that is probably something for v2.0? The idea with this PR was to at least build a bit of the infrastructure of indexed reading, and to allow some production runs ;-)

mlange05:
Oh yes, doing the buffering right (even in time) is a very complicated task. That's why I'm totally happy with this.

For the spatial buffering I was always planning to do an even more intricate scheme, where we effectively buffer in space based on some type of recursive K-tree. That is very advanced though and only necessary once we venture into real HPC.

Issue #128, opened 2016-12-02
pset.add and netcdf output; alternative output option in plain csv?


erikvansebille:
As I've started to work more with Parcels for my own research, I've ran into a serious problem with the netcdf output: it doesn't work as soon as you start adding/deleting particles...

The netcdf library assumes that there is only one dimension that can vary (which we use for time), and that all other dimensions need to be set on initialisation of the file.
This means you can't add particles to the file, or remove them, during a `.execute` if you also want to write them away.

Now, in CMS we give the option to output in simple csv/ascii format for exactly this reason. I.e. every time a `.write` is called, we would simply add all particle locations and variables to the end of a txt file. That file will be order in time, rather than ordered by particles. So we will need to 
1. provide an alternative `write` routine in ascii
2. give every particle a unique ID (so that is can be tracked within the file)
3. create a tool to postprocess the time-ordered file into a more convenient format for analysis after one is done with particle tracking

erikvansebille:
Actually, I now see that there is a cf-convention for this use case too: convention H4.4 at [http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/aphs04.html](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/aphs04.html)

In essence, rather than saving data as 2D `(traj_number, time)` matrices, you save all data as 1D vectors. And then you add a `trajectory_index` vector to the netcdf file, which relates each item in the vectors to a `traj_number`.

I'll try and implement..

erikvansebille:
Fixed using indexed ParticleFile

Issue #132, opened 2016-12-12
File output after multiple particleset.execute() calls


Jacketless:
At present it is very simple to call `particleset.execute()` within a loop (for example, when some calculation such as density is required between larger 'time-leaps'), and this results in a number of separate particle trajectory output files. Each `Particle` object in the set keeps it's lon/lat position between calls, so there is no need to create a new `particleset` object between calls to `execute()`.

However, I've realised that it's not so simple to 'stitch' these output files together post-hoc. Firstly, the starting position is re-written to the output file each time, which in this case will also be the final written position of the previous time-leap file.
In addition, the recorded timestamp of the first observation always appears to be zero, regardless of the `starttime` argument passed to the `particleset.execute()` call.
The unique particle ID suggested in issue #128 would also aid in correctly stitching multiple output files together.

Off the top of my head, this could be addressed by:

1. Leaving the user to remove duplicate observations themselves post-hoc (which is fairly trivial given the duplicate obs always has a timestamp of zero at present), or;
2. Adding a 'continue` argument to `particleset.execute()` calls, indicating the start positions do not need to be written to file for that call, or;
3. Adding a `started` boolean attribute to the `particleset` class, which is automatically flagged after the first instance of `execute()` being called, and automatically stops initial positions of particles being written in subsequent calls

erikvansebille:
I agree that this is an issue, although there is already the option to flag `initial_dump=False` when creating a new `ParticleFile`. So that should make stitching much simpler.

The other option is simply to reuse the `ParticleFile` object. I.e. to call `pfile = ParticleFile(filename)` only once, before `.execute()`, and then in the for loop, write `pset.execute(..., output_file=pfile)`

However, none of these solutions work when particles are added or deleted in between calls to `.execute()` (i.e. when the size of pset changes). That is why I have been working on a solution to this too over the past few days. See commit c24abc2.

The way I solved it in c24abc2 (part of the Particlefile-Indexed branch; soon to be a PR), is to remove the initial_dump from `ParticleFile.__init__` altogether and instead call `pfile.write()` *before* the kernel execution in the `.execute()`

Now, particleFile is updated before kernel execution. Advantages are
- ‘Time’ of particles is known, so doesn’t have to be guessed as 0.
- Particles can be added/deleted in between `.execute()` calls
- Start locations of particles are always accurately written

[...]

PR #137, opened 2016-12-22
Indexed Particle file for outputting with particles additions and removals

erikvansebille:
This PR fixes issue #128 by implementing an alternative way to store `ParticleFiles` in NetCDF, using the 'ragged' indexing of  convention H4.4 at http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/aphs04.html

This 'indexed' way of storing data means the number of particles doesn't have to be specified at the creation of the `ParticleFile`, which means experiments where particles are added and removed during execution still can be saved

Note that for this indexed writing to work properly, the ParticleFile has to be written *before* kernel execution in the loop, otherwise the initial location of added particles is never stored. This means the `initial_dump` in ParticleFile has to go, and this PR therefore also touches on issue #132

Finally, the PR includes a jupyter tutorial to show how to use the indexed `ParticleFile`


parcels/particleset.py
```diff
@@ -249,13 +249,14 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
         assert(timeleaps >= 0)
         leaptime = starttime
         for _ in range(timeleaps):
-            leaptime += interval
-            self.kernel.execute(self, endtime=leaptime, dt=dt,
-                                recovery=recovery)
+            # First write output_file, because particles could have been added
```
mlange05: If we do this, don;t we need another final output after the loop ends?

parcels/particle.py
```diff
@@ -117,6 +117,7 @@ class ScipyParticle(_Particle):
     lon = Variable('lon', dtype=np.float32)
     lat = Variable('lat', dtype=np.float32)
     time = Variable('time', dtype=np.float64)
+    trajectory = Variable('trajectory', dtype=np.int32)
```
mlange05: I wonder whether we want to call this variable `id` instead of `trajectory`, since it will potentially be used for more than indexed trajectory output in the future?

parcels/particle.py
```diff
@@ -125,6 +126,8 @@ def __init__(self, lon, lat, grid, dt=3600., time=0., cptr=None):
         type(self).lon.initial = lon
         type(self).lat.initial = lat
         type(self).time.initial = time
+        type(self).trajectory.initial = grid.lasttrajectory
```
mlange05: I'm not sure the `grid` should be counting the IDs. We might just want a global module level ID counter with a fixed initial value?

parcels/particleset.py
erikvansebille: The problem if we do another `write` at the end, is that if we then restart execution (say in a loop), we output twice. The looping means it's a choice whether we output initial or final positions.

Another option is to add a test to ParticleFile.write to check if a certain time has been written already. Then we can write both at the start and the end of the `.execute`. I'll have a look into that

parcels/particle.py
erikvansebille: OK, so this seems to be a deep bug within the JIT cptr. If I simply rename `trajectory` to `id`, the JIT goes all wonky, including segmentation faults on some tests..

One solution is to call the variable `xid`. My hypothesis now is that JIT is fine as long as this variable is copied *after* the `time` variable (i.e. has to be lexicographically later, like `trajectory` or `xid`). 

Something about alignment of the `cptr`??

parcels/particleset.py
erikvansebille: OK, so I've implemented a final write of ParticleFile at then end of .execute now (see c3bd535)

parcels/particle.py
erikvansebille: I think I've fixed this now. The trick was to, when the list of variables is compiled in particle.py, first list all the 64-bit variables, and only then concatenate the 32-bit variables. In that way, we will never get a 64-bit variable to be misaligned

PR #138, opened 2016-12-23
Error handling for extrapolating time (sampling outside time domain)

erikvansebille:
While there are some cases where it is ok to sample outside time domain (e.g. if flow is steady-state), in most production cases you would not want to sample beyond the time range provided in the Fields.

This PR provides control over whether parcels can sample outside the time range, via the `allow_time_extrapolation` boolean on `fields` and `grids`. 
- The `allow_time_extrapolation` option is by default `True` for `Grid.from_data` and also on `Fields` that have no `time` dimension. 
- The `allow_time_extrapolation` option is by default `False` for `Grid.from_netcdf` and `Grid.from_nemo`, since these are typically the large, time-evolving grids where extrapolation could lead to unexpected results.

This PR also adds a dedicated error handling class for these `TimeExtrapolationErrors`, as well as unit-tests in `grid_sampling.py`


include/parcels.h
```diff
@@ -83,6 +83,9 @@ static inline ErrorCode temporal_interpolation_linear(float x, float y, int xi,
   err = search_linear_float(x, f->xdim, f->lon, &i); CHECKERROR(err);
   err = search_linear_float(y, f->ydim, f->lat, &j); CHECKERROR(err);
   /* Find time index for temporal interpolation */
+  if (f->allow_time_extrapolation == 0 && (time < f->time[0] || time > f->time[f->tdim-1])){
+    CHECKERROR(ERROR_TIME_EXTRAPOLATION);
```
mlange05: I think you want to `return ERROR_TIME_EXTRAPOLATION` here.

parcels/kernels/error.py
```diff
@@ -46,6 +47,17 @@ def __init__(self, particle, lon=None, lat=None, field=None):
         super(OutOfBoundsError, self).__init__(particle, msg=message)
 
 
+class TimeExtrapolationError(KernelError):
```
mlange05: Not strictly necessary, but how about `OutOfTimeError` instead of `TimeExtrapolationError` for consistency?

PR #139, opened 2016-12-29
Particle set size as optional argument

erikvansebille:
Since the size of the `ParticleSet` can be obtained from the length of the `lon` and `lat` vector on `ParticleSet.__init__`, it doesn’t need to be a compulsory argument.

The PR makes `size` an optional argument, only required when using `start`/`finish` or `start_field` arguments. 

Note that this required a lot of small changes in the `tests/*.py` and `examples/*.py` files, as almost every call to `ParticleSet` assumed `size` a compulsory argument

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -69,22 +69,25 @@ class ParticleSet(object):
     Please note that this currently only supports fixed size particle
     sets.
 
-    :param size: Initial size of particle set
     :param grid: Grid object from which to sample velocity
     :param pclass: Optional class object that defines custom particle
-    :param lon: List of initial longitude values for particles
-    :param lat: List of initial latitude values for particles
+    :param lon: Optional list of initial longitude values for particles
+    :param lat: Optional list of initial latitude values for particles
     :param start: Optional starting point for initilisation of particles
                  on a straight line. Use start/finish instead of lat/lon.
     :param finish: Optional end point for initilisation of particles on a
                  straight line. Use start/finish instead of lat/lon.
     :param start_field: Optional field for initialising particles stochastically
                  according to the presented density field. Use instead of lat/lon.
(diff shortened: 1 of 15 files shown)


parcels/particleset.py
```diff
@@ -69,22 +69,25 @@ class ParticleSet(object):
...
+    :param size: Optional initial size of particle set, only required when using
+                 start/finish or start_field arguments
     """
 
-    def __init__(self, size, grid, pclass=JITParticle,
-                 lon=None, lat=None, start=None, finish=None, start_field=None):
+    def __init__(self, grid, pclass=JITParticle, lon=None, lat=None,
+                 start=None, finish=None, start_field=None, size=None):
         self.grid = grid
+        size = len(lon) if size is None else size
```
mlange05: Just FYI, a shorter version of saying that is `size = size or len(lon)`, but this is good too. :snake:

parcels/particleset.py
```diff
@@ -69,22 +69,25 @@ class ParticleSet(object):
...
+    :param lon: Optional list of initial longitude values for particles
+    :param lat: Optional list of initial latitude values for particles
     :param start: Optional starting point for initilisation of particles
                  on a straight line. Use start/finish instead of lat/lon.
     :param finish: Optional end point for initilisation of particles on a
                  straight line. Use start/finish instead of lat/lon.
     :param start_field: Optional field for initialising particles stochastically
                  according to the presented density field. Use instead of lat/lon.
+    :param size: Optional initial size of particle set, only required when using
+                 start/finish or start_field arguments
```
mlange05: I wonder if we ultimately want to create separate constructor methods, similar to the ones we have for grids/fields. Maybe something like `ParticleSet.from_array(lon, lat)` and `ParticleSet.from_line(size, lon, lat)`? But definitely not required for this PR, just a thought really.

Issue #140, opened 2016-12-30
Create ParticleSet constructor methods


erikvansebille:
It might be good to split the `ParticleSet` init into different constructor methods. E.g. 
- `ParticleSet.from_list(grid, pclass, lon, lat)`
- `ParticleSet.from_line(grid, pclass, start, finish, size)`
- `ParticleSet.from_area(grid, pclass, SW-corner, NE-corner, size)`
- `ParticleSet.from_start_field(grid, pclass, start_field, size)`

See also comments in #139

erikvansebille:
Fixed in #144

PR #144, opened 2017-01-04
Particleset constructor methods

erikvansebille:
Implementation of the ParticleSet constructor methods `from_list`, `.from_line` and `from_field`, as per Issue #140 

Note that I have also removed the `grid.ParticleSet` method. It always was a fit circular to have a `ParticleSet` class that required a `Grid` object, and at the same time a `Grid` class that had a `ParticleSet` method.

So with the new constructor methods to `ParticleSet`, I have removed the `grid.ParticleSet` method so that all calls should be `ParticleSet.from_list(grid, …)` or `ParticleSet.from_line(grid, …)` etc


parcels/particleset.py
```diff
@@ -122,6 +105,46 @@ def cptr(i):
...
+
+        :param grid: :mod:`parcels.grid.Grid` object from which to sample velocity
+        :param pclass: mod:`parcels.particle.JITParticle` or :mod:`parcels.particle.ScipyParticle`
+                 object that defines custom particle
+        :param start: Starting point for initialisation of particles on a straight line.
+        :param finish: End point for initialisation of particles on a straight line.
+        :param size: Initial size of particle set
+        """
+        lon = np.linspace(start[0], finish[0], size, dtype=np.float32)
+        lat = np.linspace(start[1], finish[1], size, dtype=np.float32)
```
Jacketless: Depending on the mesh type of `grid`, should we make sure that particles are genuinely evenly spaced if `start` and `finish` span long distances?

parcels/particleset.py
```diff
@@ -71,23 +71,16 @@ class ParticleSet(object):
...
     """
 
-    def __init__(self, grid, pclass=JITParticle, lon=None, lat=None,
-                 start=None, finish=None, start_field=None, size=None):
+    def __init__(self, grid, pclass=JITParticle, lon=None, lat=None):
+        # Convert numpy arrays to one-dimensional lists
+        lon = lon.flatten() if isinstance(lon, np.ndarray) else lon
+        lat = lat.flatten() if isinstance(lat, np.ndarray) else lat
+        size = len(lon)
         self.grid = grid
```
Jacketless: Given PARCELS current lack of smart grid loading, at present when executing the same `ParticleSet` object across multiple `grid` objects, you must either update this pointer yourself (which I think works, but manually changing pointers always makes me nervous!) or create new `ParticleSet` objects and copy particle properties one-by-one between calls to `execute()`.
Should we think about adding a `ParticleSet.switch_grid()` function that intelligently switches this pointer to a new grid object, whilst checking that all particles are still in the domain, keeping the original time_origin, and updating any appropriate `Particle` variables which may rely on grid (such as `xi` and `yi` for `JITParticle` objects) etc?

mlange05: Ok, grid-switching at runtime indeed requires a lot of sanity checking and under-the-hood updates if we want to do this safely and is therefore not really supported at the moment. If you have a good use case for this functionality though, can you please file an issue, so that we can look into this separately? I don't think this impacts this PR.

parcels/particleset.py
```diff
@@ -122,6 +105,46 @@ def cptr(i):
...
+    def from_field(cls, grid, pclass, start_field, size):
+        """Initialise the ParticleSet randomly drawn according to distribution from a field
+
+        :param grid: :mod:`parcels.grid.Grid` object from which to sample velocity
+        :param pclass: mod:`parcels.particle.JITParticle` or :mod:`parcels.particle.ScipyParticle`
+                 object that defines custom particle
+        :param start_field: Field for initialising particles stochastically according to the presented density field.
+        :param size: Initial size of particle set
+        """
+        lon, lat = positions_from_density_field(size, start_field)
```
mlange05: We could just move the code from `positions_from_density_field` into here now, so that we can sync the function arguments (I believe `mode` is missing here).

parcels/particleset.py
```diff
@@ -71,23 +71,16 @@ class ParticleSet(object):
...
+    :param lat: List of initial latitude values for particles
     """
 
-    def __init__(self, grid, pclass=JITParticle, lon=None, lat=None,
-                 start=None, finish=None, start_field=None, size=None):
+    def __init__(self, grid, pclass=JITParticle, lon=None, lat=None):
+        # Convert numpy arrays to one-dimensional lists
+        lon = lon.flatten() if isinstance(lon, np.ndarray) else lon
+        lat = lat.flatten() if isinstance(lat, np.ndarray) else lat
+        size = len(lon)
```
mlange05: For safety reasons, we might want to `assert(len(lon) == len(lat))` here.

parcels/particleset.py
erikvansebille: I think that providing for geodetic deformations (which I assume you're referring to; i.e. when a shortest line between two points on a globe is not a straight line) is actually quite complicated to compute.

My philosophy would be to only provide 'simple' functions with parcels, and have users code up (and share?) their more complicated ones themselves. Since you can always initiate a particleset from a list of lons and lats, users can create this list using as complicated functions as they want...

Issue #148, opened 2017-01-17
Tests have undocumented matplotlib dependency


dham:
Running py.test in the top level directory results in examples fails which are caused by a failure to import matplotlib. It turns out matplotlib is not in `requirements.txt`.

erikvansebille:
Parcels should fail gracefully if matplotlib is not installed (it is only necessary for the plotting; Parcels should still be runnable without it).

What's the error message you're getting? On which command?

dham:
for example:

```python
Notebook execution failed
Cell 1: no description

Input:
from parcels import Grid, ParticleSet, JITParticle
from parcels import AdvectionRK4
from datetime import timedelta as delta
from scripts import plotTrajectoriesFile
import math

Traceback:

ImportErrorTraceback (most recent call last)
<ipython-input-2-0465ed1b62bb> in <module>()
      2 from parcels import AdvectionRK4
      3 from datetime import timedelta as delta
----> 4 from scripts import plotTrajectoriesFile
      5 import math

/Users/dham/src/parcels/src/parcels/scripts/__init__.py in <module>()
----> 1 from scripts.plotParticles import *  # NOQA get flake8 to ignore unused import.

/Users/dham/src/parcels/src/parcels/scripts/plotParticles.py in <module>()
      2 from netCDF4 import Dataset
      3 import numpy as np
----> 4 import matplotlib.pyplot as plt
      5 from argparse import ArgumentParser
      6 import matplotlib.animation as animation

ImportError: No module named matplotlib.pyplot
```

dham:
this is just by running `py.test` in the top parcels directory.

erikvansebille:
Fixed in #153

PR #153, opened 2017-01-18
Graceful failing if matplotlib or Basemap is not installed

erikvansebille:
Rather than throwing errors, Parcels now displays a message when matplotlib or Basemap are not installed. This means tests don’t fail on machines without these modules pre-installed

See also #148. Thanks to @dham for reporting this Issue

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -404,7 +404,8 @@ class Field(object):
             import matplotlib.animation as animation_plt
             from matplotlib import rc
         except:
-            raise RuntimeError("Visualisation not possible: matplotlib not found!")
+            print("Visualisation is not possible. Matplotlib not found.")
+            return
 
         if with_particles or (not animation):
             idx = self.time_index(show_time)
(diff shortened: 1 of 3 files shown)


parcels/field.py
```diff
@@ -404,7 +404,8 @@ def show(self, with_particles=False, animation=False, show_time=0, vmin=None, vm
             import matplotlib.animation as animation_plt
             from matplotlib import rc
         except:
-            raise RuntimeError("Visualisation not possible: matplotlib not found!")
+            print("Visualisation is not possible. Matplotlib not found.")
```
mlange05: I'm not sure dropping the `RuntimeError` here is the right thing to do. As far as I understand it this only breaks testing when invoked from the plotting script or the notebooks. So can we not use an explicit `try/except` to simply make these more robust, but leave the low-level errors in place?

erikvansebille: I see your point about `try/except`; but that would really make the notebooks less nice to read. Since in principle it's not an error when we can't plot, I think a print statement (ideally warning, see also Issue #158) works better and keeps the notebooks clean

So am merging now, and will then rework as Warnings in #158

PR #180, opened 2017-02-16
Fields3 d

erikvansebille:
A first implementation of 3D grids and particles. In this version of Parcels, particles have a depth and fields are always sampled as 3D. This works in both Scipy and JIT

Note that the only thing that isn't implemented is 3D velocities in the advection Kernels. 

I.e. the particles can change their depth only when instructed in Kernels other than the advection ones. 

There is one unit test in test_advection, with more tests to follow


mlange05:
I only had a quick look through this, but it looks great for what we discussed. So good to go from me.

PR #217, opened 2017-08-16
Setup: add include dir to install

guidov:
Instead of adding the source to the PYTHONPATH, using "python setup.py install" will add the include directory (with the parcels.h file) to the install. This is useful for conda installs and running the example notebooks from the conda installation.  This allows the notebook to find the .h file for the examples that use JIT compile.


mlange05:
Yes, looks sane to me. Many thanks @guidov for fixing this.

Merging...

Issue #225, opened 2017-11-01
How to install a stand-alone parcels package?


willirath:
## Why?

At Geomar, we have a central Jupyter Hub allowing for running remote kernels on virtually all of our analysis machines. The remote backends rely on Conda environments. Ideally, we'd be able to include a parcels environment satisfying the `requirements.txt` and adding a parcels package.

The basic logic behind our backend setup follows a very nice blog post by @rabernat: https://medium.com/@rabernat/custom-conda-environments-for-data-science-on-hpc-clusters-32d58c63aa95

The following env file (largely based on parcel's `requirements.txt`) seems to get close to a working parcels setup:

```yml
name: py2_parcels
channels:
  - https://conda.anaconda.org/conda-forge
  - https://repo.continuum.io/pkgs/free
dependencies:
  - python=2.7
  - cachetools>=1.0.0
  - cgen
  - enum34
  - flake8>=2.1.0
  - jupyter
  - matplotlib
  - netcdf4>=1.1.9
  - numpy>=1.9.1
  - progressbar2
  - py>=1.4.27
  - pymbolic
  - python-dateutil
  - scipy>=0.16.0
  - six>=1.10.0
  - xarray>=0.5.1
  - pip:
      - pytest-ipynb
      - pytest>=2.7.0
      - "git+https://github.com/OceanParcels/parcels.git@v0.9"
```

## Problem

Running the [parcels tutorial](http://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/examples/parcels_tutorial.ipynb) with essentially the following steps works only partially:

```bash
source activate py2_parcels
git clone https://github.com/OceanParcels/parcels.git parcels
cd parcels
git fetch --tags
git checkout v0.9
python parcels/scripts/pull_data.py
jupyter notebook
```

I do, however, run into problems with `gcc` being unable to compile, because of a missing `includes/parcels.h`. Import statements seem to use the installed version (as intended) while only the tutorial itself and the `pull_data.py` from the cloned version are used.

## Going forward

I'd be fine with providing a PR for a decent `setup.py` and / or documentation. Before being able to do so, I think, I'll have to understand how a user is supposed to work with parcels:
- Keeping a full copy of the repo for each experiment?
- Or having a central installation?

(Probably related to #104)
(cc: @sruehs)

rabernat:
Probably best to package parcels via conda forge.

willirath:
[...]

Issue #104, opened 2016-08-13
Update setup.py script


erikvansebille:
The `setup.py` script in the repo root is fairly outdated. Now that parcels is becoming bigger and installation is becoming more subtle, we should update this script. 

Things that `setup.py` should do
- check for proper versions of netcdf etc
- try install `matplotlib` and `Basemap` (and gracefully fail)
- `pip install requirements`
- run the `pull_data` script (with the option not to do this if on low-bandwidth, perhaps by asking users whether they want to download the auxiliary data)
- set the `PYTHONPATH` to include parcels

mlange05:
Ok, I think this needs a little clarification: `setup.py` is a standardised install script that is required by `pip` to install packages in the local Python environment. As such there is no need for it to set `PYTHONPATH` since it will install Parcels in a central location that Python always searches during any `import`.

As for the dependencies, the standardised `setuptools` utilities allow dependencies to be specified for `pip install` but convention states that the these should be **minimal**, ie. they should not include `matplotlib`, `Basemap` or `pull_data`. `requitements.txt` on the other hand is meant to fully define a kind of "best known" environment, so we probably should add `matplotlib` and `Basemap` here. More info can be found [here](https://packaging.python.org/requirements/).

So, I don't think the "do-it-all-out-of-the-box" install script you are describing here is what `setup.py` is meant to do. One way to do this would be to add a dedicated project-specific script like `parcels-install` that does all the additional sanity checking, like netCDF versions and pulling example data. To me this sounds a bit more like renaming and adding a few features to `pull_data`, and I'll look into that.

fischcheng:
It sounds to me like a install and test script to ensure all data are ready for the tutorials, so just like the last few lines in `.travis.yml`

rabernat:
This seems like the relevant place to repeat my comment from #225.

[Conda forge](https://conda-forge.org/) will likely solve your binary dependency and distributions problems. That's why conda forge was created.

willirath:
> Conda forge will likely solve your binary dependency and distributions problems. That's why conda forge was created.

I agree. From the discussion above and from what I tried in #225, I got the impression that at this stage, a parcels package in conda-forge might not make a lot of sense, though. Users might be more likely to work with the bleeding edge from master or even some commit from a PR.

For this (and from my PoV...), a combination of a `setup.py` and a conda-env heavily relying on conda-forge would be a good compromise.

erikvansebille:
[...]

PR #227, opened 2017-11-03
WIP: Update installation guide

willirath:
I start this PR to address #225 and #104 and to discuss and maybe settle on a way of documenting different ways of installing parcels. Currently, this PR only contains a hint on how to re-organize the "Installing Parcels" section of the documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
     - conda update --yes conda
 
 install:
-    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy hdf4 netcdf4 matplotlib basemap
+    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy hdf4 netcdf4 basemap matplotlib=2.0.2
     - pip install --upgrade pip
     - pip install -r requirements.txt
     - git submodule update --init --recursive
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,18 +6,18 @@
 Welcome to Parcels
 ===================================
 
-**Parcels** (**P**\ robably **A** **R**\ eally **C**\ omputationally **E**\ fficient **L**\ agrangian **S**\ imulator) is an experimental prototype code aimed at exploring novel approaches for Lagrangian tracking of virtual ocean particles in the petascale age. 
+**Parcels** (**P**\ robably **A** **R**\ eally **C**\ omputationally **E**\ fficient **L**\ agrangian **S**\ imulator) is an experimental prototype code aimed at exploring novel approaches for Lagrangian tracking of virtual ocean particles in the petascale age.
 
 Its code is licensed under an `open source MIT license <https://github.com/OceanParcels/parcels/blob/master/LICENSE.md>`_ and can be downloaded from https://github.com/OceanParcels/parcels.
 
 .. figure:: http://oceanparcels.org/animated-gifs/globcurrent_fullyseeded.gif
    :align: center
-   
+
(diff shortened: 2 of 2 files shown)


erikvansebille:
Thanks @willirath for starting this PR! I myself have also been working on a branch yesterday and today, based on your recommendations in #225. See https://github.com/OceanParcels/parcels/tree/install_via_conda

There were some other issues with Travis that needed to be fixed too, but we now have a working(?) version of the install script in `.travis.yml`

I'll merge with your suggestions for the `index.rst` and then hopefully this fixes #225 and #104?

willirath:
> There were some other issues with Travis that needed to be fixed too, but we now have a working(?) version of the install script in .travis.yml

The tests that failed here seem to be at least partly related to a matplotlib bug which I've circumvented by pinning `matplotlib=2.0.2`.

willirath:
> I'll merge with your suggestions for the index.rst and then hopefully this fixes #225 and #104?

This would (from a more outside perspecitve) solve #104 and #225. 

What's your opinion on the space all this should take in the documentation?  I'd go for a pretty detailed (but robust) description of the following scenarios:

1. How to install master with conda?

2. How to install with pure pip? (which involves making sure that there's a working ffmpeg, compiler, netCDF-lib, etc. on ones own)

3. How to pick a different Git reference? (like the HEAD of some other branch, fork, ...)

From 02e833a9c16a50f91ebddfba87109e61056b2b7f, I gather that you'd prefer a concise documentation, though.  Not sure if an elaborated  installation guide contradicts this.

Then, there still is the possibility of providing a conda-forge package. This adds some additional maintenance tasks though.

Let's just see whatever is working for you.  I'd be happy to get involved in any of the possible solutions.

willirath:
A few suggestions regarding <https://github.com/OceanParcels/parcels/tree/install_via_conda>:

I'd remove the explicit parcels installation from the `environment.yml`. (See below for why this makes any sense at all.)

<https://github.com/OceanParcels/parcels/blob/effc1c2a61c9e4e83471df11d05b26ab2e0c10a9/.travis.yml#L27-L30>

is problematic because it first installs a Python version determined by a Travis environment variable and then ignores it by setting up a completely new env.  I'd go for

```yaml
    - conda env create --file environment.yml
    - source activate py2_parcels
    - conda install --yes python=$TRAVIS_PYTHON_VERSION
    - pip install -e .
```

which first sets up the environment as defined in `environment.yml`, activates it and only then enforces whatever Python version Travis wants. The last step then installs Parcels for whatever Git ref the job is running for. (If we'd leave an explicit `pip install parcels` in the env file, we'd always get an env that's at least in part determined what the master branch is pointing at.)

As all the hdf- and netCDF dependencies are now satisfied via conda-forge, all of the following should be safe to be omitted:

<https://github.com/OceanParcels/parcels/blob/effc1c2a61c9e4e83471df11d05b26ab2e0c10a9/.travis.yml#L11-L16>

The following is obsolete as pip now takes care of wiring everything for parcels:

<https://github.com/OceanParcels/parcels/blob/effc1c2a61c9e4e83471df11d05b26ab2e0c10a9/.travis.yml#L38>

With these changes, an installation could be done by just running:

```bash
Parcels_ref=master # or whatever branch or commit you want to install
conda env create --file environment.yml
source activate py2_parcels
pip install git+https://github.com/OceanParcels/parcels@${Parcels_ref}
```

To install from a fork or from some other ref, a user just needs to change the last line of this snippet.

willirath:
> I'd remove the explicit parcels installation from the environment.yml. (See below for why this makes any sense at all.)

That is, remove: 
<https://github.com/OceanParcels/parcels/blob/effc1c2a61c9e4e83471df11d05b26ab2e0c10a9/environment.yml#L27>

willirath:
@erikvansebille Perhaps I should close this PR and send you one for the `install_via_conda` branch?

erikvansebille:
Thanks @willirath for all these suggestions on the branch. I've just created a PR from that (#228)

Would be good if you could merge your proposed changes to the `index.rst` (and any other input you have; you're clearly much more an expert on this than I am!) to that PR. Then we can close this PR?

willirath:
As #228 is dealing with this now, I'll close here.

Issue #226, opened 2017-11-02
Swapping pytest-ipynb for nbval


erikvansebille:
We're now using `pytest-ipynb` plugin to auto-test the tutorial notebooks. However, there seems to be a newer (better?) plugin at https://github.com/computationalmodelling/nbval that might be more appropriate for what we need. For one, it seems the restrictions on file naming are loosened in `nbval`. I suggest looking into whether we want to change to this new plugin

mlange05:
:+1:

erikvansebille:
Fixed in #228

PR #228, opened 2017-11-03
Install via conda

### LINKED ISSUE #225: How to install a stand-alone parcels package?

erikvansebille:
Changing the installation scripts and notes to install via `conda-forge`, as suggested by @willirath and @rabernat. 
Aim is to fix #225 and #104, and also #226 as that was playing havoc on Travis


willirath:
Here's my suggestion for the “Installing Parcels” section of the documentation:

```rst
Installing Parcels
===================================

Parcels depends on a working Python installation, a netCDF installation, a C
compiler, and various Python packages.  These requirements can be met in
different ways:

1. Use Anaconda and Conda-Forge for all requirements.  This is the
   “batteries-included” solution probably suitable for most users.

2. Use an existing netCDF installation and an existing C compiler and just get
   the Python part of the dependencies.  This may be a suitable solution for
   more experienced users who want to control the details.


Installing Parcels and all dependencies with Anaconda and Conda-Forge
---------------------------------------------------------------------

Installing Anaconda’s Miniconda
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

(If you already have a working Anaconda installation, skip to “Creating an
environment with Parcels”.)

Download Anaconda’s Miniconda for Linux::

    wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh \
        -O miniconda.sh

or for MacOSX::

    wget https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh \
        -O miniconda.sh

and install with::

    bash miniconda.sh -b -p $HOME/miniconda


Creating an environment with Parcels
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

With a file `environment.yml`::

    name: py2_parcels
    channels:
      - https://conda.anaconda.org/conda-forge
      - https://repo.continuum.io/pkgs/free
    dependencies:
      - python=2.7
      - basemap
      - cachetools>=1.0.0
      - cgen
      - enum34
      - ffmpeg
      - flake8>=2.1.0
      - gcc
      - jupyter
      - matplotlib=2.0.2
      - netcdf4>=1.1.9
      - numpy>=1.9.1
      - progressbar2
      - py>=1.4.27
      - pymbolic
      - python-dateutil
      - scipy>=0.16.0
      - six>=1.10.0
      - xarray>=0.5.1
      - pip:
          - pytest>=2.7.0
          - nbval
          - "git+https://github.com/OceanParcels/parcels.git@master"

run::

    source $HOME/miniconda/bin/activate root
    conda env create -f environment.yml


Using the Parcels environment
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

To start working with Parcels, activate the environment with::

    source $HOME/miniconda/bin/activate py2_parcels


Just installing the python requirements
---------------------------------------

The latest version of Parcels, including tests and examples,
can be obtained directly from github via::

    git clone https://github.com/OceanParcels/parcels.git
    cd parcels
    git checkout master
    pip install -r requirements.txt
    pip install -e .

Note that a functional NetCDF installation and a working C compiler is required.
```

This is rather lengthy and still needs some work to make sure it really reflects the state of the repository (is there / should there still be a `requirements.txt`? etc.).  But maybe it's a good start to discuss how much and how to document the installation process.

environment.yml
```diff
@@ -0,0 +1,26 @@
+name: py2_parcels
+channels:
+  - https://conda.anaconda.org/conda-forge
+  - https://repo.continuum.io/pkgs/free
+dependencies:
+  - python=2.7
+  - cachetools>=1.0.0
```
willirath: I've seen `basemap` mentioned a few times. Should it be included as a requirement here?

erikvansebille: We're not enforcing `Basemap`, because on many HPC systems it's not available.  Parcels still runs without `Basemap` installed; users just won't be able to plot continents etc

erikvansebille:
Thanks @willirath for your extended proposal for new Installation guide. I think it looks good in many ways, although I would change a few things:
1. Refer simply to https://conda.io/docs/user-guide/install/ on how to install Miniconda. Advantage is that this might (?) even work on Windows?
2. Simply link to the `environment.yml` file on the master branch of the Parcels repo, rather then spell it out completely
3. Perhaps not even mentioning your option 2 of just installing the python environment? Or, if we mention it, somehow use the same `environment.yml` file? So that we don't have redundancy of both an `environment.yml` file and a `requirements.txt`. Is that even possible?

willirath:
> Refer simply to https://conda.io/docs/user-guide/install/ on how to install Miniconda. Advantage is that this might (?) even work on Windows?

Yes, that's good.  (On Windows, there's no gcc package. So I'm not sure this would lead anywhere.)

> Simply link to the environment.yml file on the master branch of the Parcels repo, rather then spell it out completely

Yes, I also like that.

> Perhaps not even mentioning your option 2 of just installing the python environment? Or, if we mention it, somehow use the same environment.yml file? So that we don't have redundancy of both an environment.yml file and a requirements.txt. Is that even possible?

Yes, option 2 could be dropped entirely. (It's probably safe to assume that anyone who insists on maintaining an own installation of any of the dependencies can figure this out themselves.

I'm not sure about deleting the `requirements.txt` though.

docs/index.rst
```diff
@@ -72,18 +72,31 @@ The best way to get started with Parcels is to have a look at the Jupyter notebo
...
+
+Installing Parcels and all dependencies with Anaconda and Conda-Forge
+-----------------------------------
+
+Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/
+
+Download the `environment.yml <https://raw.githubusercontent.com/OceanParcels/parcels/install_via_conda/environment.yml>`_ file and place it in the directory where you want to install Parcels. Then, run::
+
+    source $HOME/miniconda/bin/activate root
+    conda env create -f environment.yml
```
willirath: I'd use
```bash
conda env create -f environment.yml -n py2_parcels
```
which is redundant but works (`-n` actually overrides the `name:` given in the env file) and makes it obvious why the user is alter asked to activate an env with this name.

docs/index.rst
```diff
@@ -72,18 +72,31 @@ The best way to get started with Parcels is to have a look at the Jupyter notebo
...
+2. Use an existing netCDF installation and an existing C compiler and just get
+   the Python part of the dependencies.  This may be a suitable solution for
+   more experienced users who want to control the details.
+
+Installing Parcels and all dependencies with Anaconda and Conda-Forge
+-----------------------------------
+
+Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/
+
+Download the `environment.yml <https://raw.githubusercontent.com/OceanParcels/parcels/install_via_conda/environment.yml>`_ file and place it in the directory where you want to install Parcels. Then, run::
```
willirath: I think the "place it in the directory where you want to install Parcels" is misleading as the `environment.yml` only has to be present while the environment is created. Would it be still understandable, if the whole thing would be truncated to
> Download the ... file and run::

Another thing to keep in mind: The link should point to
<https://raw.githubusercontent.com/OceanParcels/parcels/master/environment.yml>
in the end.

docs/index.rst
willirath: Another thing:  The installation is only completed after using pip to install from the `master` branch of Parcels.  So the whole code block is:
```bash
source $HOME/miniconda/bin/activate root
conda env create -f environment.yml -n py2_parcels
pip install git+https://github.com/OceanParcels/parcels.git@master
```

environment.yml
```diff
@@ -0,0 +1,26 @@
+name: py2_parcels
+channels:
+  - https://conda.anaconda.org/conda-forge
+  - https://repo.continuum.io/pkgs/free
```
willirath: The channels field can be reduced to 
```yaml
channels:
- conda-forge
```

The second line is the default fall-back channel for anything not found in explicitly listed channels anyway. And the explicit URLs might change in the future while the short-form `conda-forge` will just refer to whatever is the current conda-forge-repo URL. (I tend to put full URLs there, because I have some setups behind firewalls where I have to point this to a mirror.)

docs/index.rst
erikvansebille: Good point. I've now also added a check of broken files in the documentation using the `sphinx linkcheck` functionality. So even if we forget, travis will complain

docs/index.rst
```diff
@@ -78,7 +78,7 @@ compiler, and various Python packages.
 The simplest way to install Parcels is to use Anaconda and Conda-Forge for all 
 requirements. This is the “batteries-included” solution probably suitable for most users.
 
-1. Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/
+1. Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/. Be sure to download the Python 2.7 version.
```
willirath: I'd rephrase this to:

> All the code below assumes that you download the Python-2 version.

Here's why: The Python version for the Parcels environment is specified in the `environment.yml`. Conda is perfectly fine with different Python versions in different envs. I usually have a Miniconda3 root env with different Python-2 and Python-3 environments for different use cases.

docs/index.rst
```diff
@@ -69,26 +69,45 @@ The best way to get started with Parcels is to have a look at the Jupyter notebo
...
-    git clone https://github.com/OceanParcels/parcels.git
-    cd parcels; pip install -r requirements.txt
-    python scripts/pull_data.py
-    export PYTHONPATH="$PYTHONPATH:$PWD"
+5. The next time you log in to your machine and want to work with Parcels, activate the environment with::
 
-In order for Parcels to work from any directory, add the following line to 
-your ~/.bash_profile::
-    export PYTHONPATH="$PYTHONPATH:$PWD"
+    source $HOME/miniconda2/bin/activate py2_parcels
```
erikvansebille: @willirath is there a reason you suggested to `source $HOME/miniconda2/bin/activate py2_parcels` whereas in most conda documentation and tutorials people use `source activate py2_parcels`?

willirath: The short version needs the Conda's `activate` executable in the Path.  During the Miniconda installation, the users have to decide if they want to add the installation to the Path. It is debatable whether this is a good idea or not.  (I prefer not to.)

The long version which explicitly includes the full path to `activate`, is working in any case.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E501,F403,E226,W503
+ignore = E501,F403,E226,E722,W503
```
delandmeterp: Should we also ignore error E731?
It makes the following error in another branch
"test_fieldset.py:186:5: E731 do not assign a lambda expression, use a def"

docs/index.rst
delandmeterp: Ok with the source comment, but you changed here the meaning of the sentence.
It was already stated that user has to source the activate file (line 87).
The point here is to say that the source can be written in the .bash_profile

erikvansebille: Well, but do we really want people to source the Parcels environment in their `.bash_rc`? That goes against the idea of conda environments, where you only source them if you need them. So I prefer not to suggest adding to `.bash_rc`

setup.cfg
erikvansebille: I didn't realise that using lambda functions is against PEP8, but actually that does make sense. Why not use a `def` indeed, for readability, in PR #228?

docs/index.rst
delandmeterp: Ok, you're right

setup.cfg
delandmeterp: I'll do the change in the other branch.

PR #232, opened 2017-11-10
parcels-windows-support

Simnator101:
# Implementing Parcels on a Windows Machine
The current master branch runs into a few problems when it is run on a Windows machine with the particles in JIT-mode. This pull request serves to solve these problems which are discussed in more detail here below.

## The Problems
1.  The original GCC compiler code, as was written in `compiler.py`, contained a small typo which made it impossible for the compiler to find the `parcels.h` header in the Windows directory system.
2. Dynamic Linked Libraries (DLL's) are somewhat more precocious on Windows systems with Python than on Linux and MacOS systems. On Windows, DLL's cannot be overwritten if there are yet programs open which have an active program handle to the library, therefore, after a `ParticleSet` is done with a Kernel object, the `Kernel` must 'properly' dispose of its connection to the DLL. This is necessary when one wishes to run same Kernel with different `ParticleSet`'s in the current Python thread.
3. Step two frees up the library to be overwritten by a another Kernel object with the same hash code. We are, however, not yet in the clear: for it seems that Python and Windows have another peculiarity when it concerns loading DLL's. When a new DLL has been compiled with the same name as the previous library, something strange happens. Python seems not to be able to access the new function's `particle_loop`. I cannot say what exactly happens here, but I think that Python has not properly unloaded the previous DLL and tries to execute the C-function at the wrong memory address. 
4. Lastly the `parcels.h` was modded with some preprocessor checks. Nothing was really wrong here, but it is considered good practice to add header defines to prevent possible double inclusions.

## How this Pull Solves these Problems
1. This is done by using native Python methods to construct the correct directory where the `parcels.h` code is located.
2. Here was introduced a destructor function and an `unload_lib` function for the `Kernel` class wherewith Python tries to remove the DLL from its system. These two functions are called respectively at the moment of class instance destruction and when the `Kernel` is asked to recompile the JIT code.
3. This is done by adding to the destructor and `unload_lib` function code which changes the hash file name of the kernel. The hash is changed by appending a POSIX timestamp to the hash calculation. To not clutter up the directory wherein the DLL are saved, sanitation code is added to remove the old logs, C-files and DLL's. 
4. For the last fix a header definition was added and a check to force C-code compilation.

## Comments on the Prerequisites for Running Parcels on Windows
This instance of parcels in able to run freely on Windows, however, a few things are assumed which the user must have on his Windows machine:
- The C-compiler gcc which can be accessed from the command line (cmd.exe). To this end I recommend getting a copy of [MinGW](http://www.mingw.org/). Please note that you probably need to modify you PATH variable to use gcc from the command line.
- A working copy of Python (version 2) with the modules required by parcels. This seems rather obvious, but a copy of an unmodified version from the Python organisation has some problems installing Numpy and SciPy on Windows. It is firstly advised to simply follow the instructions on the website [OceanParcels](http://oceanparcels.org/). If that does not work, however, try installing the Numpy and SciPy manually from [this](https://www.lfd.uci.edu/~gohlke/pythonlibs/) unofficial repository, and then continue by using the command `python -m pip install requirements.txt` in the parcels' directory.


parcels/kernel.py
```diff
@@ -96,24 +99,51 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
...
+            self.log_file = "%s.log" % basename
+
     def compile(self, compiler):
         """ Writes kernel code to file and compiles it."""
         with open(self.src_file, 'w') as f:
             f.write(self.ccode)
         compiler.compile(self.src_file, self.lib_file, self.log_file)
         logger.info("Compiled %s ==> %s" % (self.name, self.lib_file))
 
+
```
erikvansebille: Two whitelines here gives PEP8 error; within class definition functions are separated by one whiteline only

parcels/kernel.py
```diff
@@ -96,24 +99,51 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
...
-            self.lib_file = "%s.so" % basename
+            self.lib_file = "%s.%s" % (basename, 'dll' if platform == 'win32' else 'so')
             self.log_file = "%s.log" % basename
         self._lib = None
 
+    def __del__(self):
+        # Clean-up the in-memory dynamic linked libraries.
+        # This is not really necessary, as these programs are not that large, but with the new random
+        # naming scheme which is required on Windows OS'es to deal with updates to a Parcels' kernel.
+        if not self._lib == None:
```
erikvansebille: In Parcels, we're following (most) of the PEP8 formatting rules. This is what flake8 checks on Travis, and it keeps our code clean and easy-to-read. According to https://travis-ci.org/OceanParcels/parcels/builds/299949591?utm_source=github_status&utm_medium=notification, there are a few minor PEP8 violations, including here, where it's better to write `if self._lib is not None`

parcels/kernel.py
```diff
@@ -96,24 +99,51 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
...
     def _cache_key(self):
         field_keys = "-".join(["%s:%s" % (name, field.units.__class__.__name__)
                                for name, field in self.field_args.items()])
-        key = self.name + self.ptype._cache_key + field_keys
+        key = self.name + self.ptype._cache_key + field_keys + ('TIME:%f' % time.time())
         return md5(key.encode('utf-8')).hexdigest()
 
+    def remove_lib(self):
+        # Unload the currently loaded dynamic linked library to be secure
+        if not self._lib == None:
```
erikvansebille: Again, PEP8 suggests writing as `is self._lib is not None`

parcels/kernel.py
```diff
@@ -96,24 +99,51 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
...
+            _ctypes.FreeLibrary(self._lib._handle) if platform == 'win32' else _ctypes.dlclose(elf._lib._handle)
```
erikvansebille: Type here? Do you mean `_ctypes.dlclose(self._lib._handle)` (so with `self` instead of `elf`)?

parcels/kernel.py
```diff
@@ -96,24 +99,51 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
...
+        # Clean-up the in-memory dynamic linked libraries.
+        # This is not really necessary, as these programs are not that large, but with the new random
+        # naming scheme which is required on Windows OS'es to deal with updates to a Parcels' kernel.
+        if not self._lib == None:
+            _ctypes.FreeLibrary(self._lib._handle) if platform == 'win32' else _ctypes.dlclose(self._lib._handle)
+            del self._lib
+            self._lib = None
+            map(remove, [self.src_file, self.lib_file, self.log_file]) if path.isfile(self.lib_file) else None
+
+
```
erikvansebille: Two whitelines here gives PEP8 error; within class definition functions are separated by one whiteline only

parcels/kernel.py
```diff
@@ -4,9 +4,12 @@
 from parcels.field import FieldSamplingError
 from parcels.loggers import logger
 from parcels.kernels.advection import AdvectionRK4_3D
-from os import path
+from os import path, remove
 import numpy.ctypeslib as npct
+import time
 from ctypes import c_int, c_float, c_double, c_void_p, byref
+import _ctypes
```
erikvansebille: What's the reason for now importing all classes of `_ctypes`? Which ones were we missing before (`c_int, c_float, c_double, c_void_p, byref`)? Can't we just add the extra classes required by Windows (`FreeLibrary`, `dlclose`)?

Simnator101:
Thanks for the pointers! I have fixed the formatting in the `parcels\kernel.py` file, so that should now be okay. Next, concerning the `import _ctypes` statement, this is necessary because `_ctypes` does not equal `ctypes` — mind here the underscore prefixing the module. What the former does and the latter does not, is providing the user with functions to unload a DLL from Python. That the whole module has to be loaded, has got to do with the fact that `_ctypes` exposes platform dependant functions, i.e., the `FreeLibrary` on Windows machines and `dlclose` on a Linux/MacOS system. These functions are simply not provided in the module on platforms besides those on which they are available: that is why I added an import for the whole module.

erikvansebille:
OK, your point about the _ctypes is understood then. We would still need some installation instructions. I can add these if you prefer?

Simnator101:
Ah, you wanted me to add these to the file `docs/index.rst`; of course I can add these. Concerning the question whether travis can test for Windows, I do not know the answer to this question.

willirath:
Windows builds can be tested with Appveyor. This could be a good starting point: https://github.com/ogrisel/python-appveyor-demo

docs/index.rst
```diff
@@ -85,6 +85,24 @@ your ~/.bash_profile::
...
+
+To grab the latest version of Parcels with tests and examples, you can follow these steps
+::
+
+   git clone https://github.com/OceanParcels/parcels.git
+   cd parcels
+   pip install -r requirements.txt
+   python scripts/pull_data.py
+   set "PYTHONPATH=%PYTHONPATH%;%cd%"
+
```
erikvansebille: I suggest to follow the new conda-forge install instructions, as per macOS and Linus. Is there any reason why that workflow wouldn't work?

docs/index.rst
```diff
@@ -85,6 +85,24 @@ your ~/.bash_profile::
...
+* A working distribution of gcc must be installed. With working is meant that gcc an be accessed from the commandline. We suggest Windows users to install the `MinGW Package <www.mingw.org>`_. Do note that to access the gcc compiler which MinGW provides, the user must most likely add the directory path to the 'MinGW\bin' folder where gcc resides.
+* A version of python with a working copy of Numpy and SciPy. We recommend you to install a scientific Python distribution, like `anaconda <https://anaconda.org/anaconda/python>`_, which provides the easier way to install Numpy and SciPy. However, it is also possible to use a version of Python distributed by the Python organisation (make sure you grab the version 2 branch of Python). After installing Python you can install the packages manually by downloading the appropriate versions of these packages from this repository of `pre-built Windows installers <https://www.lfd.uci.edu/~gohlke/pythonlibs/>`_. Use the command: python -m pip install 'package' to install these modules.
+
```
erikvansebille: I would only mention the miniconda option. I think that we can assume that users who want to run Parcels on Windows will not want the added hassle of having to download their own scipy/numpy etc?

docs/index.rst
```diff
@@ -85,6 +85,24 @@ your ~/.bash_profile::
 
 Note that a functional NetCDF install is required.
 
+Windows specific installation instructions
+------------------------------------------
+
+To enable Parcels to run on a Windows machine, some extra steps have to be taken to install the necessary prerequisite files. Windows must have the following software before it is able to run Parcels:
+* A working distribution of gcc must be installed. With working is meant that gcc an be accessed from the commandline. We suggest Windows users to install the `MinGW Package <www.mingw.org>`_. Do note that to access the gcc compiler which MinGW provides, the user must most likely add the directory path to the 'MinGW\bin' folder where gcc resides.
```
erikvansebille: I would simply remove the sentence `With working is meant that gcc an be accessed from the commandline.` 

And can we provide a link to a site that gives instructions on how to set the PATH?

docs/index.rst
```diff
@@ -85,6 +85,24 @@ your ~/.bash_profile::
...
+To grab the latest version of Parcels with tests and examples, you can follow these steps
+::
+
+   git clone https://github.com/OceanParcels/parcels.git
+   cd parcels
+   pip install -r requirements.txt
+   python scripts/pull_data.py
+   set "PYTHONPATH=%PYTHONPATH%;%cd%"
+
+Python, however, cannot yet access Parcels in each session. To add the directory where Parcels resides as a permanent module directory to Python, add or set the 'PYTHONPATH' variable in your system's environmental variables.
```
erikvansebille: This should not be an issue anymore with the conda install method?

docs/index.rst
Simnator101: Well I also thought so, but it seems that the `index.rst` when I forked the main branch was not up-to-date. Therefore, I'll migrate my instructions to the newer version of `index.rst` in the main branch.

Simnator101:
I've updated te `docs/index.rst` file to correctly make use of the new conda install instructions. Furthermore, I've also worked in the other suggestions for what ought to be told under the Windows install guide.

Issue #233, opened 2017-11-10
Simplify steps from installation to running the examples.


willirath:
I'd like to add a `parcels_get_examples` command that
- is installed along with Parcels,
- knows where to get the examples (for the currently installed version of Parcels),
- knows where to get the example data (also for the currently installed version of Parcels), and
- puts everything in to a directory the user can pick.

Technically, this could be done by adding an entry point to the `setup.py` which puts a script to the `$PATH` that does all this.

The example notebooks can be distributed together with Parcels, installing them in whereever pip, conda or the user installs Parcels, and have the proposed `parcels_get_examples` command figure out the paths.

I am not sure, however, how to deal with the example data. This essentially boils down to the following questions:
1. How stable are the example data files?
2. How big are the example data?

Question 1 is about whether there  should be a way to make sure an example will always find exactly the data files to work with.

Question 2 is about whether it would be feasible to package the example data together with Parcels.

What do you think?

willirath:
> How big are the example data?
> [...]
> Question 2 is about whether it would be feasible to package the example data together with Parcels.

After running `pull_data.py`, my `examples/` is about 65MB. There's definitely conda-forge packages which are much bigger than that. So packaging example data together with Parcels would be possible.

erikvansebille:
Thanks @willirath, a good idea yes! 

Indeed, the data are approx 65MB or so, although if needed we could make them even smaller. 
- The GlobCurrent data is now a year, whereas the `example_globcurrent.py` only requires a few weeks of data.
- The MovingEddies is 25 days worth of data, but again could be shortened if needed. 
- The OFAM data is only used in `example_ofam.py`, which to be honest does not add much relative to e.g. `example_globcurrent.py`, so could be removed altogether if necessary 

So we might be able to get below 30MB?

As for the stability of examples, yes these are not going to change in the foreseeable future. More examples might be added, although I think the example data provided now spans the range of possible fields (stationary simple for Peninsula, time-evolving simple for MovingEddies, OGCM for GlobCurrent)

willirath:
65MB is not too big. And perhaps it's good to have some head-room for the user to expand the examples?

Including the examples in a package needs a tar file with all the examples. This could probably be provided on <http://oceanparcels.org>.

[...]

PR #235, opened 2017-11-11
Simplify getting examples

willirath:
This will become a PoC for #233.


willirath:
Note that a689e531c5c791dae5786cca9fe8f9d224240e53 is moving `include/` and  `examples/` to within the `parcels/` directory.

willirath:
Test are passing and everything works as intended. @erikvansebille can you have a look?

Still todo:
- [ ] Update documentation
- [ ] Check if this works in Windows (← @Simnator101 ?)
- [ ] Discuss if this is a good way to handle the data

willirath:
(This also does not account for the latest windows-related fixes that already are in `master`. I'll later rebase this branch and cleanly apply it on top of the latest state of `master`._

willirath:
The failing test is `tests/test_fieldset.py::test_moving_eddies_file_subsettime[indstime0]`.

@erikvansebille There seems to be new data in <http://oceanparcels.org/examples-data/MovingEddies_data/>?

erikvansebille:
Yes, sorry. I did that this morning. Not aware it would break the tests. Am
in a meeting now, will fix ASAP

erikvansebille:
@willirath, can you change line 102 in the `tests/test_fieldset.py` file to

```
@pytest.mark.parametrize('indstime', [range(2, 8), [4]])
```
That should fix the Travis error

(note that this change was related to [a branch I'm working on](https://github.com/OceanParcels/parcels/tree/shorten_example_runtimes) that shortens the CI runtime. A lot of the tests we do take unnecessarily long, which hinders rapid development)

willirath:
@erikvansebille I think this is a data point in favor of keeping track of the example data in a github repo. I've been playing around with this just now: <https://github.com/willirath/pde> tracks the current version of <http://oceanparcels.org/examples-data/> using Git LFS for the `.nc` files.

The problem is that the archives which GitHub automatically generates for a release only include the pointers and not the GIt LFS objects themselves: https://github.com/git-lfs/git-lfs/issues/1322

Having the data in a repo would, however, greatly simplify the evolution of the data files.

erikvansebille:
I see your point @willirath; indeed the fact that tests failed because I changed the example-data isn't how it should be. 

The alternative of course would be to add all the data to its own repo directly. The downside is that this eats into our github storage, which is why we didn't do it earlier. However, since we don't expect this data to change much anymore (?) perhaps we can now add it all?

@mlange05, we had discussions about this issue of where to store example-data before. What are your thoughts?

.travis.yml
```diff
@@ -30,8 +30,8 @@ before_script: # configure a headless display to test plot generation
 script:
     - flake8 parcels
     - flake8 tests
-    - flake8 examples
-    - python parcels/scripts/pull_data.py
+    - flake8 parcels/examples
```
erikvansebille: What is the need/advantage of putting the `examples` directory into the `parcels` directory? The `tests` directory isn't there either, which is why originally we also had the `examples` directory in the root of the repo

parcels/scripts/get_examples.py
```diff
@@ -0,0 +1,110 @@
...
+    copy_data_and_examples_from_package_to(target_path)
+
+    # try downloading remaining files
+    remaining_example_data_files = _still_to_download(
+        example_data_files, target_path)
+    download_files(example_data_url, remaining_example_data_files, target_path)
+
+
+if __name__ == "__main__":
+    main()
```
erikvansebille: How is this `parcels_get_examples` script different from the `pull_data.py` script? Do we need them both? Or does this replace the latter? Can we duplicate some functionality?

docs/index.rst
```diff
@@ -91,35 +91,37 @@ requirements. This is the “batteries-included” solution probably suitable fo
 
     python example_peninsula.py --fieldset 100 100
 
-4. Optionally, if you want to run all the examples and tutorials, you may want to download the example hydrodynamic fields by downloading the `pull_data.py <https://raw.githubusercontent.com/OceanParcels/parcels/master/parcels/scripts/pull_data.py>`_ file and then run::
+4. Optionally, if you want to run all the examples and tutorials, you can get tutorials and example data by running::
 
-    python pull_data.py
+    parcels_get_examples $HOME/parcels_first_steps
```
erikvansebille: Assuming users are doing all this from their $HOME directory? I reckon that most users would actually first create a different directory before they install Parcels, from which they run the commands? Can't we simply say?
```
parcels_get_examples parcels_first_steps
```
And I'm not entirely sure about the `first_steps` thing; sounds a bit too much like it's a baby-toy?? ;-)

.travis.yml
willirath: Let's discuss why I moved `include/` to `parcels/include/` (because you probably can see the problem on your own computer already):

<https://github.com/OceanParcels/parcels/blob/9b52f1f79dabd7d616adf9f2bfbceaa40e3d1f64/setup.py#L13> will install a _package_ called `include` and a package called `parcels`.  The resulting directory structure is

```bash
tree miniconda2/envs/py2_parcels/lib/python2.7/site-packages/{include/,parcels/}
```
```
miniconda2/envs/py2_parcels/lib/python2.7/site-packages/include
└── parcels.h
miniconda2/envs/py2_parcels/lib/python2.7/site-packages/parcels
├── codegenerator.py
├── codegenerator.pyc
├── compiler.py
├── compiler.pyc
├── field.py
├── field.pyc
├── fieldset.py
├── fieldset.pyc
├── __init__.py
├── __init__.pyc
├── kernel.py
├── kernel.pyc
├── kernels
│   ├── advection.py
│   ├── advection.pyc
│   ├── error.py
│   ├── error.pyc
│   ├── __init__.py
│   └── __init__.pyc
├── loggers.py
├── loggers.pyc
├── particlefile.py
├── particlefile.pyc
├── particle.py
├── particle.pyc
├── particleset.py
├── particleset.pyc
├── rng.py
├── rng.pyc
└── scripts
    ├── convert_IndexedOutputToArray.py
    ├── convert_IndexedOutputToArray.pyc
    ├── __init__.py
    ├── __init__.pyc
    ├── plotParticles.py
    ├── plotParticles.pyc
    ├── pull_data.py
    └── pull_data.pyc
```

But we want `include/` somewhere below `miniconda2/envs/py2_parcels/lib/python2.7/site-packages/parcels/`.  The same is true for `examples/` as soon as it is distributed with the package.

This was not a problem when setting `$PYTHONPATH` directly, because the `parcels.h` actually  was in `parcels/../include/`.

With <https://github.com/OceanParcels/parcels/pull/235/files#diff-2eeaed663bd0d25b7e608891384b7298R16>, I include, both, `include/` and `examples/` as `package_data` belonging to `parcels` rather than as a separate package.

docs/index.rst
willirath: Yes, this an be simplified. <https://github.com/OceanParcels/parcels/pull/235/files#diff-ff3788e5426d53da20e4414d0ad3dbbdR94> will also work with a relative path.

And maybe let's change `parcels_first_steps` to `parcels_examples`?

parcels/scripts/get_examples.py
willirath: `parcels_get_examples` (or `parcels/scripts/get_examples.py`) would make `pull_data.py` obsolete.

willirath:
With https://github.com/willirath/parcels/blob/1b74b40438fc108773b8f6c14a1c887218cf9a80/.travis.yml#L30-L43, Travis runs three concurrent tests:
1. LInkting with `flake`
2. Tests from `tests/` and examples and tutorials from `parcels/examples/`
3. Link checking. (This one is allowed to fail.)

willirath:
@erikvansebille @Simnator101 @mlange05 

I think this PR is in a state that I'd consider ready to be included in `master`.  Can you have a look?

parcels/compiler.py
```diff
@@ -10,7 +10,7 @@ def getuid():
 
 
 def get_package_dir():
-    return path.abspath(path.join(path.dirname(__file__), path.pardir))
+    return path.abspath(path.join(path.dirname(__file__)))
```
erikvansebille: What exactly does this do? Why is this necessary now? Keen to understand it's workings better

.travis.yml
```diff
@@ -27,11 +27,17 @@ before_script: # configure a headless display to test plot generation
...
-    - flake8 parcels
-    - flake8 tests
-    - flake8 examples
-    - python parcels/scripts/pull_data.py
-    - py.test -v -s tests/
-    - py.test -v -s --nbval-lax examples/
-    - make linkcheck -C docs
+jobs:
+    include:
+        - script:
```
erikvansebille: Why separate these out into different scripts? The flake8 and linkcheck take no time compared to the unit tests in examples and parcels. But every script triggers a new conda-build, which seems a waste of resources? Unless there's a very good reason, I would keep just one script

parcels/compiler.py
willirath: - `path.pardir` is whatever indicates "parent directory". In a Unix / Posix world, this is `..`.
- `path.dirname` returns the directory containing `__file__`.
- `__file__` is the full path (incl. potential symlinks etc.) of the file containing this line of code.
- `path.join` joins it inputs to a properly qualified path, taking care of extra `/` (Posix) or `\` (Windows) etc.

My change basically means (in the context of <https://github.com/willirath/parcels/blob/1b74b40438fc108773b8f6c14a1c887218cf9a80/parcels/compiler.py#L66> which is the only time it is used): "Look for `include/` in the same directory as `compiler.py` and not one level above."

And I note that I should drop the `path.join` which does not do anything here.

.travis.yml
```diff
@@ -27,11 +27,17 @@ before_script: # configure a headless display to test plot generation
...
+            - flake8 tests
+            - flake8 parcels/examples
+        - script:
+            - parcels_get_examples examples/
+            - py.test -v -s tests/
+            - py.test -v -s --nbval-lax examples/
+        - script:
+            - make linkcheck -C docs
+          env:
+            - 'ALLOW_FAIURE=true'
```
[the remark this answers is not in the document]
willirath: That's a fair point. I'll change it.

parcels/scripts/get_examples.py
```diff
@@ -0,0 +1,110 @@
+"""Get example scripts, notebooks, and data files."""
+
+import argparse
+from datetime import datetime, timedelta
+import os
+import pkg_resources
+from progressbar import ProgressBar
+try:
+    # For Python 3.0 and later
+    from urllib.request import urlopen
```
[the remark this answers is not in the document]
willirath: That's a copy-and-paste artifact from the original `pull_data.py`.  I should clean it up.

> In principle we don't aim to be Python3-compliant in Parcel.

In the long run, this may become a problem as people are moving away from "Python 2": <http://www.python3statement.org/>. (IPython, for example, already stopped implementing new features in Python 2. Bug-fixes are still backported, though.)

.travis.yml
```diff
@@ -27,11 +27,17 @@ before_script: # configure a headless display to test plot generation
...
-    - python parcels/scripts/pull_data.py
-    - py.test -v -s tests/
-    - py.test -v -s --nbval-lax examples/
-    - make linkcheck -C docs
+jobs:
+    include:
+        - script:
+            - flake8 parcels
+            - flake8 tests
+            - flake8 parcels/examples
```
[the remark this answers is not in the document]
willirath: > However, then it doesn't really make sense to keep tests outside the parcels directory. Why not put that one in too?

I don't know. As `tests/` are not meant to be distributed together with Parcels, I'd leave them where they are. (I must admit, that I don't know whether there's any proper rationale for that or if I just think this is how it should be done because I've seen it elsewhere.)

> And isn't flake8 recursive? So is flake8 parcels/examples not redundant if we already do flake8 parcels?

That's right. I'll change it.

.travis.yml
willirath: I did this, because it made it easier to locate the failing test. But I do see the point of keeping things in one script.  Should I revert this?

erikvansebille: Yes please just use one script. Let's try to limit our travis resources as much as possible ;-)

.travis.yml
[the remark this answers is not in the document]
erikvansebille: @mlange05 and @delandmeterp; any thought on whether `tests` should then also go inside the `parcels` directory or not?

parcels/scripts/get_examples.py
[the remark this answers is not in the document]
erikvansebille: OK fine. I suggest to just leave it here then. At some point we do want to move to Python3, and having this here then helps conversion for sure!

.travis.yml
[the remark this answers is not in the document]
willirath: I've noticed another problem: Currently, <https://github.com/OceanParcels/parcels/blob/467b25b110ee3bb34e4eb127586a7e7a24369bc1/tests/test_fieldset.py#L104> and <https://github.com/OceanParcels/parcels/blob/467b25b110ee3bb34e4eb127586a7e7a24369bc1/tests/test_scripts.py#L10> expect example data in `tests/../examples/`. This is not a problem, because I happened to call `parcels_get_examples examples` in Travis.  But in general, this is a problem.

willirath:
Windows tests are running here: <https://ci.appveyor.com/project/willirath/parcels/build/1.0.10>

(Sorry for the messy history.  I think I should rebase and squash this into a set of decent commits before it is merged.)

willirath:
> Windows tests are running here: https://ci.appveyor.com/project/willirath/parcels/build/1.0.10

And finished successfully. :)

willirath:
@erikvansebille You can probably just sign up on https://appveyor.com and follow instructions just as you did for Travis.  Then, Windows builds will be triggered for every branch that has the relevant configuration file `appveyor.yml`.

willirath:
In addition to the Travis build that is visible here, there also is an Appveyor build running against my fork of Parcels: https://ci.appveyor.com/project/willirath/parcels/build/1.0.12

erikvansebille:
Thanks haps for setting up the appveyor; that's great @willirath. I'm waiting on comments from @mlange05 on whether `tests` should go inside the `parcels` directory or not, and then we can merge

willirath:
JIT-Printing fails for Windows: https://ci.appveyor.com/project/willirath/parcels/build/1.0.12#L684

willirath:
@erikvansebille This branch will probably conflict with #236 because both added a windows environment file. I'd waint until #236 is merged and then make sure #235 everything works nicely.

delandmeterp:
@willirath, it seems that it's not the JIT printing itself but the read of the stdout that makes trouble in windows. Is the stdout format different in windows? I'll try to run it on a windows machine to understand it.

 out, err = capfd.readouterr()
        lst = out.split(' ')
>       assert float(lst[0]) == pset[0].id and float(lst[1]) == pset[0].p
E       ValueError: could not convert string to float:

willirath:
@delandmeterp That's good as it does not seem to be a real failure then, but only a subtlety within the test.  It should not be a reason not to merge, then. And it's probably a good idea to polish the Windows-part of the CI at some point, anyway.

docs/index.rst
erikvansebille: This still needs to change to `parcels_get_examples examples/` in the documentation too?

willirath:
> This still needs to change to parcels_get_examples examples/ in the documentation too?

Not necessarily.  I've changed it to 
```
parcels_get_examples parcels_examples
cd parcels_examples
jupyter notebook
```
as users will probably run this in their home directory. If you think, it's a good idea to keep the CI-scripts in line with the documentation, I'd rather change the Travis and Appveyor scripts to use `parcels_examples` as well.

willirath:
> I suggest to fix this last minor issue in the documentation, and then to merge. We can always decide to move tests into parcels in a later PR.

This sounds like a nice thing to finish my week. (I'll be mostly off tomorrow Monday.) I've learned a lot and had a lot of fun!

PR #236, opened 2017-11-12
Fix windows install

Simnator101:
The installation instructions for Windows were updated to provide a gcc environment which also works with 64-bit Python --- previously only 32-bit Python could interface with dynamic linked libraries created by mingw's gcc. 

Furthermore, to minimise the risk for 64/32-bit conflicts, additionnal flags were added to the gcc compiler in `parcels/compiler.py` to make compilation more secure.


environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - enum34
   - ffmpeg
   - flake8>=2.1.0
-  - gcc
+  - git
   - jupyter
```
willirath: Please don't delete gcc from the environment. It is needed to stay independent of gcc from the operating system on Linux and Mac. As there is no gcc package for Windows, there should be a separate set of dependencies. Consider to add a file environment_windows.yml?

Simnator101:
As suggested, I restored the original `environment.yml` and then created a 'Windows friendly' version, called `environment_win32.yml`. The documentation has also been updated to point users to the correct environment file for their OS.

erikvansebille:
Sorry for previously removing `gcc` from `environment.yml`, I was to blame for that. I agree with you @willirath that we should make sure *all* systems have `gcc`. 

On the other hand, maintaining two versions of the conda environment file isn't ideal, especially as they are almost identical. Is there not something like casacding environment files or something? Or a way to set OS-specific packages?

willirath:
> On the other hand, maintaining two versions of the conda environment file isn't ideal, especially as they are almost identical. Is there not something like casacding environment files or something? Or a way to set OS-specific packages?

I don't know of any simple way. For users, all this will be easier as soon as there's a package defining dependencies per OS and per Python version.

willirath:
I just came across <https://anaconda.org/anaconda/mingw> which seems to provide a full `MinGW` package for Conda. If you add `- mingw` to the `environment_win32.yml`, this would probably (can't test here due to lack of Windows machine) make Parcels work without any of the manual installation of MinGW (<https://github.com/OceanParcels/parcels/blame/ac0e100eac74133a6089dda89a9eaafe5fcec069/docs/index.rst#L112-L120>).

Simnator101:
That would certainly provide us with MinGW in Conda, but we have switched from MinGW to mingw-w64. This was done because MinGW only provides us with 32-bit compilation mode, which does not mesh well with 64-bit Python. Mingw-w64 does provide us with a 64-bit mode and it can also compile in 32-bit mode, so with that users of both the 32 as the 64-bit versions of Conda can use Parcels in JIT mode on Windows.

Simnator101:
After your hint, @willirath, that there exists a `mingw` package, I also went and tried to find a proper `mingw-w64` package which would provide Parcels with the features it needs. Luckily, it seems that Conda has a 64/32-bit implementation of the mingw toolchain, called `m2w64-toolchain`. I tried this package out in Conda and it seems to work for me, so I have changed the `environment_win32.yml` to use this package.

Anyways, I have noticed also something else within the environment files, namely that there is no requirement for Conda to install `git` in `environment.yml`, whilst there is in the Windows version. Is this on purpose? because it seems to me that other OS's also need `git` to get a copy of Parcels.

erikvansebille:
Thanks @Simnator101, installing `mingw-w64` via conda sounds like a great solution. 

And yes, `git` is only in `eniveronment_win32.yml` because I added it there at the beginning of this branch. The windows computer I was testing on didn't have `git` installed. 
I guess `git` could also be in `enivornment.yml`, although I expect all(?) linux/macOS systems to have git installed already? Yet, it shouldn't hurt to add?

willirath:
> I guess git could also be in enivornment.yml, although I expect all(?) linux/macOS systems to have git installed already? Yet, it shouldn't hurt to add?

Adding it to the environment installs Git from conda-forge.  This is a well-maintained package that usually is much newer than any Git version coming from the operating system.

PR #239, opened 2017-11-15
Better print for jit

delandmeterp:
Some improvement of the print functionalities in the python kernels for JIT.

For example, such kernel is now possible:

def UpdateP(particle, fieldset, time, dt):
    particle.p = fieldset.P[time, particle.lon, particle.lat, particle.depth]
    print(particle.p)
    b = particle.p
    xi = particle.xi
    print("I will print my Particle data")
    print("particle.p  %f %d" % (b, xi))

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -158,7 +158,7 @@ However, there are some key limitations to the Kernels that everyone who wants t
 
   * Functions from the ``maths`` standard library and from the custom ``random`` library at :mod:`parcels.rng`
   
-  * ``print`` statements. Simple print statements can be written, such as::
+  * Simple ``print`` statements, such as::
 
       print("Some print")
       print(particle.lon)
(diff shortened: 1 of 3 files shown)


erikvansebille:
Oh, and should we also unit-test this? Or is that overdoing it?

erikvansebille:
Actually, I just realised I once wrote a unit-test for printing already, but that it was xfailing. Now with this PR it works, so I've just pushed a commit so that now the printing is indeed tested

delandmeterp:
Perfect, there is still one thing I have to do.
If we advertise that we can do print in the kernels (of course it slows down the code, so it's for debugging), it has to be cleaner.

For the moment, we can do:
p = particle.p
print("check %g" % p)

but I haven't implemented that we do 
print("check %g" % particle.p)

So, I'll clean it, such that it works

erikvansebille:
Is that really true, @delandmeterp? I just simplified the print unit test, and it seems to work perfectly fine for
```
def kernel(particle, fieldset, time, dt):
    particle.p = fieldset.U[time, particle.lon, particle.lat, particle.depth]
    print("%d %f" % (particle.id, particle.p))
```

delandmeterp:
Yes indeed. I've just also checked and it works. I wasn't sure the the py_ast commands were working recursively in what I've done. But everything is right :-)

PR #241, opened 2017-11-17
Multigrid

delandmeterp:
In this PR, the StructuredGrid are fully operational. Grids and fields are different structures.

Also above StructuredGrid, there is the more general Grid class. There is now space to implement a SemiStructuredGrid class which is a sister of StructuredGrid.


parcels/fieldset.py
```diff
@@ -82,10 +86,67 @@ def from_data(cls, data, dimensions, transpose=True, mesh='spherical',
...
+            lat = dims['lat']
+            depth = np.zeros(1, dtype=np.float32) if 'depth' not in dims else dims['depth']
+            time = np.zeros(1, dtype=np.float64) if 'time' not in dims else dims['time']
+
+            fields[name] = Field(name, datafld, lon, lat, depth=depth,
+                                 time=time, transpose=transpose, units=units[name],
+                                 allow_time_extrapolation=allow_time_extrapolation, **kwargs)
+        u = fields.pop('U', None)
+        if u:
+            self.U = u
```
erikvansebille: This can be more pythonic? e.g. `self.U = u if u`? Also for `self.V` below?

parcels/fieldset.py
```diff
@@ -82,10 +86,67 @@ def from_data(cls, data, dimensions, transpose=True, mesh='spherical',
...
+        v = fields.pop('V', None)
+        if v:
+            self.V = v
+        for f in fields:
+            self.add_field(f)
+
+    def check_complete(self):
```
erikvansebille: Nice to create a `check_complete` method, but I don't think it's being called anywhere?

parcels/grid.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class StructuredGrid(Grid):
+    """Structured Grid
+
+    :param name: Name of the field
+    :param lon: Longitude coordinates of the field
+    :param lat: Latitude coordinates of the field
+    :param depth: Depth coordinates of the field
+    :param time: Time coordinates of the field
```
erikvansebille: Make explicit here in the docstring that `lon`, `lat` and `depth` can be 1D, 2D or 3D arrays, but that time has to be 1D?

parcels/grid.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    def __init__(self, name, lon, lat, depth=None, time=None):
+        assert isinstance(lon, np.ndarray), 'lon is not a numpy array'
+        sh = lon.shape
+        assert(len(sh) == 1 or len(sh) == 2 and min(sh) == 2), 'lon is not a vector'
```
erikvansebille: Why do we need to assert whether `lon.shape` is a vector? Can't we do multi-dimensional arrays yet? Is that not the purpose of this PR?

parcels/particle.py
```diff
@@ -44,6 +46,8 @@ def __repr__(self):
 
     def is64bit(self):
         """Check whether variable is 64-bit"""
+        if self.name == 'CGridIndexSet':
+            return True
         return True if self.dtype == np.float64 or self.dtype == np.int64 else False
```
erikvansebille: Or the more pythonic `return True if self.dtype == np.float64 or self.dtype == np.int64 or self.name =='CGridIndexSet' else False`

tests/test_multigrid.py
```diff
@@ -0,0 +1,76 @@
+from parcels import FieldSet, Field, ParticleSet, StructuredGrid, ScipyParticle, JITParticle, Variable
+from parcels import AdvectionRK4
+import numpy as np
+import pytest
```
erikvansebille: Not sure what the purpose of this `test_multigrid.py` function is. It looks more like an `example` than a short unit-test? 
And if we want to keep it as a unit-test, then why not within `test_fieldset_sampling`?

tests/test_fieldset_sampling.py
```diff
@@ -348,3 +348,27 @@ def test_sampling_out_of_bounds_time(mode, allow_time_extrapolation, k_sample_p,
     else:
         with pytest.raises(RuntimeError):
             pset.execute(k_sample_p, starttime=2.0, endtime=2.1, dt=0.1)
+
+
+@pytest.mark.parametrize('mode', ['jit', 'scipy'])
+def test_sampling_multiple_grid_sizes(mode):
```
erikvansebille: Do we need more unit-tests? E.g. that adding a new `Field` with a `Grid`that already exists means that no new `Grid` is added (i.e. test the `existing_grid` functionality in `GridSet`)?

parcels/codegenerator.py
```diff
@@ -568,7 +526,13 @@ def generate(self, funcname, field_args, const_args, kernel_ast):
...
 
         # Generate type definition for particle type
-        vdecl = [c.POD(v.dtype, v.name) for v in self.ptype.variables]
+        vdecl = []
+        for v in self.ptype.variables:
+            if v.name is 'CGridIndexSet':
+                vdecl.append(c.Pointer(c.POD(np.void, v.name)))
+            else:
+                vdecl.append(c.POD(v.dtype, v.name))
+
```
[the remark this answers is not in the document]
delandmeterp: Yes and No. The new six lines replace the line just above them.
All the deleted lines have a different reason. 
Before, we were searching in JIT the correct indices in the kernels, but these indices were not written back in the particles. It was only written automatically  when the lon and lat was updated. But there was still a double process of searching the correct indices. 
Now when we search (as before) the correct index, we write in straight away in the correct object (particle.grid[iGrid])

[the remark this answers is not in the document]
delandmeterp: I've put back a few lines that I have removed during the development and is not necessary

parcels/field.py
```diff
@@ -147,15 +148,19 @@ class Field(object):
...
         self.name = name
         self.data = data
-        self.lon = lon
-        self.lat = lat
-        self.depth = np.zeros(1, dtype=np.float32) if depth is None else depth
-        self.time = np.zeros(1, dtype=np.float64) if time is None else time
+        if grid:
+            self.grid = grid
+        else:
+            self.grid = StructuredGrid('auto_gen_grid', lon, lat, depth, time)
```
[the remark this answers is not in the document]
delandmeterp: Done
(+ more comments in the parameters of the constructor)

parcels/grid.py
delandmeterp: In this structured grid, all dimensions are vectors. Arrays will be defined for other types of grid.

parcels/include/parcels.h
```diff
@@ -22,23 +27,72 @@ typedef enum
...
+{
+  char *name;
+  int xdim, ydim, zdim, tdim, tidx;
   float *lon, *lat, *depth;
   double *time;
+} CStructuredGrid;
+
+typedef struct
+{
+  int xdim, ydim, zdim, tdim, tidx, allow_time_extrapolation, time_periodic;
```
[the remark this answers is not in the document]
delandmeterp: A CField has the data array double **.
I like the idea that we also store the dimensions of this array where it is defined.

parcels/gridset.py
```diff
@@ -0,0 +1,72 @@
...
+                continue
+            if not np.allclose(g.lon, grid.lon, rtol=1e-4):
+                continue
+            if not np.allclose(g.lat, grid.lat, rtol=1e-4):
+                continue
+            if not np.allclose(g.depth, grid.depth, rtol=1e-4):
+                continue
+            if not np.allclose(g.time, grid.time, rtol=1e-4):
+                continue
+            existing_grid = True
```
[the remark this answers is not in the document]
delandmeterp: No we can't.
Indeed, the main use of this testing is when user does not define itself the grids, so the grids are created at the field creation. 
This means that for every field, a different grid is created, and those are compared at the generation of a fieldset. So even if two grids are equal, they are still different objects, and have to be compared using np.allclose

parcels/field.py
```diff
@@ -131,39 +131,44 @@ class Field(object):
...
+    :param time_origin: Time origin of the time axis (only if grid is None)
     :param units: type of units of the field (meters or degrees)
     :param interp_method: Method for interpolation
     :param allow_time_extrapolation: boolean whether to allow for extrapolation
     :param time_periodic: boolean whether to loop periodically over the time component of the Field
            This flag overrides the allow_time_interpolation and sets it to False
     """
 
-    def __init__(self, name, data, lon=None, lat=None, grid=None, depth=None, time=None,
+    def __init__(self, name, data, grid=None, lon=None, lat=None, depth=None, time=None,
```
erikvansebille: I didn't realise that swapping the order here actually breaks backward compatibility for codes that call `Field` still with `lon`, `lat` etc. If that's the case, then better to revert to what you had before? Or even put the `grid` argument after the time argument?

delandmeterp: Ok, I changed it back.
But only here (I didn't reverted the tests. Indeed, lon and lat are now optional so it is a bad practice not to put explicitly lon=lon, even it it works for backward compatibility)

parcels/fieldset.py
```diff
@@ -198,7 +172,7 @@ def from_netcdf(cls, filenames, variables, dimensions, indices={},
             dims['data'] = name
             inds = indices[var] if var in indices else indices
 
-            fields[var] = Field.from_netcdf(var, dims, paths, inds, units=units[var],
```
delandmeterp: @erikvansebille 
If I am not wrong, this units argument was previously ignored. This now fixed in this version that reads the grid which has a mesh argument

PR #242, opened 2017-11-20
Asking user input when example directory already exists

erikvansebille:
Previously, `get_examples.py` did not work for users who worked off a `git clone` directly (rather than the `pip install` in Anaconda), because `parcels/examples` already exists. Now, we ask the user if they want to continue anyways

### DIFF
--- a/parcels/scripts/get_examples.py
+++ b/parcels/scripts/get_examples.py
@@ -5,6 +5,8 @@ from datetime import datetime, timedelta
 import os
 import pkg_resources
 from progressbar import ProgressBar
+import ConfigParser
+
 try:
     # For Python 3.0 and later
     from urllib.request import urlopen
@@ -40,6 +42,15 @@ def _maybe_create_dir(path):
             raise
 
 
+def _update_config_file(path):
+    config = ConfigParser.RawConfigParser()
+    config.add_section('parcels-example-data')
+    config.set('parcels-example-data', 'example-data-location',
+               os.path.join(os.getcwd(), path))
+    with open('config_parcels.yml', 'wb') as configfile:
+        config.write(configfile)
+
+
 def copy_data_and_examples_from_package_to(target_path):
     """Copy example data from Parcels directory.
 
@@ -53,6 +64,15 @@ def copy_data_and_examples_from_package_to(target_path):
     except Exception as e:
         print(e)
         pass
+
+
+def get_example_data_location():
+    """Get the location where example-data is stored,
+    from config_parcels.yml file"""
+    config = ConfigParser.RawConfigParser()
+    config.read('config_parcels.yml')
+    example_path = config.get('parcels-example-data', 'example-data-location')
+    return example_path
 
 
 def _still_to_download(file_names, target_path):
@@ -97,6 +117,9 @@ def main(target_path=None):
         answer = raw_input("Warning: {} already exists. Continue and overwrite existing example files [y/N]?".format(target_path)).lower()
         if answer not in ['y', 'Y', 'yes']:
             return
+
+    # update the location of examples data in Parcels config file
+    _update_config_file(target_path)
 
     # copy data and examples
     copy_data_and_examples_from_package_to(target_path)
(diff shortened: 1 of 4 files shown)


delandmeterp:
'parcels' should also be added to directory tree in tests_scripts.py:10

parcels/scripts/get_examples.py
```diff
@@ -94,8 +94,9 @@ def main(target_path=None):
         target_path = args.target_path
 
     if os.path.exists(target_path):
-        print("Error: {} already exists.".format(target_path))
-        return
+        answer = raw_input("Warning: {} already exists. Continue [y/n]?".format(target_path)).lower()
```
willirath: I'd suggest: "Continue and overwrite existing example files [y/N]?!"

To make sure a user knows that existing files may be lost and that the default answer (Just hitting Return.) amounts to: "Don't continue."

tests/test_fieldset.py
```diff
@@ -101,7 +101,7 @@ def test_fieldset_from_file_subsets(indslon, indslat, tmpdir, filename='test_sub
 
 @pytest.mark.parametrize('indstime', [range(2, 8), [4]])
 def test_moving_eddies_file_subsettime(indstime):
-    fieldsetfile = path.join(path.dirname(__file__), pardir, 'examples', 'MovingEddies_data', 'moving_eddies')
+    fieldsetfile = path.join(path.dirname(__file__), pardir, 'parcels', 'examples', 'MovingEddies_data', 'moving_eddies')
```
willirath: This is (of course) fine. I have a comment which I wanted to make earlier but forgot at the time, though:

The moving eddies example data seem to have been generated using a Python script anyway. Would it make sense to generate the data for the test instead of downloading it?

parcels/scripts/get_examples.py
erikvansebille: OK, I will change

tests/test_fieldset.py
erikvansebille: Actually, now with the new pathdir, the Travis CI fails, because it can't find the files anymore for `test_fieldset.py`.  So this is clearly not the solution..

erikvansebille: Would it make sense to log the location where examples data has been written to a new `config.yml` file, using the python `ConfigParser` class? So that different examples, tests, scripts etc can always find where the example data is downloaded and stored? Or is that overkill?

willirath:
I'm not sure I understand the new logic of how the example data is found.  Here's what I gathered:

- There's a new file `config.yml` which is created in the directory from which `parcels_get_examples` or `parcels/scripts/get_examples.py` is called.

- The function `parcels.scrips.get_examples.get_example_data_location` will try to open `config.yml` and read the actual location of the example data from it.  It will only succeed if `config.yml` is in `$(pwd)`.

- The test in question will search for the `config.yml` and the load a data file from the given location.

This heavily relies on the fact that `config.yml` happens to be in whatever directory the tests are called from.

erikvansebille:
That relying on `config.yml` being in the same directory is a good point yes, @willirath. I'm not sure what would be a better way though. Very open to suggestions on how other projects/software handle this kind of issue!

willirath:
From other projects, I know two ways of avoiding this:

- Ship the data with the project:<https://github.com/pydata/xarray/tree/master/xarray/tests/data>

- Generate data during the test: <https://github.com/pydata/xarray/blob/master/xarray/tests/test_backends.py#L597>

For the moving-eddies example, I'd go for the latter and incorporate the existing <http://oceanparcels.org/examples-data/MovingEddies_data/create_movingeddies_files.py> into `parcels`.

willirath:
Note that this problem really only applies for either devs who want to directly edit the installed version of parcels or during CI jobs. With a conda-forge package, we can easily make sure to include the data in parcels and let the tests directly look it up via `pkg_resources`.

willirath:
Here's how I'd do it: <https://github.com/OceanParcels/parcels/compare/master...willirath:alternative-fix-get-examples>

PR #246, opened 2017-11-22
Alternative fix get examples

willirath:
This is an alternative take at #242.

Differences in the approach:

- Do not use raw input but add a command-line arg `-o` for overwriting existing data.

- Within `tests/`, do not rely on any network connection, but generate the data that are necessary for the tests to run, during the tests.  This is an advantage when working on Parcels code on a machine with restricted network access (e.g. many high-performance computing centers).

Tests are still failing, because the script that generates the moving-eddies data files still reflects the old version of this data set.


willirath:
@erikvansebille: You can probably immediately see how https://github.com/OceanParcels/parcels/pull/246/files#diff-a9ca635146a843aee3373bda7a85768e should be adapted to the new shortened test.

erikvansebille:
Thanks @willirath for this alternative implementation. However, it seems a bit repetitive to include a new script to create the moving eddy fields, while that is already also done in `examples/example_moving_eddies.py`. We don't want to remove the code there, as it is a useful example of how to create a time-varying `fieldset`.

Perhaps the simplest (most elegant?) solution is to create a very small, very simple netcdf file (`test_data.nc`?), only to be used for `test_fieldset_from_file_subsets` and the `test_scripts.py` function, and ship that within the `tests` directory. Since it can be a very simple file, it doesn't need to be more than 10-100 Kb so we won't bloat the git repo too much. Is that a plan?

willirath:
> Perhaps the simplest (most elegant?) solution is to create a very small, very simple netcdf file (test_data.nc?), only to be used for test_fieldset_from_file_subsets and the test_scripts.py function, and ship that within the tests directory. Since it can be a very simple file, it doesn't need to be more than 10-100 Kb so we won't bloat the git repo too much. Is that a plan?

This sounds like a good solution.

willirath:
This can be closed.

Note the alternative implementation of allowing to overwrite examples in this PR, though.

PR #247, opened 2017-11-24
Single execute no time advance

delandmeterp:
The kernel will be executed only once, with dt=0.

This should then not be used for advection.

Works with both jit and scipy.

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -273,7 +273,7 @@ class ParticleSet(object):
             if interval < 0:
                 interval *= -1.
                 logger.warning("Negating interval because running in time-forward mode")
-        if endtime < starttime:  # Time-backward mode
+        elif endtime < starttime:  # Time-backward mode
             if dt > 0.:
                 dt *= -1.
                 logger.warning("Negating dt because running in time-backward mode")
@@ -281,12 +281,19 @@ class ParticleSet(object):
                 interval *= -1.
                 logger.warning("Negating interval because running in time-backward mode")
 
+        if np.allclose(endtime, starttime) or interval == 0 or dt == 0 or runtime == 0:
+            timeleaps = 1
+            dt = 0
+            runtime = 0
+            endtime = starttime
+        else:
+            timeleaps = int((endtime - starttime) / interval)
+
         # Initialise particle timestepping
         for p in self:
             p.time = starttime
             p.dt = dt
         # Execute time loop in sub-steps (timeleaps)
-        timeleaps = int((endtime - starttime) / interval)
         assert(timeleaps >= 0)
         leaptime = starttime
         for _ in range(timeleaps):
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -286,6 +286,7 @@ class ParticleSet(object):
             dt = 0
             runtime = 0
             endtime = starttime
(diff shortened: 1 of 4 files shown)


parcels/particleset.py
```diff
@@ -273,20 +273,27 @@ def execute(self, pyfunc=AdvectionRK4, starttime=None, endtime=None, dt=1.,
...
                 logger.warning("Negating dt because running in time-backward mode")
             if interval > 0.:
                 interval *= -1.
                 logger.warning("Negating interval because running in time-backward mode")
 
+        if np.allclose(endtime, starttime) or interval == 0 or dt == 0 or runtime == 0:
+            timeleaps = 1
+            dt = 0
+            runtime = 0
+            endtime = starttime
```
erikvansebille: Add a `logger.warning_once` here that kernels are executed without incrementing time?

tests/test_particle_sets.py
```diff
@@ -277,3 +277,23 @@ def test_pfile_array_remove_all_particles(fieldset, mode, tmpdir, pfile_type, np
...
+    if test == 'dt':
+        pset.execute(pset.Kernel(SetLat), starttime=0., endtime=1., dt=0.)
+    elif test == 'endtime':
+        pset.execute(pset.Kernel(SetLat), starttime=0., endtime=0., dt=1.)
+    assert np.allclose(pset.particles[0].lat, .6)
+    assert np.allclose(pset.particles[1].lat, .6)
+    assert np.allclose(pset.particles[0].lon, 0)
+    assert np.allclose(pset.particles[1].lon, 1)
+    assert np.allclose(pset.particles[1].time, 0)
+    assert np.allclose(pset.particles[0].time, 0)
```
erikvansebille: Or define first `lon = np.linspace(0, 1, npart, dtype=np.float32)` and then `pset = ParticleSet(..., lon=lon, ...)`

Then, the asserts could become
```
assert np.allclose([p.lon for p in pset], 0.6)
assert np.allclose([p.lon for p in pset], lon)
```
(untested)

tests/test_particle_sets.py
```diff
@@ -277,3 +277,23 @@ def test_pfile_array_remove_all_particles(fieldset, mode, tmpdir, pfile_type, np
...
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('test', ['dt', 'endtime'])
+def test_pset_execute_dt_0(fieldset, mode, test, npart=2):
+    def SetLat(particle, fieldset, time, dt):
+        particle.lat = .6
+    pset = ParticleSet(fieldset, pclass=ptype[mode],
+                       lon=np.linspace(0, 1, npart, dtype=np.float32),
+                       lat=np.linspace(1, 0, npart, dtype=np.float32))
+    if test == 'dt':
```
erikvansebille: If-statement can be removed by passing values for `dt` and `endtime` in pairs in the `@pytest` call. E.g.
```
@pytest.mark.parametrize('endtime', 'dt', [(1., 0.), (0., 1.)])
```
(untested)

parcels/particleset.py
delandmeterp: It is done, but git didn't see it (it's just next line)

PR #250, opened 2017-11-24
Rectilinear S grid implemented

delandmeterp:
For the moment, most of the S grid interpolation is done within one single functions (or two, one for jit, the other for scipy)

This could change when we develop curvilinear grids. All repeated parts will be put in a separate function and called. This should come for next PR of grid development


parcels/grid.py
```diff
@@ -3,12 +3,13 @@
 from ctypes import Structure, c_int, c_float, c_double, POINTER, cast, c_void_p, pointer
 from enum import IntEnum
 
-__all__ = ['GridCode', 'StructuredGrid', 'GridIndex', 'CGrid']
+__all__ = ['GridCode', 'RectilinearGrid', 'RectilinearSGrid', 'GridIndex', 'CGrid']
```
erikvansebille: Call this `RectilinearZGrid` and RectilinearSGrid`, to highlight that one is not a subset or child of the other?

parcels/grid.py
```diff
@@ -3,12 +3,13 @@
...
 
 
 class GridCode(IntEnum):
-    StructuredGrid = 0
-    SemiStructuredGrid = 1
+    RectilinearGrid = 0
+    RectilinearSGrid = 1
+    CurvilinearGrid = 2
```
erikvansebille: Already implement both `CurvilinearZGrid` and `CurvilinearSGrid`?

parcels/grid.py
```diff
@@ -87,20 +88,91 @@ def child_ctypes_struct(self):
...
+
+
+class RectilinearSGrid(Grid):
+    """Rectilinear S Grid. Same horizontal discretisation as a rectilinear grid,
+       but with s vertical coordinates
+
+    :param name: Name of the grid
+    :param lon: Vector containing the longitude coordinates of the grid
+    :param lat: Vector containing the latitude coordinates of the grid
+    :param depth: 3D array containing the vertical coordinates of the grid, which are s-coordinates
```
erikvansebille: or 4D, if time-evolving?

tests/test_grids.py
```diff
@@ -99,8 +99,63 @@ def test_avoid_repeated_grids():
...
+
+    class MyParticle(ptype[mode]):
+        temp = Variable('temp', dtype=np.float32, initial=20.)
+
+    lon = 400
+    lat = 0
+    ratio = .3
+    pset = ParticleSet.from_list(field_set, MyParticle, lon=[lon], lat=[lat], depth=[bath_func(lon)*ratio])
+
+    pset.execute(pset.Kernel(sampleTemp), runtime=1, dt=1)
```
erikvansebille: use the new `dt=0` here, once that is merged?

tests/test_grids.py
```diff
@@ -99,8 +99,63 @@ def test_avoid_repeated_grids():
...
+    field_set = FieldSet(u_field, v_field, fields=other_fields)
+    assert field_set.gridset.size == 2
+    assert field_set.U.grid.name == 'grid0py'
+    assert field_set.V.grid.name == 'grid1py'
+    assert field_set.temp.grid.name == 'grid0py'
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('z4d', ['True', 'False'])
+def test_s_grids(mode, z4d):
```
erikvansebille: Is one unit test enough here? I would ideally also like to see some advection tests. I.e. particles moving in simple flow on (time-evolving?) s-grids

parcels/grid.py
```diff
@@ -87,20 +88,91 @@ def child_ctypes_struct(self):
...
+                                   self.lon.ctypes.data_as(POINTER(c_float)),
+                                   self.lat.ctypes.data_as(POINTER(c_float)),
+                                   self.depth.ctypes.data_as(POINTER(c_float)),
+                                   self.time.ctypes.data_as(POINTER(c_double)))
+        return cstruct
+
+
+class RectilinearSGrid(Grid):
+    """Rectilinear S Grid. Same horizontal discretisation as a rectilinear grid,
+       but with s vertical coordinates
```
erikvansebille: Be more explicit here that with `s vertical coordinates`, we mean terrain-following (sigma) or density (rho) layers

parcels/grid.py
delandmeterp: No, it is not yet done. It will be done in next PR.
(I've just changed the SemiStructuredGrid line to the new names, even not available)

tests/test_grids.py
```diff
@@ -157,5 +158,65 @@ class MyParticle(ptype[mode]):
...
+    lon = np.array([0 for i in range(11)])
+    lat = [0 for i in range(11)]
+    ratio = [min(i/10., .99) for i in range(11)]
+    depth = bath_func(lon)*ratio
+    pset = ParticleSet.from_list(field_set, MyParticle, lon=lon, lat=lat, depth=depth)
+
+    kernels = AdvectionRK4_3D + pset.Kernel(sampleTemp)
+    for _ in range(10):
+        pset.execute(kernels, starttime=pset[0].time, runtime=1000, dt=1000)
+    assert np.allclose([p.temp for p in pset], ratio)
```
erikvansebille: Wouldn't it make more sense to check that `particle.depth/bath` is constant at the end of the run? Then one also wouldn't need the `temp` Variable anymore?

tests/test_grids.py
```diff
@@ -157,5 +158,65 @@ class MyParticle(ptype[mode]):
...
+    field_set = FieldSet(u_field, v_field, fields=other_fields)
+
+    def sampleTemp(particle, fieldset, time, dt):
+        particle.temp = fieldset.temp[time+dt, particle.lon, particle.lat, particle.depth]
+
+    class MyParticle(ptype[mode]):
+        temp = Variable('temp', dtype=np.float32, initial=20.)
+
+    lon = np.array([0 for i in range(11)])
+    lat = [0 for i in range(11)]
```
erikvansebille: Or use `np.zeros` in assignment of `lon` and `lat` here?

tests/test_grids.py
```diff
@@ -157,5 +158,65 @@ class MyParticle(ptype[mode]):
...
+    ratio = [min(i/10., .99) for i in range(11)]
+    depth = bath_func(lon)*ratio
+    pset = ParticleSet.from_list(field_set, MyParticle, lon=lon, lat=lat, depth=depth)
+
+    kernels = AdvectionRK4_3D + pset.Kernel(sampleTemp)
+    for _ in range(10):
+        pset.execute(kernels, starttime=pset[0].time, runtime=1000, dt=1000)
```
erikvansebille: Why run in a for-loop here? Why not simply one call to `pset.execute(kernels, runtime=10000, dt=1000)`?

parcels/grid.py
```diff
@@ -194,10 +194,6 @@ def __set__(self, instance, value):
         instance._cptr.__setitem__(self.name, value)
 
 
-class CGridIndex(Structure):
-    _fields_ = [('xi', c_int), ('yi', c_int), ('zi', c_int)]
-
-
```
erikvansebille: Correct that this `CGridIndex` structure is now completely gone? We don't need it anymore?

delandmeterp: There is a CGridIndex in parcels.h, but the formatting of this struct is defined in GridIndex.dtype

PR #252, opened 2017-11-28
Particle attributes

davidwichmann:
Can now add particle attributes, depending on the particle id ("trajectory") only and written only once in the netcdf with Variable(...,to_write='once')

### DIFF
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -28,7 +28,8 @@ class ParticleFile(object):
 
     :param name: Basename of the output file
     :param particleset: ParticleSet to output
-    :param user_vars: A list of additional user defined particle variables to write
+    :param user_vars: A list of additional user defined particle variables to write for all particles and all times
+    :param user_vars_once: A list of additional user defined particle variables to write for all particle only once at initial time
     :param type: Either 'array' for default matrix style, or 'indexed' for indexed ragged array
     """
 
@@ -90,15 +91,21 @@ class ParticleFile(object):
         self.z.positive = "down"
 
         self.user_vars = []
+        self.user_vars_once = []
+        
         for v in particleset.ptype.variables:
             if v.name in ['time', 'lat', 'lon', 'depth', 'z', 'id']:
                 continue
-            if v.to_write is True:
-                setattr(self, v.name, self.dataset.createVariable(v.name, "f4", coords, fill_value=np.nan))
+            if v.to_write:
+                if v.to_write is True:
+                    setattr(self, v.name, self.dataset.createVariable(v.name, "f4", coords, fill_value=np.nan))
+                    self.user_vars += [v.name]
+                elif v.to_write == 'once':
+                    setattr(self, v.name, self.dataset.createVariable(v.name, "f4", "trajectory", fill_value=np.nan))
+                    self.user_vars_once += [v.name]
                 getattr(self, v.name).long_name = ""
                 getattr(self, v.name).standard_name = v.name
                 getattr(self, v.name).units = "unknown"
-                self.user_vars += [v.name]
 
         self.idx = 0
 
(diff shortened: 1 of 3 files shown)


parcels/particlefile.py
```diff
@@ -28,7 +28,8 @@ class ParticleFile(object):
 
     :param name: Basename of the output file
     :param particleset: ParticleSet to output
-    :param user_vars: A list of additional user defined particle variables to write
+    :param user_vars: A list of additional user defined particle variables to write for all particles and all times
+    :param user_vars_once: A list of additional user defined particle variables to write for all particle only once at initial time
```
erikvansebille: Actually, these `users_vars` and `user_vars_once` are not parameters at all (they are not input in the `__init__`), so not sure why these lines were even here. Perhaps better to write brief comments on line 93 and 94 when `self.user_vars = []`?

parcels/particlefile.py
```diff
@@ -140,6 +147,9 @@ def write(self, pset, time, sync=True):
                     self.z[inds, self.idx] = np.array([p.depth for p in pset])
                     for var in self.user_vars:
                         getattr(self, var)[inds, self.idx] = np.array([getattr(p, var) for p in pset])
+                    if self.idx == 0:
+                        for var in self.user_vars_once:
+                            getattr(self, var)[inds] = np.array([getattr(p, var) for p in pset])
```
erikvansebille: The same type of code snippet also needs to be added to the ` elif self.type is 'indexed':` below

tests/test_particle_sets.py
```diff
@@ -291,3 +291,23 @@ def SetLat(particle, fieldset, time, dt):
...
+    class MyParticle(ptype[mode]): 
+        v_once = Variable('v_once', dtype=np.float32, initial=1.,to_write='once')
+    lon = np.linspace(0, 1, npart, dtype=np.float32)
+    lat = np.linspace(1, 0, npart, dtype=np.float32)
+    pset = ParticleSet(fieldset, pclass=MyParticle, lon=lon, lat=lat)
+    pset.execute(pset.Kernel(SetLat)+pset.Kernel(Update_v), starttime=0., endtime=1, dt=0.1, interval=interval,output_file=pset.ParticleFile(name=filepath))
+    ncfile = Dataset(filepath+".nc", 'r', 'NETCDF4') 
+    V_once = ncfile.variables['v_once'][:]
+    assert (V_once.shape == (npart,))
+    assert (V_once[0] == 1.)
```
erikvansebille: Also assert that `pset[0].v_once == 11`, just to check that the actual particle attribute has been updated?

tests/test_particle_sets.py
```diff
@@ -291,3 +291,23 @@ def SetLat(particle, fieldset, time, dt):
...
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize(('npart','interval'),[(1,0.2),(5,0.2),(1,1.0),(5,1.0)])
+def test_variable_written_once(fieldset, mode, interval, tmpdir, npart):
+    filepath = tmpdir.join("pfile_once_written_variables")
+    def Update_v(particle, fieldset, time, dt):
+        particle.v_once += 1.
+    def SetLat(particle, fieldset, time, dt):
+        particle.lat = .6
```
erikvansebille: Why do you need a `SetLat` kernel here? Would the test not work without?

tests/test_particle_sets.py
```diff
@@ -291,3 +291,23 @@ def SetLat(particle, fieldset, time, dt):
     assert np.allclose([p.lon for p in pset], lon)
     assert np.allclose([p.lat for p in pset], .6)
     assert np.allclose([p.time for p in pset], 0)
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize(('npart','interval'),[(1,0.2),(5,0.2),(1,1.0),(5,1.0)])
```
erikvansebille: If you want all combinations of `nart` and `interval`, simply put them in different `@pytest.mark.parametrize` statements?

PR #253, opened 2017-11-29
Further explanation of installation instructions

erikvansebille:
Added note that Conda works best if Python is not installed yet. And also clarified that users need to navigate to the directory where the files are downloaded (this was not clear to some users)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -85,7 +85,7 @@ compiler, and various Python packages.
 
 The simplest way to install Parcels is to use Anaconda and Conda-Forge for all
 requirements. This is the “batteries-included” solution probably suitable for most users. 
-Note that this works best if Python is not installed yet. Otherwise, see the instructions at the bottom of this section.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -84,20 +84,21 @@ Parcels depends on a working Python installation, a netCDF installation, a C
 compiler, and various Python packages.
 
 The simplest way to install Parcels is to use Anaconda and Conda-Forge for all
-requirements. This is the “batteries-included” solution probably suitable for most users.
+requirements. This is the “batteries-included” solution probably suitable for most users. 
+Note that this works best if Python is not installed yet. Otherwise, see the instructions at the bottom of this section.
```
willirath: I'd go for a formulation that cannot be confused with the operating system's Python installation present on virtually all Linux machines and most of the Macs.  What about:

> Note that if you already have an existing Python setup that you wish to use for Parcels as well, you should follow the instructions at the bottom of this section.

PR #254, opened 2017-11-30
Fix jit multi grids

delandmeterp:
Another change is also a fix of the glob.glob(file*.nc) output, which is now sorted, since the result of glob.glob function was wrong on some OS.

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -85,6 +85,23 @@ class RectilinearZGrid(Grid):
         self.mesh = mesh
         self.cstruct = None
 
+    def add_periodic_halo(self, zonal, meridional, halosize=5):
+        """Add a 'halo' to the Grid, through extending the Grid (and lon/lat)
+        similarly to the halo created for the Fields
+
+        :param zonal: Create a halo in zonal direction (boolean)
+        :param meridional: Create a halo in meridional direction (boolean)
+        :param halosize: size of the halo (in grid points). Default is 5 grid points
+        """
+        if zonal:
+            lonshift = (self.lon[-1] - 2 * self.lon[0] + self.lon[1])
+            self.lon = np.concatenate((self.lon[-halosize:] - lonshift,
+                                      self.lon, self.lon[0:halosize] + lonshift))
+        if meridional:
+            latshift = (self.lat[-1] - 2 * self.lat[0] + self.lat[1])
+            self.lat = np.concatenate((self.lat[-halosize:] - latshift,
+                                      self.lat, self.lat[0:halosize] + latshift))
+
     @property
     def child_ctypes_struct(self):
         """Returns a ctypes struct object containing all relevant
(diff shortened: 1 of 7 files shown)


tests/test_grids.py
```diff
@@ -107,6 +107,41 @@ def test_avoid_repeated_grids():
...
+    def bath_func(lon):
+        return lon / 1000. + 10
+    bath = bath_func(lon_g0)
+
+    for i in range(depth_g0.shape[0]):
+        for k in range(depth_g0.shape[2]):
+            depth_g0[i, :, k] = bath[i] * k / (depth_g0.shape[2]-1)
+
+    grid_0 = RectilinearSGrid('grid0py', lon_g0, lat_g0, depth=depth_g0)
+    grid_1 = RectilinearSGrid('grid0py', lon_g0, lat_g0, depth=depth_g0)
```
erikvansebille: Why is this second `grid_1` needed? Why not use `grid_0` for all `Fields`? Or is that one of the things you want to test?

parcels/grid.py
```diff
@@ -98,16 +102,49 @@ class CRectilinearGrid(Structure):
...
+    """
+
+    def __init__(self, name, lon, lat, depth=None, time=None, time_origin=0, mesh='flat'):
+        RectilinearGrid.__init__(self, name, lon, lat, time, time_origin, mesh)
+        if isinstance(depth, np.ndarray):
+            assert(len(depth.shape) == 1), 'depth is not a vector'
 
+        self.gtype = GridCode.RectilinearZGrid
+        self.depth = np.zeros(1, dtype=np.float32) if depth is None else depth
+        self.z4d = -1  # only used in RectilinearSGrid
```
erikvansebille: why not set `self.z4d` to `None` then?

tests/test_grids.py
```diff
@@ -190,8 +223,8 @@ def bath_func(lon):
             w_data[i, :, k] = u_data[i, :, k] * depth_g0[i, :, k] / bath[i] * 1e-3
 
     u_field = Field('U', u_data, grid=grid_0, transpose=True)
-    v_field = Field('V', v_data, grid=grid_1, transpose=True)
-    w_field = Field('W', w_data, grid=grid_2, transpose=True)
+    v_field = Field('V', v_data, grid=grid_0, transpose=True)
+    w_field = Field('W', w_data, grid=grid_0, transpose=True)
```
erikvansebille: If there's only one grid, why number them? I.e. why not call this `grid=grid`?

parcels/grid.py
delandmeterp: It is sent to C code as an int, so I gave an int (1 for z4d and 0 for z3d, and -1 if it's not even used...)

tests/test_grids.py
delandmeterp: It's a test. I'll add a comment on that

PR #255, opened 2017-12-06
Simplify gradients / central differences method

erikvansebille:
This PR simplifies and generalises some of the code in `field.py` for creating gradients of `Fields`. It also extends the unit tests and makes them more robust. 
Specifically, this PR replaces the `CentralDifferences` function from `field.py` with numpy inbuilt gradient function, while vastly shortening and simplifying the code. 

This has a few implications:
- `field.gradient()` now returns two numpy arrays, rather than `Field` objects. This is in closer agreement to the behaviour of other methods in `field.py`. Users can of course create `Field` objects from these numpy-arrays themselves
- `field.gradient()` does not take the arguments `timerange`, `lonrange` and `latrange` anymore. Again, if users want the gradients of only a subset of a field, they can first create a new (smaller) field themselves and run the `.gradient()` method on that subset

Note that this method only works for `Rectilinear` grids. For the `Curvilinear` grids which are currently being implemented by @delandmeterp, the `field.cell_edge_sizes()` method will have to be extended


parcels/field.py
```diff
@@ -288,41 +263,38 @@ def from_netcdf(cls, name, dimensions, filenames, indices={},
...
+                dy_grid[x, y] = y_conv.to_source(dy, lon, lat, self.grid.depth[0])
+        return dx_grid, dy_grid
+
+    def cell_areas(self):
+        """Method to calculate cell sizes based on cell_edge_sizes
+                Currently only works for Rectilinear Grids"""
+        dx_mesh, dy_mesh = self.cell_edge_sizes()
+        return dx_mesh * dy_mesh
+
+    def gradient(self):
```
delandmeterp: You put in the PR comment that gradient is only available for rectilinear grids, but I would also put it here in the code.

parcels/field.py
```diff
@@ -288,41 +263,38 @@ def from_netcdf(cls, name, dimensions, filenames, indices={},
...
+        """Method to calculate horizontal gradients of Field.
+                Returns two numpy arrays: the zonal and meridional gradients"""
```
delandmeterp: The user may be interested in the format of the Field obtained.
What is the size of the array obtained. How is the gradient computed?

erikvansebille:
I have just merged v1.0 into this PR. Note however, that now that we have support for Curvilinear grids in Parcels, we should also make sure that the new `Field.gradients()` methods works for these curvilinear grids too. This will need to be implemented before we merge, I reckon

PR #259, opened 2017-12-08
Two extra unit tests for the Single execute no time advance (dt=0) PR

erikvansebille:
I found this branch that I created a while ago, with two unit tests for the dt=0 issue, which at the time failed Travis testing. 
This issue was fixed in #247, and now the unit tests pass. 

While #247 also had its own unit test, I still think it would be good to merge these two unit tests into `master`


erikvansebille:
I have now folded all unit tests for the `dt=0` functionality into one unit test in `test_kernel_execution`. Note that, because only one of (`endtime`, `runtime`) can be specified, both are explicitly tested.
This new unit tests also checks for expected behaviour when `dt=1` and `endtime=1`.

PR #261, opened 2017-12-11
Repeated particle release API

### LINKED ISSUE #258: API for easier repeated (and delayed) release of particles
Currently, the `time` argument in the creation of `ParticleSets` doesn't do very much. In particular, it can not be used to schedule a particle do be released during `.execute()`.

This means it is rather cumbersome at the moment to run simulations with delayed and/or repeated releases of particles (see e.g. the [delayed-start tutorial](http://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_delaystart.ipynb)).

It would be very useful to extend the Parcels API so that particles can be scheduled for release. I now see two ways to do this:
1. Create all Particles upon initialisation, and only apply the Kernels for those where `leaptime >= particle.time`. This will be easy to implement, but it is inefficient in memory as some particles will be 'dormant' in memory for possibly large parts of the `.execute`
2. Only add the `Particles` to a `ParticleSet` when `leaptime >= particle.time`. This will be much more efficient in memory (only particles that are actually 'life' are in the `ParticleSet`), but requires a large reworking of the Parcels code as the `ParticleSet` will not be fully generated upon initialisation anymore.

I have created a new branch (https://github.com/OceanParcels/parcels/tree/repeated_particle_release_API) with unit tests for this delayed releases

erikvansebille:
This PR implements two different ways to control the time that particles are released during a `ParticleSet.execute()`, and fixes #258. 

Previously, all particles were started at the same time (`starttime`). Now, the two options are (both are shown in the updated `tutorial_delaystart.ipynb` notebook):
1. Users can control on a per-particle basis when a `Particle` is released (using the `time` argument in `ParticleSet` construction). 
2. Users can 're-release' a `ParticleSet` at a certain interval using the new `releasedt` argument in `ParticleSet` construction

Note that this means that the `starttime` argument in `ParticleSet.execute()` does not control when particles are released anymore. This may cause some (minor) backward compatibility issues.

In fact, it might be simpler and more intuitive to remove `starttime` and `endtime` altogether, and rather have only `runtime` as argument in `ParticleSet.execute()` that controls the length of a simulation? But that is probably better dealt with in a next PR.


parcels/codegenerator.py
```diff
@@ -549,23 +549,28 @@ def generate(self, funcname, field_args, const_args, kernel_ast):
...
+        sign_end_part = c.Assign("sign_end_part", "endtime - particles[p].time > 0 ? 1 : -1")
         dt_pos = c.Assign("__dt", "fmin(fabs(particles[p].dt), fabs(endtime - particles[p].time))")
         dt_0_break = c.If("particles[p].dt == 0", c.Statement("break"))
+        notstarted_continue = c.If("(sign_end_part != sign_dt) && (particles[p].dt != 0)",
+                                   c.Statement("continue"))
         body = [c.Assign("res", "%s(&(particles[p]), %s)" % (funcname, fargs_str))]
         body += [c.Assign("particles[p].state", "res")]  # Store return code on particle
-        body += [c.If("res == SUCCESS", c.Block([c.Statement("particles[p].time += sign * __dt"),
+        body += [c.If("res == SUCCESS", c.Block([c.Statement("particles[p].time += sign_dt * __dt"),
                                                  dt_pos, dt_0_break, c.Statement("continue")]))]
```
delandmeterp: Just to be sure: I think that we even did it together, but apart from checking that the test work, did you have a look at the generated c code and see that it is as expected? (it is so easy to do a typo in this generation script..)

parcels/examples/example_delay_start.py
```diff
@@ -1,59 +0,0 @@
...
-from netCDF4 import Dataset
-from os import path
-
-
-ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
-
-
-@pytest.mark.parametrize('mode', ['scipy', 'jit'])
-def test_delay_start_example(mode, npart=10, show_movie=False):
-    """Example script that shows how to 'delay' the start of particle advection.
```
delandmeterp: Do you mean you don't show any example on how to add particles with pset.add anymore?
If it is the case, I think we should still show it. Indeed, if the user wants to add the particle with some variable depending on the result of what is going on, it still needs to use this pset.add

parcels/examples/tutorial_delaystart.ipynb
```diff
@@ -42,446 +30,848 @@
...
-    "collapsed": true,
-    "deletable": true,
-    "editable": true
-   },
-   "outputs": [],
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
```
delandmeterp: Why do we have to commits pynb with the outputs inside? It makes the diff complicate to read, and add unnecessary lines with no meaning. Or do you need it in order to visualise something on oceanparcels.org?
This file is really thousands of lines!

parcels/particle.py
```diff
@@ -123,6 +124,10 @@ def __init__(self):
                 lat = self.getInitialValue(ptype, name='lat')
                 depth = self.getInitialValue(ptype, name='depth')
                 time = self.getInitialValue(ptype, name='time')
+                if time is None:
```
delandmeterp: I don't fully understand yet, where do we put a logger.error and where do we raise an error. Is it some rule for the moment?

parcels/examples/tutorial_delaystart.ipynb
erikvansebille: I know it's inconvenient that these files are so massive. But we need to save them with all these outputs to be displayed properly on nbviewer.jupyter.org (where we link to from the oceanparcels.org tutorial list); so nothing we can do bout that, I fear

parcels/particle.py
erikvansebille: Ideally we want to move to the `parcels.logger` module everywhere. But we still have some legacy `raise` statements, from before the `parcels.logger` got introduced. Also, `parcels.logger.error()` does not yet throw an `exit` by itself, as one would expect. Would be good to clean this up at some point (not this PR though)?

parcels/codegenerator.py
erikvansebille: Yes, I checked the generated C code and all looks good

parcels/examples/example_delay_start.py
erikvansebille: Good point; we indeed do not document to `pset.add()` functionality anymore, except for in the test scripts, and in the online documentation at http://oceanparcels.org/parcels.html#parcels.particleset.ParticleSet.add. 
But then again, there is a lot of functionality that is not shown (yet) in examples. I wanted to remove this specific file because with this PR it makes no sense anymore to delay the start of particles through `pset.add()` statements and could only lead to confusion. Plus, there is already the delaystart jupyter notebook at http://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_delaystart.ipynb

delandmeterp: Ok, no problem!

parcels/particle.py
delandmeterp: Such cleaning is not related to this PR. It should be done in another one I think

PR #262, opened 2017-12-13
Curvilinear

delandmeterp:
This is not ready to pull to master yet.

What is done so far (the most difficult is done):
* interpolation on curvilinear grids
* reading of NEMO grids
* rotation of UV fields
* splitting Z and S vertical discretisation
* passing fro 180W to 180E
* upload of data
* passing from i = sizeX to i = 0 in search_indices (be sure that periodic_halo works)
* tests and examples


parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+import xarray as xr
+from os import path
+from netCDF4 import Dataset
+
+
+def compute_curvilinear_rotation_angles(mesh_filename, rotation_angles_filename, variables, dimensions):
```
erikvansebille: rename this file to `compute_curvilinear_rotation_angles.py`, just like the other files in the `scripts` directory?

parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if path.isfile(rotation_angles_filename) and path.getmtime(rotation_angles_filename) > path.getmtime(mesh_filename):
+        return
+
```
erikvansebille: Provide some `logger.info` to users that a angles file is being generated, and where it is stored. Or, if the file already exists, that nothing is being done?

parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    sinUVar[:] = gsinu
+    cosVVar = subDataset.createVariable(variables['cosV'], 'f8', ('y', 'x',))
+    cosVVar[:] = gcosv
+    sinVVar = subDataset.createVariable(variables['sinV'], 'f8', ('y', 'x',))
+    sinVVar[:] = gsinv
+
+    subDataset.close()
+    # ** end netCDF4 writing **
+
+    # lonUArray = xr.DataArray(lonU[1:, 1:],
```
erikvansebille: remove all of this commented stuff?

parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+import xarray as xr
+from os import path
+from netCDF4 import Dataset
+
+
+def compute_curvilinear_rotation_angles(mesh_filename, rotation_angles_filename, variables, dimensions):
+
```
erikvansebille: Also, add a bit of documentation on what the function does and how it works?

erikvansebille: And also, add a line to `parcels/parcels/scripts/__init__.py` so that it can be directly imported?

parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    gsinu = (zxnpu*zyffu - zynpu*zxffu) / znffu
+    gcosu = (zxnpu*zxffu + zynpu*zyffu) / znffu
+    gsinv = (zxnpv*zxffv + zynpv*zyffv) / znffv
+    gcosv = -(zxnpv*zyffv - zynpv*zxffv) / znffv
+
+    # ** netCDF4 writing, since xArray is bugged **
+    lonU = lonU[1:, 1:]
+    latU = latU[1:, 1:]
+    lonV = lonV[1:, 1:]
+    latV = latV[1:, 1:]
```
erikvansebille: why strip the first indices here?

parcels/scripts/utils.py
```diff
@@ -0,0 +1,120 @@
...
+        - 2. * np.sin(rad*lonF[:-1, 1:]) * np.tan(rpi/4. - rad*latF[:-1, 1:]/2.)
+    znffu = np.sqrt(znnpu * (zxffu*zxffu + zyffu*zyffu))
+    znffu = np.maximum(znffu, 1.e-14)
+
+    zxffv = 2. * np.cos(rad*lonF[1:, 1:]) * np.tan(rpi/4. - rad*latF[1:, 1:]/2.) \
+        - 2. * np.cos(rad*lonF[1:, :-1]) * np.tan(rpi/4. - rad*latF[1:, :-1]/2.)
+    zyffv = 2. * np.sin(rad*lonF[1:, 1:]) * np.tan(rpi/4. - rad*latF[1:, 1:]/2.) \
+        - 2. * np.sin(rad*lonF[1:, :-1]) * np.tan(rpi/4. - rad*latF[1:, :-1]/2.)
+    znffv = np.sqrt(znnpv * (zxffv*zxffv + zyffv*zyffv))
+    znffv = np.maximum(znffv, 1.e-14)
```
erikvansebille: seems a lot of repetition. Use functions?

tests/test_advection.py
```diff
@@ -122,7 +122,7 @@ def test_advection_periodic_zonal(mode, xdim=100, ydim=100, halosize=3):
     assert(len(fieldset.U.lon) == xdim + 2 * halosize)
 
     pset = ParticleSet(fieldset, pclass=ptype[mode], lon=[0.5], lat=[0.5])
-    pset.execute(AdvectionRK4 + pset.Kernel(periodicBC), endtime=delta(hours=20), dt=delta(seconds=30))
+    pset.execute(AdvectionRK4 + pset.Kernel(periodicBC), endtime=0.65, dt=0.01)
```
erikvansebille: why change this? And also the other changes in this `tests/test_advection.py` file?

tests/test_grids.py
```diff
@@ -280,3 +282,158 @@ def moveEast(particle, fieldset, time, dt):
...
+    theta = theta + np.pi/6.
+
+    lon = r * np.cos(theta)
+    lat = r * np.sin(theta)
+    time = np.array([0, 86400], dtype=np.float64)
+    # import matplotlib.pyplot as plt
+    # plt.plot(xx,yy,"ob")
+    # plt.plot(lon,lat,".r")
+    # plt.axis('equal')
+    # plt.show()
```
erikvansebille: remove comments?

tests/test_grids.py
```diff
@@ -280,3 +282,158 @@ def moveEast(particle, fieldset, time, dt):
...
+        speed = Variable('speed', dtype=np.float32, initial=0.)
+        u = Variable('u', dtype=np.float32, initial=0.)
+        v = Variable('v', dtype=np.float32, initial=0.)
+        zonal = Variable('zonal', dtype=np.float32, initial=0.)
+        meridional = Variable('meridional', dtype=np.float32, initial=0.)
+
+    lonp = 175.5
+    latp = 81.5
+    pset = ParticleSet.from_list(field_set, MyParticle, lon=[lonp], lat=[latp])
+    pset.execute(pset.Kernel(eulerAdvect), runtime=86400*2, dt=3600*6)
```
erikvansebille: use `datetime` objects?

tests/test_grids.py
```diff
@@ -280,3 +282,158 @@ def moveEast(particle, fieldset, time, dt):
...
+    dimensions = {'U': {'lon': 'nav_lon_u', 'lat': 'nav_lat_u'},
+                  'V': {'lon': 'nav_lon_v', 'lat': 'nav_lat_v'},
+                  'cosU': {'lon': 'glamu', 'lat': 'gphiu'},
+                  'sinU': {'lon': 'glamu', 'lat': 'gphiu'},
+                  'cosV': {'lon': 'glamv', 'lat': 'gphiv'},
+                  'sinV': {'lon': 'glamv', 'lat': 'gphiv'}}
+
+    field_set = FieldSet.from_netcdf(filenames, variables, dimensions, mesh='spherical', allow_time_extrapolation=True)
+
+    def eulerAdvect(particle, fieldset, time, dt):
```
erikvansebille: why not use built-in advectionRK4, now that that works?

parcels/gridset.py
```diff
@@ -29,6 +29,9 @@ def add_grid(self, field):
                 if gattr.shape != gridattr.shape:
                     sameGrid = False
                     break
+                if not np.allclose(gattr, gridattr):
```
erikvansebille: combine with the previous if statement using `or`?

parcels/grid.py
```diff
@@ -205,8 +377,8 @@ class GridIndex(object):
     def __init__(self, grid, *args, **kwargs):
         self._cptr = kwargs.pop('cptr', None)
         self.name = grid.name
-        self.xi = 0
-        self.yi = 0
+        self.xi = int(grid.xdim / 2)
+        self.yi = int(grid.ydim / 2)
```
erikvansebille: is this purely for speedup? To start halfway? Does it matter? If really only for speed, then perhaps wait with this change until we really start benchmarking?

parcels/field.py
```diff
@@ -151,13 +170,17 @@ class Field(object):
 
     def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=None,
                  transpose=False, vmin=None, vmax=None, time_origin=0,
-                 interp_method='linear', allow_time_extrapolation=None, time_periodic=False):
+                 interp_method='linear', allow_time_extrapolation=None, time_periodic=False, mesh='flat'):
         self.name = name
+        if self.name == 'UV':
+            self.phantom = True
```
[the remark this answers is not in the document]
delandmeterp: Indeed, it was called `phantom` since it is an un-complete field. 
When I did it, I thought we might use such field for another reason later, it is why I didn't give this flag a so explicit name. 
But you are right, let's advance one step at a time. So for the moment, phantom is used only for 'UV', so I simply remove it, and check if self.name == 'UV' instead of self.phantom

tests/test_advection.py
delandmeterp: The change you highlighted was a consequence of the change `spherical` to `flat` just above.

I needed to do this change for previous implementation of periodic BC in curvi grids.
And I thought there was consequence since a periodic BC in spherical mode should not reconnect the 1 degree with the 0 as it is here but user might do such periodic BC so we have to be able to do it.
But indeed, it works with currenet implementation, so I reversed it as it was before.

parcels/scripts/utils.py
delandmeterp: Since to compute an angle at a node, we need the position of this node and the previous one. So it means that we lose the first line and row of a mesh.
It is important to keep in mind, and is stated in the doc of the function

parcels/scripts/utils.py
delandmeterp: It's a translation of the Fortran implementation. The link to the Fortran file is given in comments

parcels/codegenerator.py
```diff
@@ -30,10 +30,11 @@ def __getitem__(self, attr):
 
 
 class FieldEvalNode(IntrinsicNode):
-    def __init__(self, field, args, var):
+    def __init__(self, field, args, var, var2=None):
```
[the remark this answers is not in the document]
delandmeterp: done

parcels/grid.py
delandmeterp: Indeed, the final goal of this is to save some time.
But currently, if I remove it, I still have problems with the periodic boundary.
Indeed, I've just discovered the problem exists even without this.
I'm working on it

Issue #216, opened 2017-07-27
Investigate using Dask for NetCDF handling


erikvansebille:
Reading of hydrodynamic fields is currently handled through the Unidata Netcdf python library. 

However, it could be that we can gain significant speed-ups by using Dask https://dask.pydata.org/en/latest/. This is certainly worth exploring for Parcels v1.0

guidov:
Possibly consider integrating  xarray (https://github.com/pydata/xarray) into parcels.

See: http://xarray.pydata.org/en/stable/dask.html

erikvansebille:
Fully agree, see also the very good write-up of the advantages of xarray and dask at https://pangeo-data.github.io/guidelines/

This is high on the priority list for v1.0

delandmeterp:
Now, Fields are loaded using xarray.
dask should improve the way we compute specific operations on those data, but we do not currently do such operations. At a Python level, we simply load those data, which are used for interpolation at the C level. We don't need then to use dask (for the moment).
However, we are working for a smarter (and cheaper) loading of the data into memory.

willirath:
On a related note, it would be interesting to try and use `dask.bag` to call `pset.execute` for sub-sets of a particle set. This would allow for running multiple instances of Parcels on different threads, processes, or even on workers on different machines.  If then JIT code then would use OpenMP, Parcels could be scaled to a massively parallel / distributed system without much of a logistical overhead for the user.

As far as I can see (after a brief test I did a while ago), there's a few things that currently prevent this from _just working_:

- Make sure that particlesets can be serialized: Dask distributes tasks by serializing them with JSON (I think ...). A first and naive test showed that some of the C types were not understood.

- Have an easy way of splitting / merging particlesets: This would allow for splitting the particleset, exectute the kernel in a parallel way, and thenmerge everything back into one particleset.

willirath:
The idea is as follows:

```python
import dask.bag as db
import numpy as np

# a simple time step
time_step = lambda p, v: p + 0.1 * v

# initial positions and a velocity vector
initial_positions = np.linspace(0, 1, 10)
velocities = np.random.randn(*initial_positions.shape)

# apply time step
next_positions = time_step(initial_positions, velocities)

# create dask bags of positions and velocities
initial_positions_bag = db.from_sequence(initial_positions, partition_size=2)
velocities_bag = db.from_sequence(velocities, partition_size=2)

# apply time step using dask bag (no computation yet)
[...]

PR #265, opened 2017-12-20
A suggestion for multiprocessing

Simnator101:
# A simple implementation of multiprocessing
One of the big hurdles I have experienced in working with parcels is that the execution of the kernels happens only on one logical core. This is somewhat problematic, because this causes quite a bottle neck for the program as the number of particles in your `ParticleSet` increases. To find a fix for this problem a simple implementation for multiprocessing was sought for kernel execution in JIT mode. With this is simply meant that the implementation ought not to burden the end user and be done automatically in parcels itself.

To that end code was added in the c code generator and to the `parcels.h` file to allow for automatic parallelisation of the for loop which contains the execution of a particle's kernel. This concurrent running of different particle kernels at the same time across cores causes roughly a 60 per cent decrease in processing time when instead of one logical core, eight can be used. I, therefore, think this pull request is a good inital take to get parcels to run faster through accessing more of the resources available on a given system.

## How the implementation was done
The `omp.h` header which is provided in the gcc toolchain allows easy parellisation through the addition of preprocessor statements and some lines of c code. In this case the necessary commands to multiprocess a piece of code were placed around the for loop which iterates over each particle. The new lines of code effectively tell the compiler to write a program which automatically splits the to-be-done work into equal chunks to be shared among the cores. For more information about gcc's implementation of this automatic parallelisation, see the [openmp specification](https://gcc.gnu.org/wiki/openmp) on the gcc website.

# A known problem
Multitprocessing brings to the table a new problem in parcels. The big current issue, which is semi-solved in this pull request, is a problem which is related to the `random` module in parcels. As you would likely know, parcels uses a special module to interface the demands for random numbers in Python to the random number generator in standard c which is basically the functions `rand()` and `srand(int seed)`. Well, when multiprocessing is implemented, a problem will start to occur: all the different cores will be assigned their own running version seeded with the seed which was set in the main thread. This is problematic as it will cause all the different threads to run the same sequence of *random* numbers.

To fix this, each thread was given a seed as defined by the function `time(NUL) ^ omp_get_thread_num()`. This provides us again with safe random numbers which are not dependent on one and another. However, what was lost in this fix, was the ability to set a seed and then to pull the same random numbers from the `random` module and a `Kernel`. This loss is reflected in thrown assertion errors in the test file which checks the kernel's language interpreter: it will say that the numbers from `random` and a kernel are not the same; which is of course true. A true fix then ought to be sought which maybe allows the parellised loop to pull from a shared number generator which can also be accessed through the `random` module. 

Anyway, the goal here is not to dictate what ought to be done, but to sketch what is certainly amiss within the current implementation of random numbers and to make clear that some decisions must be made concerning this. However, I do think that this pull request is a nice starting point to get robust parellisation into parcels.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,7 @@ dependencies:
   - python=2.7
   - cachetools>=1.0.0
   - cgen
+  - openmp
   - enum34
   - ffmpeg
   - flake8>=2.1.0
--- a/environment.yml
+++ b/environment.yml
@@ -11,6 +11,7 @@ dependencies:
   - flake8>=2.1.0
   - git
   - gcc
+  - gsl
   - jupyter
   - matplotlib=2.0.2
   - netcdf4>=1.1.9
--- a/environment_win.yml
+++ b/environment_win.yml
@@ -5,6 +5,7 @@ dependencies:
   - python=2.7
   - cachetools>=1.0.0
   - cgen
+  - openmp
   - enum34
   - ffmpeg
   - flake8>=2.1.0
--- a/environment_win.yml
+++ b/environment_win.yml
@@ -11,6 +11,8 @@ dependencies:
   - flake8>=2.1.0
   - git
   - m2w64-toolchain
+  - gsl
+  - m2w64-gsl
   - jupyter
   - matplotlib=2.0.2
   - netcdf4>=1.1.9
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -552,7 +552,8 @@ class LoopGenerator(object):
             shared += [field]
         for const, _ in const_args.items():
             args += [c.Value("float", const)]
-        fargs_str = ", ".join(['particles[p].time', 'sign_dt * __dt'] + list(field_args.keys()) + list(const_args.keys()))
+        fargs_str = ", ".join(['particles[p].time', 'sign_dt * __dt']
+                              + list(field_args.keys()) + list(const_args.keys()))
         # Inner loop nest for forward runs
         sign_dt = c.Assign("sign_dt", "dt > 0. ? 1 : -1")
         sign_end_part = c.Assign("sign_end_part", "endtime - particles[p].time > 0 ? 1 : -1")
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -524,9 +524,10 @@ class LoopGenerator(object):
     def generate(self, funcname, field_args, const_args, kernel_ast):
         ccode = []
 
-        # Add include for Parcels and math header
+        # Add include for OpenMP, Parcels, math and time header
         ccode += [str(c.Include("parcels.h", system=False))]
         ccode += [str(c.Include("math.h", system=False))]
+        ccode += [str(c.Include("time.h", system=False))]
 
         # Generate type definition for particle type
         vdecl = []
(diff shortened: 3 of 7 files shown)


willirath:
> To fix this, each thread was given a seed as defined by the function time(NUL) ^ omp_get_thread_num(). This provides us again with safe random numbers which are not dependent on one and another. However, what was lost in this fix, was the ability to set a seed and then to pull the same random numbers from the random module and a Kernel. This loss is reflected in thrown assertion errors in the test file which checks the kernel's language interpreter: it will say that the numbers from random and a kernel are not the same; which is of course true. A true fix then ought to be sought which maybe allows the parellised loop to pull from a shared number generator which can also be accessed through the random module.

Would it make sense to use a central seed to randomly draw the seeds of all the RNG's on the different OPM threads?

Simnator101:
Yes, @willirath , that would certainly be a possibilty, however --- because I was still dissatisfied with the quality of the randomness of c's rand() function in multiprocessing, I have taken a deeper look into random numbers and whether there were any attainable implementations which were better supported in a multiprocessing environment.

 I have found a extension of GNU project called gsl (GNU scientific library) which provides random number generators which not only provide better random numbers, but can also be shared in between multiprocessing threads. Furthermore, setting the seed in parcel works again, with the caveat that the random number generator in the JIT kernel is not completely linked to the one in `rng.py`.  So you may still see some desync between these two if you use the `rng.py` generator after running a kernel in JIT mode.

willirath:
Using the gsl sounds like a sound choice. And their default RNG is the same as the default one in Numpy, Matlab etc.

willirath:
I am not sure, if one shared RNG leads to deterministic results: The order in which the different threads draw random numbers from the central rng is likely not fixed even with static scheduling.

Then, to ensure deterministic results, one would still have to have one rng per thread.

willirath:
And I'm curious about the scaling you outlined: 60% less computing time with 8x more CPU sounds like the test has already been limited by memory or IO bandwidth rather than CPU. If this was the case, then going from one to two cores should already get you close to a 50% gain.

willirath:
Just to clarify (in case my comments seem too negative): I like the approach you outlined. This looks very promising!

Simnator101:
True, I think a big bottle neck in the kernel right now, is the fact that during the kernel execution the different thread all access about the same memory range to do the fieldset evaluation. It might, therefore, be interesting to see if limiting the to-be-used number of cores to a lower amount does not significantly impact processing time.

Next, concerning what you said about getting deterministic results for a multithreaded RNG, I think you are right with your comment that whilst the seed can be set in a central RNG, the order in which numbers are pulled cannot be assured. Frankly, I do not know how realistic it is to demand that `rng.py` must provide the same random numbers in order and value as the ones pulled in a multithreaded kernel execution.

erikvansebille:
This is very nice, @Simnator101 ! Very nice that you've dug into this and have this simple proof-of-concept of how parallelisation could be implemented straight away

In response to @willirath, I wouldn't be surprised at all if I/O indeed is the bottleneck. We see this all the time, especially with bigger hydrodynamic data sets. Which is why another important step towards faster/more efficient Parcels is to implement better netcdf handling (see also #216)

After the winter break, @delandmeterp will look more closely into efficiency and parallelisation, and will see how we can implement thisPR in `master`, including the random number issue and also the problems with the Travis CI.

Great work!

willirath:
The random-number issue touches on the more general question of how much determinism to aim for. From a user perspective, I would not expect Parcels to deliver bitwise identical results on different machines, after switching from JIT to scipy, or after changing any version of any package used in the integration.  But I would probably hope that results don't change if I run an experiment twice (without changing anything I can control).

erikvansebille:
This is now implemented in #625

PR #275, opened 2018-01-23
remove the name from grid creation parameters

delandmeterp:
There is no point of keeping  such a name

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -2,7 +2,6 @@ from parcels.loggers import logger
 import numpy as np
 from ctypes import Structure, c_int, c_float, c_double, POINTER, cast, c_void_p, pointer
 from enum import IntEnum
-import hashlib
 
 __all__ = ['GridCode', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'GridIndex', 'CGrid']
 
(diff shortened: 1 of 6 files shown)


parcels/gridset.py
```diff
@@ -38,6 +38,7 @@ def add_grid(self, field):
         if not existing_grid:
             for g in self.grids:
                 if g.name == grid.name:
+                    print grid.name
```
erikvansebille: Debug statement leftover?

parcels/grid.py
```diff
@@ -49,14 +50,14 @@ class RectilinearGrid(Grid):
...
-    def __init__(self, name, lon, lat, time, time_origin, mesh):
+    def __init__(self, lon, lat, time, time_origin, mesh):
         assert(isinstance(lon, np.ndarray) and len(lon.shape) == 1), 'lon is not a numpy vector'
         assert(isinstance(lat, np.ndarray) and len(lat.shape) == 1), 'lat is not a numpy vector'
         assert (isinstance(time, np.ndarray) or not time), 'time is not a numpy array'
         if isinstance(time, np.ndarray):
             assert(len(time.shape) == 1), 'time is not a vector'
 
-        self.name = name
+        self.name = hashlib.md5().hexdigest()
```
erikvansebille: Would it make sense then to call this `self.hash` rather than `self.name`?

parcels/gridset.py
```diff
@@ -37,10 +37,10 @@ def add_grid(self, field):
 
         if not existing_grid:
```
erikvansebille: Is it not the idea of the `hash` that it is much easier also to compare the grid? Would `existing_grid` not be a matter of comparing hashes, now? If so, we also don't need the if statements here, right?

delandmeterp: All the things we do are wrong.

We should compare the object themselves, neither a string nor a hash (it's a private variable, used to compare the objects)

I did it originally, because I was worrying of breaking something while passing the objects to C. So this was the way to compare them. This is not necessary anymore. Clean up is coming

PR #278, opened 2018-01-24
C-kernels

delandmeterp:
Within kernels, we can now read a c function, which is defined in the same python file.

If the kernel is called in `Scipy` mode, this function must also be defined in Python independently.

Before merging:
* Add the option of including a header file instead of reading only string
* Auto translation of \n to \\n
* Check what @Jacketless needs


tests/test_kernel_language.py
```diff
@@ -150,3 +150,29 @@ class TestParticle(ptype[mode]):
...
+        print('%g %g' % (U.data[0, 0, 0], U.data[0, 0, 1]))
+        print('%g %g' % (U.data[0, 1, 0], U.data[0, 1, 1]))
+
+    c_func_lib = """
+             static inline void c_func(CField *f)
+             {
+               printf("Calling the C func\\n");
+               float (*data)[f->xdim] = (float (*)[f->xdim]) f->data;
+               printf("%g %g\\n", data[0][0], data[0][1]);
+               printf("%g %g\\n", data[1][0], data[1][1]);
```
erikvansebille: Nice, but would it not make more sense to have a kernel that actually *does* something? I.e. not only print, but also update the particle data? Then we can also add an `assert` to check for more than just whether compilation succeeds

tests/test_kernel_language.py
```diff
@@ -150,3 +150,29 @@ class TestParticle(ptype[mode]):
...
+             }
+             """
+
+    def ckernel(particle, fieldset, time, dt):
+        c_func(fieldset.U)
+
+    kernel = pset.Kernel(ckernel, lib=c_func_lib)
+    pset.execute(kernel, endtime=1., dt=1.)
```
erikvansebille: Note we can also run with `runtime=0` now ;-)

parcels/codegenerator.py
```diff
@@ -543,7 +546,7 @@ def __init__(self, fieldset, ptype=None):
         self.fieldset = fieldset
         self.ptype = ptype
 
-    def generate(self, funcname, field_args, const_args, kernel_ast):
+    def generate(self, funcname, field_args, const_args, kernel_ast, lib):
```
erikvansebille: Should this thing be called `lib`? Or e.g. `include` or ??

PR #280, opened 2018-01-24
Allow pset init single location

erikvansebille:
Allow for initialising of ParticleSet with single floats or ints.

Also cleaned up the code to convert numpy arrays to lists for lon, lat and depth in ParticleSet init

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -45,10 +45,17 @@ class ParticleSet(object):
         self.fieldset = fieldset
         self.fieldset.check_complete()
 
-        lon = lon.flatten() if isinstance(lon, np.ndarray) else lon
-        lat = lat.flatten() if isinstance(lat, np.ndarray) else lat
+        def convert_to_list(dim):
+            if isinstance(dim, (int, float)):
+                return [dim]
+            elif isinstance(dim, np.ndarray):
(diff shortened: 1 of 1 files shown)


parcels/particleset.py
```diff
@@ -41,14 +41,21 @@ class ParticleSet(object):
...
+            return var
+
+        lon = convert_to_list(lon)
+        lat = convert_to_list(lat)
         depth = np.ones(len(lon)) * fieldset.U.grid.depth[0] if depth is None else depth
-        depth = depth.flatten() if isinstance(depth, np.ndarray) else depth
+        depth = convert_to_list(depth)
         assert len(lon) == len(lat) and len(lon) == len(depth)
 
         time = time.tolist() if isinstance(time, np.ndarray) else time
```
delandmeterp: Why is `time` not behaving like `lon`, `lat` and `depth`?
If you change it to use `convert_to_list`, don't forget that `double` needs to be added to the type list)

erikvansebille: Because `time` can also be a `datetime` object or None (when setting time is deferred to .execute, see #261, I thought it better to keep `time` separate from the other variables. Otherwise, the function `convert_to_list` would become much more complicated. OK?

PR #282, opened 2018-01-29
Rewrite setup to match conda-forge version

willirath:
This is relative to `v1.0` for now.

PR #283, opened 2018-01-29
[WIP] Generalize gcc dependency relto v1.0

willirath:
This does not install `gcc` but the newer Anaconda compilers (see <https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html>).  Tests will fail until [`GNUCompiler`](https://github.com/OceanParcels/parcels/blob/master/parcels/compiler.py#L62) honors environment variables hinting to the GCC binary to use.


docs/index.rst
```diff
@@ -87,14 +87,14 @@ Parcels depends on a working Python installation, a netCDF installation, a C
...
+requirements. This is the “batteries-included” solution probably suitable for most users.
 Note that if you already have an existing Python setup that you wish to use for Parcels, you should follow the instructions at the bottom of this section.
 
 The five steps below are the installation instructions for Linux and macOS. Note that for Windows, steps 2 and 5 are slightly different, see :ref:`installing-windows` below.
 
 1. Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/. All the code below assumes that you download the Python-2 version.
 
-2. Download Parcels' `environment.yml <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment.yml>`_ file, navigate to the directory where you saved that file, and run::
+2. Download Parcels' environment definition for either Linux (`environment_linux.yml <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_linux.yml>`_) or for macOS (`environment_osx.yml <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_osx.yml>`_), save it as `environment.yml`, navigate to the directory where you saved that file, and run::
```
erikvansebille: The 'save it as `environment.ml` might be misread easily. How about changing 
```
conda env create -f environment.yml -n py2_parcels
```
to 
```
conda env create -f environment_OSHERE.yml -n py2_parcels
```
and then add `where OSHERE is either osx or linux, depending on your Operating System`

willirath: I'd take it even a step further: I've had the opportunity to witness a few people (by sitting next to them / talking to them on the phone). The majority already overlooked that they were supposed to download the file manually.  This is probably because they immediately focused on the inline code section.

So we could just go for:

```
curl https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_<OSHERE>.yml -o environment.yml
source $HOME/miniconda2/bin/activate root
conda env create -f environment.yml -n py2_parcels
source $HOME/miniconda2/bin/activate py2_parcels
pip install git+https://github.com/OceanParcels/parcels.git@master
```

willirath:
Turns out the compiler already falls back to an env var.  So the current status (in `master`) of `parcels/compiler.py` does not prevent us from just installing and using the newer Anaconda compilers.

In the travis log, we see that although all tests use `GNUCompiler` which specifies some variant of `gcc`,
 we end up using the binary defined in `$CC`: <https://travis-ci.org/OceanParcels/parcels/builds/334797586#L894>  This is because in [parcels/compiler.py#L37](https://github.com/willirath/parcels/blob/daf0990a07bb0c29baa4577b8eb4e1f0c3a98714/parcels/compiler.py#L37), an explicitly given compiler is only used if there is no environment variable.

As I'm not sure that this logic really reflects the intended bahaviour, I'll open an issue for some general review of the `Compiler` class.

Here, I think, just adding the separate environment files for linux and osx and adapting the documentation is sufficient to prevent users from running into problems from the deprecated `gcc` package.

willirath:
Note that #282 might quite radically change the installation instructions and at the same time simplify what's necessary to add to the documentation in this PR.

To wrap this PR up, we'll need to create three env files covering the dependencies for Linux,
 for OSX, and for Windows as detailed in the [recipe for the conda-forge package](https://github.com/conda-forge/parcels-feedstock/blob/master/recipe/meta.yaml), and then have Travis use the Linux file.

willirath:
@erikvansebille This PR now only adds / modifies the env files which use what I learned during <https://github.com/conda-forge/parcels-feedstock/pull/9> and <https://github.com/conda-forge/parcels-feedstock/commit/d2c04f79aee4d9d11299ce02f7d68303fc34eaaf>.  I've reverted changes to the `compiler.py` as these should only be made after some discussion in #284.

willirath:
@erikvansebille: Documentation updates accounting for the new env files are not done yet.  But let's wait what you think of #282.

willirath:
I think it's all done.

Changes:

- There are three env files now which define the dependencies for Linux, OSX, and Windows.  They are mirroring the dependencies of the conda-forge package.
- Travis runs everything that was done before this PR on a Linux VM.
- Travis runs `parcels_get_examples` and the units tests from `tests/` on an OSX VM (I've skipped the notebooks and flake / link checking).
- Documentation points to the three env files in developer's installation instructions.

@erikvansebille: This is quite messy now.  Could you have a look and rebase / squash before merge using the Github GUI?  I could also rebase locally and open a new PR to give you a cleaner history to merge from.

.travis.yml
```diff
@@ -1,35 +1,51 @@
...
+      # Set up display to be able to plot in linux
+      if [[ "${TRAVIS_OS_NAME}" = "linux" ]]; then
+        export DISPLAY=:99.0;
+        sh -e /etc/init.d/xvfb start;
+        sleep 3;
+      fi
     - py.test -v -s tests/
-    - py.test -v -s --nbval-lax examples/
-    - make linkcheck -C docs
+    - parcels_get_examples examples/
```
erikvansebille: parcels_get_examples is only required when examples is actually run; i.e. only on linux. Looking at travis output, even mac still downloads the data but doesn't seem to do anything with it

.travis.yml
```diff
@@ -1,35 +1,51 @@
...
-python:
-    - "2.7"
+matrix:
+  include:
+    - os: linux
+      python: 2.7
+      env: MINICONDA_NAME=Miniconda2-latest-Linux-x86_64.sh
+      sudo: required
+    - os: osx
+      language: generic
```
erikvansebille: is there a reason that language here is `generic`?

willirath: Yes. The OSX images that are provided by Travis are much less diverse. And there is no pre-installed python image available.

The reason that we first define python as a language for all jobs and then reset this for OSX is to account for weird (and probably buggy) behavior of the Travis scheduler which falls back to using ruby for everything of there no global language defined.

.travis.yml
willirath: I originally left this switched on intentionally: I wanted to make sure that at least downloading the examples also works for OSX. But that's essentially amounting to checking whether the entry point in the setup.py is working.

I think we could make this linux-only as well.

willirath:
I'll make the example download Linux-only as well and rebase in a fresh PR then.

willirath:
See #294 for a rebased version.  Closing here.

PR #288, opened 2018-01-31
Simple timer

delandmeterp:
Timer class to measure where the CPU time is spent over the code.
Examples of usage of the class shown in example_stommel.py


parcels/tools/timer.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import print_function
+import time
+
+
+class Timer():
+    def __init__(self, name, parent=None, start=True):
+        self._start = -1
+        self._end = -1
```
erikvansebille: More pythonic to initialise these to None? And then later on do e.g. the oneliner:
```
def local_time(self):
    return self._t + time.time() - self._start if self._start else self._t
```

parcels/tools/timer.py
```diff
@@ -0,0 +1,51 @@
...
+    def print_tree(self, step=0, root_time=0, parent_time=0):
+        time = self.local_time()
+        if step == 0:
+            root_time = time
+        print(('(%3d%%)' % round(time/root_time*100)), end='')
+        for i in range(step+1):
+            print('  ', end='')
+        if step > 0:
+            print('(%3d%%) ' % round(time/parent_time*100), end='')
+        print("Timer '%s': %1.1e s" % (self._name, time))
```
erikvansebille: Can we use `timedelta` objects throughout for prettier printing here? I.e. not everything in seconds, but also in minutes, hours where appropriate?

Issue #79, opened 2016-06-14
API for time-coupling / complex sub-timestepping


mlange05:
Various features in parcels need to adhere to a specific sub-timestepping, such as file I/O and animation and deriving particle densities. This becomes quite complex once we start having more than one particle set in a simulation, say a predator-prey pair with different file I/O frequencies, a global animation frequency and additional synchronisation between the sets (lockstepping). We need to (at least conceptually) define an API that allows users to specify these things easily and concisely, especially the co-execution of two particle sets in some lockstepping.

One suggestions would be to keep set/store the file I/O frequency with the `ParticleFile` object, and similarly devise a `Viewer` class of some kind that performs live animations and stores a single global animation frequency.  Similarly, diagnostic density fields might define their own update frequency, but I am open to suggestions for how we should define things for co-exection of two `ParticleSets`.

Once these components all have their own time-stepping frequency, a sort of event registry framework might then be used to allow the runtime (currently `ParticleSet.execute()`) to then derive the appropriate `interval` and then invoke the corresponding event, ie file I/O, once that interval has been performed.

As always, any thoughts and suggestions are very welcome.

PR #289, opened 2018-02-01
Simpler time loop in particleset.execution()

delandmeterp:
This will also enable a better management of updating the fieldsets


parcels/examples/parcels_tutorial.ipynb
```diff
@@ -981,7 +981,7 @@
     "editable": true
    },
    "source": [
-    "Now one of the neat features of Parcels is that the particles can be plotted as a movie *during execution*, which is great for debugging. To rerun the particles while plotting them on top of the zonal velocity field (`fieldset.U`), first reinitialise the `ParticleSet` and then re-execute. However, now rather than saving the output to a file, display a movie using the `show_movie` keyword, in this case with the zonal velocity `fieldset.U` as background"
+    "Now one of the neat features of Parcels is that the particles can be plotted as a movie *during execution*, which is great for debugging. To rerun the particles while plotting them on top of the zonal velocity field (`fieldset.U`), first reinitialise the `ParticleSet` and then re-execute. However, now rather than saving the output to a file, display a movie using the `outputdt` display frequency, in this case with the zonal velocity `fieldset.U` as background"
```
erikvansebille: should be `moviedt`?

parcels/examples/tutorial_jit_vs_scipy.ipynb
```diff
@@ -76,7 +76,7 @@
    "source": [
     "As you can see above, **Scipy mode took more than 100 times as long** (0.17 seconds versus 17.46 seconds) as the JIT mode!!\n",
     "\n",
-    "This is just an illustrative example, depending on the number of calls to `AdvectionRK4`, the size of the `FieldSet`, the size of the `pset`, the ratio between `dt` and `interval` in the `.execute` etc, the difference between JIT and Scipy can vary significantly. However, JIT will almost always be faster!\n",
+    "This is just an illustrative example, depending on the number of calls to `AdvectionRK4`, the size of the `FieldSet`, the size of the `pset`, the ratio between `dt` and `outputdt` in the `.execute` etc, the difference between JIT and Scipy can vary significantly. However, JIT will almost always be faster!\n",
```
erikvansebille: As a side-comment: now we can use the new `timer` class for this tutorial!

parcels/particleset.py
```diff
@@ -223,25 +223,34 @@ def remove(self, indices):
...
+        :param endtime: End time for the timestepping loop.
+                        It is either a datetime object or a positive double.
         :param runtime: Length of the timestepping loop. Use instead of endtime.
-        :param dt: Timestep interval to be passed to the kernel
-        :param interval: Interval for inner sub-timestepping (leap), which dictates
-                         the update frequency of file output and animation.
+                        It is either a timedelta object or a positive double.
+        :param dt: Timestep interval to be passed to the kernel.
+                   It is either a timedelta object or a double.
+                   Negative value means backward in time simulation.
```
erikvansebille: Change to `Use a negative value for a backward-in-time simulation`?

parcels/particleset.py
```diff
@@ -223,25 +223,34 @@ def remove(self, indices):
...
+        :param outputdt: Interval for inner sub-timestepping (leap), which dictates
+                         the update frequency of file output.
+                         It is either a timedelta object or a positive double.
+                         None value means no file output.
```
erikvansebille: I'm thinking; why don't we take the `outputdt` from the output_file? Would that not make more sense? Because now users have to provide *both* a `outputdt` and an `output_file` if they want output. 

If instead they specify `outputdt` on `ParticleFile.__init__`, then here it requires only a `output_file`. That's also how we do `repeatdt`, by the way

parcels/particleset.py
```diff
@@ -259,75 +268,96 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
...
-            logger.warning("Negating interval because running in time-backward mode")
-
-        if abs(endtime-_starttime) < 1e-5 or interval == 0 or dt == 0 or runtime == 0:
-            timeleaps = 1
+
+        if abs(endtime-_starttime) < 1e-5 or outputdt == 0 or dt == 0 or runtime == 0:
             dt = 0
             runtime = 0
             endtime = _starttime
+            outputdt = 0
```
erikvansebille: So means there will always be output when e.g. `runtime==0`?

parcels/particleset.py
```diff
@@ -259,75 +268,96 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
...
+        else:
+            next_prelease = np.infty * np.sign(dt)
+        next_output = time + outputdt * np.sign(dt)
+        next_movie = time + moviedt * np.sign(dt)
+        next_input = np.infty * np.sign(dt)  # Not used yet
+
+        tol = 1e-12
+        while (time < endtime and dt > 0) or (time > endtime and dt < 0) or dt == 0:
+            if dt > 0:
+                leaptime = min(next_prelease, next_input, next_output, next_movie, endtime)
```
erikvansebille: why use `leaptime`? Why not simply `time = ...` and then get rid of line 341?

erikvansebille:
This PR cleans up the timer API under the hood in `ParticleSet.execute()` (and thereby fixes #79) 

Note that we had to make some changes to the arguments of `.execute()`:
- the `interval` argument has been renamed `outputdt` and should now be set when creating the `ParticleFile` object
- the `show_movie` argument has been split into a `moviedt` argument (to set the frequency of the animation) and a `movie_background_field` argument that determines which background field to show (either a `Field` object, or the string `vector`) or the default `None` for no background field
- `endtime` cannot be a `timedelta` object anymore. Only valid formats for `endtime` are a `datetime` object or a `double`. If you want to give a `timedelta` object, use the `runtime` argument
- Except for `dt`, all arguments controlling intervals should always be positive, regardless of whether you run in forward or backward mode. Hence, to change the direction of a run, the only thing to change is to negate `dt`

PR #290, opened 2018-02-02
Write particledata on delete

### LINKED ISSUE #260: Different ways of saving particle data
Currently I only found one method to save the data of particles, which saves all trajectories of all particles.
However, could it be nice if the code incorporates different ways of saving data? Then you can save data in a more efficient way.

For instance, I am tracking a sinking particle in 3D back in time, which is released at the bottom. I only need the data of this particle once it reaches the ocean surface. I do not need all the data during the whole trajectory of any particle.

erikvansebille:
Adding the argument `write_ondelete` to `ParticleFile.__init__`. If `True`, this only writes data of particles when they are deleted.

This fixes #260, suggested by @pdnooteboom


tests/test_kernel_execution.py
```diff
@@ -54,7 +54,7 @@ def test_execution_runtime(fieldset, mode, start, end, substeps, dt, npart=10):
     pset = ParticleSet(fieldset, pclass=ptype[mode], time=start,
                        lon=np.linspace(0, 1, npart, dtype=np.float32),
                        lat=np.linspace(1, 0, npart, dtype=np.float32))
-    t_step = (end - start) / substeps
+    t_step = abs(end - start) / substeps
     for _ in range(substeps):
         pset.execute(DoNothing, runtime=abs(t_step), dt=dt)
         start += t_step
```
delandmeterp: I have already put the abs at line 59 (not at 57) because I didn't want to affect line 60.
But actually line 60 is unused.
So abs should be removed from l59 and l60 should be removed, no?

parcels/particlefile.py
```diff
@@ -165,7 +174,7 @@ def write(self, pset, time, sync=True):
                     logger.warning("Option to_write='once' is not fully functional in indexed mode! Particle properties of such variables are not written for newly added particles.")
                 ind = np.arange(pset.size) + self.idx
                 self.id[ind] = np.array([p.id for p in pset])
-                self.time[ind] = time
+                self.time[ind] = np.array([p.time for p in pset])
```
delandmeterp: This was not necessary before, because all the particles were at endtime, but now they are at the time they were deleted, right?

tests/test_particle_sets.py
```diff
@@ -329,3 +330,26 @@ class MyParticle(ptype[mode]):
...
+    lon = np.linspace(0.05, 0.95, npart, dtype=np.float32)
+    lat = np.linspace(0.95, 0.05, npart, dtype=np.float32)
+    pset = ParticleSet(fieldset, pclass=ptype[mode], lon=lon, lat=lat)
+
+    outfile = pset.ParticleFile(name=filepath, write_ondelete=True, type="indexed")
+    pset.execute(move_west, runtime=1.1, dt=0.1, output_file=outfile,
+                 recovery={ErrorCode.ErrorOutOfBounds: DeleteP})
+    ncfile = Dataset(filepath+".nc", 'r', 'NETCDF4')
+    lon = ncfile.variables['lon'][:]
+    assert (lon.size == npart)
```
delandmeterp: In this test, all the particles are deleted.
Shouldn't we keep one in the domain, and check that this one is not written?

parcels/particlefile.py
erikvansebille: yes, exactly!

PR #291, opened 2018-02-02
Use the same env-name throughout

willirath:
This introduces a minor fix to the installation instructions for an arbitrary Git reference:  The example-env name was not the same as in the installation instructions above.


willirath:
(I've cancelled the Travis job that was triggered by the push.)

PR #294, opened 2018-02-08
Use Anaconda-5 compilers and add OSX tests

willirath:
(This is a rebased version of #283.)

This uses the new compiler suites that come with Anaconda 5, splits up the
example environment files into the three OSs Linux, OSX, Windows, and adapts
the documentation accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,55 @@
+language: python
+
 notifications:
   email: false
 
-language: python
-python:
-    - "2.7"
+matrix:
+  include:
+    - os: linux
+      python: 2.7
+      env: MINICONDA_NAME=Miniconda2-latest-Linux-x86_64.sh
+      sudo: required
+    - os: osx
+      language: generic
+      env: MINICONDA_NAME=Miniconda2-latest-MacOSX-x86_64.sh
 
-sudo: false
-
-before_install:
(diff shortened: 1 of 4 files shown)


environment_win.yml
```diff
@@ -5,13 +5,14 @@ dependencies:
...
   - enum34
-  - ffmpeg
+  - ffmpeg>=3.2.3,<3.2.6
   - flake8>=2.1.0
   - git
-  - m2w64-toolchain
   - jupyter
-  - matplotlib=2.0.2
+  - m2w64-toolchain
+  - matplotlib 2.0.2
```
erikvansebille: does this not need to be a `matplotlib=2.0.2`? Is the `=` not necessary on windows?

willirath: Conda is relatively relaxed here: It will accept specification with and without a `=`.

In this particular file, the `=` was removed, because I copied the dependencies from the conda-forge recipe, where the notation matters. (Conda build does not accept an equal sign.)

willirath: Conda is relatively relaxed here: It will accept specification with and without a `=`.

In this particular file, the `=` was removed, because I copied the dependencies from the conda-forge recipe, where the notation matters. (Conda build does not accept an equal sign.)

PR #295, opened 2018-02-12
WIP: Fix travis builds for Python 3

willirath:
This introduces quite a few changes:

1. There is an env file per job now: `environment_py3_linux.yml`, ...

2. Travis runs all jobs with `language: generic` and picks the Python
version via an environment variable.

3. Travis uses Miniconda 3 for all jobs.

The resulting build fails with only one error in `test_kernel_language.py:test_print[jit]`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,26 +6,36 @@ notifications:
 matrix:
   include:
     - os: linux
-      python: 2.7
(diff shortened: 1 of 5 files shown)


willirath:
One important bit was the prioritization of the defaults channel in the env files.  This solves problems stemming from Anaconda using their own compilers in conda and pip and the fact that conda-forge has not yet moved to these new build environments.

Issue #287, opened 2018-01-31
Improve handling of netcdf output files


erikvansebille:
The `type="indexed"` version of `ParticleFile.__init__` scales poorly with number of particles, and can be extremely slow and cumbersome to work with. However, new NetCDF versions support multiple Unlimited Dimensions (https://www.unidata.ucar.edu/software/netcdf/docs/unlimited_dims.html), so it might be worth exploring if there is not a simpler way to store the particle data in array format, but with each particle having its own `time` array too

Also, there are some inconsistencies in the way that NetCDF output filenames are handled in Parcels. For example, `ParticleFile.__init__` expects a filename without the `.nc` extension. But `plotParticleFile` expects a filename with that extension. 
Ideally, we would want filenames to default to `.nc` if no extension given, but otherwise use the extension provided in the filename. And then change all examples/tutorials to use the `.nc` extension for output files throughout

erikvansebille:
Fixed in #296

PR #296, opened 2018-02-14
Particlefile updates: removing type="indexed" mode as no need for anymore

erikvansebille:
This PR implements the changes suggested in #287. It does away with the `type="indexed"` for `ParticleFiles`, since netcdf supports two unlimited dimensions. So files can be expanded both in time, as well as in the particle id dimensions. This also means we don't need the `convert_IndexedOutputToArray.py` script anymore

The PR also adds a small fix where ParticleFile names can be given either with or without extension (`.nc` or `.nc4`)


parcels/particlefile.py
```diff
@@ -31,43 +27,30 @@ class ParticleFile(object):
...
+        self.dataset = netCDF4.Dataset(fname, "w", format="NETCDF4")
         self.dataset.createDimension("obs", None)
-        if self.type is 'array':
-            self.dataset.createDimension("trajectory", particleset.size)
-            coords = ("trajectory", "obs")
-        elif self.type is 'indexed':
-            coords = ("obs")
-        else:
-            raise RuntimeError("ParticleFile type must be either 'array' or 'indexed'")
+        self.dataset.createDimension("trajectory", None)
```
delandmeterp: It is not a comment concerning the changes, but why do we call the particle dimension "trajectory", and not "particle" or "id"?

erikvansebille: That's because we want to stay CF-compliant: https://www.nodc.noaa.gov/data/formats/netcdf/v2.0/trajectoryIncomplete.cdl

parcels/particlefile.py
```diff
@@ -142,46 +125,34 @@ def write(self, pset, time, sync=True, deleted_only=False):
...
-                self.z[ind] = np.array([p.depth for p in pset])
+            if pset.size > 0:
+                fids = self.id[:]
+                if fids.shape[1] > 0:
+                    fids = fids[:, -1]
+                pids = [p.id for p in pset]
+                in_set = np.in1d(fids, pids, assume_unique=True)
+                oldinds = np.arange(len(fids))[in_set]
+
+                not_in_set = np.in1d(pids, fids, assume_unique=True, invert=True)
```
erikvansebille: There are two calls to `np.in1d` here, which I fear can be really slow once we get to millions of particles. The issue is that the `particle.id` does not need to be the `trajectory` index in the file (e.g. when multiple particlesets are defined). 
While the current implementation always works, it can be slow. Alternatives could be
1) Get rid of the global `lastID` in `particle.py`, so that the `particle.id` is always the trajectory ID in the file
2) Add an extra Variable `particle.trajinfile` or similar that stores a particle's trajectory ID in the ParticleFile

Option 2 obviously will be easy to implement, but gives extra memory overhead. So I'd first try if I can implement option 1. To be honest, I can't entirely remember anymore exactly *why* we needed the `lastID` variable in `particle.py`

PR #297, opened 2018-02-19
get rid of the dangerous reshape option in the field creation

delandmeterp:


parcels/field.py
```diff
@@ -250,14 +250,23 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
...
+        # C-contiguous memory layout for JIT mode.
         if transpose:
-            # Make a copy of the transposed array to enforce
-            # C-contiguous memory layout for JIT mode.
             self.data = np.transpose(self.data).copy()
-        if self.grid.zdim > 1:
-            self.data = self.data.reshape((self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim))
         else:
-            self.data = self.data.reshape((self.grid.tdim, self.grid.ydim, self.grid.xdim))
+            self.data = self.data.copy()
```
erikvansebille: So why do we need to copy() even if we're not transposing?

parcels/field.py
```diff
@@ -250,14 +250,23 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
...
+        if self.grid.tdim == 1:
+            if len(self.data.shape) < 4:
+                self.data = self.data.reshape(sum(((1,), self.data.shape), ()))
+        if self.grid.zdim == 1:
+            if len(self.data.shape) == 4:
+                self.data = self.data.reshape(sum(((self.data.shape[0],), self.data.shape[2:]), ()))
+        if len(self.data.shape) == 4:
+            assert self.data.shape == (self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim)
+        else:
+            assert self.data.shape == (self.grid.tdim, self.grid.ydim, self.grid.xdim)
```
erikvansebille: So we don't allow for fields to be (z, y, x)?

parcels/field.py
delandmeterp: If the user does a transpose in its own code, the same issue occurs...

parcels/field.py
delandmeterp: Yes we allow, but it does not exist. Since a field (z,y,x) is automatically changed to a (t,z,y,x) in the lines above (it was already the case before)

parcels/field.py
erikvansebille: But most people won't transpose their own code. This sounds like a hefty overhead for most usecases. Is there a way to check if a numpy array has been transposed, so that we only do it then?

PR #301, opened 2018-02-22
better error message when shape of field.data is not as expected

delandmeterp:
Error message looks like:

WARNING: Casting field data to np.float32
  File "test.py", line 37, in <module>
    field_set = FieldSet.from_data(data, dimensions,mesh='spherical', allow_time_extrapolation=True)#, transpose=False)
  File "/Users/delandmeter/sources/parcels/parcels/fieldset.py", line 73, in from_data
    allow_time_extrapolation=allow_time_extrapolation, time_periodic=time_periodic, **kwargs)
  File "/Users/delandmeter/sources/parcels/parcels/field.py", line 233, in __init__
    self.data.shape[0], self.data.shape[1], self.data.shape[2]))
AssertionError: Field U expect a data shape of a [tdim, ydim, xdim] = [1 81 131], but data is [1 131 81]

@erikvansebille , do you think we should even improve it? The message is perfect for someone using tdim > 1. But if original array has shape (131 81), it will have the following message, since we add the t dimension to the original data, the final message looks like above.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -143,7 +143,12 @@ class Field(object):
     """Class that encapsulates access to field data.
 
     :param name: Name of the field
-    :param data: 2D, 3D or 4D numpy array of field data
+    :param data: 2D, 3D or 4D numpy array of field data.
+           1. If data shape is [xdim, ydim], [xdim, ydim, zdim], [xdim, ydim, tdim] or [xdim, ydim, zdim, tdim],
+              Use the flag transpose=True
+           2. If data shape is [ydim, xdim], [zdim, ydim, xdim], [tdim, ydim, xdim] or [tdim, zdim, ydim, xdim],
+              Use the flat transpose=False
+           3. If data has any other shape, you first need to reorder it
     :param lon: Longitude coordinates (numpy vector or array) of the field (only if grid is None)
     :param lat: Latitude coordinates (numpy vector or array) of the field (only if grid is None)
     :param depth: Depth coordinates (numpy vector or array) of the field (only if grid is None)
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -145,9 +145,10 @@ class Field(object):
     :param name: Name of the field
     :param data: 2D, 3D or 4D numpy array of field data.
            1. If data shape is [xdim, ydim], [xdim, ydim, zdim], [xdim, ydim, tdim] or [xdim, ydim, zdim, tdim],
-              Use the flag transpose=True
+              whichever is relevant for the dataset,
+              use the flag transpose=True
            2. If data shape is [ydim, xdim], [zdim, ydim, xdim], [tdim, ydim, xdim] or [tdim, zdim, ydim, xdim],
-              Use the flat transpose=False
+              use the flag transpose=False
            3. If data has any other shape, you first need to reorder it
     :param lon: Longitude coordinates (numpy vector or array) of the field (only if grid is None)
     :param lat: Latitude coordinates (numpy vector or array) of the field (only if grid is None)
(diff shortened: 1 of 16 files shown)


parcels/fieldset.py
```diff
@@ -40,6 +40,11 @@ def from_data(cls, data, dimensions, transpose=True, mesh='spherical',
 
         :param data: Dictionary mapping field names to numpy arrays.
                Note that at least a 'U' and 'V' numpy array need to be given
+               1. If data shape is [xdim, ydim], [xdim, ydim, zdim], [xdim, ydim, tdim] or [xdim, ydim, zdim, tdim],
+                  Use the flag transpose=True (default value)
+               2. If data shape is [ydim, xdim], [zdim, ydim, xdim], [tdim, ydim, xdim] or [tdim, zdim, ydim, xdim],
+                  Use the flat transpose=False
```
erikvansebille: `flat` should be `flag`

Same of course for docstring in `field.py`

parcels/fieldset.py
```diff
@@ -40,6 +40,11 @@ def from_data(cls, data, dimensions, transpose=True, mesh='spherical',
...
+               3. If data has any other shape, you first need to reorder it
```
erikvansebille: add `to dimensions [ydim, xdim], [zdim, ydim, xdim], [tdim, ydim, xdim] or [tdim, zdim, ydim, xdim], whichever is relevant for your data set`

Same of course for docstring in `field.py`

parcels/field.py
```diff
@@ -222,9 +227,15 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
...
         if len(self.data.shape) == 4:
-            assert self.data.shape == (self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim)
+            assert self.data.shape == (self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim), \
+                                      ('Field %s expect a data shape of a [tdim, zdim, ydim, xdim] = [%d %d %d %d], but data is [%d %d %d %d]' %
+                                       (self.name, self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim,
+                                        self.data.shape[0], self.data.shape[1], self.data.shape[2], self.data.shape[3]))
         else:
-            assert self.data.shape == (self.grid.tdim, self.grid.ydim, self.grid.xdim)
+            assert self.data.shape == (self.grid.tdim, self.grid.ydim, self.grid.xdim), \
+                                      ('Field %s expect a data shape of a [tdim, ydim, xdim] = [%d %d %d], but data is [%d %d %d]' %
```
erikvansebille: change `Field %sexpect` to `Field %s: expecting`

Also, add that flag `transpose=True` might be used to fix this

parcels/field.py
```diff
@@ -143,7 +143,13 @@ class Field(object):
     """Class that encapsulates access to field data.
 
     :param name: Name of the field
-    :param data: 2D, 3D or 4D numpy array of field data
+    :param data: 2D, 3D or 4D numpy array of field data.
+           1. If data shape is [xdim, ydim], [xdim, ydim, zdim], [xdim, ydim, tdim] or [xdim, ydim, zdim, tdim],
+              whichever is relevant for the dataset,
+              use the flag transpose=True
```
erikvansebille: Why over three lines here? Can be two lines if you remove the return between dataset, and use

parcels/fieldset.py
```diff
@@ -34,12 +34,18 @@ def __init__(self, U, V, fields={}):
...
-    def from_data(cls, data, dimensions, transpose=True, mesh='spherical',
+    def from_data(cls, data, dimensions, transpose=False, mesh='spherical',
                   allow_time_extrapolation=True, time_periodic=False, **kwargs):
         """Initialise FieldSet object from raw data
 
         :param data: Dictionary mapping field names to numpy arrays.
                Note that at least a 'U' and 'V' numpy array need to be given
+               1. If data shape is [xdim, ydim], [xdim, ydim, zdim], [xdim, ydim, tdim] or [xdim, ydim, zdim, tdim],
+                  whichever is relevant for the dataset,
+                  use the flag transpose=True
```
erikvansebille: See comment above

PR #304, opened 2018-02-27
For s grids, depth has the same format as the fields

delandmeterp:
+ there is a check for depth format
+ there is a check depth is c_contiguous

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -191,8 +191,12 @@ class RectilinearSGrid(RectilinearGrid):
 
         self.gtype = GridCode.RectilinearSGrid
         self.depth = depth
-        self.zdim = self.depth.shape[2]
+        self.zdim = self.depth.shape[-3]
+        assert self.xdim == self.depth.shape[-1], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
+        assert self.ydim == self.depth.shape[-2], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
         self.z4d = len(self.depth.shape) == 4
+        if self.z4d:
+            assert self.tdim == self.depth.shape[0], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
         if not self.depth.dtype == np.float32:
             logger.warning_once("Casting depth data to np.float32")
             self.depth = self.depth.astype(np.float32)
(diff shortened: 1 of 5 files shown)


parcels/grid.py
```diff
@@ -191,8 +191,12 @@ def __init__(self, lon, lat, depth, time=None, time_origin=0, mesh='flat'):
 
         self.gtype = GridCode.RectilinearSGrid
         self.depth = depth
-        self.zdim = self.depth.shape[2]
+        self.zdim = self.depth.shape[-3]
+        assert self.xdim == self.depth.shape[-1], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
+        assert self.ydim == self.depth.shape[-2], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
         self.z4d = len(self.depth.shape) == 4
+        if self.z4d:
+            assert self.tdim == self.depth.shape[0], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
```
erikvansebille: In this case, we know it should be `[tdim, zdim, ydim, xdim]` (not `[zdim, ydim, xdim]`) because we're in 4d mode, right?
Shouldn't we then also change the error message on the asserts for xdim and ydim, depending on whether we're in 4d mode?

parcels/grid.py
```diff
@@ -350,8 +354,12 @@ def __init__(self, lon, lat, depth, time=None, time_origin=0, mesh='flat'):
 
         self.gtype = GridCode.CurvilinearSGrid
         self.depth = depth
-        self.zdim = self.depth.shape[2]
+        self.zdim = self.depth.shape[-3]
+        assert self.xdim == self.depth.shape[-1], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
+        assert self.ydim == self.depth.shape[-2], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
         self.z4d = len(self.depth.shape) == 4
+        if self.z4d:
+            assert self.tdim == self.depth.shape[0], 'depth dimension has the wrong format. It should be [tdim, zdim, ydim, xdim] or [zdim, ydim, xdim]'
```
erikvansebille: Same comment as above

parcels/grid.py
delandmeterp: I was with hesitating when I wrote those messages. I think there is no point of changing only the message for tdim. We should have the same message always. (especially because if the format is wrong, in 99.9%, the assert will break at xdim, not tdim). So we want one single message.

Then, should we put one or two messages? I did it like this, because it's the option we chose for field formatting. But here, it does not cost anything to put a nicer message, so I'll do it.

PR #305, opened 2018-02-27
Fixing bug in ParticleFile.write when particles are deleted

### LINKED ISSUE #303: Problem with writing particles when some of them are deleted
The write on delete of particles is wrong.
I haven't looked at it, but I guess it is this self.idx which is udpated at particle deletion, such that normal particles are not written anymore at that timestep (if all the variables are not at write_on_delete)

Below, a simple test that breaks (use ncdump to see it immediately):

```
from parcels import FieldSet, ParticleSet, JITParticle
import numpy as np

def fieldset(xdim=2, ydim=2):
    data = {'U': np.zeros((ydim, xdim), dtype=np.float32),
            'V': np.zeros((ydim, xdim), dtype=np.float32)}
    dimensions = {'lon': np.linspace(0, 1, xdim, dtype=np.float32),
                  'lat': np.linspace(0, 1, ydim, dtype=np.float32)}
    return FieldSet.from_data(data, dimensions, mesh='flat')

fset = fieldset()

pset = ParticleSet(fset, lon=[10, 0], lat=[0, 4], pclass=JITParticle)
dt = 1
pfile = pset.ParticleFile('test1', outputdt=abs(dt))

def IncrLon(particle, fieldset, time, dt):
    particle.lon += 1.
    if particle.lon > 6:
        particle.delete()

for i in range(4):
    pset.execute(IncrLon, dt=dt, runtime=1., output_file=pfile)
```

erikvansebille:
Also improved the unit test in test_particle_sets.py to include deleted particles

This fixes #303

### DIFF
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -123,7 +123,6 @@ class ParticleFile(object):
             time = time.total_seconds()
         if self.lasttime_written != time and \
            (self.write_ondelete is False or deleted_only is True):
-            self.lasttime_written = time
             if pset.size > 0:
 
                 first_write = [p for p in pset if p.fileid < 0]
@@ -148,7 +147,9 @@ class ParticleFile(object):
             else:
                 logger.warning("ParticleSet is empty on writing as array")
 
-            self.idx += 1
+            if not deleted_only:
+                self.idx += 1
+                self.lasttime_written = time
 
         if sync:
             self.sync()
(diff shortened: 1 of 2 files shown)


erikvansebille:
Yes correct @delandmeterp, the old tests were not catching the bug. The script you added on the top here now works too, but the new test I created is much more stringent. We should be able to catch this bug now if it ever pops up again

Issue #281, opened 2018-01-29
support for python3


erikvansebille:
Parcels is currently not Python3-compatible, even though it was in early development. For a number of reasons (future, compatibility with systems only having python3, #234), it would be good to support Python3 again

Note that this also means travis-testing against Python3

willirath:
Here's what I did before writing <https://github.com/OceanParcels/parcels/issues/234#issuecomment-360927637>:

Installing [`python-modernize`](https://pypi.python.org/pypi/modernize) and running

```bash
python-modernize -n -w .
```
from within a local Parcels repo (`-w` = write in place, `-n` = no backup) lead to Pytest running just fine with Python 2 and Python 3.  The example notebooks (which were not modernized) only failed where print statements were not written as function calls.

The following digests kind of a hitlist of all changes that were done to reach compatibility:

```bash
git diff | grep -v ^-- | grep -v ^++ | grep ^+  | sort | uniq -c | sort -n
```
```
      1 +    (3600., 0, 3, 25, -35, list(range(3, 9, 1))),
      1 +    (-3600., 8, 10, 20, -39, list(range(7, 2, -1)))
      1 +            dparticles = list(range(len(self.particles)))
      1 +                filebuffer.indsdepth = indices['depth'] if 'depth' in indices else list(range(depthsize))
      1 +                filebuffer.indslat = indices['lat'] if 'lat' in indices else list(range(latsize))
      1 +                filebuffer.indslon = indices['lon'] if 'lon' in indices else list(range(lonsize))
      1 +        for attr, value in six.iteritems(self.__dict__):
      1 +    for i in pbar(list(range(trajs.nid))):
      1 +from .grid import GridCode
      1 +from .grid import (RectilinearZGrid, RectilinearSGrid, CurvilinearZGrid,
      1 +    from six.moves.urllib.request import urlopen
      1 +        if self.grid.mesh is 'flat' or (name not in list(self.unitconverters.keys())):
      1 +import six
      1 +    indices = {'lon': [5], 'lat': list(range(20, 30))}
      1 +            inds = indices[dimname] if dimname in indices else list(range(dim.size))
      1 +            inds_lat = indices['lat'] if 'lat' in indices else list(range(dim.shape[-2]))
      1 +    inds = list(zip(*np.where(arr)))
      1 +            inds_lon = indices['lon'] if 'lon' in indices else list(range(dim.shape[-1]))
      1 +            lat_i = list(range(len(self.grid.lat)))
[...]

PR #307, opened 2018-02-28
Add string argument option to particleset.density()

Jacketless:
Added a check for the option of particle_val being of type basestring (note this won't work for python 3.x), in which case this string is assumed to be the name of a particle attribute used for weighting density calculation. Otherwise, the use of particle_val is as before.

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -497,7 +497,7 @@ class ParticleSet(object):
         :param field: Optional :mod:`parcels.field.Field` object to calculate the histogram
                       on. Default is `fieldset.U`
         :param particle_val: Optional numpy-array of values to weigh each particle with,
-                             or string name of particle attribute to use weigh particles with.
+                             or string name of particle variable to use weigh particles with.
                              Default is None, resulting in a value of 1 for each particle
         :param relative: Boolean to control whether the density is scaled by the total
                          weight of all particles. Default is False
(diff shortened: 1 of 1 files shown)


parcels/particleset.py
```diff
@@ -496,16 +496,20 @@ def density(self, field=None, particle_val=None, relative=False, area_scale=Fals
...
+                             Default is None, resulting in a value of 1 for each particle
         :param relative: Boolean to control whether the density is scaled by the total
                          weight of all particles. Default is False
         :param area_scale: Boolean to control whether the density is scaled by the area
                            (in m^2) of each grid cell. Default is False
         """
 
         field = field if field else self.fieldset.U
-        particle_val = particle_val if particle_val else np.ones(len(self.particles))
+        if isinstance(particle_val, basestring):
```
erikvansebille: We do want to be python3 compatible (see #281), so better to use `if isinstance(particle_val, (str, unicode))` here?

parcels/particleset.py
```diff
@@ -496,16 +496,20 @@ def density(self, field=None, particle_val=None, relative=False, area_scale=Fals
 
         :param field: Optional :mod:`parcels.field.Field` object to calculate the histogram
                       on. Default is `fieldset.U`
-        :param particle_val: Optional numpy-array of values to weigh each particle with.
-                             Default is 1 for each particle
+        :param particle_val: Optional numpy-array of values to weigh each particle with,
+                             or string name of particle attribute to use weigh particles with.
```
erikvansebille: In Parcels, a particle attribute is normally called a `Variable`. Use that word here, too?

PR #308, opened 2018-03-02
Descending latitudes

raymondedwards95:
With this small addition, data with latitudes in descending order (like data from OSCAR) can be used.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -190,7 +190,7 @@ class Field(object):
         # self.lon, self.lat, self.depth and self.time are not used anymore in parcels.
         # self.grid should be used instead.
         # Those variables are still defined for backwards compatibility with users codes.
-        if not np.all(self.grid.lat == np.sort(self.grid.lat)):
+        if np.all(self.grid.lat[:-1] > self.grid.lat[1:]):
             # lat has to be in ascending order
             self.grid.lat = np.flip(self.grid.lat, axis=0)
             # if we flip lat, we need to flip data too
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -194,7 +194,7 @@ class Field(object):
             # lat has to be in ascending order
             self.grid.lat = np.flip(self.grid.lat, axis=0)
             # if we flip lat, we need to flip data too
-            self.data = np.flip(self.data, axis=1)
+            self.data = np.flip(self.data, axis=2)
         self.lon = self.grid.lon
         self.lat = self.grid.lat
         self.depth = self.grid.depth
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -190,6 +190,11 @@ class Field(object):
         # self.lon, self.lat, self.depth and self.time are not used anymore in parcels.
         # self.grid should be used instead.
         # Those variables are still defined for backwards compatibility with users codes.
+        if not np.all(self.grid.lat == np.sort(self.grid.lat)):
+            # lat has to be in ascending order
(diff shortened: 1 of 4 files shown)


parcels/field.py
```diff
@@ -190,6 +190,11 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
         # self.lon, self.lat, self.depth and self.time are not used anymore in parcels.
         # self.grid should be used instead.
         # Those variables are still defined for backwards compatibility with users codes.
+        if np.all(self.grid.lat[:-1] > self.grid.lat[1:]):
+            # lat has to be in ascending order
+            self.grid.lat = np.flip(self.grid.lat, axis=0)
+            # if we flip lat, we need to flip data too
+            self.data = np.flip(self.data, axis=2)
```
erikvansebille: I would move these five new lines to *above* the three comment lines on self.lon etc, as these five lines are not per se related to the comment on self.lon
In other words, swap lines 190-192 and 193-197

parcels/field.py
```diff
@@ -190,6 +190,11 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
         # self.lon, self.lat, self.depth and self.time are not used anymore in parcels.
         # self.grid should be used instead.
         # Those variables are still defined for backwards compatibility with users codes.
+        if np.all(self.grid.lat[:-1] > self.grid.lat[1:]):
```
erikvansebille: Like every new feature in Parcels, this will also need a unit test. Probably in `test_fieldset.py`, it would be good to create a small test function that creates a latitude vector in 'reverse' order, then loads it using `fieldset.from_data()` and then checks if the latitude (and data array) are indeed correctly reversed

erikvansebille: Does this also work for multidimensional grids, such as CurviLinear Grids? @delandmeterp, can you have a quick look at whether that would work?

parcels/field.py
delandmeterp: I think the lat flip, should not be done in field.__init__ but in grid.__init__

Then, you can do more specific stuff, since what you did only works for lat vectors. But in curvi grids, we have lat arrays and lat arrays.

So, in RectilinearGrids, you should do the lat.flip, plus a variable grid.latFlipped = True
In RectilinearSGrids, you should have
if grid.latFlipped:
  you flip the y dimension of depth (which is axis=-2)

In Curvi grids, you do the same, but you flip also lon arrays

In field.__init__, you have 
if grid.latFlipped:
  you flip the y dimension of depth (which is axis=-2)

(Can you have a negative value for the axis? Because, ydimension is not alwasy the 2 axis, but always the -2. So what you wrote is not general)

tests/test_fieldset_sampling.py
```diff
@@ -184,16 +184,28 @@ def test_nearest_neighbour_interpolation3D(mode, k_sample_p, npart=81):
 
 
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
-def test_fieldset_sample_particle(fieldset, mode, k_sample_uv, npart=120):
+@pytest.mark.parametrize('lat_flip', [False, True])
```
erikvansebille: should we also check for curvilinear grids, explicitly?

delandmeterp: Actually, I think we shouldn't flip at all curvi grids.
Nothing requires to have ascending latitudes in that case.

So I would remove the flip from that class.

PR #309, opened 2018-03-02
Better management of non-monotonic grids in index research

delandmeterp:


parcels/field.py
```diff
@@ -566,11 +572,19 @@ def search_indices_curvilinear(self, x, y, z, xi, yi, ti=-1, time=-1, search2D=F
                          [1, -1, 1, -1]])
         maxIterSearch = 1e6
         it = 0
+        if not grid.zonal_periodic or grid.mesh == 'flat':
```
erikvansebille: So we do not support `flat` grids that are periodic? I would argue that there are many cases where flat grids can be periodic (e.g. zonal channels etc)

parcels/grid.py
```diff
@@ -86,10 +87,15 @@ def add_periodic_halo(self, zonal, meridional, halosize=5):
         """
         if zonal:
             lonshift = (self.lon[-1] - 2 * self.lon[0] + self.lon[1])
+            if not np.allclose(self.lon[1]-self.lon[0], self.lon[-1]-self.lon[-2]):
+                logger.warning_once("To zonal halo is located at the east and west of current grid, with a dx = lon[1]-lon[0] between the last nodes of the original grid and the first ones of the halo. In your grid, lon[1]-lon[0] != lon[-1]-lon[-2]. Is the halo computed as you expect?")
```
erikvansebille: Not sure the grammar is correct in this warning. `To zonal halo`?

parcels/grid.py
```diff
@@ -86,10 +87,15 @@ def add_periodic_halo(self, zonal, meridional, halosize=5):
...
             self.lon = np.concatenate((self.lon[-halosize:] - lonshift,
                                       self.lon, self.lon[0:halosize] + lonshift))
             self.xdim = self.lon.size
+            self.zonal_periodic = True
         if meridional:
+            if not np.allclose(self.lat[1]-self.lat[0], self.lat[-1]-self.lat[-2]):
+                logger.warning_once("To meridional halo is located at the north and south of current grid, with a dy = lat[1]-lat[0] between the last nodes of the original grid and the first ones of the halo. In your grid, lat[1]-lat[0] != lat[-1]-lat[-2]. Is the halo computed as you expect?")
```
erikvansebille: See comment above

parcels/grid.py
```diff
@@ -275,22 +297,32 @@ def advancetime(self, grid_new):
         else:
             raise RuntimeError("Time of field_new in Field.advancetime() overlaps with times in old Field")
 
+    def check_zonal_periodic(self):
```
erikvansebille: Why is this function duplicated? Make it globally accessible?

parcels/field.py
delandmeterp: We do support periodic flat grids. (the test in test_fieldset is an example of such a grid). It is just that the out-of-bound tests here shouldn't be used in spherical, zonal-periodic grids.

parcels/grid.py
delandmeterp: Those functions are actually not the same
`dx = self.lon[0, 1:] - self.lon[0, :-1]`
vs
`dx = self.lon[1:] - self.lon[:-1]`

But since they are almost equal, we should maybe still define the function in the parent class `Grid` and manage the difference with a simple if. 
Do you have a preference for the style? I would put the function in `Grid`

erikvansebille: Yes, I agree

erikvansebille:
Looks good to me; happy for this to be merged

PR #310, opened 2018-03-08
Reducing file size of ParticleFile when particles are deleted and not all started at same time

erikvansebille:
This PR reduces the file set of ParticleFile, by making the time index where each particle Variable is written dependent on the Particle. This means that if Particles are added during `.execute()`, they are still written from index 0, reducing the file size if particles are also deleted.

Since for each Particle, we also store `time`, we still store all information required for post-processing.

For example, in the function `test_pset_repeated_release_delayed_adding_deleting(fieldset(), 'jit', repeatdt=2, tmpdir='', dt=1, maxvar=4, runtime=10)`, we go from
```
ncfile.variables['time'][:] =
[[0.0 1.0 2.0 3.0 4.0 5.0 -- -- -- -- --]
 [-- -- 2.0 3.0 4.0 5.0 6.0 7.0 -- -- --]
 [-- -- -- -- 4.0 5.0 6.0 7.0 8.0 9.0 --]
 [-- -- -- -- -- -- 6.0 7.0 8.0 9.0 10.0]
 [-- -- -- -- -- -- -- -- 8.0 9.0 10.0]
 [-- -- -- -- -- -- -- -- -- -- 10.0]]
```
To
```
ncfile.variables['time'][:] =
[[0.0 1.0 2.0 3.0 4.0 5.0]
 [2.0 3.0 4.0 5.0 6.0 7.0]
 [4.0 5.0 6.0 7.0 8.0 9.0]
 [6.0 7.0 8.0 9.0 10.0 --]
 [8.0 9.0 10.0 -- -- --]
 [10.0 -- -- -- -- --]]
```


parcels/particlefile.py
```diff
@@ -125,25 +125,25 @@ def write(self, pset, time, sync=True, deleted_only=False):
...
-                self.z[inds, self.idx] = np.array([p.depth for p in pset])
-                for var in self.user_vars:
-                    getattr(self, var)[inds, self.idx] = np.array([getattr(p, var) for p in pset])
-                for var in self.user_vars_once:
-                    if np.any(first_write):
-                        vals = [getattr(p, var) for p in first_write]
-                        newinds = [p.fileid for p in first_write]
-                        getattr(self, var)[newinds] = np.array(vals)
+                self.idx = np.append(self.idx, np.zeros(len(first_write)))
+
```
delandmeterp: Nice development! It's great to indeed have the nc files written by Parcels as small and simple as possible.

But it seems that we don't document anywhere what is the format of those files. Shouldn't we do it?
For example, I received an email from @Jacketless yesterday that had problems to read an old file generated by Parcels in the new code. It may be related to the commit with arrays and indexed pfiles.
Should we explain the file format somewhere?

PR #311, opened 2018-03-09
Xarray read

delandmeterp:
Pass from netcdf4 to xarrays to read netcdf files when loading a `Field`

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -268,9 +268,9 @@ class Field(object):
 
         if not isinstance(filenames, Iterable) or isinstance(filenames, str):
             filenames = [filenames]
-        with FileBuffer(filenames[0], dimensions) as filebuffer:
-            lon, lat = filebuffer.read_lonlat(indices)
-            depth = filebuffer.read_depth(indices)
+        with FileBuffer(filenames[0], dimensions, indices) as filebuffer:
+            lon, lat = filebuffer.read_lonlat
+            depth = filebuffer.read_depth
             if name in ['cosU', 'sinU', 'cosV', 'sinV']:
                 warning = False
                 try:
@@ -286,7 +286,7 @@ class Field(object):
         # across multiple files
         timeslices = []
         for fname in filenames:
-            with FileBuffer(fname, dimensions) as filebuffer:
+            with FileBuffer(fname, dimensions, indices) as filebuffer:
                 timeslices.append(filebuffer.time)
         timeslices = np.array(timeslices)
         time = np.concatenate(timeslices)
@@ -304,7 +304,7 @@ class Field(object):
         data = np.empty((time.size, depthdim, latdim, londim), dtype=np.float32)
         ti = 0
         for tslice, fname in zip(timeslices, filenames):
-            with FileBuffer(fname, dimensions) as filebuffer:
+            with FileBuffer(fname, dimensions, indices) as filebuffer:
                 depthsize = depth.size if len(depth.shape) == 1 else depth.shape[-3]
                 latsize = lat.size if len(lat.shape) == 1 else lat.shape[-2]
                 lonsize = lon.size if len(lon.shape) == 1 else lon.shape[-1]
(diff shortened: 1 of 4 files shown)


parcels/field.py
```diff
@@ -994,33 +988,34 @@ def subset(self, dim, dimname, indices):
...
     @property
     def data(self):
-        if len(self.dataset[self.name].shape) == 2:
-            data = self.dataset[self.name][self.indslat, self.indslon]
-        elif len(self.dataset[self.name].shape) == 3:
+        data = getattr(self.dataset, self.name)
+        if len(data.shape) == 2:
+            data = data[self.indslat, self.indslon]
+        elif len(data.shape) == 3:
             data = self.dataset[self.name][:, self.indslat, self.indslon]
```
erikvansebille: why not use `data = data[:, self.indslat, self.indslon]` here?

parcels/field.py
```diff
@@ -994,33 +988,34 @@ def subset(self, dim, dimname, indices):
...
         else:
             data = self.dataset[self.name][:, self.indsdepth, self.indslat, self.indslon]
```
erikvansebille: Again, why not use `data = data[:, self.indsdepth, self.indslat, self.indslon]`

parcels/fieldset.py
```diff
@@ -92,8 +92,12 @@ def add_field(self, field):
     def check_complete(self):
         assert(self.U), ('U field is not defined')
         assert(self.V), ('V field is not defined')
+        ugrid = self.U.grid
         for g in self.gridset.grids:
             g.check_zonal_periodic()
+            if g is not ugrid and g.time_origin != 0:
```
erikvansebille: why is this `g is not ugrid` needed? I don't understand the rationale

parcels/particleset.py
```diff
@@ -9,6 +9,7 @@
 import bisect
 from collections import Iterable
 from datetime import timedelta as delta
+import pandas as pd
```
erikvansebille: So we're using `pandas` now? Is that really necessary? That's a pretty hefty requirement. I think you said on Slack you wanted to stick with `datetime` and `np`? I'd agree with that

parcels/particleset.py
```diff
@@ -467,10 +470,10 @@ def nearest_index(array, value):
         if time_origin is 0:
             timestr = ' after ' + str(delta(seconds=show_time)) + ' hours'
         else:
-            timestr = ' on ' + str(time_origin + delta(seconds=show_time))
+            timestr = ' on ' + str(time_origin + pd.Timedelta(show_time, 's'))
```
erikvansebille: Can't we do this without `pandas`? If we need pandas only for the plotting of a figure title, then I think we really need to reconsider whether we're not better off not doing these figure titles

parcels/field.py
erikvansebille: Also, what happens if field is in `(depth, lat, lon)`, instead of `(time, lat, lon)`? Shouldn't we be using `self.indsdepth` too, then? Note that in the previous version this wasn't possible either, and we might get rid of explicit indexing in the near future, so if there's no easy fix then probably better to leave as is for now

parcels/particleset.py
delandmeterp: The only reason for pandas was that we could use float number of seconds.
(also, str(np.datetime64)) is quite ugly in pset.show(), so I think I'll clean it.
But yes, I'll change this (there are still a few things I wanna clean)

willirath:
Pandas is a runtime dependency of xarray anyway. See, e.g., the conda-forge package: https://github.com/conda-forge/xarray-feedstock/blob/master/recipe/meta.yaml

parcels/fieldset.py
delandmeterp: We want to re-shift time from all grids to the one of ugrid.
So no shift for ugrid.
But what happens if ugrid is independent of time and not the other grids? For the moment, it crashes (instead of giving non-sense result)

parcels/particleset.py
delandmeterp: @willirath, yes so pandas is available, but loading a module to have in one single location (the title of a plot)
2000-01-01 03:04:04.57
instead of 
2000-01-01 03:04:04
Is it worth it?

erikvansebille:
OK, looks good to me now. Happy for you to merge as soon as you have green ticks on travis

PR #313, opened 2018-03-11
Python3 support

### LINKED ISSUE #281: support for python3

erikvansebille:
This PR adds support for python3 to Parcels, and thereby fixes #281 

Two issues:
- While printing in kernels works in python3, it can't be properly tested because of an issue with py.test. The Travis testing of print statements in kernels is therefore turned off in python3
- Jupyter notebooks can be tested only in python2, so are not travis-tested in python3


willirath:
Looks good!

One comment: The notebooks are not evaluated during the Python 3 tests. But they could be changed to be compatible at least. (If I remember correctly, there only were a few print statements that needed to be adapted.)

erikvansebille:
Thanks for the comment, @willirath. Ideally, we would indeed test the jupyter notebooks on python3 too, but `py.test` can't seem to test notebooks made on python2 for python3. E.g., the error below
```
(py3_parcels) $ py.test --nbval-lax prcels/examples/tutorial_jit_vs_scipy.ipynb 
========================= test session starts =========================
platform darwin -- Python 3.6.3, pytest-3.4.0, py-1.5.2, pluggy-0.6.0
rootdir: /Users/erik/Codes/ParcelsCode, inifile: setup.cfg
plugins: nbval-0.9.0
collected 1 item                                                      

parcels/examples/tutorial_jit_vs_scipy.ipynb E                  [100%]

=============================== ERRORS ================================
 ERROR at setup of parcels/examples/tutorial_jit_vs_scipy.ipynb::Cell 0 

self = <nbval.kernel.NbvalKernelspecManager object at 0x10b0f2320>
kernel_name = 'python2'

    def get_kernel_spec(self, kernel_name):
        """Returns a :class:`KernelSpec` instance for the given kernel_name.
    
            Raises :exc:`NoSuchKernel` if the given kernel name is not found.
            """
        d = self.find_kernel_specs()
        try:
>           resource_dir = d[kernel_name.lower()]
E           KeyError: 'python2'

../../anaconda3/lib/python3.6/site-packages/jupyter_client/kernelspec.py:201: KeyError
```
I did update parcels/examples/parcels_tutorial.ipynb to work on python3, and just confirmed again that all notebooks work for now on both python2 and python3. 

However, it is not ideal indeed that notebooks are not standard tested. If you know a way to have python2-generated notebooks tested in python3, then let me know!

willirath:
#314 solves this by writing the desired kernels specifications to the notebooks when they are copied with `parcels_get_examples`.

erikvansebille:
Thanks @willirath for providing this solution, it's great!

environment_py2_linux.yml
```diff
@@ -1,5 +1,6 @@
-name: py2_parcels
+name: parcels
```
delandmeterp: I don't understand why all the environments have the same name.
We would like to have different names such that we can load what we want, no?

parcels/field.py
```diff
@@ -320,8 +320,8 @@ def from_netcdf(cls, name, dimensions, filenames, indices={},
                 filebuffer.indslon = indices['lon'] if 'lon' in indices else range(lonsize)
                 filebuffer.indsdepth = indices['depth'] if 'depth' in indices else range(depthsize)
                 for inds in [filebuffer.indslat, filebuffer.indslon, filebuffer.indsdepth]:
-                    if not isinstance(inds, list):
-                        raise RuntimeError('Indices sur field subsetting need to be a list')
+                    if type(inds) not in [list, range]:
```
delandmeterp: isinstance does not work for more than an argument? I thought we used this already in the code!

delandmeterp: (just a comment by the way. nothing to do with that)

willirath:
>> @@ -1,5 +1,6 @@
> -name: py2_parcels
> +name: parcels
>
> I don't understand why all the environments have the same name.
> We would like to have different names such that we can load what we want, no?

As far as I understand, these environments are only used by Travis. And there, only one env is installed at a time. In this context, having the same env name simplifies the code below.

delandmeterp:
But for example I also use such environment.
A user who uses anaconda but wants to control parcels version via github will use it, no?

If it helps a lot for Travis, then ok and the user changes the file before calling it (only once, so we don’t care so much), otherwise, better to have multiple names I would say




---
Philippe Delandmeter, PhD
Utrecht University
Institute for Marine and Atmospheric Research (IMAU)
E-mail: p.b.delandmeter@uu.nl<mailto:p.b.delandmeter@uu.nl>



On 12 Mar 2018, at 08:25, Willi Rath <notifications@github.com<mailto:notifications@github.com>> wrote:


>> @@ -1,5 +1,6 @@
> -name: py2_parcels
> +name: parcels
>
> I don't understand why all the environments have the same name.
> We would like to have different names such that we can load what we want, no?

As far as I understand, these environments are only used by Travis. And there, only one env is installed at a time. In this context, having the same env name simplifies the code below.

willirath:
> But for example I also use such environment.
> A user who uses anaconda but wants to control parcels version via github 
> will use it, no?
> 
> If it helps a lot for Travis, then ok and the user changes the file 
> before calling it (only once, so we don’t care so much), otherwise, 
> better to have multiple names I would say

Ah.  I think, we can override the env-name by passing an additional `-n 
parcels` to `conda env create`.  Then, it would be most convenient to 
have the envs called `py[2,3]_parcels` in the YAML file and set the name 
to the simpler `parcels` when creating envs in Travis.

willirath:
See #315 which also adds Python-3 tests for Windows.

PR #315, opened 2018-03-12
Use meaningful env names and test Python 3 on Windows

willirath:
b64058d uses more meaningful env names in the YAML files and overrides them in Travis.

b2685be sets up Appveyor to also test Python 3.

06defd9 allows failure of the print test on Windows.

Appveyor is running here: <https://ci.appveyor.com/project/willirath/parcels>

PR #316, opened 2018-03-12
Use meaningful names for environments, and test Python 2 and 3 on Windows

willirath:
This is a clean version of #315.


willirath:
Appveyr builds: <https://ci.appveyor.com/project/willirath/parcels/>

PR #318, opened 2018-03-13
Use xfail rather than skip

willirath:
This will run the tests that we know (or expect) to fail under the given circumstances rather than skipping them right away.  Acutally running the tests and allowing them to fail has the advantage that if the reason for the failure disappears in the future, we'll notice that the tests pass.


willirath:
I've canceled one of the two Travis builds.

PR #319, opened 2018-03-14
Auto advance time

delandmeterp:
The tests that do not pass are tests using time subsetting, which shouldn't be used anymore.


parcels/fieldset.py
```diff
@@ -343,3 +343,97 @@ def advancetime(self, fieldset_new):
...
+                    if signdt >= 0 and time >= g.time[1] and g.ti < len(g.time_full)-3:
+                        g.ti += 1
+                        g.time = g.time_full[g.ti:g.ti+3]
+                        g.advanced = 1
+                        if g.ti != len(g.time_full)-3:
+                            nextTime_loc = g.time[2]
+                    if signdt == -1 and time <= g.time[1] and g.ti > 0:
+                        g.ti -= 1
+                        g.time = g.time_full[g.ti:g.ti+3]
+                        g.advanced = 1
```
erikvansebille: This code is quite long and confusing. Can be cleaned up and pythonized a bit? E.g. both if statements have `g.advanced=1`, so can be taken out of if loops?

christensen5:
Any thoughts on leaving some form of manual advancetime in as a legacy option? Particularly for the benefit of users doing "live" simulations (ie. with the velocity field being updated at each timestep rather than loaded from disk)?

It would be a shame for example to have to save a firedrake field to an nc file at each timestep in order to then pass it to parcels. I worry that this would significantly slow things down.

Having said this, Eric as you're aware, in my current simulations I'm not using advancetime at all but rather calling .at() on the firedrake field to pass velocity field data directly to individual particles. I'm just thinking about a possible future in which manually updating fields 'live' might remain desirable.

delandmeterp:
Hi @christensen5,

For the moment, we won't remove the avanceTime function, but it is deprecated, and we advice not to use it. (The new system deals with time advancing).

For the cases where new nc files appear during the simulation (as you said, it is not the best way to communicate between two processes working together, but it could happen..), you can still use advanceTime.

But the best way would be to have a function fieldset.update() adding the new existing files.
It shouldn't be difficult to implement, but there is no point to do it now without a real application to use and test it. If you need such function in the future and want some help or hints, don't hesitate to ask.

parcels/field.py
```diff
@@ -265,11 +258,15 @@ def from_netcdf(cls, name, dimensions, filenames, indices={},
                1. spherical (default): Lat and lon in degree, with a
                   correction for zonal velocity U near the poles.
                2. flat: No conversion, lat/lon are assumed to be in m.
+        :param full_load: boolean whether to fully load the data or only pre-load them. (default: False)
+               It is advised not to fully load the data, since in that case Parcels deals with
+               a better memory management during particle set execution.
+               full_load is however necessary for plotting the fields.
```
erikvansebille: Add 'sometimes' to this sentence. I.e. `full_load is however sometimes necessary for plotting the fields`

parcels/fieldset.py
```diff
@@ -127,6 +127,10 @@ def from_netcdf(cls, filenames, variables, dimensions, indices={},
                (i.e. beyond the last available time snapshot)
         :param time_periodic: boolean whether to loop periodically over the time component of the FieldSet
                This flag overrides the allow_time_interpolation and sets it to False
+        :param full_load: boolean whether to fully load the data or only pre-load them. (default: False)
+               It is advised not to fully load the data, since in that case Parcels deals with
+               a better memory management during particle set execution.
+               full_load is however necessary for plotting the fields.
```
erikvansebille: Again, add 'sometimes' here

parcels/field.py
```diff
@@ -946,8 +969,28 @@ def advancetime(self, field_new, advanceForward):
             self.data = np.concatenate((field_new.data[:, :, :], self.data[:-1, :, :]), 0)
             self.time = self.grid.time
 
+    def computeTimeChunk(self, data, tindex):
```
erikvansebille: I don't entirely understand why `data` needs to be an input argument here. Isn't that just `self.data`?

The reason I'm asking is because I think it would be useful for users to be able to do e.g. `fieldset.U.computeTimeChunk(tindex=0)` is e.g. they want to plot that. Now I see that users can do `fieldset.U.fieldset.computeTimeChunk(tindex=0, dt=1)`, but it's a bit redundant to do `Field.FieldSet`?

parcels/field.py
```diff
@@ -916,23 +917,39 @@ def add_periodic_halo(self, zonal, meridional, halosize=5, data=None):
         data = self.data if dataNone else data
         if zonal:
             if len(data.shape) is 3:
-                data = np.concatenate((data[:, :, -halosize:], data,
-                                       data[:, :, 0:halosize]), axis=len(data.shape)-1)
+                if hasattr(self.grid, 'ti') and self.grid.ti == -1:
```
erikvansebille: @delandmeterp, I have now used the statement `hasattr(self.grid, 'ti') and self.grid.ti == -1` to check if `Field.data` is fully loaded or not. However, this might not be the most compact/intuitive way to check that. 

Is there a more elegant way to confirm if the `Field` is fully loaded or not?

PR #322, opened 2018-03-16
Use `timedelta64` in kernel errors as well

willirath:
This makes sure that `OutOfBounds` errors can be raised again.

### DIFF
--- a/parcels/kernels/error.py
+++ b/parcels/kernels/error.py
@@ -1,6 +1,6 @@
 """Collection of pre-built recovery kernels"""
 from enum import IntEnum
-from datetime import timedelta
+from numpy import timedelta64
 
 
 __all__ = ['ErrorCode', 'KernelError', 'OutOfBoundsError', 'recovery_map']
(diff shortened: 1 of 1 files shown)


willirath:
As the Agulhas example uses the `OutOfBounds` exception, the runtime of the example notebook is probably too short to actually have a particle reaching the boundary.

parcels/kernels/error.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, particle, fieldset=None, msg=None):
 def parse_particletime(time, fieldset):
     if fieldset is not None and fieldset.U.grid.time_origin != 0:
         # TODO assuming that error was thrown on U field
-        time = fieldset.U.grid.time_origin + timedelta(seconds=time)
+        time = fieldset.U.grid.time_origin + timedelta64(seconds=time)
```
delandmeterp: Yes, this part wasn't tested in the tests and I forgot it. I've also modified it in the current branch autoAdvanceTime2.
But it has to be changed to `timedelta64(int(time), 's')`. That is the was timedelta64 works.

willirath: With #319 also fixing this, I'll close #322.

willirath:
Closing as <https://github.com/OceanParcels/parcels/pull/319/files#diff-dc4092e8c3dabbadd7981b1540fc3ef2R34> already takes care of this.

PR #324, opened 2018-03-23
New grid index

delandmeterp:
Pset creation is now linear in time!

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -185,6 +185,7 @@ class Field(object):
             self.grid = grid
         else:
             self.grid = RectilinearZGrid(lon, lat, depth, time, time_origin=time_origin, mesh=mesh)
+        self.iGrid = -1
         # self.lon, self.lat, self.depth and self.time are not used anymore in parcels.
         # self.grid should be used instead.
         # Those variables are still defined for backwards compatibility with users codes.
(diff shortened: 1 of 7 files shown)


parcels/field.py
```diff
@@ -820,7 +808,7 @@ def ctypes_struct(self):
         # Ctypes struct corresponding to the type definition in parcels.h
         class CField(Structure):
             _fields_ = [('xdim', c_int), ('ydim', c_int), ('zdim', c_int),
-                        ('tdim', c_int),
+                        ('tdim', c_int), ('iGrid', c_int),
```
erikvansebille: none of the other variables have capitalisation. Use `igrid` instead?

parcels/include/parcels.h
```diff
@@ -459,49 +447,47 @@ static inline ErrorCode spatial_interpolation_nearest3D(double xsi, double eta,
 
 /* Linear interpolation along the time axis */
 static inline ErrorCode temporal_interpolation_structured_grid(float x, float y, float z, double time, CField *f, 
-                                                               GridCode gcode, CGridIndex *gridIndex,
+                                                               GridCode gcode, int *xis, int *yis, int *zis, int *tis,
                                                                float *value, int interp_method)
 {
   ErrorCode err;
   CStructuredGrid *grid = f->grid->grid;
+  int iGrid = f->iGrid;
```
erikvansebille: same here with `igrid` instead of `iGrid`?

parcels/particleset.py
```diff
@@ -516,11 +516,11 @@ def density(self, field=None, particle_val=None, relative=False, area_scale=Fals
         density = np.zeros((field.grid.lat.size, field.grid.lon.size), dtype=np.float32)
 
         for pi, p in enumerate(self.particles):
-            try:  # breaks if either p.gridIndexSet does not exist (in scipy) or field not in fieldset
-                if p.gridIndexSet[field.grid].ti < 0:  # xi, yi, ti, not initialised
+            try:  # breaks if either p.xis do not exist (in scipy) or field not in fieldset
+                if p.tis[field.iGrid] < 0:  # xis, yis, tis, not initialised
```
erikvansebille: So the comment says it's checking for `p.xis` but the actual check is on `p.tis`?

PR #327, opened 2018-04-02
Website into separate repository

erikvansebille:
The code for the https://www.oceanparcels.org site has now spun off into its own repository (https://github.com/OceanParcels/oceanparcels_website) and can therefore be removed from the main Parcels repo here.

A few comments:
- The new website can be viewed at https://oceanparcels.github.io/oceanparcels_website/. If we're happy with that, I'll ask the administrators of the https://www.oceanparcels.org domain to update the DNS so that it points to this new URL
- I've designed the new website to have a more modern feel. Keen to hear any comments on that (by filing issues and PRs at https://github.com/OceanParcels/oceanparcels_website)
- The new site includes a FAQ section, where we can start adding tips etc on how to use Parcels
- The documentation in `parcels.html` has not migrated over to the new website. However, it was hard to keep that updated anyways, and with the different versions (conda-forge always lagging `master`) was not universally useful anyway. If someone has an idea how to document individual functions and classes, let me know


willirath:
I did not have a thorough look yet. But one thing comes to mind: The download location for the example data is hard coded. Will this data still be accessible?

erikvansebille:
Thanks for the comment about the `examples-data` directory, @willirath. We will make sure that this will stay accessible yes

Issue #334, opened 2018-04-09
Dimension `trajectory` conflicts with variable `trajectory`


willirath:
Currently, <https://github.com/OceanParcels/parcels/blob/master/parcels/particlefile.py#L44_L52> defines a dimension `trajectory` and a variables `trajectory` that has more than one dimension.  This collides with the concept of "coordinate variables" defined in [the CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#terminology). This causes problems whenever a tools relies on the notion of coordinate variables (xarray does, and NCO does partly).

Opening a particlefile with Xarray results in 
```
MissingDimensionsError: 'trajectory' has more than 1-dimension and the same name as one of its dimensions ('trajectory', 'obs'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.
```

Current workaround (rename the **dimension** to, e.g., `"traj"`):

```python
from shutil import copy
import netCDF4 as nc4
import xarray as xr

orig_file = "some_particlefile.nc"
dest_file = "some_particlefile_with_renamed_dim.nc"

copy(source_file, dest_file)

with nc4.Dataset(dest_file, mode="a") as tmp_ds:
    tmp_ds.renameDimension("trajectory", "traj")

particle_dataset = xr.open_dataset(dest_file)
```

As far as I see, fixing this could be as easy as modifying `"trajectory"` to `"traj"` in `particlefile.py`.  But I'm not sure if I missed something.

PR #335, opened 2018-04-09
WIP: Rename trajectory dimension to avoid conflict

willirath:
This addresses #334


erikvansebille:
Green ticks for CI now. @willirath, can you check whether indeed this resolves #335? If so, we can merge

willirath:
I'll have a look.

willirath:
Looks good.  Thanks!

Issue #332, opened 2018-04-06
Error with Compiler


michellemho:
Hi Ocean Parcels team!

David Lindo-Atichati (@dlindo) has been running into an error when trying to execute a run of particles using the Brownian Motion 2D kernel. I have not been able to replicate the error on my machine, so I suspect there might be a problem with our different operating systems and some Parcels dependencies. Any advice? Thank you!

**Code:**
```
# Single lat_long
lat = [27.952]
lon = [178.991]

# number of times to repeat the particles
npart = 5

# Set random seed
random.seed(123456)

days = 50
runtime = timedelta(days=days)

random.seed(1234)
pset = ParticleSet(fieldset=fieldset, pclass=JITParticle,
                  lon=np.tile(lon,npart), lat=np.tile(lat,npart))
pset.execute(pset.Kernel(BrownianMotion2D),
            runtime=runtime,
            dt=timedelta(hours=1),
            recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle},
            output_file=pset.ParticleFile(name=“{}Drifters_Brownian_npart{}_{}_days”.format(len(lat),npart,days),
                                          outputdt=timedelta(hours=1)))
```

**And the error:**
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-5a75059c49ce> in <module>()
    19              dt=timedelta(hours=1),
    20              recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle},
---> 21              output_file=pset.ParticleFile(name=“{}Drifters_Brownian_npart{}_{}_days”.format(len(lat),npart,days), outputdt=timedelta(hours=1)))

/Users/davidlindo-atichati/miniconda2/envs/py2_parcels/lib/python2.7/site-packages/parcels/particleset.pyc in execute(self, pyfunc, endtime, runtime, dt, moviedt, recovery, output_file, movie_background_field)
   262             if self.ptype.uses_jit:
   263                 self.kernel.remove_lib()
--> 264                 self.kernel.compile(compiler=GNUCompiler())
   265                 self.kernel.load_lib()
   266

/Users/davidlindo-atichati/miniconda2/envs/py2_parcels/lib/python2.7/site-packages/parcels/kernel.pyc in compile(self, compiler)
   155         with open(self.src_file, ‘w’) as f:
   156             f.write(self.ccode)
--> 157         compiler.compile(self.src_file, self.lib_file, self.log_file)
[...]

PR #341, opened 2018-04-13
WIP: Also test against a newer OSX image

willirath:
This will probably fail because of #332

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ matrix:
     - os: osx
       language: generic
       env: PY_VERSION=3 MINICONDA_NAME=Miniconda3-latest-MacOSX-x86_64.sh
+    - os: osx
+      osx_image: xcode9.3
+      language: generic
+      env: PY_VERSION=3 MINICONDA_NAME=Miniconda3-latest-MacOSX-x86_64.sh
 
 install:
     - wget http://repo.continuum.io/miniconda/${MINICONDA_NAME} -O miniconda.sh;
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ install:
     - conda env create --name parcels --file environment_py${PY_VERSION}_${TRAVIS_OS_NAME}.yml
     - source activate parcels
     - python setup.py install
-    - ln -s ${LD} ${HOME}/miniconda/bin/ld # workaround avoiding os linker
+    - ln -s ${LD} ${HOME}/miniconda/envs/parcels/bin/ld # workaround avoiding os linker
+    - which ld
 
 script:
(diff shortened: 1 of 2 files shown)


willirath:
This PR is mostly to test if the error reported in #332 can be tracked down to the version of xcode.

willirath:
(I've cancelled all but the two OSX builds that only differ in the OSX image.)

willirath:
JIT fails with newest XCODE: <https://travis-ci.org/OceanParcels/parcels/jobs/366004275#L519>

willirath:
21eac8c has no effect as the linker is never called explicitly.

.travis.yml
```diff
@@ -23,6 +27,7 @@ install:
     - conda env create --name parcels --file environment_py${PY_VERSION}_${TRAVIS_OS_NAME}.yml
     - source activate parcels
     - python setup.py install
+    - ln -s ${LD} ${HOME}/miniconda/bin/ld # workaround avoiding os linker
```
willirath: Note that this should go to the env's bin directory ...

willirath:
With the last commit, I think I got what I wanted: Override the call to $CC with a call that makes sure the Anaconda linker is used. But that does not seem to help with xcode 9.3.

.travis.yml
```diff
@@ -23,6 +27,7 @@ install:
     - conda env create --name parcels --file environment_py${PY_VERSION}_${TRAVIS_OS_NAME}.yml
     - source activate parcels
     - python setup.py install
+    - export CC=gcc
```
erikvansebille: Should we not do this only when in `osx-xcode9.3` mode?

parcels/compiler.py
```diff
@@ -28,13 +28,14 @@ class Compiler(object):
...
-        overridden by exporting the environment variable ``LDSHARED``).
+        overridden by exporting the environment variables ``LDSHARED`` or
+        ``LD``).
     :arg cppargs: A list of arguments to the C compiler (optional).
     :arg ldargs: A list of arguments to the linker (optional)."""
 
     def __init__(self, cc, ld=None, cppargs=[], ldargs=[]):
         self._cc = environ.get('CC', cc)
-        self._ld = environ.get('LDSHARED', ld)
+        self._ld = environ.get('LDSHARED', environ.get('LD', ld))
```
erikvansebille: These changes to `compiler.py` are not necessary, right? Do we still want to merge them into master?

willirath:
I think we should close and drop this PR and just include an xcode 9.3 setup that is allowed to fail in the Travis config. The changes to the compiler.py did not lead anywhere.

PR #344, opened 2018-04-16
Pset time step

delandmeterp:


parcels/particleset.py
```diff
@@ -324,9 +324,12 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
...
             if abs(time-next_prelease) < tol:
-                self.add(ParticleSet(fieldset=self.fieldset, time=time, lon=self.repeatlon,
-                                     lat=self.repeatlat, depth=self.repeatdepth,
-                                     pclass=self.repeatpclass))
+                pset_new = ParticleSet(fieldset=self.fieldset, time=time, lon=self.repeatlon,
+                                       lat=self.repeatlat, depth=self.repeatdepth,
+                                       pclass=self.repeatpclass)
+                for p in pset_new:
+                    p.dt = dt
+                self.add(pset_new)
```
erikvansebille: Why not have `dt` as an argument to `ParticleSet.__init__`, which is then directly passed to `Particle.__init__`? Then, this could simply be
```
self.add(ParticleSet(fieldset=self.fieldset, time=time, lon=self.repeatlon,
                                   lat=self.repeatlat, depth=self.repeatdepth,
                                   pclass=self.repeatpclass, dt=dt))
```
Would that not be cleaner?

parcels/particle.py
```diff
@@ -169,7 +168,7 @@ class ScipyParticle(_Particle):
     dt = Variable('dt', dtype=np.float32, to_write=False)
     state = Variable('state', dtype=np.int32, initial=ErrorCode.Success, to_write=False)
 
-    def __init__(self, lon, lat, fieldset, depth=0., dt=1., time=0., cptr=None):
+    def __init__(self, lon, lat, fieldset, depth=0., time=0., cptr=None):
```
erikvansebille: Or set `dt=None` as default to `Particle.__init__`?

parcels/particleset.py
delandmeterp: If dt is an argument of ParticleSet constructor, there is no point on having a dt argument at pset.execute(), because they would conflict.

parcels/particle.py
delandmeterp: Indeed, this is what I did (Actually, I put it to nan, but None makes more sense. I'll check if it is possible)

PR #348, opened 2018-04-18
also shifting time_full to correct time_origin

delandmeterp:
This PR fixes issue #351

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -97,6 +97,8 @@ class FieldSet(object):
             g.check_zonal_periodic()
             if g is not ugrid and g.time_origin != 0:
                 g.time = g.time + (g.time_origin - ugrid.time_origin) / np.timedelta64(1, 's')
+                if g.defer_load:
+                    g.time_full = g.time
                 g.time_origin = ugrid.time_origin
 
     @classmethod
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -267,3 +267,24 @@ def test_fieldset_defer_loading_with_diff_time_origin(tmpdir, filename='test_par
                                  # time=[np.datetime64('2018-04-20 01:00')])
                                  time=[datetime.datetime(2018, 4, 20, 1)])
     pset.execute(AdvectionRK4_3D, runtime=delta(hours=4), dt=delta(hours=1))
+
+
+@pytest.mark.xfail(strict=True)
+def test_fieldset_defer_loading_with_diff_time_origin_should_fail(tmpdir, filename='test_parcels_defer_loading2'):
+    filepath = tmpdir.join(filename)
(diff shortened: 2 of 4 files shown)


parcels/fieldset.py
```diff
@@ -97,6 +97,8 @@ def check_complete(self):
             g.check_zonal_periodic()
             if g is not ugrid and g.time_origin != 0:
                 g.time = g.time + (g.time_origin - ugrid.time_origin) / np.timedelta64(1, 's')
+                if g.defer_load:
+                    g.time_full = g.time
```
erikvansebille: Doesn't this also need a simple unit test, by the way? So that we can catch #351 if it appears again?

delandmeterp: Done

tests/test_fieldset.py
```diff
@@ -242,3 +247,23 @@ class MyParticle(ptype[mode]):
     elif dt_sign == -1:
         temp_theo = temp_vec[0]
     assert np.allclose(temp_theo, pset.particles[0].temp, atol=1e-5)
+
+
+def test_fieldset_defer_loading_with_diff_time_origin(tmpdir, filename='test_parcels_defer_loading'):
```
erikvansebille: Don't we want to have an `xfail` for pytest when the two `time_origins` _don't_ match? This would really catch #351

tests/test_fieldset.py
```diff
@@ -242,3 +247,23 @@ class MyParticle(ptype[mode]):
...
+    data1, dims1 = generate_fieldset(10, 10, 1, 10)
+    dims1['time'] = np.arange(0, 10, 1) * 1800 + (24+25)*3600
+    gridW = RectilinearZGrid(dims1['lon'], dims1['lat'], dims1['depth'], dims1['time'])
+    fieldW = Field('W', np.zeros(data1['U'].shape), grid=gridW)
+    fieldW.grid.time_origin = np.datetime64('2018-04-18')
+    fieldset_out.add_field(fieldW)
+    fieldset_out.write(filepath)
+    fieldset = FieldSet.from_parcels(filepath, extra_fields={'W': 'W'})
+    pset = ParticleSet.from_list(fieldset, pclass=JITParticle, lon=[0.5], lat=[0.5], depth=[0.5],
+                                 # time=[np.datetime64('2018-04-20 01:00')])
```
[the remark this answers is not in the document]
delandmeterp: I would like to start using np.datetime64 in dates. This comment is waiting for next branch which is coming. It will be uncommented and replace the next line.

tests/test_fieldset.py
delandmeterp: I did the mirror test.
In #351, two different `time_origin` were making the test pass while it was supposed to crash.
Here I defined two time vectors such that it crashes if the bug is still there.
I'll add the other config as well (it is a different time_vector with a different time_origin) to show both options.

PR #353, opened 2018-04-23
Xarray no decode cf

delandmeterp:


parcels/field.py
```diff
@@ -281,16 +281,17 @@ def from_netcdf(cls, name, dimensions, filenames, indices={},
         timeFiles = []
         for fname in filenames:
             with NetcdfFileBuffer(fname, dimensions, indices) as filebuffer:
-                timeslices.append(filebuffer.time)
-                timeFiles.append([fname for i in range(len(filebuffer.time))])
+                ftime = filebuffer.time
+                timeslices.append(ftime)
+                timeFiles.append([fname for i in range(len(ftime))])
```
erikvansebille: Since `i` is not actually used, would it be better to do `[fname for _ in range(len(ftime))]`? Would that work?

parcels/field.py
```diff
@@ -993,6 +994,7 @@ def computeTimeChunk(self, data, tindex):
             filebuffer.name = self.dimensions['data'] if 'data' in self.dimensions else self.name
             time_data = filebuffer.time
             if isinstance(time_data[0], np.datetime64):
+                assert isinstance(time_data[0], type(g.time_origin))
```
erikvansebille: Does the assert here need an error message if it fails?

parcels/field.py
```diff
@@ -1101,13 +1108,17 @@ def data(self):
...
+            if self.dataset['decoded'] and 'Unit' not in time_da.attrs:
+                time = np.array(time_da)
+            else:
+                if 'units' not in time_da.attrs and 'Unit' in time_da.attrs:
+                    time_da.attrs['units'] = time_da.attrs['Unit']
+                ds = xr.Dataset({self.dimensions['time']: time_da})
+                ds = xr.decode_cf(ds)
+                time = np.array(getattr(ds, self.dimensions['time']))
+            if isinstance(time[0], datetime.datetime):
+                raise NotImplementedError('Parcels does currently only parses date ranging from 1678 AD to 2262 AD, which are stored by xarray as np.datetime64. A larger time range could be implemented on demand.')
```
erikvansebille: Slight grammatical improvement to error message: ''Parcels currently only parses dates ranging from 1678 AD to 2262 AD, which are stored by xarray as np.datetime64. If you need a wider date range, please open an Issue on the parcels github page."

parcels/grid.py
```diff
@@ -37,6 +38,10 @@ def __init__(self, lon, lat, time, time_origin, mesh):
             logger.warning_once("Casting time data to np.float64")
             self.time = self.time.astype(np.float64)
         self.time_origin = time_origin
+        if self.time_origin:
+            if isinstance(self.time_origin, datetime.datetime):
+                self.time_origin = np.datetime64(self.time_origin)
+            assert isinstance(self.time_origin, np.datetime64)
```
erikvansebille: Again, assert needs a short error message?

PR #359, opened 2018-05-01
interpolation at correct timestep

delandmeterp:


delandmeterp:
Yes I confirm it. 
This works properly in JIT mode.

Issue #357, opened 2018-04-25
Written particle files enormously large due to two 'unlimited' dimensions


pdnooteboom:
Currently I am writing extremely small files (say 'file.nc') (11 particles at 11 time steps). Nevertheless, the written files are 45 Mb each. 

I can create a file with fixed dimensions, with a size of 55 Kb with: 'ncecat file.nc filefixed.nc.' 

How can I write a file with lower size in first instance?

erikvansebille:
I've looked into this a bit today, and it appears to be an issue with the 'chunk-size' of netcdf files. See e.g. the blog piece at https://www.unidata.ucar.edu/blogs/developer/entry/chunking_data_choosing_shapes

I've now added an option `chunksizes` to `ParticleFile.__init__` in https://github.com/OceanParcels/parcels/commit/0770eb74b5fa8da166e0d4c2d9b195b8c386b156

Indeed, the test file reduces from approx 24MB before controlling to <60KB for `chunksizes = [1, 1]`

However, this will likely lead to much slower I/O... @pdnooteboom, can you test this and see whether 1) this reduces file sizes for you and 2) doesn't lead to much increased time spent on file writing?

If all good, then we'll need to think about the best default values for the `chunksizes` argument. Perhaps `[len(pset), 1]`?

pdnooteboom:
I checked the time and the file size when "chunksizes = [1, 1]" and "chunksizes = None" @erikvansebille. Interestingly, in both cases the written file sizes I found are now the same (both 105 Kb which is reasonably low). The "chunksizes = [1, 1]" is slightly faster even.

erikvansebille:
Fixed in #366

PR #366, opened 2018-05-03
Chunksize control particlefile

erikvansebille:
Adding `chunksize` argument to `ParticleFile` class, to allow for control of the netcdf chunk size, and hence the size of the output file. See also https://www.unidata.ucar.edu/blogs/developer/entry/chunking_data_choosing_shapes

In the test, the file size goes from approx 27MB without chunking to less than 60KB with `chunksize` set to the `[len(pset), 1]` value. However, this may lead to slower I/O (to be confirmed).

I would be very grateful if people could run their 'production runs' with this PR and report whether
1) Their output files indeed reduce in size
2) Their walltime doesn't dramatically increase

Once we've got some reports that all is ok, then we can merge

This PR fixes issue #357


willirath:
@ericvansebille @pdnooteboom Would be interesting to see the full netCDF (`ncdump -hs <filename>` ) header of the files with and without chunking.

My understanding is that the chunk-size itself should not determine the file size. But deflation / compression does.  As netCDF compresses the info contained in a chunk of data.  And as compression needs some statistics, chunks of size 1 won't allow for any transformation of the data that represents the information in a cheaper way.

The large benefits from chunking reported here and in #357 (O(10 MB) → O(100 KB)) likely stem from two things:

1.  Chunks that contain more than 1 data point making it possible to deflate the data at all.

2.  With all particles being close together in these short experiments, the deflation can be very efficient. For large psets and / or for long runs, I'd guess that we're down to maybe a factor of 5 or 10 between the original and the chunked format.

A note on the chunk size: As soon as you start looking at single and potentially long trajectories, a chunking that covers the _whole_ list of particles (`len(pset)`) may mean that for each trajectory, _all_ data has to be read from disk, deflated, and then all but the desired particle is forgotten.

tests/test_particle_sets.py
```diff
@@ -93,34 +94,42 @@ def IncrLon(particle, fieldset, time, dt):
...
-    assert samplevar.shape == (runtime // repeatdt+1, min(maxvar+1, runtime)+1)
-    if repeatdt == 0:
-        # test whether samplevar[i, i+k] = k for k=range(maxvar)
-        for k in range(maxvar):
-            for i in range(runtime-k):
-                assert(samplevar[i, i+k] == k)
+    ncfile.close()
+    if type == 'releasedt':
+        assert samplevar.shape == (runtime // repeatdt+1, min(maxvar+1, runtime)+1)
+        assert np.allclose([p.sample_var for p in pset], np.arange(maxvar, -1, -repeatdt))
```
delandmeterp: Is it normal that we check p.sample_var for p in pset for `type == 'releasedt'` but not `type == 'timearr'` ?

PR #375, opened 2018-05-30
Scipy s grid search fix

delandmeterp:


parcels/field.py
```diff
@@ -511,7 +513,9 @@ def search_indices_vertical_s(self, x, y, z, xi, yi, xsi, eta, ti, time):
                     (1-xsi)*eta * grid.depth[ti:ti+2, :, yi+1, xi]
                 t0 = grid.time[ti]
                 t1 = grid.time[ti + 1]
-                depth_vector = dv2[0, :] + (dv2[1, :]-dv2[0, :]) * (time - t0) / (t1 - t0)
+                tt = (time - t0) / (t1-t0)
+                assert tt >= 0 and tt <= 1
```
erikvansebille: needs an error message with the assert?

parcels/field.py
```diff
@@ -511,7 +513,9 @@ def search_indices_vertical_s(self, x, y, z, xi, yi, xsi, eta, ti, time):
                     (1-xsi)*eta * grid.depth[ti:ti+2, :, yi+1, xi]
                 t0 = grid.time[ti]
                 t1 = grid.time[ti + 1]
-                depth_vector = dv2[0, :] + (dv2[1, :]-dv2[0, :]) * (time - t0) / (t1 - t0)
+                tt = (time - t0) / (t1-t0)
```
erikvansebille: since `t0` and `t1` are only used in the scope here, would it not be clearer to remove lines 514-515 and here write `tt = (time - grid.time[ti]) /(grid.time[ti+1] - grid.time[ti])`?

PR #380, opened 2018-06-11
Improved netcdf filebuffer

delandmeterp:


parcels/field.py
```diff
@@ -267,9 +267,11 @@ def from_netcdf(cls, filenames, variable, dimensions, indices={},
 
         if not isinstance(filenames, Iterable) or isinstance(filenames, str):
             filenames = [filenames]
+        indices = indices.copy()
```
erikvansebille: why do we need to copy? If this is necessary, perhaps add a short comment on the rationale

parcels/field.py
```diff
@@ -267,9 +267,13 @@ def from_netcdf(cls, filenames, variable, dimensions, indices={},
 
         if not isinstance(filenames, Iterable) or isinstance(filenames, str):
             filenames = [filenames]
+        # since indices will be modified through filebuffer.indices, we need to make a copy
+        # (especially nasty with default argument ({}) )
```
willirath: Mutable default args should be avoided anyway: <http://docs.python-guide.org/en/latest/writing/gotchas/#what-you-should-do-instead>

So I'd go for the `indices=None` default and then use the suggested construct

```python
if indices is None:
    indices = {}
```

erikvansebille: Good point, @willirath. I agree that this would be the proper solution to this issue (and it might also fix some other py.test issues we've experienced before?)

Issue #384, opened 2018-06-12
Arithmetic operators for Fields


erikvansebille:
Following comments in #371, it would be useful to be able to add, subtract and perhaps use other operators directly on `Fields`. This should also check that `grids` are consistent etc

erikvansebille:
This has now been implemented with `FieldLists` (#393) so closing this Issue for now

PR #385, opened 2018-06-14
Field arithmetic operators

erikvansebille:
This PR introduces support for the `+` and `-` arithmetic operators on `Field` objects. An error is thrown if the `Fields` do not have the same `Grid`

This implements #384


erikvansebille:
Closing this PR, as arithmetic operators can now be dealt with more cleanly through FieldLists (#393)

PR #388, opened 2018-06-25
Vector fields

delandmeterp:


parcels/fieldset.py
```diff
@@ -109,6 +108,8 @@ def check_complete(self):
                 if g.defer_load:
                     g.time_full = g.time_full + (g.time_origin - ugrid.time_origin) / np.timedelta64(1, 's')
                 g.time_origin = ugrid.time_origin
+        if getattr(self, 'UV') is None:
+            self.add_field(VectorField('UV', self.U, self.V))
```
erikvansebille: It's somewhat confusing to add a `VectorField` with a `add_field` method. Should we create a new method `add_vectorfield`? Especially because half of the lines of `add_field` only apply to `Fields`

tests/test_fieldset.py
```diff
@@ -201,6 +201,32 @@ def test_fieldset_constant(mode):
...
+def test_vector_fields(mode, reverse):
+    lon = np.linspace(0., 10., 12, dtype=np.float32)
+    lat = np.linspace(0., 10., 10, dtype=np.float32)
+    U = np.ones((10, 12), dtype=np.float32)
+    V = np.zeros((10, 12), dtype=np.float32)
+    data = {'U': U, 'V': V}
+    dimensions = {'U': {'lat': lat, 'lon': lon},
+                  'V': {'lat': lat, 'lon': lon}}
+    fieldset = FieldSet.from_data(data, dimensions, mesh='flat')
+    if reverse:  # we test that we can freely edit whatever UV field
```
erikvansebille: `reverse` might be a confusing term here. Would be `swapUV` be better, or so?

Issue #326, opened 2018-03-28
How to override velocities passed to the advection kernels


willirath:
_(My example is for adding Stokes drift to the ocean velocities used for advecting particles.  The general idea of combining velocity fields in a non-invasive way is very interesting for various applications, though.)_

We were working on the inclusion of Stokes drift. Ideally, we'd just make sure that the velocities used for advection are `U_ocean + U_stokes`. The following is a first stab at sub-classing `FieldSet` to make `FieldSet.UV` return a sum of `(U, V)` and the Stokes-drift velocities:

```python
from parcels import FieldSet, Field
import numpy as np

class StokesFieldSet(FieldSet):
    """FieldSet with additional stokes velocities.

    :param U: :class:`parcels.field.Field` object for zonal velocity component
    :param V: :class:`parcels.field.Field` object for meridional velocity component
    :param US: :class:`parcels.field.Field` object for zonal stokes velocity component
    :param VS: :class:`parcels.field.Field` object for meridional stokes velocity component
    :param fields: Dictionary of additional :class:`parcels.field.Field` objects

    """
    def __init__(self, U, V, US=None, VS=None, fields={}):
        super(StokesFieldSet, self).__init__(U, V, fields=fields)

        if US:
            self.add_field(US)
        if VS:
            self.add_field(VS)

        self._UV = self.UV

        UV = Field('UV', None)
        UV.fieldset = self
        self.UV = UV
        self.UV.getUV = self.getUV_with_stokes

    def getUV_with_stokes(self, time, x, y, z):
        _U, _V = self._UV[time, x, y, z]
        if "US" in [f.name for f in self.fields]:
            _U = _U + self.US[time, x, y, z]
        if "VS" in [f.name for f in self.fields]:
            _V = _V + self.VS[time, x, y, z]
        return (_U, _V)


def test_stokes_fieldset():
    np.random.seed(1)

    xdim = 2
    ydim = 2

    lon = np.linspace(-20, 20, xdim, dtype=np.float32)
    lat = np.linspace(-20, 20, ydim, dtype=np.float32)

    dimensions = {'lon': lon, 'lat': lat}

    U = Field('U', np.random.randn(ydim, xdim), lon=lon, lat=lat)
    V = Field('V', np.random.randn(ydim, xdim), lon=lon, lat=lat)
    US = Field('US', np.random.randn(ydim, xdim), lon=lon, lat=lat)
[...]

PR #393, opened 2018-06-29
List of velocity fields for advection

erikvansebille:
This PR introduced a new feature to Parcels: `FieldLists`.  With these `FieldLists`, you can run a kernel over multiple `Fields` and the values of the individual `Field` interpolations are automatically added. Note that the `Fields` can be on different `Grids`

This is useful for example if particles need to be advected with a combination of different velocity fields such as ocean currents, windage, etc). See also #326, where @willirath suggested such a feature

If the velocity fields are `(U_ocean, V_ocean)` for the currents and `(U_stokes, V_stokes)` for Stokes drift, then they can be combined as
```
fset = FieldSet([U_ocean, U_stokes], [V_ocean, V_stokes])
```
And then simply use the built-in `AdvectionRK4` kernel as normal

Lists can be of arbitrary lengths. This also works in 3D and for fields other than velocity `Fields`.

I'd really like some feedback on whether this works. Note that before merging a few minor things need to be fixed
1) Confirmation that the implementation also works on `mesh=spherical` and Curvilinear `Grids` (now tested only for `mesh=flat`)
2) Some under-the-hood cleaning of the `fieldset.ugrid` attribute, which is a rather hacky implementation
3) Create a tutorial to explain how to use these `FieldLists`. This tutorial can then also explain how to work with different `Grids`

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -294,7 +294,7 @@ class KernelGenerator(ast.NodeVisitor):
                         if f not in self.field_args:
                             args += [c.Pointer(c.Value("CField", "%s" % f))]
                     except:
-                        if fieldset.U.grid.gtype in [GridCode.CurvilinearZGrid, GridCode.CurvilinearSGrid]:
+                        if fieldset.ugrid.gtype in [GridCode.CurvilinearZGrid, GridCode.CurvilinearSGrid]:
                             raise RuntimeError("cosU, sinU, cosV and sinV fields must be defined for a proper rotation of U, V fields in curvilinear grids")
         for const, _ in self.const_args.items():
             args += [c.Value("float", const)]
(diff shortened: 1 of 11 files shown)


willirath:
This looks very interesting!

> If the velocity fields are `(U_ocean, V_ocean)` for the currents and `(U_stokes, V_stokes)` for Stokes drift, then they can be combined as
> ```
> fset = FieldSet([U_ocean, U_stokes], [V_ocean, V_stokes])
> ```

Is this also meant to work for `U_ocean` and `U_stokes` living on different grids / time axes?

erikvansebille:
Yes, this works on Fields on different grids too.

Under the hood, what happens is that the particle location is interpolated on each Field separately, and that the different values are then added together. So where normally we would have
```
particle.p = fieldset.P[time, particle.lon, particle.lat, particle.depth]
```
Now we have for a FieldList `P=[P1, P2]`
```
tmp1 = fieldset.P1[time, particle.lon, particle.lat, particle.depth]
tmp2 = fieldset.P2[time, particle.lon, particle.lat, particle.depth]
particle.p = tmp1 + tmp2
```
And because the interpolations of `P1` and `P2` are done separately, they can be done on different grids too

Issue #389, opened 2018-06-25
Allow setting of any Variable from lists in ParticleSet.from_list()


erikvansebille:
It would be a nice feature to be able to set the value of any (custom) `Variable` for a `Particle` directly from the `ParticleSet.from_list()` constructor, just like how it now works for `lon`;, `lat`, `depth`, and `time`.

For example, if a `Particle` class has a variable `v`, then one should be able to set the values for `particle.v` with 
```
pset = ParticleSet.from_list(fset, pclass, lon=[..], lat=[..], time=[..], v=[..])
```

PR #397, opened 2018-07-26
Initialisation of custom Variables directly in ParticleSet.from_list()

erikvansebille:
This PR implements #389, so that now the value of custom `Variables` can be set in `ParticleSet.from_list()`, similar to how `lon` and `lat` are set there

This PR also improves  `print()` for `ParticleSets`, so that all custom `Variables` are also printed

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -99,6 +99,8 @@ class ParticleSet(object):
                 self.particles[i] = pclass(lon[i], lat[i], fieldset=fieldset, depth=depth[i], cptr=cptr(i), time=time[i])
                 # Set other Variables if provided
                 for kwvar in kwargs:
+                    if not hasattr(self.particles[i], kwvar):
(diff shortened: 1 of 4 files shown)


parcels/particleset.py
```diff
@@ -96,11 +97,14 @@ def cptr(i):
 
             for i in range(size):
                 self.particles[i] = pclass(lon[i], lat[i], fieldset=fieldset, depth=depth[i], cptr=cptr(i), time=time[i])
+                # Set other Variables if provided
+                for kwvar in kwargs:
+                    setattr(self.particles[i], kwvar, kwargs[kwvar][i])
```
delandmeterp: Since kwargs gives a lot of freedom,  it would be nice to test here if kwvar is a particle attribute, with a if(hasattr)
or a try/except, with a nice print in case of error
(a difficulty for new users remains the reading of the error message when there is a problem)

Issue #284, opened 2018-01-29
Compiler class might need some cleanup


willirath:
Currently, [`parcels/compiler.py`](https://github.com/OceanParcels/parcels/blob/d283b44fbb73fed0698a6625ba4d968f0f092c12/parcels/compiler.py) has some problems:

## 1. The `ld` argument is never used

In <https://github.com/OceanParcels/parcels/blob/d283b44fbb73fed0698a6625ba4d968f0f092c12/parcels/compiler.py#L37>, there is a variable `self._ld` which is set from either an env var or from whatever the input arg `ld` contains. (Note the order: `ld` only is used if there is no env var.)  I'd suggest to solve this by just dropping the case with a separate linker.

## 2. `$CC` in the env overrides everything

The more specific `GNUCompiler` sets `gcc` specific flags and also initializes `Compiler` with `cc="gcc"`, the initialization of `Compiler` disregards any speicific compiler given via `cc` if there is an environment variable called `CC` in https://github.com/OceanParcels/parcels/blob/d283b44fbb73fed0698a6625ba4d968f0f092c12/parcels/compiler.py#L36.

I'm not sure this is what the user expects.  I'd probably expect the compiler to fall back to come generic `cc` or `$CC` only if the call of the compiler did not specify anything.

(See also #234 and #283.)

PR #399, opened 2018-07-26
Compiler cleanup

erikvansebille:
Cleaning up the Compiler class, as suggestion by @willirath in #284 

Note that this also fixes an old issue (#332) where users had to explicitly set `export CC=gcc`. That is not necessary anymore with this change, and therefore also works now with xcode 9.3 (#343).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,6 @@ matrix:
       osx_image: xcode9.3
       language: generic
       env: PY_VERSION=3 MINICONDA_NAME=Miniconda3-latest-MacOSX-x86_64.sh
-  allow_failures:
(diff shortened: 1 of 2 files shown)


willirath:
I have one additional remark: Currently, it's difficult to test for the compile command. If https://github.com/OceanParcels/parcels/blob/76ff82aaf663b4738d72822ed187a66fa16f29f9/parcels/compiler.py#L42 was made a property of the compiler class, it would be accessible from a unit test without actually compiling.

willirath:
(Writing a test for this would be a nice exercise to get somebody like a new student started with test-driven development.)

PR #400, opened 2018-07-30
Minor bug fixes to Cgrid interpolation

erikvansebille:
This PR fixes some minor issues to the C-grid interpolation in Scipy mode. 
It also adds .pytest_cache directory to .gitignore

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -200,7 +200,7 @@ class Field(object):
             self.units = self.unitconverters[fieldtype]
         else:
             raise ValueError("Unsupported mesh type. Choose either: 'spherical' or 'flat'")
-        self.interp_method = interp_method
+        self.interp_method = interp_method[name] if name in interp_method else interp_method
         self.fieldset = None
         if self.name in ['cosU', 'sinU', 'cosV', 'sinV']:
             self.allow_time_extrapolation = True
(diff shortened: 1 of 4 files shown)


parcels/field.py
```diff
@@ -200,7 +200,7 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
             self.units = self.unitconverters[fieldtype]
         else:
             raise ValueError("Unsupported mesh type. Choose either: 'spherical' or 'flat'")
-        self.interp_method = interp_method
+        self.interp_method = interp_method[name] if name in interp_method else interp_method
```
delandmeterp: interp_method should be tested to see if it is a dict or not

Example that does not work currently: `name = 'i'` and `interp_method = 'linear'`
In this case `name in interp_method` returns `True`, then `interp_method[name]` breaks

parcels/fieldset.py
```diff
@@ -239,8 +239,15 @@ def from_nemo(cls, filenames, variables, dimensions, indices=None, mesh='spheric
 
         dimension_filename = filenames.pop('mesh_mask')
 
+        interp_method = {}
+        for v in variables:
+            if v in ['U', 'V', 'W']:
+                interp_method[v] = 'cgrid_linear'
+            else:
+                interp_method[v] = 'linear'
```
delandmeterp: Should we let freedom to the user?
I am thinking if he wants to interpolate the tracer with `'nearest'`
(This is a valid choice. I guess tracer are considered constant per cell when the integral over the domain is computed, but I'm not sure)

Watch out, a tracer field does not live on the same grid as U and V
=> line 223 should be modified giving `dimensions['U'], dimensions['Salt']` (in the doc)

PR #401, opened 2018-08-02
Refactoring of plotting routines

### LINKED ISSUE #221: Generalize the plotting interface
The standard basemap Mercator projection when working with a global grid might be improved. A feature enhancement will involve a more adaptive plotting interface.
The ability to pass different basemap or cartopy projections to ParticleSet.show() could be useful. I can make an attempt to enhance the plotting interface. 

Another plotting feature might include changing an (x,y,z) particle trajectory color to indicated depth in a 2-D plot.

erikvansebille:
This PR (which fixes #221) is a refactoring of the plotting routines for Parcels. Specifically
- `Basemap` is replaced for the newer and more versatile `cartopy` library
- All plotting of `Fields` and `Fieldsets` is moved to a new file `plotting.py`, to reduce clutter
- Continents (`land=True`) can now be shown on any map, and are set by default depending on `mesh` type
- the `plotTrajectoryFile` function now also uses the `plotting.py` routines to set axes and plot background fields etc

Note that after this PR is released, the oceanparcels.org website will need to be updated to change the `conda install` instruction from using `Basemap` to `cartopy`


parcels/scripts/plotParticles.py
```diff
@@ -1,17 +1,20 @@
 #!/usr/bin/env python
-from netCDF4 import Dataset
+import xarray as xr
```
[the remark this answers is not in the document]
erikvansebille: No, plotParticles.py is still its own function (in the `scripts` directory). `plotting.py` only provides the helper functions of creating a `cartopy` axis, plotting a `Field`, and plotting a `ParticleSet` (which is different from plotting a `particleFile`)

parcels/scripts/plotParticles.py
```diff
@@ -24,71 +27,91 @@ def plotTrajectoriesFile(filename, mode='2d', tracerfile=None, tracerfield='P',
...
-        X = tfile.variables[tracerlon]
-        Y = tfile.variables[tracerlat]
-        P = tfile.variables[tracerfield]
-        plt.contourf(np.squeeze(X), np.squeeze(Y), np.squeeze(P))
-        tfile.close()
+    if tracerfile is not None and mode is not 'hist2d':
+        tracerfld = Field.from_netcdf(tracerfile, tracerfield, {'lon': tracerlon, 'lat': tracerlat})
+        plt, fig, ax, cartopy = plotfield(tracerfld)
+        if plt is None:
+            return  # creating axes was not possible
```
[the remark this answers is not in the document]
erikvansebille: The message is already thrown in `create_parcelsfig_axis()` in plotting.py. But because we don't error, we have to catch any `None` return

parcels/examples/tutorial_FieldLists.ipynb
```diff
@@ -24,6 +24,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "%matplotlib inline\n",
```
delandmeterp: Do you know what does that line?

parcels/examples/tutorial_jit_vs_scipy.ipynb
```diff
@@ -74,7 +76,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As you can see above, **Scipy mode took more than 20 times as long** (17.2 seconds versus 0.79 seconds) as the JIT mode!!\n",
+    "As you can see above, **Scipy mode took almost 3 times as long** (7.3 seconds versus 2.6 seconds) as the JIT mode!!\n",
```
delandmeterp: We want to highlight that JIT is way faster than Scipy. Why don't take we more particles such that JIT time is not mostly initialisation.

parcels/scripts/plotParticles.py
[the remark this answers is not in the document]
delandmeterp: I mean there is a function 
`plotparticles()` in `plotting.py`
and in `scripts/plogParticles.py` there is a function `plotTrajectoryFiles()`.

parcels/plotting.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def plotfield(field, show_time=None, domain=None, projection=None, land=None,
+              vmin=None, vmax=None, savefile=None, **kwargs):
+    """Function to plot a Parcels Field
+
+    :param show_time: Time at which to show the Field
+    :param domain: Four-vector (latN, latS, lonE, lonW) defining domain to show
+    :param projection: type of cartopy projection to use (default PlateCarree)
+    :param land: Boolean whether to show land
```
delandmeterp: Here (and everywhere else where there is a `land` flag).
The default value for `land` is `None`, while it is a boolean.
I guess it is because you want `land=True` if there is a geomap, and `False` otherwise?
I would still put it to `True` by default, since it is clear, and simply put in the docstrings that this value is ignored for flat mesh, no?

parcels/plotting.py
```diff
@@ -0,0 +1,272 @@
...
+    :param vmin: minimum colour scale (only in single-plot mode)
+    :param vmax: maximum colour scale (only in single-plot mode)
+    :param savefile: Name of a file to save the plot to
+    :param animation: Boolean whether result is a single plot, or an animation
+    """
+
+    if type(field) is VectorField:
+        geomap = True if field.U.grid.mesh == 'spherical' else False
```
delandmeterp: `geomap = field.U.grid.mesh == 'spherical'`

By why calling it geomap? Let's keep it `spherical` to be consistent with previous uses?

parcels/examples/tutorial_FieldLists.ipynb
erikvansebille: If we don't have this line, then the first plot of the notebook is not shown. Something strange with matplotlib initialisation...

PR #402, opened 2018-08-02
Field write extension to CurvilinearGrids

### LINKED ISSUE #378: Field.write() does not work for curvilinear grids
Field.write() was originally written for rectilinear grids.
Nothing was done for curvilinear grids when those were developed.

erikvansebille:
This PR implements `Field.write()` for `CurvilinearZGrids` (so fixes #378) 

It also adds support for non-dictionary filenames, if the filenames for all variables are the same (i.e. if all variables are in one file, then no need to specify a dictionary entry of filenames for for each variable).

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -935,17 +935,16 @@ class Field(object):
         vname_depth = 'depth%s' % self.name.lower()
 
         # Create DataArray objects for file I/O
-        t, d = (self.grid.time.size, self.grid.depth.size)
         if type(self.grid) is RectilinearZGrid:
-            x, y = (self.grid.lon.size, self.grid.lat.size)
-            nav_lon = xr.DataArray(self.grid.lon + np.zeros((y, x), dtype=np.float32),
+            nav_lon = xr.DataArray(self.grid.lon + np.zeros((self.grid.ydim, self.grid.xdim), dtype=np.float32),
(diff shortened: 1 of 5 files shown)


parcels/field.py
```diff
@@ -935,12 +935,20 @@ def write(self, filename, varname=None):
...
-                               coords=[('y', self.grid.lat), ('x', self.grid.lon)])
+        t, d = (self.grid.time.size, self.grid.depth.size)
+        if type(self.grid) is RectilinearZGrid:
+            x, y = (self.grid.lon.size, self.grid.lat.size)
+            nav_lon = xr.DataArray(self.grid.lon + np.zeros((y, x), dtype=np.float32),
+                                   coords=[('y', self.grid.lat), ('x', self.grid.lon)])
+            nav_lat = xr.DataArray(self.grid.lat.reshape(y, 1) + np.zeros(x, dtype=np.float32),
+                                   coords=[('y', self.grid.lat), ('x', self.grid.lon)])
+        elif type(self.grid) is CurvilinearZGrid:
+            y, x = self.grid.lat.shape
```
delandmeterp: self.grid.xdim and self.grid.ydim exist (in a unique way, indpendent of Rectilinear and Curvi grids.
(also zdim)

PR #403, opened 2018-08-03
Updating example meshes

erikvansebille:
This PR adds `flat` meshes to the `moving_eddies` and `brownian` examples, and updates the FieldSet generation for the `peninsula` example. Furthermore, it adds information on whether the mesh was `flat` or `spherical` to `Field.write`. 

That last feature is important for the plotting of Fields (see also #401) as it will allow us to show the examples in `flat` meshes rather than `spherical` ones

### DIFF
--- a/parcels/examples/example_moving_eddies.py
+++ b/parcels/examples/example_moving_eddies.py
@@ -12,29 +12,40 @@ ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
 method = {'RK4': AdvectionRK4, 'EE': AdvectionEE, 'RK45': AdvectionRK45}
 
 
-def moving_eddies_fieldset(xdim=200, ydim=350):
+def moving_eddies_fieldset(xdim=200, ydim=350, mesh='flat'):
     """Generate a fieldset encapsulating the flow field consisting of two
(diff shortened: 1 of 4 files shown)


parcels/examples/example_moving_eddies.py
```diff
@@ -117,35 +132,44 @@ def test_moving_eddies_fwdbwd(mode, npart=2):
     pset.execute(method, endtime=0, dt=-dt,
                  output_file=pset.ParticleFile(name="EddyParticlebwd", outputdt=outputdt))
 
-    assert(pset[0].lon > 3.2 and 45.9 < pset[0].lat < 46.1)
-    assert(pset[1].lon > 3.2 and 47.7 < pset[1].lat < 47.9)
-
+    assert np.allclose([p.lon for p in pset], lons)
+    assert np.allclose([p.lat for p in pset], lats)
```
delandmeterp: Why is this new? It is not related to the other changes.
We were not checking the particle comes back to where it initially started?

erikvansebille: The locations used to be hardcoded; now we check whether the final locations of the fwd_bwd run are equal to the start locations. So it's not really new; it's just more general

PR #408, opened 2018-08-13
Adding FieldLists to FieldSet.fields()

erikvansebille:
Fixing a bug where `FieldLists` were not returned by the `FieldSet.fields()` function


parcels/fieldset.py
```diff
@@ -280,7 +280,14 @@ def from_parcels(cls, basename, uvar='vozocrtx', vvar='vomecrty', indices=None,
...
         associated with this FieldSet"""
-        return [v for v in self.__dict__.values() if isinstance(v, Field)]
+        fields = []
+        for v in self.__dict__.values():
+            if isinstance(v, Field):
+                fields.append(v)
+            elif isinstance(v, FieldList):
+                for v2 in v:
+                    fields.append(v2)
+        return fields
```
delandmeterp: Where do we use this function?
If the fields of a FieldList are already defined in FieldSet, then, they will appear twice in FieldSet.fields(). Do we want that?

Also, I don't remember. Isn't a requirement that the fields in the FieldList are in FieldSet (for JIT mode)? I have something like this in mind, but I'm not sure.

erikvansebille: The bug came up in `FieldSet.computeTimeChunk`, where we twice use `for f in self.fields:`

And no, the Fields in a FieldList are not also in the FieldSet by themselves. For JIT, we add them to the `KernelGenerator.field_args` list in `visit_FieldListNode()`, perhaps that's what you're thinking of?

delandmeterp: Oh, maybe that's this! In `FieldSet.computeTimeChunk()`, we looped over self.fields so we used to need that all fields are defined in the FieldSet.
So we don't need this anymore thanks to your commit. But still we want to avoid adding twice a field, no?

PR #410, opened 2018-08-15
closing Dataset in plotTrajectoriesFile

delandmeterp:
plotTrajectoriesFile() was opening a Dataset without closing it


parcels/scripts/plotParticles.py
```diff
@@ -92,10 +92,12 @@ def animate(t):
 
     if mode == 'movie2d_notebook':
         plt.close()
+        pfile.close()
         return anim
     else:
         if show_plt:
             plt.show()
+        pfile.close()
```
erikvansebille: I would close right after last use of `pfile`

PR #416, opened 2018-08-17
Grid takes into account a time argument which is an int or a float.

delandmeterp:
We add a test that asserts it is the case

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -35,7 +35,7 @@ class Grid(object):
             logger.warning_once("Casting lat data to np.float32")
             self.lat = self.lat.astype(np.float32)
         if not self.time.dtype == np.float64:
-            assert isinstance(self.time[0], int) or isinstance(self.time[0], float), 'Time vector must be an array of int or floats'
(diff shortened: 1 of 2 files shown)


parcels/grid.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, lon, lat, time, time_origin, mesh):
             logger.warning_once("Casting lat data to np.float32")
             self.lat = self.lat.astype(np.float32)
         if not self.time.dtype == np.float64:
+            assert isinstance(self.time[0], int) or isinstance(self.time[0], float), 'Time vector must be an array of int or floats'
```
erikvansebille: should we also allow np.float32? From [checking stackexchange](https://stackoverflow.com/questions/28292542/how-to-check-if-a-number-is-a-np-float64-or-np-float32-or-np-float16?lq=1), I think we can do `assert isinstance(self.time[0], (int, float, np.floating))`, since `isinstance` can also take a tuple of types

delandmeterp: Cool, I tried with a list, but not with a tuple. I'll change it.
I thought `int` and `float` were the general types. But you are right it breaks.

General types are np.floating and np.integer.

PR #427, opened 2018-09-03
Allowing variable names to be a list

erikvansebille:
In the GlobCurrent v3.0 data set, the variable names change halfway (between delay-time and near-real-time mode) from `eastward_eulerian_current_velocity` to `u`. 

Here, we allow that the `variable` names can be a `list`, e.g.:
```
variables = {'U': ['eastward_eulerian_current_velocity', 'u'], 'V': ['northward_eulerian_current_velocity', 'v']}
```
Parcels will then work with either of these two names, whichever is in the file

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -335,7 +335,7 @@ class Field(object):
                 with NetcdfFileBuffer(fname, dimensions, indices) as filebuffer:
                     # If Field.from_netcdf is called directly, it may not have a 'data' dimension
                     # In that case, assume that 'name' is the data dimension
-                    filebuffer.name = dimensions['data'] if 'data' in dimensions else variable
+                    filebuffer.name = filebuffer.parse_name(dimensions, variable)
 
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -1174,6 +1174,14 @@ def __enter__(self):
     def __exit__(self, type, value, traceback):
         self.dataset.close()
 
+    def parse_name(self, dimensions, variable):
+        name = dimensions['data'] if 'data' in dimensions else variable
+        if isinstance(name, list):
+            for nm in name:
+                if hasattr(self.dataset, nm):
+                    name = nm
```
delandmeterp: You would like to put a `break` statement here. Such that if you find the name in your list, you stop searching (then it will also make a priority: first name in the list = preferred name

delandmeterp: We should add a piece of code like this 

`if isinstance(name, list): raise ...`

such that if nothing was found we throw an error

PR #429, opened 2018-09-04
Curvilinear pset fromfield

delandmeterp:


tests/test_grids.py
```diff
@@ -343,6 +343,9 @@ def test_nemo_grid(mode):
     dimensions = {'lon': 'glamf', 'lat': 'gphif'}
     field_set = FieldSet.from_nemo(filenames, variables, dimensions)
 
+    # test ParticleSet.from_field on curvilinear grids
+    ParticleSet.from_field(field_set, ptype[mode], start_field=field_set.U, size=5)
+
```
erikvansebille: Does this still need an assert statement to check if the particles are indeed roughly homogenous? Because the `ParticleSet.from_field()` statement above was simply to check if there was no error, for now

parcels/particleset.py
```diff
@@ -156,24 +157,31 @@ def from_field(cls, fieldset, pclass, start_field, size, mode='monte_carlo', dep
...
+            j, i = np.unravel_index(inds, p_interior.shape)
+            grid = start_field.grid
+            if grid.gtype in [GridCode.RectilinearZGrid, GridCode.RectilinearSGrid]:
+                lon = grid.lon[i] + xsi * (grid.lon[i + 1] - grid.lon[i])
+                lat = grid.lat[j] + eta * (grid.lat[j + 1] - grid.lat[j])
+            else:
+                lons = np.array([grid.lon[j, i], grid.lon[j, i+1], grid.lon[j+1, i+1], grid.lon[j+1, i]])
+                if grid.mesh == 'spherical':
+                    lons[1:] = np.where(lons[1:] - lons[0] > 180, lons[1:]-360, lons[1:])
+                    lons[1:] = np.where(-lons[1:] + lons[0] > 180, lons[1:]+360, lons[1:])
```
erikvansebille: So this assumes that the grid is from -180 to 180? I don't think we want to assume/hard-code any longitude wrapping boundary. Can we make this more general?

delandmeterp: No, if lons are for example [269, -90, -89, -88]
It will (locally) rewrite them into
[269, 270, 271, 272]

Otherwise, interpolating between 269 and -90 will make nonsense

This isn't perfect, but it follows the same structure as in other parts of the code. All this should indeed be improved (see TODO comment in PR #428

tests/test_grids.py
delandmeterp: Done in test_particle_sets.py

tests/test_particle_sets.py
```diff
@@ -66,6 +66,32 @@ def test_pset_create_field(fieldset, mode, npart=100):
...
+    ufield = Field('U', u, grid=grid)
+    vfield = Field('V', v, grid=grid)
+    fieldset = FieldSet(ufield, vfield)
+    pset = ParticleSet.from_field(fieldset, size=npart, pclass=ptype['scipy'], start_field=fieldset.V)
+
+    lons = [p.lon+1 for p in pset]
+    lats = [p.lat+1 for p in pset]
+    thetas = np.arctan2(lats, lons)
+    assert np.all(np.pi/4-dtheta < thetas)
+    assert np.all(thetas < np.pi/3+dtheta)
```
erikvansebille: So we don't assert that the `r`'s are correct too?

PR #431, opened 2018-09-04
Ignoring single-snapshot Grids in calculating of dimrange(`time`)

erikvansebille:
Fixing a bug where `dimrange('time_full')` was wrong when `FieldSet` included a `Field` with one snapshot

### DIFF
--- a/parcels/gridset.py
+++ b/parcels/gridset.py
@@ -40,14 +40,15 @@ class GridSet(object):
            in a gridset. Useful for finding e.g. longitude range that
            overlaps on all grids in a gridset"""
 
-        maxleft, minright = (0, np.inf)
+        maxleft, minright = (-np.inf, np.inf)
         for g in self.grids:
-            if dim == 'time_full' and len(getattr(g, dim)) == 1:
-                continue  # not including grids where only one time_full entry, as with allow_time_extrapolation=True
+            if len(getattr(g, dim)) == 1:
+                continue  # not including grids where only one entry
             else:
                 maxleft = max(maxleft, getattr(g, dim)[0])
                 minright = min(minright, getattr(g, dim)[-1])
-        minright = 0 if minright == np.inf else minright  # if all len(time_full) == 1
+        maxleft = 0 if maxleft == -np.inf else maxleft  # if all len(dim) == 1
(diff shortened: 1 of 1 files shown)


parcels/gridset.py
```diff
@@ -40,10 +40,14 @@ def dimrange(self, dim):
            in a gridset. Useful for finding e.g. longitude range that
            overlaps on all grids in a gridset"""
 
-        maxleft, minright = (0, np.infty)
+        maxleft, minright = (0, np.inf)
```
delandmeterp: As you suggested, we should range from -inf to +inf

parcels/gridset.py
```diff
@@ -40,10 +40,14 @@ def dimrange(self, dim):
...
         for g in self.grids:
-            maxleft = max(maxleft, getattr(g, dim)[0])
-            minright = min(minright, getattr(g, dim)[-1])
+            if dim == 'time_full' and len(getattr(g, dim)) == 1:
+                continue  # not including grids where only one time_full entry, as with allow_time_extrapolation=True
+            else:
+                maxleft = max(maxleft, getattr(g, dim)[0])
+                minright = min(minright, getattr(g, dim)[-1])
+        minright = 0 if minright == np.inf else minright  # if all len(time_full) == 1
```
delandmeterp: Then we have to consider -inf here as well

parcels/gridset.py
```diff
@@ -40,10 +40,14 @@ def dimrange(self, dim):
            in a gridset. Useful for finding e.g. longitude range that
            overlaps on all grids in a gridset"""
 
-        maxleft, minright = (0, np.infty)
+        maxleft, minright = (0, np.inf)
         for g in self.grids:
-            maxleft = max(maxleft, getattr(g, dim)[0])
-            minright = min(minright, getattr(g, dim)[-1])
+            if dim == 'time_full' and len(getattr(g, dim)) == 1:
+                continue  # not including grids where only one time_full entry, as with allow_time_extrapolation=True
```
delandmeterp: Should we do exactly the same for depth dimension?

This can be interesting if we want to read both 2d and 3d fields
(example velocity 3d and EEZ 2d)

Issue #432, opened 2018-09-06
Changing API for Field summing from [] to + operator


erikvansebille:
PR #393 introduced FieldLists as a way to specify that a Field interpolation needed to be done as the sum of a set of Fields (e.g. flow = currents+wind+stokes)

However, it might be more intuitive to use the `+` operator for this feature, just like kernels can be concatenated with the `+` operator.

In the future, this will then also allow us to re-use the list operator `[]` to specify a hierarchy or nested domains

erikvansebille:
Fixed in #435

PR #435, opened 2018-09-10
Changing FieldLists to SummedFields

erikvansebille:
This PR introduces a change in the API for combining of different `Fields` for interpolation (see also Issue #432). 

Using the `+` operator is much more intuitive if the data in different `Fields` need to be summed at a particle location.

Also, this means that we can use the list operator `[]` for a future feature such as possibly nested `Fields`

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -981,10 +981,18 @@ class Field(object):
         if isinstance(self, Field) and isinstance(field, Field):
             return SummedField([self, field])
         elif isinstance(field, SummedField):
-            return field.__add__(self)
+            field.insert(0, self)
+            return field
 
 
 class SummedField(list):
+    """Class SummedField is a list of Fields over which Field interpolation
+    is summed. This can e.g. be used when combining multiple flow fields,
+    where the total flow is the sum of all the individual flows.
+    Note that the individual Fields can be on different Grids.
+    Also note that, since SummedFields are lists, the individual Fields can
+    still be queried through their list index (e.g. SummedField[1]).
+    """
     def eval(self, time, x, y, z, applyConversion=True):
         tmp = 0
         for fld in self:
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -978,7 +978,10 @@ class Field(object):
         return data
 
     def __add__(self, field):
-        return SummedField([self, field])
+        if isinstance(self, Field) and isinstance(field, Field):
+            return SummedField([self, field])
+        elif isinstance(field, SummedField):
+            return field.__add__(self)
 
 
 class SummedField(list):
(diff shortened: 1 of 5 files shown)


parcels/field.py
```diff
@@ -976,6 +977,34 @@ def computeTimeChunk(self, data, tindex):
...
         return data
 
+    def __add__(self, field):
+        if isinstance(self, Field) and isinstance(field, Field):
+            return SummedField([self, field])
+        elif isinstance(field, SummedField):
+            return field.__add__(self)
+
+
+class SummedField(list):
```
delandmeterp: Could you add some doc here?
Especially advert that even now that we have sums, we still keep the info on the different sum terms 
(that we can say SummedField[4])

parcels/field.py
```diff
@@ -976,6 +977,34 @@ def computeTimeChunk(self, data, tindex):
 
         return data
 
+    def __add__(self, field):
+        if isinstance(self, Field) and isinstance(field, Field):
+            return SummedField([self, field])
+        elif isinstance(field, SummedField):
+            return field.__add__(self)
```
delandmeterp: Here you said that we do support
Field + SummedField
But this will simply compute
SummedField + Field
right? Should we put a warning? Since the indexing is not really used by the users? (maybe?)
Or only support SummedField + Field

tests/test_fieldset_sampling.py
```diff
@@ -426,27 +426,29 @@ def test_list_of_fields(mode, with_W, k_sample_p, mesh):
...
     assert np.allclose(fieldset.U.eval(0, 0, 0, 0)*conv, 0.3)
     assert np.allclose(fieldset.U[0, 0, 0, 0]*conv, 0.3)
 
     P1 = Field('P1', 30*np.ones((zdim*gf, ydim*gf, xdim*gf), dtype=np.float32), grid=U1.grid)
     P2 = Field('P2', 20*np.ones((zdim, ydim, xdim), dtype=np.float32), grid=U2.grid)
-    fieldset.add_field([P1, P2], name='P')
-    assert np.allclose(fieldset.P[0, 0, 0, 0], 50)
+    P3 = Field('P3', 10*np.ones((zdim, ydim, xdim), dtype=np.float32), grid=U2.grid)
+    P4 = Field('P4', 0*np.ones((zdim, ydim, xdim), dtype=np.float32), grid=U2.grid)
+    fieldset.add_field((P1+P4)+(P2+P3), name='P')
```
delandmeterp: So adding fields is neither commutative (was already the case) nor associative, right?

delandmeterp: Cool, I see it is now associative since the change on Field.__add__
By 'not commutative', I was only referring that FA + FB creates [FA, FB] and FB + FA creates [FB, FA]
But of course the result of the interpolation is the same at the end.

Issue #436, opened 2018-09-10
Combining different fields and indices fails


pdnooteboom:
If one combines 3D and 2D fields and makes use of indices, only the surface velocity currents are used to advect particles.  A quick fix is to add a line indices = indices.copy() in field.py:

diff --git a/parcels/field.py b/parcels/field.py
index 163c325..bb67c86 100644
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -273,19 +273,32 @@ class Field(object):
                full_load is however sometimes necessary for plotting the fields.
         """
 
+        indices = indices.copy()
         if not isinstance(filenames, Iterable) or isinstance(filenames, str):
             filenames = [filenames]
         dimension_filename = dimension_filename if dimension_filename else filenames[0]
         if indices is None:
             indices = {}
         with NetcdfFileBuffer(dimension_filename, dimensions, indices) as filebuffer:

erikvansebille:
See also PR #437

erikvansebille:
Wouldn't another solution to this Issue be to change line 191 in fieldset.py:
```
inds = indices[var] if (indices and var in indices) else indices
```
The problem here is that if `indices` is a dictionary but `var` is not in that dictionary, that `inds` then becomes the whole dictionary. That doesn't seem right. 

Should it be something like
```
if type(indices) is dict:
    inds = indices[var].copy() if var in indices else None
else:
    inds = indices.copy()
```

delandmeterp:
I don't understand.
A typical indices is a dictionary:
`indices = {'lon': indslon, 'lat': indslat}`

What line 191 does it to process indices like:
`indices = {'U':{'lon': indslonU, 'lat': indslatU}, 'V':{'lon':indslonV, 'lat':indslatV} }`

So indices is always a dictionary, no? `var` in line 191 is not 'lon', or 'lat' (the dimensions) but the fields names. So this proposed solution couldn't work if I'm not mistaken

erikvansebille:
Fixed in #437

PR #437, opened 2018-09-10
field.from_netcdf: indices argument in read-only

delandmeterp:
This PR solves #436


parcels/field.py
```diff
@@ -279,6 +279,8 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         dimension_filename = dimension_filename if dimension_filename else filenames[0]
         if indices is None:
             indices = {}
+        else:
+            indices = indices.copy()
```
erikvansebille: Or as a one-liner?
```
indices = {} if indices is None else indices.copy()
```

Or is that too confusing? If so, then perhaps also add some comments here?

PR #440, opened 2018-09-11
Lonlat minmax values

delandmeterp:
To review this PR, it is maybe better to read separately the two commits:
(1) does cleaning
(2) process the lon lat min max

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -52,7 +52,7 @@ class Grid(object):
         self.meridional_halo = 0
         self.lat_flipped = False
         self.defer_load = False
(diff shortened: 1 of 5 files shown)


parcels/grid.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, lon, lat, time, time_origin, mesh):
         self.meridional_halo = 0
         self.lat_flipped = False
         self.defer_load = False
+        self.lonlat_minmax = np.array([np.min(lon), np.max(lon), np.min(lat), np.max(lat)])
```
erikvansebille: use `np.nanmin` and `np.nanmax`? 
(this is related to another issue we have with grids where some lon/lat over land are set to Nan)

Issue #425, opened 2018-08-27
Using from_netcdf getting a TypeError


michaeldenes:
Hi all,
I'm fairly new to this, so it could be just user error. I'm using model output from a mom01v5 model, and trying to use the FieldSet.from_netcdf method to create a fieldset, but I'm getting the error
`TypeError: '>' not supported between instances of 'datetime.timedelta' and 'int'`

My code:
`filenames = "/g/data3/hh5/tmp/cosima/mom01v5/KDS75/output*/ocean__*_*.nc"`
`variables = {'U': 'u',
             'V': 'v'}`
`dimensions = {'lat': 'geolat_c',
              'lon': 'geolon_c',
              'time': 'time',
              'depth': 'st_ocean'}`
`fieldset = FieldSet.from_netcdf(filenames, variables, dimensions)`

The issue appears to be in parcels/field.py line 310
`assert(np.all((time[1:]-time[:-1]) > 0))`
when called from parcels/fieldset.py line 201-203
`fields[var] = Field.from_netcdf(paths, var, dims, inds, grid=grid, mesh=mesh, allow_time_extrapolation=allow_time_extrapolation, time_periodic=time_periodic, full_load=full_load, **kwargs)`

Any help would be great! Thanks.

erikvansebille:
Hi Michael,

Thanks for reporting this Issue; it appears to be a bug yes. For some reason, your `time` is a `datetime`, rather than a float, as `Field.from_netcdf` is always supposed to return a float. Not really sure how that could happen..

Would you be able to share one of your output files with me? Then it will be easier for me to debug and provide a fix

michaeldenes:
Hi Erik, thanks for getting back so quickly. Unfortunately the files are rather large (~5.6GB each). Are there any diagnostics I could run which would help?

erikvansebille:
Can't you share one file over e.g. wetransfer with me? I'm at e.vansebille@uu.nl. That really is easier (and faster!) than trying to debug over this Issue tracker

erikvansebille:
Whatever you can send is useful. But make sure it has the correct
dimensions (including `time`) as the problem seems to be in reading that
dimension

erikvansebille:
Hi Michael,

The good news is that I downloaded your files and could reproduce the bug. I also know now that the issue is with the `time:calendar_type = "NOLEAP"` attribute in your netcdf file.

[...]

Issue #220, opened 2017-09-07
Model data starting at year 0000-01-01


guidov:
days since 0000-01-01 00:00:00' ValueError: year is out of range

I ran into this problem with ocean model data starting at year 0000 with a standard noleap calendar.

Don't have a good work around for this right now. 

See description of the issue here: https://github.com/Unidata/netcdf4-python/issues/442

erikvansebille:
This looks much like a bug in the unidata netcdf python library, and might be resolved when we move to dask (Issue #216)?

In the meantime, a work-around could be to read in the data using the netcdf Dataset directly, and then use `FieldSet.from_data` (instead of the shortcut `FieldSet.from_netcdf` method)

erikvansebille:
This is now explicitly dealt with through PR #353; Parcels indeed only supports `time_origins` between 1678 and 2262, and throws an error if time_origin is outside. Closing this issue and #220 for now, until we hear about a actual use-case for extending this date range (which would be a lot of work!)

PR #456, opened 2018-09-19
Time_origin of calendars as TimeConverter class

erikvansebille:
This PR introduces the `TimeConverter()` class, that converts between dates and the number of seconds that Parcels uses under the hood. For this, we require a `time_origin`, but after Issue #425, we realised that it is not easy to implement a new calendar (such as `NOLEAP`). This PR therefore should make it easier to support many different date formats and calendars (and thereby might also aid support of e.g. Issue #220)


parcels/particleset.py
```diff
@@ -285,9 +285,9 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
         if isinstance(endtime, datetime):
             endtime = np.datetime64(endtime)
         if isinstance(endtime, np.datetime64):
-            if not self.time_origin:
+            if self.time_origin.calander is None:
```
delandmeterp: `calander` -> `calendar`

Issue #453, opened 2018-09-17
fieldset time dimension


dencan06:
Hello
I'm just a new user trying to read my model outputs from a netcdf file with following:
variables={'U': 'UVEL', 'V' :'VVEL', 'W':'WVEL'}
dimensions['U']={'time':'T','depth':'diag_levels','lat':'Y','lon':'Xp1'}  
dimensions['V']={'time':'T','depth':'diag_levels','lat':'Yp1','lon':'X'} 
dimensions['W']={'time':'T','depth':'diag_levels','lat':'Y','lon':'X'} 
fset = FieldSet.from_netcdf(filenames, variables, dimensions,full_load=True)

however I'm getting this error:
field.pyc in from_netcdf(cls, filenames, variable, dimensions, indices, grid, mesh, allow_time_extrapolation, time_periodic, full_load, dimension_filename, **kwargs)
    221                             data[ti:ti+len(tslice), 0, :, :] = filebuffer.data[:, :, :]
    222                     else:
--> 223                         data[ti:ti+len(tslice), :, :, :] = filebuffer.data[:, :, :, :]
    224                 ti += len(tslice)
    225         else:

ValueError: could not broadcast input array from shape (583,50,200,176) into shape (1,50,200,176)

T is the unlimited dimension so it's a masked array. could this be causing the problem?

erikvansebille:
I indeed expect that the problem is related to your time dimension. Could you copy the header of your netcdf file here, so that we can have a look? 

By the way, is there a reason that you use `full_load=True`? Do you also get an error when you use `full_load=False`?

dencan09:
Actually I tried first without full_load and then with, and also with specified indices but i got the same error. 
"
<type 'netCDF4._netCDF4.Dataset'>
root group (NETCDF3_CLASSIC data model, file format NETCDF3):
    MITgcm_version: checkpoint66g
    ...
    MITgcm_mnc_ver: 0.9
    sNx: 22
    sNy: 25
    OLx: 5
    OLy: 5
    nSx: 1
    nSy: 1
    nPx: 8
    nPy: 8
    Nx: 176
    Ny: 200
    Nr: 50
    dimensions(sizes): T(583), Zmd000050(50), Xp1(177), Y(200), X(176), Yp1(201), Zld000050(50)
[...]

PR #460, opened 2018-09-25
Allowing for different netcdf engine in Field.from_netcdf

erikvansebille:
Some files (e.g. the MITgcm files of Issue #453) lead to errors when openend by Field.from_netcdf() because the default netcdf4 engine of `xarray` can't read them.

For these files, this PR now provides the option to specify `netcdf_engine='scipy'` in the Field.from_netcdf and FieldSet.from_netcdf constructors

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1117,12 +1117,12 @@ class NetcdfFileBuffer(object):
 
     def __enter__(self):
         try:
-            self.dataset = xr.open_dataset(str(self.filename), decode_cf=True)
+            self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine='scipy')
             self.dataset['decoded'] = True
         except:
             logger.warning_once("File %s could not be decoded properly by xarray (version %s).\n         It will be opened with no decoding. Filling values might be wrongly parsed."
                                 % (self.filename, xr.__version__))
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -50,6 +50,8 @@ class Field(object):
            (i.e. beyond the last available time snapshot)
     :param time_periodic: boolean whether to loop periodically over the time component of the Field
            This flag overrides the allow_time_interpolation and sets it to False
+    :param netcdf_engine: engine to use for netcdf reading in xarray. Default is 'netcdf',
+           but in cases where this doesn't work, setting netcdf_engine='scipy' could help
```
delandmeterp: Why is this doc here in `Field` constructor? Shouldn't it be in `Field.from_netcdf()` and same for `FieldSet.from_netcdf()`?

parcels/field.py
```diff
@@ -198,6 +202,7 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                     grid = CurvilinearSGrid(lon, lat, depth, time, time_origin=time_origin, mesh=mesh)
             grid.timeslices = timeslices
             kwargs['dataFiles'] = dataFiles
+            kwargs['netcdf_engine'] = netcdf_engine
```
delandmeterp: why is this in `if grid is None` and not later with other kwargs (at the end of the function)

PR #463, opened 2018-10-02
Nemo 3d

delandmeterp:


parcels/field.py
```diff
@@ -156,12 +156,15 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
...
             lon, lat = filebuffer.read_lonlat
-            depth = filebuffer.read_depth
             indices = filebuffer.indices
             # Check if parcels_mesh has been explicitly set in file
             if 'parcels_mesh' in filebuffer.dataset.attrs:
                 mesh = filebuffer.dataset.attrs['parcels_mesh']
 
+        dimension_filename = filenames[0]
+        with NetcdfFileBuffer(dimension_filename, dimensions, indices) as filebuffer:
+            depth = filebuffer.read_depth
```
erikvansebille: This isn't a very pretty solution, I'd say. This only works for nemo, but may leave users of other datasets out. Even if they would put `depth` in their `dimension_filename`, that would be overloaded here. Should we implement this in a way that is more flexible for different use cases, not only nemo?

delandmeterp: Why don't we replace the `mesh_mask` file which is not super explicit by a dimesion_file, which can be a file (the mesh_mask) or a dict of files, per dimension.
This follows the logic as everywhere else, no?

erikvansebille: yes, that is a good idea. Although we need to support also that `dimensions` can also be a dictionary of dictionaries, for each variable. So then we could have two dictionaries: `dimension_names` and `dimension_files`. 

The former is what we now have as `dimensions`, and the latter would default to `filenames[0]` (i.e. assume that all dimensions information is simply in the variable files) but can be overloaded as either a single dictionary (`dimension_files['lon']) or a dictionary-of-dictionaries (`dimension_files['U']['lon']). 
Makes sense?

parcels/examples/tutorial_nemo_curvilinear.ipynb
```diff
@@ -52,9 +52,13 @@
...
-    "filenames = {'U': 'NemoCurvilinear_data/U_purely_zonal-ORCA025_grid_U.nc4',\n",
-    "             'V': 'NemoCurvilinear_data/V_purely_zonal-ORCA025_grid_V.nc4',\n",
-    "             'mesh_mask': 'NemoCurvilinear_data/mesh_mask.nc4'}\n",
+    "data_path = 'NemoCurvilinear_data/'\n",
+    "filenames = {'U': {'lon': data_path + 'mesh_mask.nc4',\n",
+    "                   'lat': data_path + 'mesh_mask.nc4',\n",
+    "                   'data': data_path + 'U_purely_zonal-ORCA025_grid_U.nc4'},\n",
+    "             'V': {'lon': data_path + 'mesh_mask.nc4',\n",
+    "                   'lat': data_path + 'mesh_mask.nc4',\n",
+    "                   'data': data_path + 'V_purely_zonal-ORCA025_grid_V.nc4'}}\n",
```
erikvansebille: Does this new API for `filenames` also need to be documented? Perhaps explain here in the nemo tutorial, and also add in the docstring of FieldSet.from_netcdf etc?

parcels/field.py
```diff
@@ -152,25 +152,44 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
 
         if not isinstance(filenames, Iterable) or isinstance(filenames, str):
             filenames = [filenames]
-        dimension_filename = dimension_filename if dimension_filename else filenames[0]
+
+        data_filenames = filenames['data'] if type(filenames) is dict else filenames
+        if type(filenames) == dict:
+            assert len(filenames['lon']) == 1
+            assert filenames['lon'] == filenames['lat']
```
erikvansebille: why can't 'lon' and 'lat' not be in different files? Seems too strict to me? Or, if we need to be this strict, then at least add an error message

parcels/fieldset.py
```diff
@@ -178,15 +178,26 @@ def from_netcdf(cls, filenames, variables, dimensions, indices=None,
...
+            else:
+                for dim, p in paths.items():
+                    if not isinstance(p, list):
+                        paths[dim] = sorted(glob(str(p)))
+                    if len(paths[dim]) == 0:
+                        notfound_paths = filenames[var] if type(filenames) is dict and var in filenames else filenames
+                        raise IOError("FieldSet files not found: %s" % str(notfound_paths))
+                    for fp in paths[dim]:
+                        if not path.exists(fp):
+                            raise IOError("FieldSet file not found: %s" % str(fp))
```
erikvansebille: Here, we have two copies of almost the same code. Create a function instead, for conciseness an ease of maintenance?

parcels/examples/tutorial_nemo_curvilinear.ipynb
```diff
@@ -33,7 +33,17 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can create a `FieldSet` just like we do for normal grids. Note that we need to provide an extra filename for the `mesh_mask` file that holds information on the Curvilinear grid."
+    "We can create a `FieldSet` just like we do for normal grids.\n",
+    "Note that NEMO is discretised on a C-grid. U and V velocities are not located on the same grid (see https://www.nemo-ocean.eu/doc/node19.html).\n",
```
erikvansebille: Is 'grid' the right word here, is 'not located on the same grid'? Perhaps 'location'?

parcels/examples/tutorial_nemo_curvilinear.ipynb
```diff
@@ -33,7 +33,17 @@
...
+    "\n",
+    "```\n",
+    " __V1__\n",
+    "|      |\n",
+    "U0     U1\n",
+    "|__V0__|\n",
+    "```\n",
+    "\n",
+    "To interpolate U, V velocities on the C-grid, Parcels needs to read the f-nodes, which are located on the corners of the cells."
```
erikvansebille: change to 'lower-left corners' to be more explicit which corner?

parcels/fieldset.py
```diff
@@ -160,6 +160,9 @@ def from_netcdf(cls, filenames, variables, dimensions, indices=None,
         :param filenames: Dictionary mapping variables to file(s). The
                filepath may contain wildcards to indicate multiple files,
                or be a list of file.
+               filenames can be a list [files], a dictionary {var:[files]},
+               a dictionary {dim:[files]} (if lon, lat, depth not stored in same files as data),
+               a dictionary of dictionaries {var:{dim:[files]}}
```
erikvansebille: add 'or' before the dictionary-of-dictionaries?

parcels/fieldset.py
```diff
@@ -232,19 +235,27 @@ def from_nemo(cls, filenames, variables, dimensions, indices=None, mesh='spheric
...
         """Initialises FieldSet object from NetCDF files of Curvilinear NEMO fields.
-        Note that this assumes there is a variable mesh_mask that is used for the dimensions
 
         :param filenames: Dictionary mapping variables to file(s). The
                filepath may contain wildcards to indicate multiple files,
-               or be a list of file. At least a 'mesh_mask' needs to be present
+               or be a list of file.
+               filenames can be a list [files], a dictionary {var:[files]},
+               a dictionary {dim:[files]} (if lon, lat, depth not stored in same files as data),
+               a dictionary of dictionaries {var:{dim:[files]}}
```
erikvansebille: see comment above

parcels/fieldset.py
```diff
@@ -232,19 +235,27 @@ def from_nemo(cls, filenames, variables, dimensions, indices=None, mesh='spheric
...
-               (e.g. dimensions['U'], dimensions['V'], etc).
+               dimension names are different for each variable.
+               Watch out: NEMO is discretised on a C-grid: U and V are not located
+               U and V velocities are not located on the same grid (see https://www.nemo-ocean.eu/doc/node19.html).
+                __V1__
+               |      |
+               U0     U1
+               |__V0__|
+               To interpolate U, V velocities on the C-grid, Parcels needs to read the f-nodes,
+               which are located on the corners of the cells.
```
erikvansebille: see minor comments on the jupyter notebook

PR #469, opened 2018-10-10
Nemo north plost

delandmeterp:


parcels/include/index_search.h
```diff
@@ -244,7 +246,8 @@ static inline ErrorCode search_indices_curvilinear(float x, float y, float z, CS
 
   *xsi = *eta = -1;
   int maxIterSearch = 1e6, it = 0;
-  while ( (*xsi < 0) || (*xsi > 1) || (*eta < 0) || (*eta > 1) ){
+  double tol = 1e-12;
```
erikvansebille: what is this value for tolerance based on? Why 1e-12? Have you checked that this actually fixes the lost particles?

delandmeterp: Due to rounding errors, it happens that the particle search oscillates between two cells, leading to errors such as:

`Relative particle position:  (xsi, eta) -8.4619131356012846e-14 6.0000000000008880e-01`

So here, the particle is at the left of the current cell, but if I go to the left cell, it will be located as too much to the right, and the search is stuck between those two cells.

I have observed this error 4 times on the last 5 runs (three particles out of ~500,000), and adding this small tolerance has fixed the 4 problems.

Issue #458, opened 2018-09-19
First time not written in particle file


davidwichmann:
I am running parcels 1.1.0 with nemo and noticed that the particle files have as first time entry 1.5 days after the time of the first velocity file (129600s=1.5 days). Also, I managed to start and execute a run which has as initial particle time 1 day before the first velocity file and it does not crash. 2 days before the first velocity file crashes.  It seems like there is a time shift happening somewhere in parcels of around 1.5 days.

erikvansebille:
This is strange, I can't reproduce your error. From your description, I assume that the code below would work for you?
```
from parcels import FieldSet, ParticleSet, JITParticle, AdvectionRK4
import numpy as np
from glob import glob


def set_nemo_fieldset(ufiles, vfiles):
    filenames = {'U': ufiles, 'V': vfiles, 'mesh_mask': '/Users/erik/Desktop/ORCA0083-N006/domain/mesh_hgr.nc'}
    indices = {'lon': range(0, 4320)}

    variables = {'U': 'uo', 'V': 'vo'}
    dimensions = {'lon': 'glamf', 'lat': 'gphif', 'time': 'time_centered'}

    return FieldSet.from_nemo(filenames, variables, dimensions, indices, vmax=1e10)


ufiles = sorted(glob('/Users/erik/Desktop/ORCA0083-N006/means/ORCA0083-N06_2000*d05U.nc'))
vfiles = sorted(glob('/Users/erik/Desktop/ORCA0083-N006/means/ORCA0083-N06_2000*d05V.nc'))

fieldset = set_nemo_fieldset(ufiles, vfiles)
fieldset.add_periodic_halo(zonal=True)

X, Y = np.meshgrid(np.arange(-175, 176, 5), np.arange(-70, 81, 5))
pset = ParticleSet(fieldset, JITParticle, lon=X, lat=Y, time=fieldset.U.grid.time[0]-86400)
pset.execute(AdvectionRK4, dt=3600)
```
(note the `time=fieldset.U.grid.time[0]-86400` in the `ParticleSet` creation)

Because I get an error `Field sampled outside time domain at time 2000-01-02T12:00:00.000000000`

Can you add a minimalist code that reproduces this bug?

davidwichmann:
Hi Erik,

you are right, I noticed that the times in the names of the nemo files (which I took as time of the velocity file) are not the actual times in the "time_centered" variable - my mistake. So everything is fine!

Thanks!

[...]

PR #470, opened 2018-10-11
Warning when exporting particle when  dt not set

delandmeterp:
See issue #458 

The dt check is only done for pset[0].
This ignores then if some particles have a dt=0 (for example when we add particles later on)

Maybe we should rather check for every particle, and putting a `warning_once` stating "at least one p.dt is not set. Either some or all of the particles will be ignored for exporting at t = %g"


parcels/particlefile.py
```diff
@@ -172,7 +174,7 @@ def write(self, pset, time, sync=True, deleted_only=False):
                     for var in self.user_vars_once:
                         getattr(self, var)[p.fileid] = getattr(p, var)
             else:
-                logger.warning("ParticleSet is empty on writing as array")
+                logger.warning("ParticleSet is empty on writing as array at tiem %g" % time)
```
erikvansebille: type-o: tiem -> time

parcels/particlefile.py
```diff
@@ -158,6 +158,8 @@ def write(self, pset, time, sync=True, deleted_only=False):
 
                 self.idx = np.append(self.idx, np.zeros(len(first_write)))
 
+                if np.isnan(pset[0].dt):
+                    logger.warning("particle dt not defined yet at particle file writing. Writing at time %g will be ignored." % time)
                 for p in pset:
                     if p.dt*p.time <= p.dt*time:  # don't write particles if they haven't started yet
```
erikvansebille: The alternative is to make this statement (`if p.dt*p.time <= p.dt*time`) more precise.

Remember that the goal is just to prevent writing a particle when they haven't started yet (i.e. when `time < p.time` in time-forward mode, and `time > p.time` in time-backward mode.

However, if p.dt is NaN while pfile.write is called, then it means that this happens _before_ the pset.execute. So in that case we can assume that the user wants the whole particleset to be written?

So what about changing to `if p.dt*p.time <= p.dt*time or np.isnan(p.dt)` and removing the first warning?

parcels/particlefile.py
```diff
@@ -158,6 +158,8 @@ def write(self, pset, time, sync=True, deleted_only=False):
 
                 self.idx = np.append(self.idx, np.zeros(len(first_write)))
 
+                if np.isnan(pset[0].dt):
```
erikvansebille: I agree that this should be inside the `for p in pset` call, as it might be that only some particles still have dt set to NaN?

parcels/particlefile.py
delandmeterp: If we do that it means you could have exports like:
p0: t0, t1, t2, t3, t4
p1: t0,            t3, t4

If p1 is release at t3, then p1(t0) and p1(t3) will show same ideas.

But we don't have to forget that we can as well release particles at t = (t2+t3)/2, so in this case we would have initial position at least. 
But this will be only the case with pfile manual write as described in issue #458.
Not having the initial position of particles released between 2 timesteps is and will stay like this.

So, we go for exporting any p for which `p.dt=nan`? 
At the end, I would say it is simply a choice.

parcels/particlefile.py
delandmeterp: Then we shouldn't use warning.
warning_once is not good either since we want the warning each time the export might be not full, right? I'll see if we already have somewhere a warning_once that we can reset after a loop to authorise a next warning again, or do something like this

parcels/particlefile.py
erikvansebille: I agree that this is simply a choice we need to make. Perhaps the design choice should be
- If pfile is written as part of pset.execute, do as before
- If pfile is written by an explicit call by users, then write out all particles in the pset, independent of what their start_time is

Note also that there may be a complication that if `dt` is not set yet, then `time` may also not be set; in which case we would not know which time to write for the particles. I guess that we should throw an error then?

delandmeterp: I agree for first point.
For second point, I would like not to save everything, but yes we can include the nan `dt`.
There is no problem with nan times. Indeed, the `p.time` is only used for the check I wrote earlier.

The time at which to write the particle is the function argument `pfile.write(pset, time)`
And actually, this is strange. We should write the particles at their own time. pfile.write should not have a time argument.
In the way we call the function (`pfile.write(pset, pset[0].time)`), it is redundant, but it might be conflicting.
I think we should clean that part. I'll work on that asap.

erikvansebille: Actually, I think that we do need to have `time` as an argument in `pfile.write()`, particularly for this bit of code
https://github.com/OceanParcels/parcels/blob/39ffc77cfdeb7e664cb2ff044c11c4b815db369a/parcels/particlefile.py#L164
and also for
https://github.com/OceanParcels/parcels/blob/39ffc77cfdeb7e664cb2ff044c11c4b815db369a/parcels/particlefile.py#L150-L151
So I don't think it is redundant, right?

delandmeterp: Ok, but then shouldn't we have rather `self.time[i, self.idx[i]] = p.time` to avoid
that if all particles are at time 10, and you write pfile.write(pset, time=5) you break everything

https://github.com/OceanParcels/parcels/blob/39ffc77cfdeb7e664cb2ff044c11c4b815db369a/parcels/particlefile.py#L166

Or actually, is it wrong to simply test `p.time == time or np.isnan(p.dt)` instead of ` p.dt*p.time <= p.dt*time` ?

parcels/particlefile.py
```diff
@@ -162,17 +163,20 @@ def write(self, pset, time, sync=True, deleted_only=False):
...
                         self.id[i, self.idx[i]] = p.id
-                        self.time[i, self.idx[i]] = time
+                        self.time[i, self.idx[i]] = p.time
                         self.lat[i, self.idx[i]] = p.lat
                         self.lon[i, self.idx[i]] = p.lon
                         self.z[i, self.idx[i]] = p.depth
                         for var in self.user_vars:
                             getattr(self, var)[i, self.idx[i]] = getattr(p, var)
+                        if p.state != ErrorCode.Delete:
+                            assert np.allclose(p.time, time), \
```
erikvansebille: Shouldn't this be a warning? Do we really want Parcels to crash if this happens?

PR #471, opened 2018-10-13
Fixing windows python3 jupyter appveyor error

erikvansebille:
The issue where py3_win always gives an error on Appveyor CI turns out to be related to an upgrade of the `ipython` package. Setting it to version<5.0 fixes the CI again. Note that we need to do this only for the `py3_win` environment (i.e. we are still testing latest `ipython` versions on the other environments)

Of course not ideal to limit a package version this way, but it's the easiest fix to get us back to green ticks for CI. And both windows and python3 are not our core environments, so I suggest we use this for a while and then revisit in a few months to see if updated versions of `ipython` fix the issue


willirath:
> [...] both windows and python3 are not our core environments [...]

Note that the end of life of Python 2 is approaching, though.  Xarray will stop releasing new versions for Python 2 at the end of this year.  [Others will do so next year.](https://python3statement.org/#sections40-timeline)

PR #472, opened 2018-10-15
[WIP] Un-pin matplotlib

willirath:
This accounts for https://github.com/conda-forge/parcels-feedstock/issues/29 and https://github.com/conda-forge/parcels-feedstock/pull/30

I'm testing this here before merging the PR in conda-forge.


willirath:
Only problem seems to be in `ax.scatter(...)` of the Argo example: Matplotlib has problems handling the masked arrays, takes a different road to still produce a plot and then fails at dealing with the singleton dims of `x`, `y`, and `z`.  66b3a89 should fix this.

willirath:
I think parcels also pins ffmpeg. See my comment on the conda-forge feedstock about maybe removing the matplotlib and the ffmpeg dependency altogether. (But note that this should go together with an update of the installation instructions on the website that then would need to point out that the users need to install matplotlib and ffmpeg manually (like they do with cartopy already).)


Am 15. Oktober 2018 17:02:50 schrieb Philippe Delandmeter <notifications@github.com>:

PR #474, opened 2018-10-16
Mom model fixes

erikvansebille:
Fixing two bugs in running on MOM6 with NOLEAP calendar (see also #456).
1) Since the curvilinear grid in MOM starts quite far northward, the test 
```if np.allclose(lon_subset[0, :], lon_subset[int(xdim/2), :]) and np.allclose(lat_subset[:, 0], lat_subset[:, int(ydim/2)])```
did not catch the grid as curvilinear. So changed this to 
```if np.allclose(lon_subset[0, :], lon_subset[xdim-1, :]) and np.allclose(lat_subset[:, 0], lat_subset[:, ydim-1])```
2) If ParticleFile is written as `NOLEAP` calendar, then creating an animation was not possible.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1190,7 +1190,7 @@ class NetcdfFileBuffer(object):
         elif len(lon.shape) == 4:  # some lon, lat have a time and depth dimension 1
             lon_subset = np.array(lon[0, 0, self.indices['lat'], self.indices['lon']])
             lat_subset = np.array(lat[0, 0, self.indices['lat'], self.indices['lon']])
-         if len(lon.shape) > 1:  # Tests if lon, lat are rectilinear but were stored in arrays
+        if len(lon.shape) > 1:  # Tests if lon, lat are rectilinear but were stored in arrays
             rectilinear = True
             # test if all columns and rows are the same for lon and lat (in which case grid is rectilinear)
             for xi in range(1, lon_subset.shape[0]):
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1191,9 +1191,18 @@ class NetcdfFileBuffer(object):
             lon_subset = np.array(lon[0, 0, self.indices['lat'], self.indices['lon']])
             lat_subset = np.array(lat[0, 0, self.indices['lat'], self.indices['lon']])
         if len(lon.shape) > 1:  # if lon, lat are rectilinear but were stored in arrays
-            xdim = lon_subset.shape[0]
-            ydim = lat_subset.shape[1]
-            if np.allclose(lon_subset[0, :], lon_subset[xdim-1, :]) and np.allclose(lat_subset[:, 0], lat_subset[:, ydim-1]):
+            rectilinear = True
+            # test if all columns and rows are the same for lon and lat (in which case grid is rectilinear)
+            for x in range(1, lon_subset.shape[0]):
+                if not np.allclose(lon_subset[0, :], lon_subset[x, :]):
+                    rectilinear = False
+                    break
+            if rectilinear:
+                for y in range(1, lat_subset.shape[1]):
+                    if not np.allclose(lat_subset[:, 0], lat_subset[:, y]):
+                        rectilinear = False
+                        break
+            if rectilinear:
                 lon_subset = lon_subset[0, :]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: parcels/field.py [hunk 1]
```diff
@@ -1193,7 +1193,7 @@ def read_lonlat(self):
         if len(lon.shape) > 1:  # if lon, lat are rectilinear but were stored in arrays
             xdim = lon_subset.shape[0]
             ydim = lat_subset.shape[1]
-            if np.allclose(lon_subset[0, :], lon_subset[int(xdim/2), :]) and np.allclose(lat_subset[:, 0], lat_subset[:, int(ydim/2)]):
+            if np.allclose(lon_subset[0, :], lon_subset[xdim-1, :]) and np.allclose(lat_subset[:, 0], lat_subset[:, ydim-1]):
```
delandmeterp: The point of this check was to quickly assess if a mesh was curvilinear.
If that test is not sufficient, we should maybe have a more complete one checking all the rows and lines, or one out of XX rows and lines, but checking 1 and last can be problematic, especially if a mesh is periodic.

erikvansebille: OK, well maybe we need to check all rows and columns then, because the mi-way rows and columns failed for MOM.
But perhaps we should even aks ourselves why we need this test. Even if a grid is RectiLinear, treating it as Curvilinear here would not slow down the code by much, would it? Shouldn't we just assume that any 2d array of Longitudes or latitudes always signals a CurviLinear Grid?

delandmeterp: I don't really know here about timing. We could quantify that.

erikvansebille: I've now written a code that checks for all lons and lats, but does it quite efficient to stop as soon at it finds the first lon that is different from the first column. Since we don't expect to see 2D Rectilinear Grids often (in which case we have to search all lons and lats), I think this is a good solution. Agree?

parcels/field.py
```diff
@@ -1191,9 +1191,18 @@ def read_lonlat(self):
             lon_subset = np.array(lon[0, 0, self.indices['lat'], self.indices['lon']])
             lat_subset = np.array(lat[0, 0, self.indices['lat'], self.indices['lon']])
         if len(lon.shape) > 1:  # if lon, lat are rectilinear but were stored in arrays
```
delandmeterp:
```suggestion
         if len(lon.shape) > 1:  # Tests if lon, lat are rectilinear but were stored in arrays
```

parcels/field.py
```diff
@@ -1191,9 +1191,18 @@ def read_lonlat(self):
...
-            xdim = lon_subset.shape[0]
-            ydim = lat_subset.shape[1]
-            if np.allclose(lon_subset[0, :], lon_subset[int(xdim/2), :]) and np.allclose(lat_subset[:, 0], lat_subset[:, int(ydim/2)]):
+            rectilinear = True
+            # test if all columns and rows are the same for lon and lat (in which case grid is rectilinear)
+            for x in range(1, lon_subset.shape[0]):
```
delandmeterp:
```suggestion
            for xi in range(1, lon_subset.shape[0]):
```

parcels/field.py
```diff
@@ -1191,9 +1191,18 @@ def read_lonlat(self):
...
+                if not np.allclose(lon_subset[0, :], lon_subset[x, :]):
```
delandmeterp:
```suggestion
                if not np.allclose(lon_subset[0, :], lon_subset[xi, :]):
```

parcels/field.py
```diff
@@ -1191,9 +1191,18 @@ def read_lonlat(self):
...
+                    rectilinear = False
+                    break
+            if rectilinear:
+                for y in range(1, lat_subset.shape[1]):
```
delandmeterp:
```suggestion
                for yi in range(1, lat_subset.shape[1]):
```

parcels/field.py
```diff
@@ -1191,9 +1191,18 @@ def read_lonlat(self):
...
+                    if not np.allclose(lat_subset[:, 0], lat_subset[:, y]):
```
delandmeterp:
```suggestion
                    if not np.allclose(lat_subset[:, 0], lat_subset[:, yi]):
```

PR #476, opened 2018-10-17
Implementing Fieldset.from_xarray_dataset() method for integration with xarray Datasets

### LINKED ISSUE #467: Implement FieldSet.from_xarray_dataset()
Following the suggestion by @rabernat in #462, it would be good to implement a simple way to convert xarray `Datasets` to Parcels `FieldSets`.

erikvansebille:
This PR implements a new method, `FieldSet.from_ds()` to run Parcels directly on xarray `Dataset` objects. Following the suggestion by @rabernat, this fixes #467 

An example script would be
```python
from parcels import FieldSet, AdvectionRK4, ParticleSet, JITParticle
from parcels import plotTrajectoriesFile
import numpy as np
import xarray as xr
import gcsfs


def periodicBC(particle, fieldset, time, dt):
    if particle.lon < 0:
        particle.lon += 360
    elif particle.lon > 360:
        particle.lon -= 360


# create xarray Dataset of AVISO geostrophic velocities using gcsfs
fname = "pangeo-data/dataset-duacs-rep-global-merged-allsat-phy-l4-v3-alt"
gcsmap = gcsfs.mapping.GCSMap(fname)
ds = xr.open_zarr(gcsmap)

# convert Dataset to Parcels FieldSet
variables = {'U': 'ugos', 'V': 'vgos'}
dimensions = {'lon': 'longitude', 'lat': 'latitude', 'time': 'time'}
fset = FieldSet.from_xarray_dataset(ds, variables, dimensions)
fset.add_periodic_halo(zonal=True, meridional=False, halosize=2)

# Create ParticleSet and output file
lons, lats = np.meshgrid(np.arange(60, 80, 2), np.arange(-50, -30, 2))
pset = ParticleSet(fset, JITParticle, lon=lons, lat=lats)
ofile = pset.ParticleFile('aviso_particles.nc', outputdt=delta(days=2))

# Advect ParticleSet with RK4 and periodic Boundary conditions
pset.execute(AdvectionRK4+pset.Kernel(periodicBC), runtime=delta(days=30),
             dt=delta(hours=1), output_file=ofile)

plotTrajectoriesFile('aviso_particles.nc')
```

The key statement here is `fset = FieldSet.from_xarray_dataset(ds, variables, dimensions)`. Note that users will still need to define what the `U` and `V` Fields are named in the Dataset, and that also the dimensions need to be specified.


environment_py2_linux.yml
```diff
@@ -23,6 +23,7 @@ dependencies:
   - scipy>=0.16.0
   - six >=1.10.0
   - xarray>=0.10.8
+  - dask
```
delandmeterp: Why is dask being in the requirements (and only in linux) but not imported in the code?

delandmeterp: Ok, it is because of xr.mf_dataset(), right? But why only in linux?

parcels/field.py
```diff
@@ -864,19 +880,20 @@ def advancetime(self, field_new, advanceForward):
...
+                filebuffer.name = filebuffer.parse_name(self.dimensions, self.name)
+            if len(filebuffer.data.shape) == 2:
+                data[tindex, 0, :, :] = filebuffer.data
+            elif len(filebuffer.data.shape) == 3:
                 if g.zdim > 1:
-                    data[tindex, :, :, :] = filebuffer.data[:, :, :]
+                    data[tindex, :, :, :] = filebuffer.data
                 else:
-                    data[tindex, 0, :, :] = filebuffer.data[ti, :, :]
+                    data[tindex, 0, :, :] = filebuffer.data
```
delandmeterp: We can do that?

environment_py2_linux.yml
erikvansebille: To be honest, I don't understand it either. I was hoping perhaps @rabernat would know more. I added it because otherwise I would get an error in xarray/backends/api.py (https://travis-ci.org/OceanParcels/parcels/jobs/441398583#L1805); but only in python3. Since Windows wasn't working anyway (#471), I didn't add it there. We could though, if we understand better

parcels/fieldset.py
```diff
@@ -331,6 +331,46 @@ def from_parcels(cls, basename, uvar='vozocrtx', vvar='vomecrty', indices=None,
                                dimensions=dimensions, allow_time_extrapolation=allow_time_extrapolation,
                                time_periodic=time_periodic, full_load=full_load, **kwargs)
 
+    @classmethod
+    def from_ds(cls, ds, variables, dimensions, indices=None, mesh='spherical', allow_time_extrapolation=None,
+                time_periodic=False, full_load=False, **kwargs):
```
rabernat: I would probably prefer the name `from_xarray_dataset`, which is more explicit.

parcels/fieldset.py
```diff
@@ -331,6 +331,46 @@ def from_parcels(cls, basename, uvar='vozocrtx', vvar='vomecrty', indices=None,
...
+        :param ds: xarray Dataset
+        :param dimensions: Dictionary mapping data dimensions (lon,
+               lat, depth, time, data) to dimensions in the xarray Dataset.
+               Note that dimensions can also be a dictionary of dictionaries if
+               dimension names are different for each variable
+               (e.g. dimensions['U'], dimensions['V'], etc).
+        :param indices: Optional dictionary of indices for each dimension
+               to read from file(s), to allow for reading of subset of data.
+               Default is to read the full extent of each dimension.
+        :param extra_fields: Extra fields to read beyond U and V
```
rabernat: the `mesh` keyword does not appear to be documented in the docstring

tests/test_fieldset.py
```diff
@@ -365,3 +366,31 @@ def DoNothing(particle, fieldset, time, dt):
...
+    ds = generate_dataset(3, 3, 2, 10)
+    variables = {'U': 'Uxr', 'V': 'Vxr'}
+    dimensions = {'lat': 'lat', 'lon': 'lon', 'depth': 'depth', 'time': 'time'}
+    indices = {'lat': range(0, maxlatind)}
+    fieldset = FieldSet.from_ds(ds, variables, dimensions, indices, mesh='flat')
+
+    pset = ParticleSet(fieldset, JITParticle, 0, 0)
+
+    pset.execute(AdvectionRK4, dt=1)
+    assert pset[0].lon == 4.5 and pset[0].lat == 10
```
rabernat: This is a good test. It makes it quite clear how to use an xarray dataset with parcels.

One way to address the comment about laziness would be to create a special dask-backed array that raises an error when it is computed.

parcels/examples/example_globcurrent.py
```diff
@@ -4,64 +4,68 @@
...
 import numpy as np
 import pytest
+import xarray as xr
 
 
 ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
 
 
-def set_globcurrent_fieldset(filename=None, indices=None, full_load=False):
+def set_globcurrent_fieldset(filename=None, indices=None, full_load=False, from_ds=False):
```
rabernat: I believe an explicit `use_xarray` would probably be better thatn `from_ds` here.

environment_py2_linux.yml
rabernat: Dask is not be *required* for xarray. It is, however, required for `open_mfdataset`.

erikvansebille:
> The key question from my point of view, which I can't completely figure out from reading your code, is when exactly xarray's various forms of lazy array-like things (including dask arrays as well as the other array wrappers used by xarray) are coerced to actual numpy arrays. How does this work? How does parcels decide which data along the space and time axes to load?

The loading happens in the `Field.computeTimeChunk()` method (https://github.com/OceanParcels/parcels/blob/fieldset_from_ds_xarray/parcels/field.py#L880) and the `NetcdfFileBuffer.data` property (https://github.com/OceanParcels/parcels/blob/fieldset_from_ds_xarray/parcels/field.py#L1243). 
Note that we only load in the snapshots that are needed, and that if users give an `indices` dictionary with subsets of lon,lat,depth indices, we will also only read in these.
I'm not sure though whether we need any extra `.load()` functions etc for dask. I'm completely new to all that.s 

> Have you actually tried the code in your comment above? If not, would you like me to give this a spin on some other xarray datasets?

Yes, I have tested the code on my own notebooks here, but it's quite slow, probably mostly due to the latency of the connection to your server. Could you try also running it on a server closer to where the data is stored, and report how fast it goes?

tests/test_fieldset.py
erikvansebille: Thanks! So how would we create a "special dask-backed array that raises an error when it is computed"? Do you have a suggestion for that?

environment_py2_linux.yml
erikvansebille: OK, I understand then. We use `open_mfdataset` in `example_globcurrent.py`, which we only test on linux (not on mac). We do test it on windows though, where apparently we don't need to explicitly add `dask` as a requirement?

rabernat:
Sorry to pester with questions, but I'm trying to understand the design of parcels better in order to understand how xarray / dask will interact with it. A key routine seems to be `computeTimeChunk`

https://github.com/OceanParcels/parcels/blob/1020717d363426c49a2425d445b298d65f910adb/parcels/field.py#L880-L898

I'm confused because this function takes `data` as an input, modifies it in place, and then returns it. If you're going to modify it in place, why return anything at all?

When you call this function, e.g.

https://github.com/OceanParcels/parcels/blob/119f6df8e1353ba88be677d57f4487e366e85518/parcels/fieldset.py#L459

I'm pretty sure it would work exactly the same if you just did without any return at all
```
f.computeTimeChunk(data, f.loaded_time_indices[0])
```

For example
```python
import numpy as np
def modify_data(data_in):
    data_in[0] = 1
data = np.zeros(2)
modify_data(data)
np.testing.assert_array_equal(data, [1, 0])
```

How often is `computeTimeChunk` called? Once per timestep for the whole set of particles? Or once per timestep per particle?

parcels/field.py
```diff
@@ -864,19 +880,20 @@ def advancetime(self, field_new, advanceForward):
     def computeTimeChunk(self, data, tindex):
         g = self.grid
         with NetcdfFileBuffer(self.dataFiles[g.ti+tindex], self.dimensions, self.indices, self.netcdf_engine) as filebuffer:
```
rabernat: what is `self.dataFiles` in the case of an xarray dataset? What happens if there are no data files, like in your test below?

erikvansebille: These `self.dataFiles`refer to the Dataset (`ds`) itself. `Filebuffer.dataFiles` is set in field.py 
https://github.com/OceanParcels/parcels/blob/1020717d363426c49a2425d445b298d65f910adb/parcels/field.py#L208-L211

And `dataset` is set at the top of the file as 
https://github.com/OceanParcels/parcels/blob/1020717d363426c49a2425d445b298d65f910adb/parcels/field.py#L158-L162

This may not be the cleanest way to do it (and the variable names may be confusing, as they originally referred to actual NetCDF files), but this was a simple way to implement xarray handling too.

erikvansebille:
> I'm confused because this function takes `data` as an input, modifies it in place, and then returns it. If you're going to modify it in place, why return anything at all?

You're right, I just tested and it also works without the `return data` at the end of `Field.computeTimeChunk()`. An oversight, I guess. I'll remove for clarity

> How often is `computeTimeChunk` called? Once per timestep for the whole set of particles? Or once per timestep per particle?

It's called once per timestep for the whole set of particles. The particle loop itself runs in C (the JIT compilation), and we exit that to e.g. call computeTimeChunk, but also to write output etc. 
https://github.com/OceanParcels/parcels/blob/1020717d363426c49a2425d445b298d65f910adb/parcels/particleset.py#L384

environment_py2_linux.yml
delandmeterp: If it is the case, maybe dask is automatically in xarray dependencies on Windows?
But even if we don't test it on mac, we want the example to run there, so we should put dask in the mac yml files as well, no?

delandmeterp:
I've checked the NEMO runs who pass fine as well.

PR #478, opened 2018-10-18
Adding support for depth_level argument in Field.show()

### LINKED ISSUE #475: Field.show() does not work for Fields with multiple depths
`Field.show()` throws an error for Fields with more than one depth level. Should be a argument, with default to zero?

erikvansebille:
This fixes #475

### DIFF
--- a/parcels/plotting.py
+++ b/parcels/plotting.py
@@ -93,12 +93,13 @@ def plotparticles(particles, with_particles=True, show_time=None, field=None, do
         plt.close()
 
 
-def plotfield(field, show_time=None, domain=None, projection=None, land=True,
+def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None, land=True,
               vmin=None, vmax=None, savefile=None, **kwargs):
     """Function to plot a Parcels Field
 
     :param show_time: Time at which to show the Field
     :param domain: Four-vector (latN, latS, lonE, lonW) defining domain to show
+    :param depth_level: depth level to be plotted (default 0)
     :param projection: type of cartopy projection to use (default PlateCarree)
     :param land: Boolean whether to show land. This is ignored for flat meshes
     :param vmin: minimum colour scale (only in single-plot mode)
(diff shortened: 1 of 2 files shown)


parcels/plotting.py
```diff
@@ -186,10 +193,14 @@ def plotfield(field, show_time=None, domain=None, projection=None, land=True,
 
     timestr = parsetimestr(field[0].grid.time_origin, show_time)
     titlestr = kwargs.pop('titlestr', '')
+    if fld.grid.zdim > 1:
+        depthstr = ' at depth %f ' % fld.grid.depth[depth_level]
```
delandmeterp: This works for z-levels.
For s-levels, maybe you could put a warning or something that says that you show a constant level, which does not correspond to a constant depth.

parcels/plotting.py
```diff
@@ -193,8 +193,9 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None
 
     timestr = parsetimestr(field[0].grid.time_origin, show_time)
     titlestr = kwargs.pop('titlestr', '')
-    if fld.grid.zdim > 1:
-        depthstr = ' at depth %f ' % fld.grid.depth[depth_level]
+    if field[0].grid.zdim > 1:
+        gphrase = 'depth' if field[0].grid.gtype in [GridCode.CurvilinearZGrid, GridCode.RectilinearZGrid] else 'level'
+        depthstr = ' at %s %f ' % (gphrase, field[0].grid.depth[depth_level])
```
delandmeterp: Ok with `gphrase` but `depthstr` shouldn't be like following?

```
depth_or_level = field[0].grid.depth[depth_level] if field[0].grid.gtype in [GridCode.CurvilinearZGrid, GridCode.RectilinearZGrid] else depth_level
depthstr = ' at %s %g ' % (gphrase, depth_or_level) 
```

erikvansebille: No I don't think so. Say that the model is density-coordinate, then the value of `field[0].grid.depth[depth_level]` is still a relevant parameter (e.g. 1027.3 kg/m3 or something like that). That's why in both cases I would print the value (in physical units) of the depth level

delandmeterp: I'm not sure to understand.
Yes a density level is interesting, but in s-grid, field[0].grid.depth is an array with shape len 3 or 4. So printing `field[0].grid.depth[depth_level]` does not work (This is not a scalar, but an array).

But indeed, plotting a depth level is also meaningful in s-grids.

PR #479, opened 2018-10-19
Better support of parallel runs when spawning multiple .executes

erikvansebille:
Until parallel execution is properly implemented in Parcels (see #17), one way that users can run multiple processes on a core is to simply spawn a number of `.execute()` statements, for example in a bash script.

However, the creation of the temporary directory in compiler.py can then go wrong, as all `.execute()` will create the same directory. While before this PR we used to check whether this temporary directory already existed, the slight lag time it takes to call `makedirs(directory)` sometimes meant that the directory was created while another process was creating it too. This caused an error.

To fix this, this PR uses the `exist_ok` argument to directory creation. Since that isn't default on python2, we use a solution suggested at
https://stackoverflow.com/questions/45283093/how-to-walk-around-exist-ok-missing-on-python-2-7

### DIFF
--- a/parcels/compiler.py
+++ b/parcels/compiler.py
@@ -1,7 +1,11 @@
 import subprocess
-from os import path, getenv, makedirs
+from os import path, getenv
 from tempfile import gettempdir
 from struct import calcsize
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path  # python 2 backport
 try:
     from os import getuid
 except:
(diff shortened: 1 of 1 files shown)


erikvansebille:
> And we won't have the problem of writing the same file just afterwards?

The writing issue can relatively easily be circumvented by making sure that each .execute() process writes to its own file (with a different filename. 
When we support 'proper' parallelisation, then I think we need to look into the writing more closely. But for now I don't think it's a priority

PR #480, opened 2018-10-19
C code cleanup

delandmeterp:
@pdnooteboom , could you check that this solves your bug with nemo data?


pdnooteboom:
Nice job @delandmeterp! This fixes both the NotImplementedError and the assertion error.

PR #482, opened 2018-10-22
Removing netcdftime module

### LINKED ISSUE #481: Importing netcdftime which is a deprecated module
In the code `netcdftime` is imported:

https://github.com/OceanParcels/parcels/blob/master/parcels/tools/converters.py#L4

This is a deprecated name for the `cftime` module (which is also imported). 

The `netcdftime` import should be removed.

erikvansebille:
Removing `netcdftime` module, which is deprecated and replaced by `cftime`. This fixes #481, as suggested by @aidanheerdegen

Note that we also changed flake8 PEP W504 error, where binary operators on multiple lines should be placed at beginning of line, rather than end (see https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1003,14 +1003,14 @@ class VectorField(object):
         meshJac = (deg2m * deg2m * cos(rad * y)) if grid.mesh == 'spherical' else 1
         jac = self.jacobian(xsi, eta, px, py) * meshJac
 
-        u = ((-(1-eta) * U - (1-xsi) * V) * px[0] +
-             ((1-eta) * U - xsi * V) * px[1] +
-             (eta * U + xsi * V) * px[2] +
-             (-eta * U + (1-xsi) * V) * px[3]) / jac
-        v = ((-(1-eta) * U - (1-xsi) * V) * py[0] +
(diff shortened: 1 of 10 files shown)


parcels/field.py
```diff
@@ -1003,14 +1003,14 @@ def spatial_c_grid_interpolation2D(self, ti, z, y, x, time):
...
-        u = ((-(1-eta) * U - (1-xsi) * V) * px[0] +
-             ((1-eta) * U - xsi * V) * px[1] +
-             (eta * U + xsi * V) * px[2] +
-             (-eta * U + (1-xsi) * V) * px[3]) / jac
-        v = ((-(1-eta) * U - (1-xsi) * V) * py[0] +
-             ((1-eta) * U - xsi * V) * py[1] +
-             (eta * U + xsi * V) * py[2] +
-             (-eta * U + (1-xsi) * V) * py[3]) / jac
+        u = ((-(1-eta) * U - (1-xsi) * V) * px[0]
+             + ((1-eta) * U - xsi * V) * px[1]
```
delandmeterp: They changed something in flake8?

Issue #468, opened 2018-10-09
Extracting sub-domains with indices, but how to handle the longitude periodicity ?   


slawchune:
Hello Erik and Phillipe, 

I run Parcels with 1/12° global currents and my velocity data are so big that I feel convenient to extract a subdomain of velocity fields to work with. But some of the drifters that I model cross the [-180°,180°] boundary, that are in fact at the edges of my grid (centred at 0° of longitude). 

I tried to set negative indexes to my extractions (for instance : indices_extract= {'depth': [0], 'lat': range(863, 975), 'lon': range(-67, 67)}), which seems to work when plotting the initial state of the particles : 
![parcels_longitude_periodicity_0](https://user-images.githubusercontent.com/43207640/46663452-698eef80-cbbe-11e8-964e-757fb7130609.png)

But finally after some hours of drift, the particle doesn't want to cross the western boundary to the reach the eastern one. 

![parcels_longitude_periodicity_50](https://user-images.githubusercontent.com/43207640/46663560-b7a3f300-cbbe-11e8-8ff1-fd0794960d62.png)

Instead, the particle falls out of the domain and is removed thanks the DeleteParticle kernel given as example in the tutorial. 

Am I doing things wrong ? Sorry if the question was already post, I didn't find anything about this.  

Best regards, 
Stéphane

slawchune:
Closing the issue because the answer might be there : https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_periodic_boundaries.ipynb

slawchune:
It actually works with a periodic halo. I only add the halo if the -180/180° longitude axis is inside my extraction region : 

        # add a zonal periodic halo at the (180/-180) longitude axis. 
        # We only perform this if the longitude axis (180/-180) is contained inside the subgrid
        if ( fieldset.U.grid.lon[0] > 0  and fieldset.U.grid.lon[-1] < 0 ):
            fieldset.add_constant('halo_west', fieldset.U.grid.lon[fieldset.U.grid.lon.argmin()])
            fieldset.add_constant('halo_east', fieldset.U.grid.lon[fieldset.U.grid.lon.argmax()])
            fieldset.add_periodic_halo(zonal=True)

![parcels_longitude_periodicity_72](https://user-images.githubusercontent.com/43207640/46671172-63a40900-cbd4-11e8-8a13-9e63e80c17ad.png)

Maybe a more elegant solution exists...

[...]

PR #483, opened 2018-10-23
inform that negative indices are not allowed in fields

delandmeterp:
See issue #468

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -136,7 +136,8 @@ class Field(object):
         :param variable: Name of the field to create. Note that this has to be a string
         :param dimensions: Dictionary mapping variable names for the relevant dimensions in the NetCDF file
         :param indices: dictionary mapping indices for each dimension to read from file.
-               This can be used for reading in only a subregion of the NetCDF file
+               This can be used for reading in only a subregion of the NetCDF file.
+               Note that negative indices are not allowed.
         :param mesh: String indicating the type of mesh coordinates and
                units used during velocity interpolation:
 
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -588,6 +589,7 @@ def interpolator2D(self, ti, z, y, x):
         xi = 0
         yi = 0
         (xsi, eta, _, xi, yi, _) = self.search_indices(x, y, z, xi, yi)
+        return self.data[ti, yi+1, xi+1]
```
erikvansebille: Why `return` here? Shouldn't we instead do an `assert` of the indices themselves?

parcels/field.py
```diff
@@ -136,7 +136,8 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         :param variable: Name of the field to create. Note that this has to be a string
         :param dimensions: Dictionary mapping variable names for the relevant dimensions in the NetCDF file
         :param indices: dictionary mapping indices for each dimension to read from file.
-               This can be used for reading in only a subregion of the NetCDF file
+               This can be used for reading in only a subregion of the NetCDF file.
+               Note that negative indices are not allowed.
```
erikvansebille: Are they really 'not allowed'? Or are they just discouraged as it makes the code much slower?

parcels/field.py
delandmeterp: Woups! This is a completely unrelated line that shouldn't have been commited...

PR #486, opened 2018-11-06
WIP: New writing routine for Parcels

pjpetersik:
The new writing routine for Parcels, ParticleFile.wirte(), writes one tempory output file each time step using the method np.save() that saves data to a binary pickle format. After the simulation is done, the temporary output files are converted to the output NetCDF file by calling the method ParticleFile.convert_npy() upon deletion of the ParticleFile instance (ParticleFile.__del__()). The conversion is done by first loading the temporary files into the memory with np.load() and writing them afterward into the ParticleFile.dataset instance. 

A fast solution to not exceed the free memory of a system:
I included an option to use batch processing in the ParticleFile.conver_npy() method that will automatically adjust the batch size such that the free memory of the system should not be exceeded and throw a MemoryError if even one output file is larger than the free Memory. For this, the package psutil is used.


erikvansebille:
PR has been cleaned up and progressed in #497

PR #488, opened 2018-11-12
Nested fields

delandmeterp:


tests/test_fieldset_sampling.py
```diff
@@ -453,3 +453,54 @@ def test_summedfields(mode, with_W, k_sample_p, mesh):
     assert np.isclose(pset[0].p, 60)
     assert np.isclose(pset[0].lon*conv, 0.6, atol=1e-3)
     assert np.isclose(pset[0].lat, 0.9)
+
+
+@pytest.mark.parametrize('mode', ['jit', 'scipy'])
+def test_nestedfields(mode, fieldset, k_sample_p):
+    xdim = 10
+    ydim = 20
+    gf = 10  # factor by which the resolution of grid1 is higher than of grid2
```
erikvansebille: The comment here is not true anymore, as now the range of the two grids is different. Clarify better?

PR #491, opened 2018-11-20
New C-grid tracer interpolation method

IsoldeGlissenaar:
Increasing the accuracy of the C-grid Tracer interpolation. We suggest a new interpolation method called uniform_value, written for both 2D and 3D fields. 

This new interpolation method assigns the tracer value of the gridcell the particle is in to the particle. Everywhere in the cell, the particle will be assigned the same tracer value. This is an improvement for particles located in a C-grid like the NEMO fields, which are currently interpolated with nearest neighbour interpolation.

Written by Anne Kruijt and Isolde Glissenaar


erikvansebille:
Moved this PR to the parcels repo, for easier maintenance. See #494

PR #495, opened 2018-11-26
Field plotting

delandmeterp:


parcels/plotting.py
```diff
@@ -175,10 +185,26 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None
...
+        if field[0].interp_method == 'cgrid_tracer':
+            d = data[0][1:, 1:]
+        elif field[0].interp_method == 'cgrid_velocity':
+            if field[0].fieldtype == 'U':
+                d = np.empty_like(data[0])
+                d[:-1, :-1] = (data[0][1:, :-1] + data[0][1:, 1:]) / 2.
+            elif field[0].fieldtype == 'V':
+                d = np.empty_like(data[0])
+                d[:-1, :-1] = (data[0][:-1, 1:] + data[0][1:, 1:]) / 2.
+        else:  # if A-grid
```
erikvansebille: This is quite similar to the code in #496, right? Make a function of this, so that we can reuse and maintain in one location?

Issue #473, opened 2018-10-16
Performance of IO-Loop


willirath:
Parcels spends a lot of time in <https://github.com/OceanParcels/parcels/blob/master/parcels/particlefile.py#L161_L170>.

For 12-hourly output and a 5-minute time step, the Agulhas example spends 95% of the time in this loop. [(See this gist for details.)](https://gist.github.com/willirath/dbdd1ee9578fb42718e96de7597caeb4)

To me, the root of the problem seems to be the fact that this loop supports the most general case (particles being killed / created at runtime) in an explicit way.

willirath:
(cc: @ohlmann and @phandmann This is the IO-bottleneck I was referring to.)

willirath:
(cc: @handmannp)

delandmeterp:
This should be fixed in the near future. We haven't attacked that bottleneck since two students are working on it for a project. Outcomes should come within two weeks.

willirath:
That's great to hear.  Then as a next step, we can talk about OpenMP parallelization of the particle loop. :)

erikvansebille:
This has been fixed now in #486 and #497

PR #497, opened 2018-11-27
Faster output writing by merging temporary files into one large netcdf file at end of run

erikvansebille:
This PR builds upon the code developed in #486 by @pjpetersik and Frederik Felius. It is a complete redesign of the ParticleFile writing routine in Parcels, and thus addresses #473. Instead of writing all particle data to a NetCDF file with two unlimited dimensions directly, this PR makes writing a two-step process:

1) During execution, every `ParticleFile.write()` creates a temporary `.npy` file, stored in a temporary directory
2) When the ParticleFile is closed (when the script terminates, or by issuing `ParticleFile.close()` or `ParticleFile.export()`), all temporary `.npy` files are combined in one large NetCDF file. Since dimensions are known by that time, the file creation is much faster and more efficient

Things still to do:

- [x] Confirm that `to_write=once` variables are aways written to correct Particle ID
- [x] Improve doc-strings of new functions/methods
- [x] Test in realistic runs, and confirm speed-up


willirath:
A quick report on the impact  of this on #473: Looks like the cost for IO has been slashed to about 15% of the implementation currently in `master`.

erikvansebille:
> A quick report on the impact of this on #473: Looks like the cost for IO has been slashed to about 15% of the implementation currently in `master`.

Hmm, that's somewhat disappointing. We were hoping for order(s) of magnitude faster runs. Could you let us know what type of run you've done (how many particles, how many integration steps, how many output time steps)?

erikvansebille:
> > A quick report on the impact of this on #473: Looks like the cost for IO has been slashed to about 15% of the implementation currently in `master`.
> 
> Hmm, that's somewhat disappointing. We were hoping for order(s) of magnitude faster runs. Could you let us know what type of run you've done (how many particles, how many integration steps, how many output time steps)?

Oh sorry, I now see that all that information is actually already in #473. So ignore the last request. I'll do some performance checking too

willirath:
> Oh sorry, I now see that all that information is actually already in #473. So ignore the last request. I'll do some performance checking too

I think you can almost completely re-run this gist: https://gist.github.com/willirath/dbdd1ee9578fb42718e96de7597caeb4

erikvansebille:
I've now also done an analysis based on the gist from @willirath. As can be seen below, I'm getting easily a 7x speed-up, for 8000 particles. So that agrees nicely with the 15% slash that he saw

```python
import numpy as np
import time
import os
from matplotlib import pyplot as plt
```


```python
%%writefile profiling_script.py
import parcels
from datetime import timedelta
import numpy as np
from argparse import ArgumentParser

def DeleteParticle(particle, fieldset, time, dt):
    particle.delete()

def run_experiment(runtime=timedelta(days=5),
                   output_timedelta=timedelta(hours=12),
                   num_of_particles=1e3):
    filenames = "/Users/erik/Codes/ParcelsCode/parcels/examples/GlobCurrent_example_data/20*.nc"
    variables = {'U': 'eastward_eulerian_current_velocity',
                 'V': 'northward_eulerian_current_velocity'}
    dimensions = {'lat': 'lat',
                  'lon': 'lon',
                  'time': 'time'}
    fieldset = parcels.FieldSet.from_netcdf(filenames, variables, dimensions)
    
    num_of_particles_sqrt = int(np.round(num_of_particles ** 0.5))
    lons, lats = np.meshgrid(np.linspace(15, 35, num_of_particles_sqrt),
                             np.linspace(-40, -30, num_of_particles_sqrt))
    pset = parcels.ParticleSet(fieldset=fieldset, pclass=parcels.JITParticle, lon=lons, lat=lats)

    output_file = parcels.ParticleFile(
                 name=("agulhas_profiling"),
                 particleset=pset,
                 outputdt=output_timedelta)
    
    pset.execute(parcels.AdvectionRK4,
                 runtime=runtime,
                 dt=timedelta(minutes=5),
                 output_file=output_file,
                 recovery={parcels.ErrorCode.ErrorOutOfBounds: DeleteParticle})

if __name__ == "__main__":
    p = ArgumentParser()
    p.add_argument('num_of_particles', default=500)
    args = p.parse_args()
    run_experiment(num_of_particles=int(args.num_of_particles))
```

    Overwriting profiling_script.py



```python
import os
func_time = {}
for branch in ['master', 'pr/486']:
    print('Changing to branch '+branch)
    os.system("git --git-dir='/Users/erik/Codes/ParcelsCode/.git' --work-tree='/Users/erik/Codes/ParcelsCode/' checkout "+branch)
    func_time[branch] = []
    nops = [10, 50, 100, 500, 1000, 2000, 4000, 8000]
    for nop in nops:
        foo = %timeit -r1 -n1 -o os.system("python profiling_script.py "+str(nop))
        func_time[branch].append(foo)

    plt.plot(nops, [f.best for f in func_time[branch]], 'o-', label=branch)
plt.legend()
plt.xlabel('Number of particles')
plt.ylabel('Time spent [s]')
plt.show()
```

    Changing to branch master
    1 loop, best of 1: 3.27 s per loop
    1 loop, best of 1: 3.79 s per loop
    1 loop, best of 1: 4.66 s per loop
    1 loop, best of 1: 11.1 s per loop
    1 loop, best of 1: 20.4 s per loop
    1 loop, best of 1: 38.8 s per loop
    1 loop, best of 1: 1min 13s per loop
    1 loop, best of 1: 2min 22s per loop
    Changing to branch pr/486
    1 loop, best of 1: 3.44 s per loop
    1 loop, best of 1: 4.22 s per loop
    1 loop, best of 1: 3.86 s per loop
    1 loop, best of 1: 4.46 s per loop
    1 loop, best of 1: 5.65 s per loop
    1 loop, best of 1: 7.29 s per loop
    1 loop, best of 1: 11.6 s per loop
    1 loop, best of 1: 20.2 s per loop



output_2_1


```python
plt.plot(nops, [fold.best/fnew.best for fold, fnew in zip(func_time["master"], func_time["pr/486"])], 'o-')
plt.xlabel('Number of particles')
plt.ylabel('Speed-up factor [time master / time new]')
plt.show()
```


output_3_0

parcels/particlefile.py
```diff
@@ -3,12 +3,20 @@
 import netCDF4
 from datetime import timedelta as delta
 from parcels.tools.loggers import logger
+import os
+from tempfile import gettempdir
+import psutil
```
delandmeterp: psutil not found, both on my laptop and cartesius.
I don't know why this wasn't catched by travis.
Maybe another dependency needs it, I haven't updated my env recently. I'll try tomorrow

erikvansebille: Might be difference between anaconda and miniconda? In any case, I now added `psutil` to the dependency list

willirath:
A few comments on the design of `ParticleFile`:

The [`.write` methdod](https://github.com/OceanParcels/parcels/blob/fae198f94e2373ed2535fe958c600884f6f58523/parcels/particlefile.py#L167_L229) currently serves two purposes at the same time:

- Mapping particle positions from the `ParticleSet` dictionary (`data`)
- Writing the contents of this dictionary to disk.

As is evident from the [`.export` method](https://github.com/OceanParcels/parcels/blob/fae198f94e2373ed2535fe958c600884f6f58523/parcels/particlefile.py#L255_L270), `data` can be mapped to a dataset.  This is great, because we could easily add a method to export to an xarray dataset!

But I think it would be a good idea to think about disentangling or simplifying the structure by splitting the logic into:

get raw data from the `ParticleSet`
→ buffer raw data to disk or / and to memory
→ get raw data from disk or from memory
→ map raw data to a dataset (nc / xarray / ...)
→ write dataset to disk

Make each of these steps configurable at the initialization of the `ParticleFile`.

The benefits would be:
- support systems without (sufficiently large) local storage
- provide easy interfaces for getting the data out of Parcels
- make it easier to further evolve the logic

ignasivalles:
After update into this branch I get a KeyError with the psutil package. I'm releasing different number of particles and executing them in each time stept (runtime) within a while loop using pse.add and pset.write in each loop. The error appears at the end of the run when it's time to close everything into the netcdf output file.


Exception ignored in: <function ParticleFile.__del__ at 0x2ba13da16b70>
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/parcels/particlefile.py", line 147, in __del__
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/parcels/particlefile.py", line 152, in close
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/parcels/particlefile.py", line 277, in export
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/psutil/__init__.py", line 1946, in virtual_memory
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/psutil/_pslinux.py", line 397, in virtual_memory
  File "/home/valles/miniconda3/envs/py3_parcels/lib/python3.7/site-packages/psutil/_pslinux.py", line 217, in get_procfs_path
KeyError: ('psutil',)

erikvansebille:
Thanks for reporting this, @ignasivalles. Seems to be an error in the implementation of `psutil`? I had a quick google search of the error, but couldn't find much. There was something about problems when calling `psutil` on `__del__` (https://github.com/yahoo/redislite/issues/93).

Could you try again, but then add `pset.close()` at the end of the run? And can you then report back if that fixes the issue?

ignasivalles:
Hi @erikvansebille adding pset.close() at the end of the run doesn't solve the problem. It appears the same type of error. I'll keep looking around if I find what's going on.

parcels/particlefile.py
```diff
@@ -24,46 +34,60 @@ def _is_particle_started_yet(particle, time):
...
+        if len(self.var_names_once) > 0:
+            self.written_once = []
+            self.file_list_once = []
+
+        self.file_list = []
+        self.time_written = []
+        self.maxid_written = -1
+        self.dataset_open = True
+
+        self.npy_path = os.path.join(os.path.dirname(str(self.name)), "out-%s" % ''.join(random.choice(string.ascii_uppercase) for _ in range(8)))
```
willirath: It would be great for the user to be able to change this path.  This would be useful for, e.g., having a local SSD for NPY files and only build the final netCDF in distributed storage.  (We, e.g., ran into an inode-quota problem when writing the many temporary NPY files to where the final netCDF file is meant to be.)

Ways to implement this would be:

- Use env vars like `$TMP`.  But not sure if this is a good choice, because the NPY files can be very big.
- Add kwarg to `ParticleFile` and have some good default.

I'd prefer the latter.

willirath: cc: @schmidt-christina

erikvansebille: Thanks @willirath! I've now implemented the kwarg version (via a new optional argument `tempwritedir` in the `ParticleFile` constructor) in a89bdaf

environment_py2_linux.yml
```diff
@@ -25,5 +25,6 @@ dependencies:
   - xarray>=0.10.8
   - dask
   - cftime
+  - psutil
```
delandmeterp: Since the PR has been open for a while, should we check if this dependency is still necessary?

erikvansebille: We need this package to check on the amount of free virtual memory: https://github.com/OceanParcels/parcels/blob/ab2976a63584ca3f8590282a862abd674fd3ddfa/parcels/particlefile.py#L324 

Unless anyone knows a python-core method to do this, I think we'll have to keep this dependency

parcels/particlefile.py
```diff
@@ -31,177 +41,301 @@ def _set_calendar(origin_calendar):
...
         self.dataset.feature_type = "trajectory"
         self.dataset.Conventions = "CF-1.6/CF-1.7"
         self.dataset.ncei_template_version = "NCEI_NetCDF_Trajectory_Template_v2.0"
         self.dataset.parcels_version = parcels_version
-        self.dataset.parcels_mesh = self.particleset.fieldset.gridset.grids[0].mesh
+        self.dataset.parcels_mesh = self.parcels_mesh
 
         # Create ID variable according to CF conventions
-        self.id = self.dataset.createVariable("trajectory", "i4", coords, chunksizes=self.chunksizes)
+        self.id = self.dataset.createVariable("trajectory", "i4", coords, fill_value=-2147483647, chunksizes=data_shape)
```
delandmeterp: What is this fill_value?

parcels/particlefile.py
```diff
@@ -31,177 +41,301 @@ def _set_calendar(origin_calendar):
...
+            self.close()
 
-    def sync(self):
-        """Write all buffered data to disk"""
-        self.dataset.sync()
+    def close(self):
+        """Close the ParticleFile object by exporting and then deleting
+        the temporary npy files"""
+        self.export()
+        self.delete_tempwritedir()
```
delandmeterp: the temp write dir can sometimes be useful (especially if something goes wrong in a big run)
Should that be an option to delete the temp write dir?
(not sure though, especially since if something goes wrong, the code should break so the dir is not deleted)

parcels/particlefile.py
```diff
@@ -31,177 +41,301 @@ def _set_calendar(origin_calendar):
...
+        return data_dict, data_dict_once
+
+    def dump_dict_to_npy(self, data_dict, data_dict_once):
+        """Buffer data to set of temporary numpy files, using np.save"""
+
+        if not os.path.exists(self.tempwritedir):
+            os.mkdir(self.tempwritedir)
+
+        if len(data_dict) > 0:
+            tmpfilename = os.path.join(self.tempwritedir, str(len(self.file_list) + 1) + ".npy")
```
delandmeterp: Why the +1 here?
In Python iterators start at 0

parcels/particlefile.py
```diff
@@ -31,177 +41,301 @@ def _set_calendar(origin_calendar):
...
+            with open(tmpfilename, 'wb') as f:
+                np.save(f, data_dict)
+            self.file_list.append(tmpfilename)
+
+        if len(data_dict_once) > 0:
+            tmpfilename = os.path.join(self.tempwritedir, str(len(self.file_list) + 1) + '_once.npy')
```
delandmeterp: also here

parcels/particlefile.py
erikvansebille: Because it's the next number in line. So if there are already x files in the list, then the next filename is x+1

parcels/particlefile.py
erikvansebille: See above

parcels/particlefile.py
erikvansebille: Yes, indeed it is quite useful that the output-dir exists if the run is somehow broken off. We delete it only after a successful `export()`. 

In the future, we might even want a function to restart a run by creating a particleSet from the last numpy file in a temporary output directory.

parcels/particlefile.py
delandmeterp: I meant that when list is [], file should be 0.npy,
then you add 1.npy to [0.npy]
no?

parcels/particleset.py
```diff
@@ -411,6 +411,10 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
         while (time < endtime and dt > 0) or (time > endtime and dt < 0) or dt == 0:
             if verbose_progress is None and time_module.time() - walltime_start > 10:
                 # Showing progressbar if runtime > 10 seconds
+                if output_file:
```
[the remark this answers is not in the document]
erikvansebille: True, but the idea is that if a run is so short that Parcels doesn't show a progressbar (i.e. shorter than 10 seconds), it also doesn't need to show where the temporary files are stored. But I agree with you that the verbosity levels of Parcels could be rethought and refactored

willirath:
:tada:

Issue #447, opened 2018-09-14
Support for other grids that RectilinearZGrids in FieldSet.from_data()


erikvansebille:
It is currently not easy to create Fields on other Grids than RectilinearZGrids using `FieldSet.from_data()` or `Field.__init__()`. The only way is to first create the Grid, and only then call `Field.__init__`. It is not possible by providing `lon`, `lat` etc arrays.

erikvansebille:
@delandmeterp, is this now fixed in #496?

erikvansebille:
Fixed in #498

PR #498, opened 2018-11-29
From data v2.0

delandmeterp:
solves issue #447

### DIFF
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -413,3 +413,53 @@ def test_fieldset_from_xarray(maxlatind):
 
     pset.execute(AdvectionRK4, dt=1)
     assert pset[0].lon == 4.5 and pset[0].lat == 10
+
+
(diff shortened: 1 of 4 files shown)


tests/test_fieldset.py
```diff
@@ -413,3 +413,50 @@ def generate_dataset(xdim, ydim, zdim=1, tdim=1):
...
+    plon = pset[0].lon
+    plat = pset[0].lat
+
+    # Rectilinear S grid
+    dimensions['depth'] = depth_s
+    fieldset = FieldSet.from_data(data, dimensions, mesh='flat')
+    pset = ParticleSet(fieldset, ScipyParticle, 0, 0)
+    pset.execute(AdvectionRK4, runtime=1, dt=.5)
+    plon = pset[0].lon
+    plat = pset[0].lat
```
erikvansebille: Assert here that particles have actually moved (i.e. that `plon == 1`)? Because if for some reason there is no movement, then all these asserts will go well

delandmeterp: Done, I check analytical value.

PR #499, opened 2018-12-04
Cgrid full interpolation

delandmeterp:


parcels/field.py
```diff
@@ -628,8 +629,8 @@ def interpolator2D(self, ti, z, y, x):
             raise RuntimeError(self.interp_method+" is not implemented for 2D grids")
 
     def interpolator3D(self, ti, z, y, x, time):
-        xi = int(self.grid.xdim / 2)
-        yi = int(self.grid.ydim / 2)
+        xi = int(self.grid.xdim / 2) - 1
+        yi = int(self.grid.ydim / 2) - 1
```
erikvansebille: Why this change? Just for future reference, would be good to mention why we need to subtract 1 here, and then add it later in this function

parcels/field.py
```diff
@@ -967,15 +968,10 @@ def __init__(self, name, U, V, W=None):
...
-            x1 = r*np.cos(rad*lon1) * np.cos(rad*lat1)
-            y1 = r*np.sin(rad*lon1) * np.cos(rad*lat1)
-            z1 = r*np.sin(rad*lat1)
-            x2 = r*np.cos(rad*lon2) * np.cos(rad*lat2)
-            y2 = r*np.sin(rad*lon2) * np.cos(rad*lat2)
-            z2 = r*np.sin(rad*lat2)
-            return np.sqrt((x2-x1)**2 + (y2-y1)**2 + (z2-z1)**2)
+            deg2m = 1852 * 60.
+            lat = (lat1+lat2)/2.
+            return np.sqrt(((lon2-lon1)*deg2m*cos(rad * lat))**2 + ((lat2-lat1)*deg2m)**2)
```
erikvansebille: So we are using the assumption that distances are simple L2 norm on the plane tangental to the sphere? That only works for small distances between points. Why make this change then, if the former method was more general?

erikvansebille: Also, use `np.cos` instead to normal `cos`?

parcels/field.py
```diff
@@ -1041,6 +1037,107 @@ def spatial_c_grid_interpolation2D(self, ti, z, y, x, time):
...
+        if grid.gtype in [GridCode.RectilinearSGrid, GridCode.RectilinearZGrid]:
+            px = np.array([grid.lon[xi], grid.lon[xi+1], grid.lon[xi+1], grid.lon[xi]])
+            py = np.array([grid.lat[yi], grid.lat[yi], grid.lat[yi+1], grid.lat[yi+1]])
+        else:
+            px = np.array([grid.lon[yi, xi], grid.lon[yi, xi+1], grid.lon[yi+1, xi+1], grid.lon[yi+1, xi]])
+            py = np.array([grid.lat[yi, xi], grid.lat[yi, xi+1], grid.lat[yi+1, xi+1], grid.lat[yi+1, xi]])
+
+        if grid.mesh == 'spherical':
+            px[0] = px[0]+360 if px[0] < x-225 else px[0]
+            px[0] = px[0]-360 if px[0] > x+225 else px[0]
```
erikvansebille: what's special about the `225` here?

parcels/include/parcels.h
```diff
@@ -201,15 +212,10 @@ static inline ErrorCode temporal_interpolation_structured_grid(float x, float y,
...
-    double x1 = r*cos(rad*lon1) * cos(rad*lat1);
-    double y1 = r*sin(rad*lon1) * cos(rad*lat1);
-    double z1 = r*sin(rad*lat1);
-    double x2 = r*cos(rad*lon2) * cos(rad*lat2);
-    double y2 = r*sin(rad*lon2) * cos(rad*lat2);
-    double z2 = r*sin(rad*lat2);
-    return sqrt((x2-x1)*(x2-x1) + (y2-y1)*(y2-y1) + (z2-z1)*(z2-z1));
+    double deg2m = 1852 * 60.;
+    double lat = (lat1+lat2)/2.;
+    return sqrt((lon2-lon1)*(lon2-lon1) * deg2m * deg2m * cos(rad * lat) * cos(rad * lat) + (lat2-lat1)*(lat2-lat1) * deg2m * deg2m);
```
erikvansebille: This is a different `dist` function than the one in field.py?

tests/test_grids.py
```diff
@@ -385,3 +385,84 @@ def test_advect_nemo(mode):
...
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+def test_cgrid_uniform_2dvel(mode):
+    lon = np.array([[0, 2], [.4, 1.5]])
+    lat = np.array([[0, -.5], [.8, .5]])
+    U = np.array([[-99, -99], [4.4721359549995793e-01, 1.3416407864998738e+00]])
+    V = np.array([[-99, 1.2126781251816650e+00], [-99, 1.2278812270298409e+00]])
+
+    dimensions = {'lat': lat, 'lon': lon}
+    data = {'U': np.array(U, dtype=np.float32), 'V': np.array(V, dtype=np.float32)}
+    fieldset = FieldSet.from_data(data, dimensions, mesh='flat')
```
erikvansebille: also test this on spherical mesh?

delandmeterp:
Indeed, there is a lot of repetition between py and C code.
The positive way to see it is that it is independent so double check up
The negative is that everything is duplicated.
We could work on reducing all this load of code. Should that be here? Or part of the efficiency work to also reduce the code and call faster c function?
(I think it shouldn't be only the interpolation_utils.h, but most of the interpolation code in field.py)

I would do it as a separate branch.

parcels/field.py
delandmeterp: We don't add it later in the function.
This is the initial xi, yi guess before looking really in which cell we are (in scipy mode, xi and yi are not attached to the particle)

For the smaller possible element, xdim=ydim=2
So you can see that initial xi and yi are 1, which does not make sense since it ranges from 0 to 1. This is why there is this -1

erikvansebille:
Ok, I agree with the strategy to postpone the cleanup of the code to a next
PR, post v2.0, as it's under-the-hood work only

parcels/field.py
delandmeterp: This change is more complicate.
Basically, we pass twice from deg/s to m/s: the first time here when we compute the fluxes; the second time in the interpolation while we divide by the jacobian, which takes into account the changes from curvi to parent element + the spheric mesh.
This new change is more consistent because we use the same transformation twice, while before the transformation where independent one from each other.

This way is also consistent with what we do for 3D interpolation. And in 3D, it is the only way to do it.

As you said, it is less accurate (theoretically). It is true, but only because what I do is still currently wrong. For the moment I evaluate at the lat as the mean lat of the edge. If I take the correct one (not the particle lat, but the position as it is done in 3D (see fig. 5 of the draft), it will be consistent and accurate). For the moment, the two transformations we do are not consistent.

Practically, I observed that it does not change anything.
I'll check with more details to see if it is a bit more accurate with the new scheme (while corrected).

delandmeterp: Oups, this is wrong (bad copy paste from the .h version). Flake8 should have found that...

parcels/field.py
delandmeterp: This is nothing new. It's as before how we try to find in which cell our particle is located. It was meant to make it work even for coarse mesh (we want to mesh the 360 shift if distance is more than 180 but less than 270, alias 90 degrees)

parcels/include/parcels.h
delandmeterp: It shouldn't.
It is the same as ` return np.sqrt(((lon2-lon1)*deg2m*cos(rad * lat))**2 + ((lat2-lat1)*deg2m)**2)`, no?

erikvansebille: Ah yes, you're right

parcels/field.py
erikvansebille: OK, is fine then

tests/test_grids.py
delandmeterp: I can design such test. The problem is that I don't have a velocity dataset. I did create this one.
I'll test by using the velocities from the 2d nemo zonal velocity (so it will be z-grid which I'll artificially move to s-grid).

erikvansebille: Do you still want to add this test of `mesh=spherical`? I think we should, also because spherical meshes are more common in our applications than flat meshes. 
Once that gets tested, I'm happy for this to be merged into `master`

PR #512, opened 2018-12-21
correcting error message

delandmeterp:
when interpolating a scalar field with interp_method=cgrid_velocity

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -626,6 +626,8 @@ class Field(object):
             return val
         elif self.interp_method is 'cgrid_tracer':
             return self.data[ti, yi+1, xi+1]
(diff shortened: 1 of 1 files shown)


parcels/field.py
```diff
@@ -626,6 +626,8 @@ def interpolator2D(self, ti, z, y, x):
             return val
         elif self.interp_method is 'cgrid_tracer':
             return self.data[ti, yi+1, xi+1]
+        elif self.interp_method is 'cgrid_velocity':
+            raise RuntimeError(self.interp_method+" must be interpolated as VectorField")
```
erikvansebille: Also add name of Field to error message? And be more explicit by adding `(e.g. fieldset.UV[])`?

PR #514, opened 2018-12-21
[Do not merge!] This shows a way out of the performance issues with netcdf4=1.4.1

willirath:
Changes 

- Set automasking of netCDF data to False.
- Unpin netCDF4 max version again
- Measrure timing of pytests (which might help with diagnosing similar issues in the future)

Note that my change in 

https://github.com/OceanParcels/parcels/commit/378c5ee5437a6e0aa71d6b3ee75ec9adbb4332e6#diff-a68ef84b67d3b404f48daf2ccc3ca441R1337

is a relatively dirty hack.  So please don't just merge this. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
         sh -e /etc/init.d/xvfb start;
         sleep 3;
       fi
-    - py.test -v -s tests/
+    - py.test -v -s --durations=0 tests/
     - |
       # only get examples on linux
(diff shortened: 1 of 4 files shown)


willirath:
With e001ae6, the individual test timings should show where exactly things go wrong.

erikvansebille:
The latest version of netcdf4 is fast again, so we don't need this workaround anymore. Thanks for your help, @willirath

PR #521, opened 2019-01-10
summedfields cgrid unitconversion bugfix

erikvansebille:
Fixing a bug where unit converters were still applied to `cgrid_velocity` in summed fields, even though that is not necessary. 

While this was correctly implemented for normal VectorFields and NestedVectorFields, it was not yet implemented for SummedVectorFields

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -761,10 +761,11 @@ class KernelGenerator(ast.NodeVisitor):
         for U, V, W, var, var2, var3 in zip(node.field.obj.U, node.field.obj.V, Wlist, node.var, node.var2, node.var3):
             vfld = VectorField(node.field.obj.name, U, V, W)
             ccode_eval = vfld.ccode_eval(var, var2, var3, U, V, W, *node.args.ccode)
-            ccode_conv1 = U.ccode_convert(*node.args.ccode)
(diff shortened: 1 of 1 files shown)


parcels/codegenerator.py
```diff
@@ -761,10 +761,11 @@ def visit_SummedVectorFieldEvalNode(self, node):
...
             ccode_eval = vfld.ccode_eval(var, var2, var3, U, V, W, *node.args.ccode)
-            ccode_conv1 = U.ccode_convert(*node.args.ccode)
-            ccode_conv2 = V.ccode_convert(*node.args.ccode)
-            statements = [c.Statement("%s *= %s" % (var, ccode_conv1)),
-                          c.Statement("%s *= %s" % (var2, ccode_conv2))]
+            if U.interp_method != 'cgrid_velocity':
+                ccode_conv1 = U.ccode_convert(*node.args.ccode)
+                ccode_conv2 = V.ccode_convert(*node.args.ccode)
+                statements = [c.Statement("%s *= %s" % (var, ccode_conv1)),
+                              c.Statement("%s *= %s" % (var2, ccode_conv2))]
```
delandmeterp: an `else` part is missing (see other vector meta fields). Otherwise now `statement` is not defined for A-grids

PR #527, opened 2019-02-01
Temporary fix for appveyor while conda 4.5 has permission issues

erikvansebille:
See also https://github.com/conda/conda/issues/8155

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,7 @@ matrix:
 install:
   - "powershell ./appveyor/install_miniconda.ps1"
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - "conda install --yes conda=4.4"
+  - "conda install --yes conda<4.6"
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -19,7 +19,7 @@ matrix:
 install:
   - "powershell ./appveyor/install_miniconda.ps1"
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - "conda update --yes conda"
+  - "conda install --yes conda=4.4"
```
willirath: I think this is more restrictive than necessary.  

```yaml
  - "conda install --yes 'conda<4.6'"
```
should do.

willirath: (According to conda/conda#8155, this is a problem with 4.6 and not 4.5.)

erikvansebille:
New release of conda seems to have fixed appveyor issue, so this PR is not needed anymore

PR #531, opened 2019-02-12
Nemo3d example

delandmeterp:


parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,161 @@
...
+    "\n",
+    "lonp = lon_r.flatten()\n",
+    "latp = lat_r.flatten()\n",
+    "depthp = np.ones(latp.shape)\n",
+    "\n",
+    "pset = ParticleSet.from_list(fieldset, JITParticle, lon=lonp, lat=latp, depth=depthp)\n",
+    "kernels = pset.Kernel(AdvectionRK4_3D)\n",
+    "pset.execute(kernels, runtime=delta(days=4), dt=delta(hours=6))\n",
+    "\n",
+    "for p in pset:\n",
```
erikvansebille: I think you can just do `print(pset)`, instead of this for-loop

parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,161 @@
...
+    "fieldset = FieldSet.from_nemo(filenames, variables, dimensions)\n",
+    "fieldset.U.vmax = 5\n",
+    "fieldset.V.vmax = 5\n",
+    "fieldset.W.vmax = 5\n",
+    "\n",
+    "# meshgrid containing 5x5 points uniformly distributed in a [0,1]x[0,1] quad\n",
+    "vec = np.linspace(0, 1, 5)\n",
+    "xsi, eta = np.meshgrid(vec, vec)\n",
+    "\n",
+    "# get particles in Rhine estuary\n",
```
erikvansebille: Explain a bit more what this is, and why it's so complicated? If there's a moment to explain something important about dealing with 3D NEMO grids, then I suggest to split it into its own cell. Otherwise, just release one point?

parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,161 @@
...
+    "             'V': 'vo',\n",
+    "             'W': 'wo'}\n",
+    "dimensions = {'U': {'lon': 'glamf', 'lat': 'gphif', 'depth': 'depthw', 'time': 'time_counter'},\n",
+    "              'V': {'lon': 'glamf', 'lat': 'gphif', 'depth': 'depthw', 'time': 'time_counter'},\n",
+    "              'W': {'lon': 'glamf', 'lat': 'gphif', 'depth': 'depthw', 'time': 'time_counter'}}\n",
+    "\n",
+    "fieldset = FieldSet.from_nemo(filenames, variables, dimensions)\n",
+    "fieldset.U.vmax = 5\n",
+    "fieldset.V.vmax = 5\n",
+    "fieldset.W.vmax = 5\n",
```
erikvansebille: explain why these `vmax` are necessary?

parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,161 @@
...
+    "wfiles = sorted(glob(data_path+'ORCA025-N06_200001??d05W.nc'))\n",
+    "mesh_mask = data_path + 'coordinates.nc'\n",
+    "\n",
+    "filenames = {'U': {'lon': mesh_mask, 'lat': mesh_mask, 'depth': wfiles[0], 'data': ufiles},\n",
+    "             'V': {'lon': mesh_mask, 'lat': mesh_mask, 'depth': wfiles[0], 'data': vfiles},\n",
+    "             'W': {'lon': mesh_mask, 'lat': mesh_mask, 'depth': wfiles[0], 'data': wfiles}}\n",
+    "\n",
+    "variables = {'U': 'uo',\n",
+    "             'V': 'vo',\n",
+    "             'W': 'wo'}\n",
```
erikvansebille: Are we sure that `wo` is positive downward, as is assumed by the `AdvectionRK4_3D` kernel? Otherwise, should we add a `set_scaling_factor(-1)`?

parcels/fieldset.py
```diff
@@ -293,12 +293,69 @@ def from_nemo(cls, filenames, variables, dimensions, indices=None, mesh='spheric
...
         :param tracer_interp_method: Method for interpolation of tracer fields. It is recommended to use 'cgrid_tracer' (default)
-               Note that in the case of from_nemo(), the velocity fields are default to 'cgrid_velocity'
+               Note that in the case of from_nemo() and from_cgrid(), the velocity fields are default to 'cgrid_velocity'
+
+        """
+
+        fieldset = cls.from_c_grid_dataset(filenames, variables, dimensions, mesh=mesh, indices=indices, time_periodic=time_periodic,
+                                  allow_time_extrapolation=allow_time_extrapolation, tracer_interp_method=tracer_interp_method, **kwargs)
+        if hasattr(fieldset, 'W'):
+            fieldset.W.set_scaling_factor(-1.)
```
erikvansebille: But the warning in `AdvectionRK4_3D` is still that vertical velocity is assumed DOWNWARD; even though that is not the case with `from_nemo`. Might confuse users? Shall we change the warning message?

parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,138 @@
...
+   ],
+   "source": [
+    "from parcels import FieldSet, ParticleSet, JITParticle, AdvectionRK4_3D\n",
+    "from glob import glob\n",
+    "import numpy as np\n",
+    "from datetime import timedelta as delta\n",
+    "from os import path\n",
+    "\n",
+    "data_path = 'NemoNorthSeaORCA025-N006_data/'\n",
+    "ufiles = sorted(glob(data_path+'ORCA025-N06_200001??d05U.nc'))\n",
```
erikvansebille: use `ORCA*.nc` so that it would out of the box also work for 1/12 NEMO?

parcels/examples/tutorial_nemo_3D.ipynb
```diff
@@ -0,0 +1,138 @@
...
+    "pset = ParticleSet.from_line(fieldset=fieldset, pclass=JITParticle,\n",
+    "                             size=10,\n",
+    "                             start=(2.96, 51.6),\n",
+    "                             finish=(3.26, 51.8),\n",
+    "                             depth=1)\n",
+    "\n",
+    "kernels = pset.Kernel(AdvectionRK4_3D)\n",
+    "pset.execute(kernels, runtime=delta(days=4), dt=delta(hours=6))\n",
+    "\n",
+    "print(pset)\n"
```
erikvansebille: What really is the point of this `print` statement? What do we want the users to take away? Why not a figure (that would also be useful for the card that we will create on http://oceanparcels.org/#tutorials)?

parcels/fieldset.py
delandmeterp: This warning is still true. `AdvectionRK4_3D` assumes that positive `w` means increasing `z`.
I would not remove it. Maybe we should add "(increasing z)" after "downward", then it's more complete.

Also what we can do is to have a variable `fieldset.creation_log` or something like this.
And in `Fieldset.from_nemo` we set this variable to `from_nemo`.
And we can disable the warning if `creation_log == "from_nemo"`

Maybe we'll use this log for various purposes later on (not sure if it's a good idea though)

erikvansebille: > This warning is still true. `AdvectionRK4_3D` assumes that positive `w` means increasing `z`. I would not remove it. Maybe we should add "(increasing z)" after "downward", then it's more complete.

Well, if a user looks into their netcdf file, they may see something else; Since they may not realise we're already applying the `set_scaling_factor(-1)` for them, this could confuse them?

>Also what we can do is to have a variable `fieldset.creation_log` or something like this.
And in `Fieldset.from_nemo` we set this variable to `from_nemo`.
And we can disable the warning if `creation_log == "from_nemo"`

I like this idea yes. But I would do it on a per-field basis (since different `Fields` in a `FieldSet` can have a different `creation_log`, e.g. when combining models)

delandmeterp: Ok, I'll do it

PR #534, opened 2019-02-18
Support for custom variable initialisation from ParticleSet.__init__() with repeatdt

erikvansebille:
It was already possible to initialise a custom variable from `ParticleSet.__init()__` using a `kwargs` (e.g. `pset = ParticleSet(...,custom_var=3)`). However, this was not possible when also using repeated releases with `repeatdt` argument. This PR fixes that

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -64,6 +64,9 @@ class ParticleSet(object):
             raise NotImplementedError('If fieldset.time_origin is not a date, time of a particle must be a double')
         time = [self.time_origin.reltime(t) if isinstance(t, np.datetime64) else t for t in time]
 
+        for kwvar in kwargs:
+            kwargs[kwvar] = convert_to_list(kwargs[kwvar])
+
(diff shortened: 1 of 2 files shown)


parcels/particleset.py
```diff
@@ -64,6 +64,9 @@ def convert_to_list(var):
             raise NotImplementedError('If fieldset.time_origin is not a date, time of a particle must be a double')
         time = [self.time_origin.reltime(t) if isinstance(t, np.datetime64) else t for t in time]
 
+        for kwvar in kwargs:
+            kwargs[kwvar] = convert_to_list(kwargs[kwvar])
```
delandmeterp: Should we use a copy of kwargs or use it directly?

erikvansebille: Why do you think we need a copy? This won't change, would it?

delandmeterp: I'm always double checking when editing a dictionary, but here you're right it would not change anything.

Issue #537, opened 2019-02-19
fieldset.U.show() only displaying positive lon values for HYCOM


MahiMahi90:
When calling the fieldset.U.show() for a HYCOM dataset, only the positive lon values are displayed. (Hycom Lon values range from 0 to 360 deg.)

![hycom_w_oceanparcels](https://user-images.githubusercontent.com/44847330/53047440-b4851180-3492-11e9-843c-d1b0608b4251.png)

erikvansebille:
Thanks for reporting this! It's indeed a bug. I will fix it in a PR, but in the meantime a workaround seems to be to explicitly give the `projection` type. I.e.
```
import cartopy
fieldset.U.show(projection=cartopy.crs.PlateCarree())
```
Could you confirm that this works, @MahiMahi90 ?

MahiMahi90:
Yep that works! Thanks!!


<img width="637" alt="screenshot 2019-02-20 at 20 30 08" src="https://user-images.githubusercontent.com/44847330/53118869-61709480-354e-11e9-889e-a25e480b6399.png">

erikvansebille:
Fixed in #539

PR #539, opened 2019-02-20
Fix plotting domain

erikvansebille:
Changing the call to the `domain` option in plotting from a vector `[max lat, min lat, max lon, min lon]` to a dictionary with entries `{'S', 'N', 'E', 'W'}` for South, North, East and West extent, respectively. Note that the vectors notation is still backward-compatible

This PR also fixes the Issue with the hycom plotting #537

### DIFF
--- a/parcels/plotting.py
+++ b/parcels/plotting.py
@@ -210,11 +210,11 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None
         else:
             cs = ax.pcolormesh(plotlon[0], plotlat[0], d)
 
-    if cartopy is None or projection is None:
+    if cartopy is None:
         ax.set_xlim(np.nanmin(plotlon[0]), np.nanmax(plotlon[0]))
         ax.set_ylim(np.nanmin(plotlat[0]), np.nanmax(plotlat[0]))
     elif domain is not None:
-        ax.set_extent([np.nanmin(plotlon[0]), np.nanmax(plotlon[0]), np.nanmin(plotlat[0]), np.nanmax(plotlat[0])])
+        ax.set_extent([np.nanmin(plotlon[0]), np.nanmax(plotlon[0]), np.nanmin(plotlat[0]), np.nanmax(plotlat[0])], crs=cartopy.crs.PlateCarree())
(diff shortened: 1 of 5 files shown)


parcels/plotting.py
```diff
@@ -291,8 +291,10 @@ def create_parcelsfig_axis(spherical, land=True, projection=None, central_longit
 def parsedomain(domain, field):
     field.grid.check_zonal_periodic()
     if domain is not None:
-        _, _, _, lonW, latS, _ = field.search_indices(domain[3], domain[1], 0, 0, 0, search2D=True)
-        _, _, _, lonE, latN, _ = field.search_indices(domain[2], domain[0], 0, 0, 0, search2D=True)
+        if not isinstance(domain, dict) and len(domain)  == 4:
```
delandmeterp: Add small comment here saying this is backward compatibility with version < 2.0 ?

parcels/plotting.py
```diff
@@ -210,11 +210,11 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None
         else:
             cs = ax.pcolormesh(plotlon[0], plotlat[0], d)
 
-    if cartopy is None or projection is None:
+    if cartopy is None:
         ax.set_xlim(np.nanmin(plotlon[0]), np.nanmax(plotlon[0]))
         ax.set_ylim(np.nanmin(plotlat[0]), np.nanmax(plotlat[0]))
     elif domain is not None:
-        ax.set_extent([np.nanmin(plotlon[0]), np.nanmax(plotlon[0]), np.nanmin(plotlat[0]), np.nanmax(plotlat[0])])
+        ax.set_extent([np.nanmin(plotlon[0]), np.nanmax(plotlon[0]), np.nanmin(plotlat[0]), np.nanmax(plotlat[0])], crs=cartopy.crs.PlateCarree())
```
delandmeterp: This the change for #537 ?

erikvansebille: No, it was the change `if cartopy is None or projection is None:` to `if cartopy is None:`

The issue is that `projection` is always `None` if it's not explicitly passed as an argument

PR #540, opened 2019-02-20
Fieldset.from_netcdf with manual timestamps

christensen5:
This PR creates an option to provide time data as a separate 1D numpy array (called 'timestamps') when creating a fieldset from a netcdf file - this is useful when time is not explicitly included within the 'dimensions' dictionary of the numpy array, but only in the 'variables' dictionary. Note the following comments on the current implementation:

- All the `variables` (u, v, w, temp, etc...) must be contained in the same files, and each file must correspond to exactly one timestep. I check if the number of files is equal to the length of `timestamps` and throw an exception if not.

- Filenames must be provided as a list or as a dict; I throw an error if its a dict of dicts.

- Works with deferred loading.


erikvansebille:
Thanks for this Pull, @christensen5; a nice feature for some users who have netcdf files without time dimension

I just tried to create a unit test that tests this new feature, but as you see it currently breaks. Can you have a look at it (see https://github.com/OceanParcels/parcels/pull/540/commits/d5e750499d22161ccb099bec36a050e074054206)? Perhaps you can find why it breaks?

christensen5:
Unittest should now pass. The issue was that when working with the user-provided array of timestamps, netcdfFileBuffer.time() was returning the entire array, rather than the current time for a given file.

PR #545, opened 2019-02-25
Test correct keys in dimensions dictionary

erikvansebille:
This PR tests that the `dimensions` dictionary has no other keys than `lon`, `lat`, `depth` and `time`. Since other keys are not used, this prevents errors where users use the wrong key for e.g. `depth` and inadvertently use in 2D mode

Furthermore, this PR also cleans up the code by removing the `dimensions['data']` entry for the name of the variable in the filebuffer

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -55,8 +55,11 @@ class Field(object):
     def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=None, mesh='flat',
                  fieldtype=None, transpose=False, vmin=None, vmax=None, time_origin=None,
                  interp_method='linear', allow_time_extrapolation=None, time_periodic=False, **kwargs):
-        self.name = list(name)[0] if isinstance(name, dict) else name
-        self.filebuffername = name[list(name)[0]] if isinstance(name, dict) else name
+        if not isinstance(name, tuple):
+            self.name = name
+            self.filebuffername = name
+        else:
+            self.name, self.filebuffername = name
         self.data = data
         time_origin = TimeConverter(0) if time_origin is None else time_origin
         if grid:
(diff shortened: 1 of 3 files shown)


parcels/field.py
```diff
@@ -55,7 +55,8 @@ class Field(object):
     def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=None, mesh='flat',
                  fieldtype=None, transpose=False, vmin=None, vmax=None, time_origin=None,
                  interp_method='linear', allow_time_extrapolation=None, time_periodic=False, **kwargs):
-        self.name = name
+        self.name = list(name)[0] if isinstance(name, dict) else name
+        self.filebuffername = name[list(name)[0]] if isinstance(name, dict) else name
```
delandmeterp: It's quite hard to follow. Why is `name` a dict?
If you make `name` a tuple,  you can do
```
if not isinstance(name, tuple):
    self.name = name
    self.filebuffername = name
else:
   self.name, self.filebuffername = name
```
or as you did
```
self.name = name[0] if isinstance(name, tuple) else name
self.filebuffername = name[1] if isinstance(name, tuple) else name
```

parcels/field.py
```diff
@@ -133,7 +134,7 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                filenames can be a list [files]
                or a dictionary {dim:[files]} (if lon, lat, depth and/or data not stored in same files as data)
                time values are in filenames[data]
-        :param variable: Name of the field to create. Note that this has to be a string
+        :param variable: Dictionary mapping name of the field to create to variable in file.
```
delandmeterp: So here I would put a tuple. You will never have more than one key, right?

parcels/fieldset.py
```diff
@@ -34,6 +34,12 @@ def __init__(self, U, V, fields=None):
 
         self.compute_on_defer = None
 
+    @classmethod
```
delandmeterp: This should be a static method

PR #563, opened 2019-03-13
Adding check for duplicate Field name in FieldSet.add_field()

### LINKED ISSUE #562: FieldSet.add_field() should throw error if Field name already exist
Currently, `FieldSet.add_field()` doen's't throw an error when a user tries to add a field with a name that already exists. This can be very confusing, so Parcels should check for this

erikvansebille:
This fixes #562

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -13,7 +13,7 @@ import math
 from .grid import Grid, CGrid, GridCode
 
 
-__all__ = ['Field', 'VectorField', 'SummedField', 'SummedVectorField', 'NestedField']
+__all__ = ['Field', 'VectorField', 'SummedField', 'NestedField']
 
 
 class Field(object):
@@ -939,39 +939,11 @@ class Field(object):
 
     def __add__(self, field):
         if isinstance(self, Field) and isinstance(field, Field):
-            return SummedField([self, field])
+            return SummedField('_SummedField', [self, field])
         elif isinstance(field, SummedField):
+            assert isinstance(self, type(field[0])), 'All fields in a SummedField should either all scalars or  all vectors'
             field.insert(0, self)
             return field
-
-
-class SummedField(list):
-    """Class SummedField is a list of Fields over which Field interpolation
-    is summed. This can e.g. be used when combining multiple flow fields,
-    where the total flow is the sum of all the individual flows.
-    Note that the individual Fields can be on different Grids.
-    Also note that, since SummedFields are lists, the individual Fields can
-    still be queried through their list index (e.g. SummedField[1]).
-    """
-    def eval(self, time, z, y, x, applyConversion=True):
-        tmp = 0
-        for fld in self:
-            tmp += fld.eval(time, z, y, x, applyConversion)
-        return tmp
-
-    def __getitem__(self, key):
-        if isinstance(key, int):
-            return list.__getitem__(self, key)
-        else:
-            return self.eval(*key)
-
-    def __add__(self, field):
-        if isinstance(field, Field):
-            self.append(field)
-        elif isinstance(field, SummedField):
-            for fld in field:
-                self.append(fld)
-        return self
 
 
 class VectorField(object):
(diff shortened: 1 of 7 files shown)


parcels/fieldset.py
```diff
@@ -103,6 +103,8 @@ def add_field(self, field, name=None):
         :param name: Name of the :class:`parcels.field.Field` object to be added
         """
         name = field.name if name is None else name
+        if name in [fld.name for fld in self.fields]:
```
delandmeterp: So, in here, you don't check that all `fld` in `fset.fields` have a different name.
You check that a field that you add with `add_field` has a different name than all `fset.fields`.

That's why it does not conflict with #555 
I don't see why this should crash, so I think it's cool to do it that way. But it would be good to add a comment explaining this

delandmeterp: I can't comment two other lines, since they are not part of the commit, but this PR should also fix some inconsistent ways of adding NestedFields and SummedFields:

In line 114, we add all fields of a NestedField to the fset. So basically, the test you did with same name for `SummedFields` would not pass for `NestedFields`. So I would remove that line.

In function fset.fields (line 500, with property decorator):
`NestedFields` should be added to fields in the same way as `SummedFields`

Also, this function does list all fields in subfields, but it does not list scalar components of vector fields. Is this what we want?

delandmeterp:
Still three points to do:
(1) rename fieldset.fields_TMP (I've renamed it to check easily where it was used). Check if they are used properly?
(2) getting similar structure as `NestedField` for `SummedVectorField` ? This would enable to call something like `SummedVectorField[0][time, depth, lat, lon]`
(3) Better search for correct field in Kernel constructor (see comment above)

parcels/kernel.py
```diff
@@ -115,14 +115,18 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
             self.vector_field_args = kernelgen.vector_field_args
             fieldset = self.fieldset
             for fname in self.vector_field_args:
-                f = getattr(fieldset, fname)
+                fields = fieldset.fields_TMP
+                found = False
+                for f in fields:
+                    if f.name == fname:
```
delandmeterp: I was thinking this was not efficient.
But on top of that, it's wrong, since fields can have same names. so this `vector_field_args` that works using name is not correct. This needs to be changed

parcels/field.py
```diff
@@ -939,41 +939,13 @@ def computeTimeChunk(self, data, tindex):
 
     def __add__(self, field):
         if isinstance(self, Field) and isinstance(field, Field):
-            return SummedField([self, field])
+            return SummedField('_SummedField', [self, field])
         elif isinstance(field, SummedField):
+            assert isinstance(self, type(field[0])), 'All fields in a SummedField should either all scalars or  all vectors'
```
erikvansebille: Should we also check that, if VectorField, they are all either 2D or 3D?

parcels/field.py
```diff
@@ -1253,47 +1225,61 @@ def __getitem__(self, key):
...
+    Note that the individual Fields can be on different Grids.
+    Also note that, since SummedFields are lists, the individual Fields can
+    still be queried through their list index (e.g. SummedField[1]).
+    SummedField is composed of either Fields or VectorFields.
 
-    :param name: Name of the vector field
-    :param U: SummedField defining the zonal component
-    :param V: SummedField defining the meridional component
-    :param W: SummedField defining the vertical component (default: None)
+    :param name: Name of the Summed field
```
erikvansebille: Use 'SummedField' instead of 'Summed field'?

parcels/field.py
```diff
@@ -1253,47 +1225,61 @@ def __getitem__(self, key):
...
+    :param U: List of fields. U can be a scalar Field, a VectorField, or the zonal component of the VectorField
+    :param V: List of fields defining the meridional component (default: None)
+    :param W: List of fields defining the vertical component (default: None)
```
erikvansebille: Somewhat clunky to explicitly call these parameters `U`, `V` and `W`, even though they could also be e.g. temperature? Why not use `Field` as first argument, and deal with the second and third using `kwargs`?

parcels/field.py
```diff
@@ -1253,47 +1225,61 @@ def __getitem__(self, key):
...
-            return self.eval(*key)
+            vals = []
+            val = None
+            for iField in range(len(self)):
+                    val = list.__getitem__(self, iField).eval(*key)
+                    vals.append(val)
+            if isinstance(val, tuple):
+                return tuple(np.sum(vals, 0))
+            else:
+                return np.sum(vals)
```
erikvansebille: These four lines can also be written more pythonic as `return tuple(np.sum(vals, 0))  if isinstance(val, tuple) else np.sum(vals)`?

parcels/field.py
```diff
@@ -1253,47 +1225,61 @@ def __getitem__(self, key):
...
+
+    def __add__(self, field):
+        if isinstance(field, Field):
+            assert isinstance(self[0], type(field)), 'All fields in a SummedField should either all scalars or all vectors'
+            self.append(field)
+        elif isinstance(field, SummedField):
+            assert isinstance(self[0], type(field[0])), 'All fields in a SummedField should either all scalars or  all vectors'
```
erikvansebille: Assert messages here miss a verb ('should be either all')

parcels/field.py
delandmeterp: Usually, kwargs in a function are arguments which were real arguments in another function that calls this function.
They are then neither referenced explicitly, not present in the docstring, no?

I would keep V, W explicitly. I don't really see the advantage of having a hidden kwargs.
But should we call the argument F, V, W ? And in changing this docstring 
```
    :param U: List of fields (order matters). U can be a scalar Field, a VectorField, or the zonal component of the VectorField
```
to 
```
    :param F: List of fields (order matters). F can be a scalar Field, a VectorField, or the zonal component (U) of the VectorField
```

?

delandmeterp: And also add
```
       :param V: List of fields defining the meridional velocity component, if F is the zonal component. (default: None)
    :param W: List of fields defining the vertical velocity component, if F and U are the zonal and meridional components (default: None)
```

erikvansebille: OK I agree with your argument about kwargs. 
But then I would add to the docstring for `V` and `W` that they are only used if `F` is the zonal component of a VectorField

parcels/field.py
```diff
@@ -1305,20 +1291,27 @@ class NestedField(list):
     NestedField is composed of either Fields or VectorFields.
 
     :param name: Name of the Nested field
-    :param U: List of fields (order matters). U can be a scalar Field, a VectorField, or the zonal component of the VectorField
-    :param V: List of fields defining the meridional component (default: None)
-    :param W: List of fields defining the vertical component (default: None)
+    :param F: List of fields (order matters). F can be a scalar Field, a VectorField, or the zonal component (U) of the VectorField
```
erikvansebille: Same docstring for `SummedFields` above?

PR #573, opened 2019-04-03
Pop interpolation

pdnooteboom:
This branch contains code to apply parcels (in 'scipy' and 'jit' mode) to output of the Parallel Ocean Program (POP) model, which uses a staggered B grid. 
Therefore, fieldset.py now contains a 'from_pop' function which calls 'from_b_grid_dataset' to read the B grid. 

Notice that in field.py, a vertical layer of zeros is added to the velocity and tracer fields at the bottom, because the z-dimension length does not match the amount of layers in the data otherwise.

test_grids.py now contains a function which applies the B grid interpolation on simplified B grid fields.


delandmeterp:
I don't think shaved cells are currently correct.  The depths are wrongly interpolated to know in which cell we are. We have to check that

PR #574, opened 2019-04-09
Changing interpolating sanity check to throw FieldSamplingError

### LINKED ISSUE #572: Error in Scipy mode when OutOfBounds in depth dimension.
It seems that in 3D Scipy mode, a particle moving outside of the grid in the depth dimension is not handled correctly, throwing a generic KernelError rather than passing to the recovery kernel as expected.

In the minimal working example below, a 10cell x 10cell x 10cell grid is created, followed by a velocity field with constant flow in the lon and depth directions (and zero flow in the lat direction). I also define a recovery kernel called `DeleteParticle` which simply deletes a particle from its particleset when called.

When a particle `p` crosses the lon limit (i.e. when `p.lon` > 10), it is correctly passed to the recovery kernel and deleted. But when a particle crosses depth limit (`p.depth` > 10), Parcels throws a generic KernelError, rather than calling the recovery kernel.

This issue is not present in JIT mode, where the second particle is deleted as expected.

```
#!/usr/bin/env
from parcels import *

# define recovery kernel
def DeleteParticle(particle, fieldset, time):  # delete particles who run out of bounds.
    print("Particle %d deleted at (%f, %f, %f)" % (particle.id, particle.lon, particle.lat, particle.depth))
    particle.delete()

# create grid
lon = np.linspace(0, 10, 11, dtype=np.float32)
lat = np.linspace(0, 10, 11, dtype=np.float32)
depth = np.linspace(0, 10, 11, dtype=np.float32)
grid = RectilinearZGrid(lon=lon, lat=lat, depth=depth, time=np.zeros(1), mesh='flat')

# create fields
U = Field(name='U', data=np.ones([len(depth), len(lat), len(lon)]), grid=grid,
          allow_time_extrapolation=True)
V = Field(name='V', data=np.zeros([len(depth), len(lat), len(lon)]), grid=grid,
          allow_time_extrapolation=True)
W = Field(name='W', data=np.ones([len(depth), len(lat), len(lon)]), grid=grid,
          allow_time_extrapolation=True)
fieldset = FieldSet(U=U, V=V, fields={'W': W})

# show expected OutOfBounds behaviour when particle crosses lon limit.
particleset_lon = ParticleSet.from_list(fieldset=fieldset, pclass=ScipyParticle,
                                    lon=[9.9],
		                    lat=[5],
		                    depth=[5])
particleset_lon.execute(AdvectionRK4_3D, runtime=1., dt=.5,
                        recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle})

# show erroneous OutOfBounds behaviour when particle crosses depth limit.
particleset_depth = ParticleSet.from_list(fieldset=fieldset, pclass=ScipyParticle,
                                          lon=[5],
		                          lat=[5],
		                          depth=[9.9])
particleset_depth.execute(AdvectionRK4_3D, runtime=1., dt=.25,
                          recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle})
```

erikvansebille:
Also adding unit test (and removing old ‘main’ in test_advection.py to bring in line with all other unit tests)

This fixes #572

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -560,9 +560,8 @@ class Field(object):
             zi = -1
             zeta = 0
 
-        assert(xsi >= 0 and xsi <= 1)
-        assert(eta >= 0 and eta <= 1)
-        assert(zeta >= 0 and zeta <= 1)
+        if not ((0 <= xsi <= 1) and (0 <= eta <= 1) and (0 <= zeta <= 1)):
+            raise FieldSamplingError(x, y, z, field=self)
 
         return (xsi, eta, zeta, xi, yi, zi)
 
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -429,7 +429,7 @@ class Field(object):
         grid = self.grid
         z = np.float32(z)
         if z < grid.depth[0] or z > grid.depth[-1]:
-            return (-999, 0)
+            raise FieldOutOfBoundError(0, 0, z, field=self)
         depth_index = grid.depth <= z
         if z >= grid.depth[-1]:
             zi = len(grid.depth) - 2
@@ -555,8 +555,9 @@ class Field(object):
         if grid.zdim > 1 and not search2D:
             if grid.gtype == GridCode.RectilinearZGrid:
                 # Never passes here, because in this case, we work with scipy
-                (zi, zeta) = self.search_indices_vertical_z(z)
-                if zi == -999:
+                try:
+                    (zi, zeta) = self.search_indices_vertical_z(z)
+                except FieldOutOfBoundError:
                     raise FieldOutOfBoundError(x, y, z, field=self)
             elif grid.gtype == GridCode.RectilinearSGrid:
                 (zi, zeta) = self.search_indices_vertical_s(x, y, z, xi, yi, xsi, eta, ti, time)
@@ -632,8 +633,9 @@ class Field(object):
 
         if grid.zdim > 1 and not search2D:
             if grid.gtype == GridCode.CurvilinearZGrid:
-                (zi, zeta) = self.search_indices_vertical_z(z)
-                if zi == -999:
+                try:
+                    (zi, zeta) = self.search_indices_vertical_z(z)
+                except FieldOutOfBoundError:
                     raise FieldOutOfBoundError(x, y, z, field=self)
             elif grid.gtype == GridCode.CurvilinearSGrid:
                 (zi, zeta) = self.search_indices_vertical_s(x, y, z, xi, yi, xsi, eta, ti, time)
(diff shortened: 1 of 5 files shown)


parcels/tools/error.py
```diff
@@ -1,8 +1,8 @@
-"""Collection of pre-built recovery kernels"""
+"""Collection of pre-buil recovery kernels"""
```
erikvansebille: Why remove the `t` in `built` here?

parcels/kernel.py
```diff
@@ -226,7 +226,7 @@ def execute_python(self, pset, endtime, dt):
                 try:
                     p.dt = sign_dt * dt_pos
                     res = self.pyfunc(p, pset.fieldset, p.time)
-                except FieldSamplingError as fse:
+                except FieldOutOfBoundError as fse:
```
erikvansebille: Also catch `FieldSamplingError` here?

parcels/field.py
```diff
@@ -625,20 +628,21 @@ def search_indices_curvilinear(self, x, y, z, xi, yi, ti=-1, time=-1, search2D=F
             it += 1
             if it > maxIterSearch:
                 print('Correct cell not found after %d iterations' % maxIterSearch)
-                raise FieldSamplingError(x, y, 0, field=self)
+                raise FieldOutOfBoundError(x, y, 0, field=self)
 
         if grid.zdim > 1 and not search2D:
             if grid.gtype == GridCode.CurvilinearZGrid:
                 (zi, zeta) = self.search_indices_vertical_z(z)
+                if zi == -999:
```
erikvansebille: Why do we need this -999 here? Is it because we don't know `x` and `y` in `self.search_indices_vertical_z`? 
If that's the case, would it not be cleaner to add `x` and `y` as arguments to that function too, so we can throw the error there?
Or, alternatively and perhaps more correctly coded, throw a `VerticalSamplingError` in `self.search_indices_vertical_z`, and catch that error here, then throw the `FieldOutOfBoundError`.

parcels/field.py
```diff
@@ -1,7 +1,7 @@
 from parcels.tools.loggers import logger
 from parcels.tools.converters import unitconverters_map, UnitConverter, Geographic, GeographicPolar
 from parcels.tools.converters import TimeConverter
-from parcels.tools.error import FieldSamplingError, TimeExtrapolationError
+from parcels.tools.error import FieldSamplingError, FieldOutOfBoundError, TimeExtrapolationError
```
erikvansebille: Do we need to implement the same separation between SamplingError and OutOfBoundError in `index_search.h` too?

parcels/kernel.py
delandmeterp: It's not needed (at least for now).
FieldSamplingError should not really happen. If it happens, there is a specific message for FieldSamplingError.
If user wants to treat this error, it can do it with the Error recovery kernel (not really used so far, so no need to add a specific one for this error)

parcels/field.py
delandmeterp: See 7ecfc02

erikvansebille:
PR looks good to me know. @delandmeterp, is it ready for merging?

PR #575, opened 2019-04-11
New ErrorThroughSurface error object

erikvansebille:
In many use cases, particles that cross the ocean surface need to be treated differently from particles that corss the lateral boundaries of the domain.
Therefore, this PR adds a new type of Error (`ErrorThroughSurface`) for particles that go OutofBounds at the upper-most layer. 

The advantage is that this allows separate handling of particles leaving the domain through the surface, using e.g. a kernel like
```python
    def SubmergeParticle(particle, fieldset, time):
        particle.depth = 0
        AdvectionRK4(particle, fieldset, time)  # perform a 2D advection
        particle.time = time + particle.dt  # to not trigger kernels again, otherwise infinite loop
```
And then in the `pset.execute()` a 
```python
recovery={ErrorCode.ErrorThroughSurface: SubmergeParticle}
```


parcels/codegenerator.py
```diff
@@ -911,11 +911,10 @@ def generate(self, funcname, field_args, const_args, kernel_ast, c_include):
         body += [pdt_eq_dt_pos]
         body += [c.Assign("res", "%s(&(particles[p]), %s)" % (funcname, fargs_str))]
         body += [c.Assign("particles[p].state", "res")]  # Store return code on particle
-        body += [c.If("res == SUCCESS", c.Block([c.Statement("particles[p].time += sign_dt * __dt"),
-                                                 dt_pos, dt_0_break, c.Statement("continue")]))]
-        body += [c.If("res == REPEAT",
+        body += [c.If("res == SUCCESS | res == DELETE", c.Block([c.Statement("particles[p].time += sign_dt * __dt"),
```
delandmeterp: `|` should be `||`, no?

parcels/field.py
```diff
@@ -428,7 +428,9 @@ def gradient(self, update=False, tindex=None):
     def search_indices_vertical_z(self, z):
         grid = self.grid
         z = np.float32(z)
-        if z < grid.depth[0] or z > grid.depth[-1]:
+        if z < grid.depth[0]:
+            raise FieldOutOfBoundError_Surface(0, 0, z, field=self)
```
delandmeterp: `FieldOutOfBounrError_Surface` should be also in `search_indices_vertical_s` (line 476)

parcels/field.py
```diff
@@ -1,7 +1,7 @@
 from parcels.tools.loggers import logger
 from parcels.tools.converters import unitconverters_map, UnitConverter, Geographic, GeographicPolar
 from parcels.tools.converters import TimeConverter
-from parcels.tools.error import FieldSamplingError, FieldOutOfBoundError, TimeExtrapolationError
+from parcels.tools.error import FieldOutOfBoundError, FieldOutOfBoundError_Surface, FieldSamplingError, TimeExtrapolationError
```
delandmeterp: Shouldn't `Error` be the last word? `FieldSurfaceOutOfBoundError` ?

parcels/tools/error.py
```diff
@@ -43,6 +45,20 @@ def __init__(self, x, y, z, field=None):
...
 
+class FieldOutOfBoundError_Surface(RuntimeError):
+    """Utility error class to propagate out-of-bound field sampling at the surface in Scipy mode"""
+
+    def __init__(self, x, y, z, field=None):
+        self.field = field
+        self.x = x
+        self.y = y
+        self.z = z
+        message = "%s sampled out-of-bound, at (%f, %f, %f)" % (
```
delandmeterp: Edit the message to say it's the surface?

delandmeterp: Maybe not, I don't know

tests/test_advection.py
```diff
@@ -102,22 +102,36 @@ def test_advection_3D(mode, npart=11):
...
-            'W': np.ones((xdim, ydim, zdim), dtype=np.float32)}
+            'W': wfac * np.ones((xdim, ydim, zdim), dtype=np.float32)}
     fieldset = FieldSet.from_data(data, dimensions, mesh='flat')
 
     def DeleteParticle(particle, fieldset, time):
         particle.delete()
 
+    def SubmergeParticle(particle, fieldset, time):
+        particle.depth = 0
+        AdvectionRK4(particle, fieldset, time)  # perform a 2D advection
```
delandmeterp: Explain why in comment? Or not necessary?

PR #576, opened 2019-04-17
Adding lonlatdepth_dtype argument to ParticleSet constructor methods too

erikvansebille:
This follows #552

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -166,9 +166,8 @@ class ParticleSet(object):
                and np.float64 if the interpolation method is 'cgrid_velocity'
         """
 
-        lonlat_type = cls.lonlatdepth_dtype_from_field_interp_method(fieldset.U)
-        lon = np.linspace(start[0], finish[0], size, dtype=lonlat_type)
-        lat = np.linspace(start[1], finish[1], size, dtype=lonlat_type)
+        lon = np.linspace(start[0], finish[0], size)
+        lat = np.linspace(start[1], finish[1], size)
         if type(depth) in [int, float]:
             depth = [depth] * size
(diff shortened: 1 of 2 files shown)


parcels/particleset.py
```diff
@@ -158,17 +161,20 @@ def from_line(cls, fieldset, pclass, start, finish, size, depth=None, time=None,
         :param depth: Optional list of initial depth values for particles. Default is 0m
         :param time: Optional start time value for particles. Default is fieldset.U.time[0]
         :param repeatdt: Optional interval (in seconds) on which to repeat the release of the ParticleSet
+        :param lonlatdepth_dtype: Floating precision for lon, lat, depth particle coordinates.
+               It is either np.float32 or np.float64. Default is np.float32 if fieldset.U.interp_method is 'linear'
+               and np.float64 if the interpolation method is 'cgrid_velocity'
```
delandmeterp: Do we want all those changes?
I thought we could simply set `lonlatdepth_dtype` through the `**kwargs`. 
(but they were missing in from_line)
So both methods are obviously valid, but I don't know what's the most common one. And if the param is not in the argument list, then should it be in the docstring, or only in the docstring of pset constructor? Let's have a look how such parameters usually work)

parcels/particleset.py
```diff
@@ -158,17 +161,20 @@ def from_line(cls, fieldset, pclass, start, finish, size, depth=None, time=None,
...
         """
 
         lonlat_type = cls.lonlatdepth_dtype_from_field_interp_method(fieldset.U)
```
delandmeterp: if `lonlatdepth_type` is defined, it should overwrite the default value of `lonlat_type`

PR #577, opened 2019-04-17
Uv and uvw jit

delandmeterp:
This commit enables to call `fieldset.UV` and `fieldset.UVW` in a same kernel.
Before, vectorFields containing the same components were conflicting in JIT mode


erikvansebille:
Do we need a unit test too for this change? Something that broke before but works with this PR?

PR #578, opened 2019-04-17
Fix pop grid interpolation

pdnooteboom:
This branch fixes a bug in the POP interpolation. Before, no layer of zeros was added to the Bgrid fields in the computeTimeChunk function. 

Moreover, the 'bgrid_tracer' was forgotten in the 2D interpolator (see line 671/675), but this is unrelated to the computeTimeChunk problem.


erikvansebille:
Is it possible to expand the B-grid unit test in `test_grids.py` to also test for the `computeTimeChunk`, for example by putting four time slices in the test files and trigger `deferred_loading=True`?

parcels/field.py
```diff
@@ -53,7 +53,7 @@ class Field(object):
     :param time_periodic: boolean whether to loop periodically over the time component of the Field
            This flag overrides the allow_time_interpolation and sets it to False
     """
-
+
```
erikvansebille: spurious spaces? Can be removed?

tests/test_grids.py
```diff
@@ -559,7 +559,7 @@ def test_popgrid(mode, vert_discretisation):
                  'V': 'V',
                  'W': 'W',
                  'T': 'T'}
-    dimensions = {'lon': 'lon', 'lat': 'lat', 'depth': w_dep}
+    dimensions = {'lon': 'lon', 'lat': 'lat', 'depth': w_dep, 'time': 'time'}
```
erikvansebille: So are we sure that this now tests for `deferred_load=True`? Should we add that as a pytest option (`@pytest.mark.parametrize('deferred_load', [True, False])`)?

parcels/field.py
```diff
@@ -133,6 +133,8 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
         self.loaded_time_indices = []
         self.creation_log = kwargs.pop('creation_log', '')
 
+        self.datadepthsize = kwargs.pop('datadepthsize', None)
```
erikvansebille: Is `datadepthsize` the best name for this variable? Should we explain a bit more what this is for?

tests/test_grids.py
delandmeterp: Yes, I would do that. @pdnooteboom, you can add it as a flag to `from_pop`

parcels/field.py
```diff
@@ -1489,7 +1487,11 @@ def data(self):
             else:
                 data = data[ti, self.indices['lat'], self.indices['lon']]
         else:
-            data = data[ti, self.indices['depth'], self.indices['lat'], self.indices['lon']]
+            if self.indices['depth'][-1] == self.datadepthsize-1 and data.shape[1] == self.datadepthsize-1 and self.interp_method in ['bgrid_velocity', 'bgrid_w_velocity', 'bgrid_tracer']:
+                data = np.concatenate((data[ti, self.indices['depth'][:-1], self.indices['lat'], self.indices['lon']],
+                                       np.zeros((1, len(self.indices['lat']), len(self.indices['lon'])))), axis=0)
```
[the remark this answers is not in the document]
pdnooteboom: The 1 is there for the depth dimension, not for the time dimension.

tests/test_grids.py
```diff
@@ -545,8 +545,9 @@ class MyParticle(ptype[mode]):
 
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
 @pytest.mark.parametrize('vert_discretisation', ['zlevel', 'slevel', 'slevel2'])
-def test_popgrid(mode, vert_discretisation):
-    mesh = path.join(path.dirname(__file__), 'test_data/') + 'POPtestdata_.nc'
+@pytest.mark.parametrize('deferred_load', [True, False])
+def test_popgrid(mode, vert_discretisation, deferred_load):
+    mesh = path.join(path.dirname(__file__), 'test_data/') + 'POPtestdata_time.nc'
```
erikvansebille: Should the `+` here be a `path.join()` too? This may not work on Windows, where directory structure is with `\` instead of `/` (which is why we use `path.join()` instead)

PR #581, opened 2019-04-29
Unpinning appveyor conda

erikvansebille:
This undoes #551, which was needed because of a bug in conda/appveyor that is fixed in the latest conda


erikvansebille:
Well, I think it's better to leave these as separate PRs. But feel free to make a new one!

PR #583, opened 2019-04-29
DO NOT MERGE: Parallel open mp

delandmeterp:


delandmeterp:
TO DO:
* py3_win breaks mid way (even if tick is green)
* random node is compiled even if not necessary.

parcels/include/random.h
```diff
@@ -4,31 +4,52 @@
 extern "C" {
 #endif
 
+#ifdef USE_OPENMP
+#include <omp.h>
+#endif
+#include <gsl/gsl_rng.h>
+
+//#define GSL_RNG_TYPE mt19937
+//#define GSL_RNG_SEED 0x1UL
```
erikvansebille: Why commented?

erikvansebille:
This PR can be closed, as we now have #625?

PR #595, opened 2019-05-21
doc in fieldset.computeTimeChunk()

delandmeterp:


parcels/fieldset.py
```diff
@@ -731,6 +731,13 @@ def advancetime(self, fieldset_new):
             f.advancetime(fnew, advance == 1)
 
     def computeTimeChunk(self, time, dt):
+        """Load a chunk of three data time steps into the FieldSet.
+        This is used when FieldSet uses data imported from netcdf,
+        with default option deferred_load. The loaded time steps are the one immediatly before time
+        and the two following immediatly time if dt is positive (and inversely for negative dt)
+        :param time: Time around which the FieldSet chunks are to be loaded. Time is provided as a double, relatively to Fieldset.time_origin
+        :param dt: time step of the integration scheme
```
erikvansebille: add `(note that only the sign of dt matters for computeTimeChunk)`?

parcels/fieldset.py
```diff
@@ -731,6 +731,13 @@ def advancetime(self, fieldset_new):
             f.advancetime(fnew, advance == 1)
 
     def computeTimeChunk(self, time, dt):
+        """Load a chunk of three data time steps into the FieldSet.
+        This is used when FieldSet uses data imported from netcdf,
+        with default option deferred_load. The loaded time steps are the one immediatly before time
```
erikvansebille: `at or immediately before`?

parcels/fieldset.py
```diff
@@ -731,6 +731,13 @@ def advancetime(self, fieldset_new):
...
+        and the two following immediatly time if dt is positive (and inversely for negative dt)
```
erikvansebille: `and the two time steps immediately following time if`...

parcels/fieldset.py
delandmeterp: I was going to put it, but this is not completely true.
Apart from computing the time chunks, this function also returns `timef`, the last available time (basically, close to `grid.time[2]`, but that also takes into account `time` and `dt`, such that a full simulation does always run with the desired time step, and no cutting of `dt` at the end of the chunk)

So because of this, `dt` is also important, not really when the user uses this function in its code, but in the particleset.execution loop, it is.

Maybe we should also document what is returned by `computeTimeChunk`, but it's of no use outside the time loop. What do you think?

erikvansebille: OK, I agree with you, let's then leave as is. No need to make it even more confusing than it already is ;-)

PR #596, opened 2019-05-22
Adding support for wildcards in Field.from_netcdf()

### LINKED ISSUE #588: Field.from_netcdf() with wild cards
While `FieldSet.from_netcdf()` makes use of wildcards to load a list of files, `Field.from_netcdf()` does not provide such utility, even if it is documented in the docstring.

erikvansebille:
This fixes #588

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -10,7 +10,6 @@ from ctypes import Structure, c_int, c_float, POINTER, pointer
 import xarray as xr
 import datetime
 import math
-from glob import glob
 from .grid import Grid, CGrid, GridCode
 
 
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -138,6 +139,17 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
         # since some datasets do not provide the deeper level of data (which is ignored by the interpolation).
         self.data_full_zdim = kwargs.pop('data_full_zdim', None)
 
+    @classmethod
+    def parse_wildcards(cls, filenames):
+        if isinstance(filenames, str):
+            filenames = sorted(glob(str(filenames)))
+        if len(filenames) == 0:
+            raise IOError("Field files not found: %s" % str(filenames))
+        for fp in filenames:
```
delandmeterp: Is this loop necessary? 
I understand the previous test, but this one isn't already tested later when you try to open the file?

erikvansebille: This is in case a user provides a list with filenames themselves (rather than wildcards) where not all files actually exist. An unlikely situation probably, but still good to check?

delandmeterp: Yes I understood, but such problem should be caught in lines 283,284 (new indices), avoiding to loop twice over the files, no?

Issue #602, opened 2019-06-05
cftime's calendar type is not passed to the particlefile


jbusecke:
I am using three different xarray datasets to run parcels, all of them use cftime but with different calendars.
I noticed that when saving the parcel results to file, the calendar attribute in the file is `cftime`(see example below), which xarray does not understand upon reading. Is there a way to retain the `calendar` attribute of the original xarray dataset, when using `FieldSet.from_xarray_dataset`?

Example output file:
```
$ ncdump -h eddy_drops_auto_strongest_aviso_seedradius-0.05_0.nc
netcdf eddy_drops_auto_strongest_aviso_seedradius-0.05_0 {
dimensions:
	obs = UNLIMITED ; // (29 currently)
	traj = UNLIMITED ; // (8 currently)
variables:
	int trajectory(traj, obs) ;
		trajectory:long_name = "Unique identifier for each particle" ;
		trajectory:cf_role = "trajectory_id" ;
	double time(traj, obs) ;
		time:_FillValue = NaN ;
		time:long_name = "" ;
		time:standard_name = "time" ;
		time:units = "seconds since 1993-01-01 00:00:00" ;
		time:calendar = "cftime" ;
		time:axis = "T" ;
	float lat(traj, obs) ;
		lat:_FillValue = NaNf ;
		lat:long_name = "" ;
		lat:standard_name = "latitude" ;
		lat:units = "degrees_north" ;
		lat:axis = "Y" ;
	float lon(traj, obs) ;
		lon:_FillValue = NaNf ;
		lon:long_name = "" ;
		lon:standard_name = "longitude" ;
		lon:units = "degrees_east" ;
		lon:axis = "X" ;
	float z(traj, obs) ;
		z:_FillValue = NaNf ;
		z:long_name = "" ;
		z:standard_name = "depth" ;
		z:units = "m" ;
		z:positive = "down" ;

// global attributes:
		:feature_type = "trajectory" ;
		:Conventions = "CF-1.6/CF-1.7" ;
		:ncei_template_version = "NCEI_NetCDF_Trajectory_Template_v2.0" ;
		:parcels_version = "2.0.0b2" ;
		:parcels_mesh = "spherical" ;
}
```

erikvansebille:
The calendar of the ParticleFile object is set at https://github.com/OceanParcels/parcels/blob/master/parcels/particlefile.py#L87

What if you set `ParticleSet.pset.time_origin.calendar = 'standard'`? Does that fix your problem?

[...]

PR #604, opened 2019-06-11
Calendar parsing

jbusecke:
I attempted to enable parcels to carry the cftime calendars through to the output (#602).
I added two utility functions `_get_cftime_calendars` and `_get_cftime_datetimes` which return lists of all available calendars and corresponding datetime instances. 
The other changes mostly just implement these two functions and I added some tests. 

I am not very familiar with the code structure, and am happy to rearrange things.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -279,12 +279,11 @@ class Field(object):
                 time[0] = 0
             time_origin = TimeConverter(time[0])
             time = time_origin.reltime(time)
-            
+
             if not np.all((time[1:]-time[:-1]) > 0):
                 id_not_ordered = np.where(time[1:] < time[:-1])[0][0]
-                raise AssertionError('Please make sure your netCDF files are ordered in time. First pair of non-ordered files: %s, %s' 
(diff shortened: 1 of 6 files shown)


willirath:
@jbusecke I've fixed a small typo in an import statement in the tests see if Travis makes it then.  Didn't have a closer look yet, though.

jbusecke:
There was a bug in `test_set_calendar`, but that import statement is curious. My version works for me locally. Did I start from the wrong branch?

Ughh I also just noticed that I messed up the merge. Let me correct that...

willirath:
Looks like it's gone wrong again.  For ca42fbb99114d9de2eeb83103243f9d9f7b8d872 travis jobs seem to do fine up to flake errors: <https://travis-ci.org/OceanParcels/parcels/jobs/544224870#L1619>

jbusecke:
Sorry for the style errors. I usually use black but I wanted to avoid it reformatting the whole file when I save. The flake errors should be fixed now.

PR #605, opened 2019-06-18
WIP: Chunked fields

delandmeterp:
Field.data is now an dask.array, from which only necessary chunked are converted into np.array before C call.

But something is wrong with the dask.array: example_peninsula test breaks, it seems that something is still into memory such that the files related to the fieldset are still attached even after fieldset is completely deleted.
@willirath, if you wanna have a look at the dask construction of the field: Basically, we don't use indices, that are not available in dask, but concatenate the different time steps of the data. Maybe something is wrong there.

So far, the tests are by construction more expensive, since we interpolate everything twice in C lib, and compare that the chunked interpolation is equal to the original one.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -877,6 +877,8 @@ class Field(object):
                 self.data_chunks[block_id] = np.array(self.data.blocks[block])
                 #self.load_chunk[block_id] = 1
 
+        # self.chunk_info format: number of dimensions; typical chunksizes; number of chunks per dimensions;
+        #                         chunksizes (the 0th dim sizes for all chunk of dim[0], then so on for next dims
         self.chunk_info = [[len(nchunks)], list(chunksize), list(nchunks), sum(list(list(ci) for ci in chunks), [])]
         self.chunk_info = sum(self.chunk_info, [])
         #print("chunk_info", self.chunk_info)
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -139,6 +139,8 @@ class Field(object):
         # since some datasets do not provide the deeper level of data (which is ignored by the interpolation).
         self.data_full_zdim = kwargs.pop('data_full_zdim', None)
 
+        self.load_chunk = None
+
     @classmethod
     def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                     mesh='spherical', timestamps=None, allow_time_extrapolation=None, time_periodic=False,
@@ -834,6 +836,62 @@ class Field(object):
     def ccode_convert(self, _, z, y, x):
         return self.units.ccode_to_target(x, y, z)
 
+
+    def find_in_block(self, chunksize, ti, zi, yi, xi):
+        #index = (ti, zi, yi, xi)
+        index = (ti, yi, xi)
+        block = tuple((int(index[i]/chunksize[i]) for i in range(len(index))))
+        local_index = tuple(index[i] - chunksize[i]*block[i] for i in range(len(index)))
+        return block, local_index
+
+
+    def get_block_id(self, nchunks, block):
+        return np.ravel_multi_index(block, nchunks)
+
+
+    def get_block(self, nchunks, bid):
+        return np.unravel_index(bid, nchunks)
+
+
+    def chunk_data(self):
+        #print('Chunking data')
+        # Rechunking
+        self.data = self.data.rechunk(sum(((self.grid.tdim,), self.data.chunksize[1:]), () ))
+
+
+        npartitions = self.data.npartitions
+        chunksize = self.data.chunksize
+        chunks = self.data.chunks
+        nchunks = tuple(len(chunks[i]) for i in range(len(chunks)))
+        self.data_chunks = [None] * npartitions
+        self.c_data_chunks = [None] * npartitions
+
+        if self.load_chunk is None:
+            self.load_chunk = np.zeros(npartitions, dtype=c_int)
+        print(self.load_chunk)
+
+        for block_id in range(len(self.load_chunk)):
+            if self.load_chunk[block_id] == 1:
+                #block, local_index = self.find_in_block(chunksize, 0, 0, 20, 40)
+                #block_id = self.get_block_id(nchunks, block)
+                block = self.get_block(nchunks, block_id)
+                self.data_chunks[block_id] = np.array(self.data.blocks[block])
+                #self.load_chunk[block_id] = 1
+
+        self.chunk_info = [[len(nchunks)], list(chunksize), list(nchunks), sum(list(list(ci) for ci in chunks), [])]
+        self.chunk_info = sum(self.chunk_info, [])
+        #print("chunk_info", self.chunk_info)
+        #print(self.data.chunks)
+        #print(self.data.chunksize)
+        #print("block_id", block, block_id)
+
+
+
+        #To put it as it use to be
+        self.datatmp = np.array(self.data)
+        if not self.datatmp.flags.c_contiguous:
+            self.datatmp = self.data.copy()
+
     @property
     def ctypes_struct(self):
         """Returns a ctypes struct object containing all relevant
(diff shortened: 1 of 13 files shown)


parcels/include/parcels.h
```diff
@@ -94,6 +146,194 @@ static inline ErrorCode spatial_interpolation_tracer_c_grid_3D(int xi, int yi, i
...
+  int ndim = chunk_info[0];
+  if (ndim != 3)
+    exit(-1);
+  int chunksize, i;
+  int shape[ndim];
+  int index[3] = {ti, yi, xi};
+  for(i=0; i<ndim; ++i){
+    shape[i] = chunk_info[ndim+1+i];
+    chunksize = chunk_info[1+i];
+    block[i] = (int) index[i] / chunksize;
```
angus-g: If you apply a periodic halo to the fieldset, you will probably get uneven chunking along a dimension. For example, my `lon` dimension had block sizes of `5, 260, ..., 260, 5`. Calculating the block using the nominal "chunksize" doesn't work in this case. Instead, I put in a loop over the block sizes:
```suggestion
    int chunk_sum = 0, shift = 0;
    // shift index into blocksizes
    for (j = 0; j < i; j++) shift += chunk_info[1+ndim+j];

    for (j = 0; j < shape[i]; j++) {
      chunk_sum += chunk_info[1+2*ndim+shift+j];
      if (index[i] < chunk_sum) {
        chunk_sum -= chunk_info[1+2*ndim+shift+j];
        break;
      }
    }
    block[i] = j;
    index_local[i] = index[i] - chunk_sum;
```

A small loop over blocks is probably not a huge loss of efficiency. Something is still blowing memory usage up significantly, so I can't test further after applying this change...

parcels/field.py
```diff
@@ -826,6 +839,73 @@ def ccode_eval(self, var, t, z, y, x):
...
+                self.data = self.data.rechunk(sum(((self.grid.tdim,), self.data.chunksize[1:]), ()))
+            npartitions = self.data.npartitions
+            chunksize = self.data.chunksize
+            chunks = self.data.chunks
+        else:
+            npartitions = 1
+            chunksize = self.data.shape
+            chunks = tuple((t,) for t in chunksize)
+
+        self.nchunks = tuple(len(chunks[i]) for i in range(len(chunks)))
```
angus-g: In the case of dask arrays:
```suggestion
        self.nchunks = self.data.numblocks
```
and `self.nchunks = (1,) * len(chunksize)` or something to that effect for the non-dask case.

parcels/field.py
```diff
@@ -826,6 +839,73 @@ def ccode_eval(self, var, t, z, y, x):
...
+        # 1: was asked to load
+        # 2: is loaded and was touched last C call
+        # 3: is loaded
+        if isinstance(self.data, da.core.Array):
+            for block_id in range(len(self.grid.load_chunk)):
+                if self.grid.load_chunk[block_id] == 1:
+                    #block, local_index = self.find_in_block(chunksize, 0, 0, 20, 40)
+                    #block_id = self.get_block_id(nchunks, block)
+                    block = self.get_block(block_id)
+                    self.data_chunks[block_id] = np.array(self.data.blocks[block])
```
angus-g: Can't you index `dask.array.blocks` by a scalar anyway to get this same behaviour?

parcels/field.py
```diff
@@ -826,6 +839,73 @@ def ccode_eval(self, var, t, z, y, x):
...
+    def get_block_id(self, block):
+        return np.ravel_multi_index(block, self.nchunks)
+
+    def get_block(self, bid):
+        return np.unravel_index(bid, self.nchunks)
+
+    def chunk_data(self):
+        if isinstance(self.data, da.core.Array):
+            if len(self.data.chunks[0]) > 1:
+                self.data = self.data.rechunk(sum(((self.grid.tdim,), self.data.chunksize[1:]), ()))
```
angus-g: Is there some kind of design documentation for this chunking scheme? I don't really understand the rechunking that goes on, particularly this rechunk to span the full time dimension. Maybe it makes sense given how `computeTimeChunk` operates. Do loaded chunks remain loaded, and are just stepped forward in time? It's a bit hard to piece this all together as a newcomer to the project.

delandmeterp: Hi @angus-g,

First of all, thank you for your constructive comments!
As you've already guessed, I was on holidays and didn't see your posts. I'm now back to business: I will take into account and answer your comments, thanks!

parcels/include/parcels.h
delandmeterp: You are right. I'll clean that such that chunk indexing to take into account halos in particular and non-constant chunksizes in general.

parcels/field.py
delandmeterp: I'll clean the rechunking and document it.

Something to point out already is the fact that we don't chunk the same way in space and time.
Indeed, while we take only the chunks that we need in space and let the rest being lazily open, we can't do that in time, basically because all the time steps are generally written on different files, and we don't want to open all files at once. So we manually control which files to open to have the correct time steps loaded. In space, we use the chunks as developed in this branch.

But, there is something wrong in dask and xarray (see https://github.com/pydata/xarray/issues/3041) such that files that should have been closed remain actually open.
That brings us a lot of problems, and explains while the chunking fails to be quick on fields with many time steps.

I'm now manually closing the files when I'm sure I won't use them anymore (I'll push that work somewhen next week)

parcels/include/parcels.h
delandmeterp: Thanks @angus-g, I'm doing those changes
+ `tdim` is not in the chunks anymore, so we don't look at it
+ `self.data.chunksize` is not used anymore, so it's removed from the `chunk_info` list

parcels/kernel.py
```diff
@@ -207,16 +207,24 @@ def execute_jit(self, pset, endtime, dt):
...
+        #    if not f.data.flags.c_contiguous:
+        #        f.data = f.data.copy()
         for f in self.field_args.values():
-            if not f.data.flags.c_contiguous:
-                f.data = f.data.copy()
+            f.chunk_data()
+
         for g in pset.fieldset.gridset.grids:
+            if len(g.load_chunk) > 0:  # not the case if a field in not called in the kernel
+                g.load_chunk = np.where(g.load_chunk > 0, 3, g.load_chunk)
```
erikvansebille: what's so special about the value '3' here? I don't understand this statement

parcels/kernels/advection.py
```diff
@@ -19,6 +19,7 @@ def AdvectionRK4(particle, fieldset, time):
     (u4, v4) = fieldset.UV[time + particle.dt, particle.depth, lat3, lon3]
     particle.lon += (u1 + 2*u2 + 2*u3 + u4) / 6. * particle.dt
     particle.lat += (v1 + 2*v2 + 2*v3 + v4) / 6. * particle.dt
+    #print("plon plat %g %g" % (particle.lon, particle.lat))
```
erikvansebille: remove debug statement?

tests/test_fieldset.py
```diff
@@ -469,6 +469,7 @@ def test_fieldset_defer_loading_function(zdim, scale_fac, tmpdir, filename='test
     def compute(fieldset):
         # Calculating vertical weighted average
         for f in [fieldset.U, fieldset.V]:
+            f.data = np.array(f.data)
```
erikvansebille: why is it needed to convert `f.data` to numpy array here? Is that always needed in compute functions? Should we make that clear somewhere?

parcels/include/parcels.h
```diff
@@ -529,11 +1146,20 @@ static inline ErrorCode temporal_interpolationUVW_c_grid(type_coord x, type_coor
...
+    err = getCell3D(W, xi[igrid], yi[igrid], zi[igrid], ti[igrid], data3D_W, 1); CHECKERROR(err);
     if (grid->zdim==1){
       return ERROR;
     }
     else{
       err = spatial_interpolation_UVW_c_grid(xsi, eta, zet, xi[igrid], yi[igrid], zi[igrid], ti[igrid], grid, gcode, (float**)(dataU[ti[igrid]]), (float**)(dataV[ti[igrid]]), (float**)(dataW[ti[igrid]]), u, v, w); CHECKERROR(err);
+
+      err = spatial_interpolation_UVW_c_grid2(xsi, eta, zet, xi[igrid], yi[igrid], zi[igrid], ti[igrid], grid, gcode, data3D_U[0], data3D_V[0], data3D_W[0], &up, &vp, &wp); CHECKERROR(err);
+      if ( (fabs(*u-up) > 1e-12) || (fabs(*v-vp) > 1e-12) || (fabs(*w-wp) > 1e-12) ) {
+        printf("error uvw c 1\n");
```
erikvansebille: not sure that this is a clear error message. Or is it leftover from debugging?

erikvansebille:
Can this PR be closed now, since #632 has merged into `master`?

delandmeterp:
As #632 was merged, there is no purpose for this PR and branch anymore

PR #610, opened 2019-06-21
Sgrids and periodic_halo together

delandmeterp:
Previously, the periodic halo was added to the field.data, lon, lat, but
not grid.depth in case of S grids

Watch out: This still does not really work with Deferred 4D s grids (we
don't have that for the moment)

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -140,6 +140,26 @@ class Grid(object):
         dx = np.where(dx > 180, dx-360, dx)
         self.zonal_periodic = sum(dx) > 359.9
 
+    def add_Sdepth_periodic_halo(self, zonal, meridional, halosize):
+        if zonal:
+            if len(self.depth.shape) == 3:
+                self.depth = np.concatenate((self.depth[:, :, -halosize:], self.depth,
+                                             self.depth[:, :, 0:halosize]), axis=len(self.depth.shape) - 1)
+                assert self.depth.shape[2] == self.xdim, "Third dim must be x."
+            else:
+                self.depth = np.concatenate((self.depth[:, :, :, -halosize:], self.depth,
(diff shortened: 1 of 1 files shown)


parcels/grid.py
```diff
@@ -223,6 +223,25 @@ def add_periodic_halo(self, zonal, meridional, halosize=5):
...
         self.lonlat_minmax = np.array([np.nanmin(self.lon), np.nanmax(self.lon), np.nanmin(self.lat), np.nanmax(self.lat)], dtype=np.float32)
+        if isinstance(self, RectilinearSGrid):
+            if zonal:
+                if len(self.depth.shape) == 3:
+                    self.depth = np.concatenate((self.depth[:, :, -halosize:], self.depth,
+                                                 self.depth[:, :, 0:halosize]), axis=len(self.depth.shape)-1)
+                    assert self.depth.shape[2] == self.xdim, "Third dim must be x."
+                else:
+                    self.depth = np.concatenate((self.depth[:, :, :, -halosize:], self.depth,
+                                                 self.depth[:, :, :, 0:halosize]), axis=len(self.depth.shape) - 1)
```
erikvansebille: Spaces around `- 1` here, but not on line 230?

parcels/grid.py
```diff
@@ -352,6 +371,25 @@ def add_periodic_halo(self, zonal, meridional, halosize=5):
             self.xdim = self.lon.shape[1]
             self.ydim = self.lat.shape[0]
             self.meridional_halo = halosize
+        if isinstance(self, CurvilinearSGrid):
```
erikvansebille: Since code for `RectilinearSGrid` and `CurvilinearSGrid` is (almost?) the same, create a method?

PR #622, opened 2019-08-14
Removing appveyor py2.7 check

erikvansebille:
For a few weeks, appveyor fails on python 2.7. I looked into it, and it was partly appveyor.yml, but also a mysterious failed unit-test in RK45 (test_stommel) that only appeared in py27. 

Since it doesn't happen on any other platform or python version, and since we have now moved to python 3 as default for Parcels, I suggest that we suspend python2.7 support on Windows for now?


willirath:
+1 for dropping Py27 for Windows.

PR #629, opened 2019-08-21
Removing support for adding Particles via ParticleSet.add()

erikvansebille:
Since adding `Particles` directly to a `ParticleSet` breaks the `ParticleID` in MPI mode, for now we only support adding of `ParticleSets` to `ParticleSets`. 

Note that it is very easy to create a ParticleSet out of a single `Particle`, so this doesn’t fundamentally change Parcels capabilities.

### DIFF
--- a/tests/test_particle_sets.py
+++ b/tests/test_particle_sets.py
@@ -187,7 +187,8 @@ def test_pset_add_explicit(fieldset, mode, npart=100):
     lat = np.linspace(1, 0, npart)
     pset = ParticleSet(fieldset, lon=[], lat=[], pclass=ptype[mode], lonlatdepth_dtype=np.float64)
     for i in range(npart):
-        particle = ptype[mode](lon=lon[i], lat=lat[i], fieldset=fieldset)
+        particle = ParticleSet(pclass=ptype[mode], lon=lon[i], lat=lat[i],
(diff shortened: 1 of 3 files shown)


tests/test_particle_sets.py
```diff
@@ -187,7 +187,8 @@ def test_pset_add_explicit(fieldset, mode, npart=100):
     lat = np.linspace(1, 0, npart)
     pset = ParticleSet(fieldset, lon=[], lat=[], pclass=ptype[mode], lonlatdepth_dtype=np.float64)
     for i in range(npart):
-        particle = ptype[mode](lon=lon[i], lat=lat[i], fieldset=fieldset)
+        particle = ParticleSet(pclass=ptype[mode], lon=lon[i], lat=lat[i],
+                               fieldset=fieldset, lonlatdepth_dtype=np.float64)
```
delandmeterp: It's the purpose of the test to add particles one by one?
I guess somewhere else, they are already added at once, right?

Because we do this 3 times. Is there still a purpose to this test?

erikvansebille: Well, they each test slightly different things. I'd suggest to keep. We did't test for adding ParticleSets to ParticleSets, so with this change at least we test that functionality

PR #632, opened 2019-08-22
Chunked fields operational

delandmeterp:
This PR splits the field_data into chunks.

`Field.data` is not a `np.array` anymore but a lazily loaded `da.array`.
If a JIT kernel needs to interpolate data within a chunk, this chunk is loaded into memory.
In Scipy, `pset.execute()` first converts all `Field.data` into `np.array`s.

To do
- [x] What to do with `Field.gradient()`
- [x] Cut `Field.chunk_data()` into an initial call and what is called during simu
- [x] Cleaning
- [x] `Field.chunk_data` should be moved to `Field.grid.chunk_data`
- [x] We should catch the following situation:
```
pset.execute(kernel=InterpOnlyU, ...)
pset.execute(kernel=InterpOnlyUandV, ...)
```
where `grid.load_chunk_data` will not be matching with loaded `V` chunks
-> solution here is to attach `load_chunk_data` to the field itself...
     NO, it does not work, since really not efficient
-> we keep it on grid, but in chunk_data, we also load the data also if it's not loaded yet, but then we also need to set chunk_data back to None at the pset.execute() if it't not used by kernels! 

- [x] What to do with data_array interpolated in scipy. it can happen, especially when called out of a pset.execute (at particle.variable init for example). Putting tests for da.array and compute everywhere?
- [x] repeat kernel are correctly interpreted with `pset.execute(dt=0)` ?
- [x] add 3D example plot to documentation notebook


parcels/examples/example_moving_eddies.py
```diff
@@ -161,6 +162,7 @@ def fieldsetfile(mesh):
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
 @pytest.mark.parametrize('mesh', ['flat', 'spherical'])
 def test_moving_eddies_file(mode, mesh):
+    gc.collect()
```
erikvansebille: Why exactly is this necessary? Does it have something to do with pytest?

parcels/examples/example_peninsula.py
```diff
@@ -140,6 +141,7 @@ def fieldsetfile(mesh):
 @pytest.mark.parametrize('mesh', ['flat', 'spherical'])
 def test_peninsula_file(mode, mesh):
     """Open fieldset files and execute"""
+    gc.collect()
```
erikvansebille: Again, why exactly is this `gc` necessary?

parcels/field.py
```diff
@@ -184,6 +196,7 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                fully load them (default: True). It is advised to deferred load the data, since in
                that case Parcels deals with a better memory management during particle set execution.
                deferred_load=False is however sometimes necessary for plotting the fields.
+        :param field_chunksize: size of the chunks in dask loading
```
erikvansebille: Do we need to also copy this docstring to `FieldSet.from_netcdf` and similar methods in `fieldset.py`?

parcels/field.py
```diff
@@ -350,9 +368,15 @@ def reshape(self, data, transpose=False):
             logger.warning_once("Casting field data to np.float32")
             data = data.astype(np.float32)
         if transpose:
-            data = np.transpose(data)
+            if isinstance(data, np.ndarray):
+                data = np.transpose(data)
```
erikvansebille: can we use the same pythonic writing as further down here and below: `lib = np if isinstance(data, np.ndarray) else da`?

parcels/field.py
```diff
@@ -683,6 +707,9 @@ def interpolator2D(self, ti, z, y, x):
                 xsi*(1-eta) * self.data[ti, yi, xi+1] + \
                 xsi*eta * self.data[ti, yi+1, xi+1] + \
                 (1-xsi)*eta * self.data[ti, yi+1, xi]
+            if isinstance(val, da.core.Array):
+                val = val.compute()
+                # raise RuntimeError("This happens at p init from field")
```
erikvansebille: leftover debug comment?

parcels/field.py
```diff
@@ -844,15 +920,24 @@ class CField(Structure):
...
-                        ('data', POINTER(POINTER(c_float))),
+                        ('load_chunk', POINTER(c_int)),
+                        ('data_chunks', POINTER(POINTER(POINTER(c_float)))),
                         ('grid', POINTER(CGrid))]
 
         # Create and populate the c-struct object
         allow_time_extrapolation = 1 if self.allow_time_extrapolation else 0
         time_periodic = 1 if self.time_periodic else 0
+        for i in range(len(self.load_chunk)):
+            assert(self.load_chunk[i] != 1)
```
erikvansebille: This assert needs an error message? When would this happen?

parcels/fieldset.py
```diff
@@ -772,41 +773,80 @@ def computeTimeChunk(self, time, dt):
...
-            else:
-                f.loaded_time_indices = []
-
-            # do built-in computations on data
-            for tind in f.loaded_time_indices:
+                    f.filebuffers[2].dataset.close()
+                    f.filebuffers[1:] = f.filebuffers[:2]
+                    data = f.computeTimeChunk(data, 0)
+                # do built-in computations on data
                 if f._scaling_factor:
```
erikvansebille: Some of the code below is copied from above. Put into a function so that we don't have repetition?

parcels/kernel.py
```diff
@@ -230,6 +235,11 @@ def execute_python(self, pset, endtime, dt):
         # back up variables in case of ErrorCode.Repeat
         p_var_back = {}
 
+        for f in self.fieldset.get_fields():
+            if type(f) in [VectorField, NestedField, SummedField]:
+                continue
+            f.data = np.array(f.data)
```
erikvansebille: I guess that in the future we might also do something smarter with chunking in SciPy mode here? Because now we're just moving the whole field to a numpy array?

parcels/examples/example_moving_eddies.py
delandmeterp: It's somehow related with pytest, but more generally with Python.

In our computeTimeChunk, we are closing our netcdf files as we open new ones. But we always stay with 3 open files. We don't want to close them at the end of pset.execute(), which does not make sense.
We only close them at field deletion (`see new func __del__()`).

While a variable in Python is supposed to leave within its scope (variable defined within a function does not exist elsewhere), but there is no guarantee the variable is really deleted until the end of the full Python program.
-> But we can afford such issue in our pytests, since we need a field from previous test to be deleted to recover access to the nc files which are created. This is where we use the garbage collector (`gc`) that triggers object deletion

delandmeterp: While this makes sense for moving_eddies and peninsula, I still don't know why we need it for the ofam benchmark (see my last commit 1584056524e54410659c8f9b9ab05794cf9a559e).

Since I added that we close the datasets when reseting computeTimeChunk, ofam produces an Time out-of-bound if I don't call the gc. 
This happens Only in pytest and only if I test the other example_* before.
I'm wondering why...

parcels/examples/example_peninsula.py
delandmeterp: See above

parcels/plotting.py
```diff
@@ -169,7 +169,10 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection=None
             data[1] = d
 
         spd = data[0] ** 2 + data[1] ** 2
-        speed = np.sqrt(spd, where=spd > 0)
+        try:
+            speed = np.sqrt(spd, where=spd > 0, out=np.zeros(spd.shape))
```
delandmeterp: @erikvansebille, did you see this strange error?
It was breaking on py3 linux setup.

Anyway, I would suggest to only keep the except part of the fix, which is valid for everyone.

By the way, old code was not filling values where the `where` condition was not met (see [here](https://docs.scipy.org/doc/numpy/reference/generated/numpy.sqrt.html))
So there was some uninitialized values, out argument is important here)

erikvansebille: Yes, I agree to keep only the `speed = np.where(spd > 0, np.sqrt(spd), 0)` statement. Can you change that?

PR #635, opened 2019-09-05
Small fix dt0

erikvansebille:
small fix for a warning that `np.infty * signdt` is not well-defined if `dt = 0`, including some unit tests

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -167,7 +167,7 @@ class Grid(object):
                 assert self.depth.shape[2] == self.ydim, "Third dim must be y."
 
     def computeTimeChunk(self, f, time, signdt):
-        nextTime_loc = np.infty * signdt
+        nextTime_loc = np.infty if signdt >= 0 else -np.infty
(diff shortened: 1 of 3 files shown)


parcels/grid.py
```diff
@@ -167,7 +167,7 @@ def add_Sdepth_periodic_halo(self, zonal, meridional, halosize):
                 assert self.depth.shape[2] == self.ydim, "Third dim must be y."
 
     def computeTimeChunk(self, f, time, signdt):
-        nextTime_loc = np.infty * signdt
+        nextTime_loc = np.infty if signdt >= 0 else -np.infty
```
delandmeterp: 0*inf was nan. It was breaking? I thought that the nan, was just ignored later in the min(next_times) of pset.execute()

erikvansebille:
True it wasn't breaking it was just throwing a warning 
```python
RuntimeWarning: invalid value encountered in multiply 
    nextTime_loc = np.infty * signdt
```
This PR removes that warning

PR #641, opened 2019-09-11
Time periodic neg dt

delandmeterp:
It is important to note that with the current code, first and last time step of the periodic series must be the same:
Time needs to go from 0 to 24h, year from Jan 1st to Jan 1st year+1

So far, Windows is breaking, I guess because in the time series we provide, we have Jan 1st 2002 and 2003, with their corresponding data. And depending if user chooses `defer_load` or not, one or the other time will be chosen.

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -179,7 +179,7 @@ class Grid(object):
                     self.ti += 1
                     self.time = self.time_full[self.ti:self.ti+3]
                     self.update_status = 'updated'
-                elif signdt == -1 and time - periods*(self.time_full[-1]-self.time_full[0]) <= self.time[1] and self.ti > 0:
+                elif signdt == -1 and time - periods*(self.time_full[-1]-self.time_full[0]) < self.time[1] and self.ti > 0:
                     self.ti -= 1
                     self.time = self.time_full[self.ti:self.ti+3]
                     self.update_status = 'updated'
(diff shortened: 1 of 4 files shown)


erikvansebille:
> It is important to note that with the current code, first and last time step of the periodic series must be the same: Time needs to go from 0 to 24h, year from Jan 1st to Jan 1st year+1

Hmm, this is not ideal. I don't think we can expect users to realise this. I know that the solution is absolutely non-trivial, but we should at least open an Issue that tracks and discusses this 'bug'. And shall we also add a warning when users use `time_periodic=True`?

delandmeterp:
Indeed let's put an issue and also a warning for the moment

parcels/grid.py
```diff
@@ -173,20 +173,26 @@ def computeTimeChunk(self, f, time, signdt):
...
                     self.update_status = 'updated'
-                elif signdt == -1 and time - periods*(self.time_full[-1]-self.time_full[0]) <= self.time[1] and self.ti > 0:
+                elif signdt == -1 and time - periods*(self.time_full[-1]-self.time_full[0]) < self.time[1] and self.ti > 0:
                     self.ti -= 1
                     self.time = self.time_full[self.ti:self.ti+3]
                     self.update_status = 'updated'
             if self.ti == -1:
                 self.time = self.time_full
-                self.ti, _ = f.time_index(time)
+                self.ti, _ = f.time_index(time)#, signdt=signdt)
```
erikvansebille: debugging statement after `#`?

PR #647, opened 2019-09-17
Change particle.id

delandmeterp:
Code to have a smooth evolution of the particle.id within and between the particlesets.
Also enables to keep original order of particle for mpi runs

Currently, lastID is a class variable of `_Particle`. Maybe it should be passed to a class variable of `ParticleSet` ?


tests/test_mpirun.py
```diff
@@ -10,8 +10,9 @@
 def test_mpi_run():
     if MPI:
         repeatdt = 200*86400
-        os.system('mpirun -np 2 python parcels/examples/example_stommel.py -p 4 -o StommelMPI.nc -r %d' % repeatdt)
-        os.system('python parcels/examples/example_stommel.py -p 4 -o StommelNoMPI.nc -r %d' % repeatdt)
+        parcels_path = os.path.dirname(__file__) + '/../parcels'
```
erikvansebille: Will this work on Windows with `/` as directory splitter? Use `path.join` instead?

parcels/particleset.py
```diff
@@ -37,10 +37,12 @@ class ParticleSet(object):
     :param lonlatdepth_dtype: Floating precision for lon, lat, depth particle coordinates.
            It is either np.float32 or np.float64. Default is np.float32 if fieldset.U.interp_method is 'linear'
            and np.float64 if the interpolation method is 'cgrid_velocity'
+    :param partitions: List of cores on which distribute the particles for MPI runs. Default: None, in which case particles
```
erikvansebille: 'List of core numbers on which to distribute ...'

tests/test_mpirun.py
delandmeterp: MPI does not work on Windows, so this won't work anyway.
But I'll try to write nicely the paths anyway.. (not sure __file__ works for example on Windows though. For example parcels.__file__ does not work it seems)

PR #650, opened 2019-09-20
Management of grid.load_chunk

delandmeterp:
The idea behind `grid.load_chunk`
A chunk can have 4 different status:
* 0: it is not loaded
* 1: the chunk is requested to be loaded (a C-call should never start with a chunk having this status)
* 2: chunk loaded and at least a particle on the chunk
* 3: chunk loaded, not used

A few important points:
* At `field.computeTimeChunk()`, all chunks with status 3 should be reset to 0 (This was not done) AND the actual chunks, reset to None, to avoid errors (to be done)
* During `kernel.execution()`, particles with status different than `SUCCESS` are rerun. This was affecting the chunks, which could pass from status 2 to 3, because of the multiple calls to JIT kernels for the same time steps. We have changed this by moving the 2->3 status conversion from `kernel.execute_jit()` (called multiple times) to `kernel.execute()` (called once)

### DIFF
--- a/parcels/examples/example_globcurrent.py
+++ b/parcels/examples/example_globcurrent.py
@@ -1,4 +1,4 @@
-from parcels import FieldSet, ParticleSet, ScipyParticle, JITParticle, AdvectionRK4, Variable
+from parcels import FieldSet, ParticleSet, ScipyParticle, JITParticle, AdvectionRK4, Variable, ErrorCode
 from datetime import timedelta as delta
 from os import path
 from glob import glob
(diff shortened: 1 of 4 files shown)


parcels/examples/example_globcurrent.py
```diff
@@ -183,3 +183,37 @@ class MyParticle(ptype[mode]):
...
+
+    pset0.execute(pset0.Kernel(DeleteP0)+AdvectionRK4,
+                  runtime=delta(days=rundays),
+                  dt=delta(minutes=5),
+                  recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle})
+
+    pset1 = ParticleSet(fieldset, pclass=JITParticle,
+                        lon=[25, 25],
+                        lat=[-35, -35],
+                        time=time0)
```
erikvansebille: This unit test should also work without explicitly setting `time`, right? If so, would be good to remove so that that feature is also tested.

delandmeterp: If you don't put explicitly a time, `particle.time` is set to `fieldset.U.grid.time[0]`, but with the defer_loading, `fieldset.U.grid.time` has only 3 time steps, which are moving.

So we can either keep it as it is, or set the `particle.time` to fieldset.U.grid.time_full[0]`, which actually also exists for non defer mode (in which case it is equal to time)

The second option actually makes sense. Otherwise, it's not very clear when do we set the particle...

erikvansebille: Good point. I would say that it's a bug that we don't default to grid.full_time[0] for particle initialisation, as that is what users would expect (they shouldn't have to care about the three time steps in deferred_mode)

Actually, would it make sense to change the naming of `time` to e.g. `_defer_time` (signalling that it's a private attribute) and of `time_full` to `time`? A major refactoring, I know.. Perhaps better for a separate Issue?

PR #652, opened 2019-09-24
Group imports according to pep8 and prevent merge conflicts

J535D165:
Group imports according to pep8 guidelines for better dependency management. https://www.python.org/dev/peps/pep-0008/#imports. The second group consists of external deps.

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -1,12 +1,17 @@
-from parcels.field import Field, VectorField, SummedField, NestedField
-from parcels.tools.loggers import logger
 import ast
-import cgen as c
 import collections
 import math
-import numpy as np
 import random
 from copy import copy
+
+import cgen as c
+import numpy as np
+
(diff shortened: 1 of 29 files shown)


delandmeterp:
Thanks Jonathan! It's good to have some cleaning in the dependencies!

I have a question though: You've cut many multiple import lines into shorter imports, but this does not look necessary in the pep8 doc you've sent:

They wrote that we should NOT do something like:
```
import os, sys
```
but rather
```
import os
import sys
```

BUT that doing:
```
from subprocess import Popen, PIPE
```
was ok. So we could keep all the lines such as `from parcels.field import Field, VectorField, SummedField, NestedField`, no? Or am I missing something?

erikvansebille:
OK, I see the point of separating onto different lines to avoid merge conflicts. Although we've never encountered that before, it's better to be prepared. So I'm happy for this PR to be merged

Issue #648, opened 2019-09-19
`FieldSet.from_netcdf()` timestamps for one NetCDF file containing multiple timesteps


daanreijnders:
NetCDF files don't always have a time axis that is formatted in a way that Parcels can read them. In some cases, the time axis is just an index. Parcels already has the `timestamps` parameter built into `Fieldset.from_netcdf()`, which allows users to pass arrays with `np.datetime64` corresponding to the actual date-times of a snapshot. However, it is currently implemented in such a way that each array element corresponds to one field file which contains a single timestep. Therefore, this does not work for NetCDF files containing multiple timesteps.

Proposed solution: ability to pass a nested arrays with `np.datetime64` objects. Outer index corresponds to individual field files, while the inner index corresponds to time indices within a file. 

It is also useful to implement this for `FieldSet.from_pop`, `FieldSet.from_nemo` etc.

erikvansebille:
This has been fixed in #653, thanks @daanreijnders for finding this bug _and_ suggesting a fix!

PR #653, opened 2019-09-24
Multiple timestamps

daanreijnders:
Solution for issue #648 

`timestamps` must now always be a nested list or array. The outer list indices correspond to each datafile, while the inner list indices correspond to time indices within a NetCDF file. This allows `timestamps` to work for:
- multiple snapshot (i.e. one time index) files: inner array length is 1, outer array length corresponds to number of snapshots
- one NetCDF file with time indices that Parcels cannot readily process: outer array length is 1, inner array length corresponds to number of time indices
- combination of the above


daanreijnders:
@delandmeterp Just saw PR #654 to merge with master. This PR tries to merge with `grid_loaded_chunks`. Should I change that to `master`?

delandmeterp:
When #654 goes to master, your PR should be reattached to master, so nothing to do so far I think

PR #657, opened 2019-09-26
warning if dt is modified during kernel returning SUCCESS

delandmeterp:
This PR captures if a kernel has modified `particle.dt`.
This leads to problems in many situations:
* In our code, `particle.time` will be updated using the `dt` before the kernel call, ignoring the `dt`, which will be used at next kernel call (so one time step later)
* everything called after the `particle.dt` will then be executed with the wrong dt

-> There is no problem if after modifying dt, the kernel returns `REPEAT`, in which case, all variables are reset to their value before the kernel call (except dt) and the kernel is rerun

Problems:
* In current version of the PR, JIT particles call a warning for each occurence of p.dt modification. This can lead to a lot of prints
* the test does not capture the scipy particle warning. Nothing is really tested there
* Should JIT particle return a different error code in case of p.dt modifications, and a true warning_once be printed? There will be only one print, but the C loop will be broken all the time...

Solution? Why do we want to raise a warning? Why don't deleting the particle (or even killing the run?). Modifying dt during kernel execution is dangerous. Some users could use such operation correctly. But they could also do the same job without modifying the dt like this. It's better to completely disable such operation

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -279,7 +279,6 @@ class Kernel(object):
                     res = self.pyfunc(p, pset.fieldset, p.time)
                     if (res is None or res == ErrorCode.Success) and not np.isclose(p.dt, pdt):
                         res = ErrorCode.Repeat
-                    p.update_next_dt()
                 except FieldOutOfBoundError as fse:
                     res = ErrorCode.ErrorOutOfBounds
                     p.exception = fse
@@ -294,7 +293,8 @@ class Kernel(object):
                 # Handle particle time and time loop
                 if res is None or res == ErrorCode.Success:
                     # Update time and repeat
-                    p.time += sign_dt * dt_pos
+                    p.time += p.dt
+                    p.update_next_dt()
                     dt_pos = min(abs(p.dt), abs(endtime - p.time))
                     if dt == 0:
                         break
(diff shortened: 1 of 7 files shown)


delandmeterp:
So we decided that the code should crash if dt is modified without REPEAT error.
This makes sense.

Now, we have a problem with RK45. RK45 does either divide dt by 2 if accuracy is bad, or multiply it by 2 if accuracy really good. We were not repeating the kernel in the last case (which was wrong).

But if RK45 requests to multiply dt by 2, but dt can't be increased since t+dt > tend, them we are stuck in an infinite loop.

delandmeterp:
Solution might be the following:

We wanted to do "if dt is modified, code crashes"
Why not: "if dt is modified*, then kernel.repeat"

by modified*, I mean modified between "before kernel call" and "after kernel call + adjustment due to tend"

I think this would make sense

erikvansebille:
> I think this would make sense

Yes, I agree.

delandmeterp:
Limitation is that if kernel does `particle.dt *=2` then it loops forever, but
(1) it would have crashed anyway with first suggestion
(2) we already have possible infinite loops in current code. So it's a different weakness

I'll implement this then.

parcels/kernel.py
```diff
@@ -275,8 +275,11 @@ def execute_python(self, pset, endtime, dt):
                 for var in ptype.variables:
                     p_var_back[var.name] = getattr(p, var.name)
                 try:
-                    p.dt = sign_dt * dt_pos
+                    pdt = sign_dt * dt_pos
+                    p.dt = pdt
```
erikvansebille: These two lines are somewhat unclear. Would it help to rename `pdt` to something clearer, e.g.
```
p.dt =  sign_dt * dt_pos
pdt_prekernels = p.dt
```
and then check `... not np.isclose(p.dt, pdt_prekernels)` below?

Issue #651, opened 2019-09-20
Renaming grid.time and grid.time_full


erikvansebille:
The naming of `grid.time` and `grid.time_full` is now somehwat confusing, because users would typically use `grid.time` when they need `grid.time_full` (see also https://github.com/OceanParcels/parcels/pull/650#discussion_r326561584)

Suggestion is to change the naming of `time` to e.g. `_defer_time` (signalling that it's a private attribute) and of `time_full` to `time`

CKehl:
Agreed - this may be subject to change in the soon future.

PR #659, opened 2019-10-02
In periodic time series, no need for last time dt

delandmeterp:
`time[-1] = time[0] + period` works
`time[-1] < time[0] + period` also works (last data time step is the
same as first, and is not provided)

We have removed the option `time_periodic=True` that existed before.
It is now either `False` or providing `period`
For backward compatibility, we could still allow `True`, but I prefer to explicitly provide the period, to avoid confusion for the user..

### DIFF
--- a/parcels/examples/example_globcurrent.py
+++ b/parcels/examples/example_globcurrent.py
@@ -21,7 +21,7 @@ ptype = {'scipy': ScipyParticle, 'jit': JITParticle}
 def set_globcurrent_fieldset(filename=None, indices=None, deferred_load=True, use_xarray=False, time_periodic=False, timestamps=None):
     if filename is None:
         filename = path.join(path.dirname(__file__), 'GlobCurrent_example_data',
-                             '20*-GLOBCURRENT-L4-CUReul_hs-ALT_SUM-v02.0-fv01.0.nc')
+                             '2002*-GLOBCURRENT-L4-CUReul_hs-ALT_SUM-v02.0-fv01.0.nc')
     variables = {'U': 'eastward_eulerian_current_velocity', 'V': 'northward_eulerian_current_velocity'}
     if timestamps is None:
         dimensions = {'lat': 'lat', 'lon': 'lon', 'time': 'time'}
@@ -94,7 +94,7 @@ def test_globcurrent_particles(mode, use_xarray):
 def test_globcurrent_time_periodic(mode, rundays):
     sample_var = []
     for deferred_load in [True, False]:
-        fieldset = set_globcurrent_fieldset(time_periodic=True, deferred_load=deferred_load)
+        fieldset = set_globcurrent_fieldset(time_periodic=(86400*365), deferred_load=deferred_load)
 
         class MyParticle(ptype[mode]):
             sample_var = Variable('sample_var', initial=fieldset.U)
--- a/parcels/examples/example_globcurrent.py
+++ b/parcels/examples/example_globcurrent.py
@@ -94,7 +94,7 @@ def test_globcurrent_particles(mode, use_xarray):
 def test_globcurrent_time_periodic(mode, rundays):
     sample_var = []
     for deferred_load in [True, False]:
-        fieldset = set_globcurrent_fieldset(time_periodic=(86400*365), deferred_load=deferred_load)
+        fieldset = set_globcurrent_fieldset(time_periodic=delta(days=365), deferred_load=deferred_load)
 
         class MyParticle(ptype[mode]):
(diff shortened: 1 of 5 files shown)


parcels/examples/example_globcurrent.py
```diff
@@ -94,7 +94,7 @@ def test_globcurrent_particles(mode, use_xarray):
 def test_globcurrent_time_periodic(mode, rundays):
     sample_var = []
     for deferred_load in [True, False]:
-        fieldset = set_globcurrent_fieldset(time_periodic=True, deferred_load=deferred_load)
+        fieldset = set_globcurrent_fieldset(time_periodic=(86400*365), deferred_load=deferred_load)
```
erikvansebille: use `delta(days=365)` here for clarity?

parcels/field.py
```diff
@@ -64,7 +64,7 @@ class Field(object):
     :param interp_method: Method for interpolation. Either 'linear' or 'nearest'
     :param allow_time_extrapolation: boolean whether to allow for extrapolation in time
            (i.e. beyond the last available time snapshot)
-    :param time_periodic: boolean whether to loop periodically over the time component of the Field
+    :param time_periodic: To loop periodically over the time component of the Field. It is either set to False or to the value of the period. The last value of the time series can be provided (which is the same as the initial one) or not (Default: False)
```
erikvansebille: `or the length of the period, in seconds.`

parcels/field.py
```diff
@@ -116,13 +116,18 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
...
-        if self.time_periodic and self.allow_time_extrapolation:
+        if self.time_periodic is not False and self.allow_time_extrapolation:
             logger.warning_once("allow_time_extrapolation and time_periodic cannot be used together.\n \
                                  allow_time_extrapolation is set to False")
             self.allow_time_extrapolation = False
-        if self.time_periodic:
-            logger.warning_once("When using time_periodic=True, it is necessary that the first and last time steps\n \
-                                 of the series are the same, with time[-1] = time[0] + T")
+        if self.time_periodic is True:
+            raise ValueError("Unsupported time_periodic=True. time_periodic must now be either False or the period value")
```
erikvansebille: `or the length of the period in seconds'

parcels/field.py
```diff
@@ -116,13 +116,18 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
...
+        if self.time_periodic is not False and not np.isclose(self.grid.time[-1] - self.grid.time[0], self.time_periodic):
+            if self.grid.time[-1] - self.grid.time[0] > self.time_periodic:
+                raise ValueError("Time series provided is longer than the time_periodic parameter")
+            self.grid._add_last_periodic_data_timestep = True
+            self.grid.time = np.append(self.grid.time, self.grid.time[0] + self.time_periodic)
+            self.grid.time_full = self.grid.time
```
erikvansebille: This shows how confusing `grid.time` and `grid.time_full` are. We really should at some point fix this, see also #651 (although not in this PR)

parcels/field.py
```diff
@@ -351,7 +362,8 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                     else:
                         data_list.append(buffer_data)
                 ti += len(tslice)
-            data = da.concatenate(data_list, axis=0)
+            lib = np if isinstance(data_list[0], np.ndarray) else da
+            data = lib.concatenate(data_list, axis=0)
```
erikvansebille: This was a bug before, that it was always using `da`?

parcels/fieldset.py
```diff
@@ -80,7 +80,7 @@ def from_data(cls, data, dimensions, transpose=False, mesh='spherical',
         :param allow_time_extrapolation: boolean whether to allow for extrapolation
                (i.e. beyond the last available time snapshot)
                Default is False if dimensions includes time, else True
-        :param time_periodic: boolean whether to loop periodically over the time component of the FieldSet
+        :param time_periodic: To loop periodically over the time component of the Field. It is either set to False or to the value of the period (Default: False)
```
erikvansebille: Again, use `length of the period in seconds` (also all the ones below)

parcels/field.py
delandmeterp: Yes indeed!

parcels/field.py
delandmeterp: Indeed I've just found it yesterday. It was not a problem, but basically it was converting np data to dask where it was not necessary (maybe bringing some memory overhead)

PR #660, opened 2019-10-02
4D (time evolving) depth S-grids

delandmeterp:
In this PR, we provide a `Field.from_netcdf` that allows to read vertical grids evolving in time. The `depth` dimension must be `Field.depthfield` where 'depthfield` is the name of the field on which the data of the depth are provided, such that `U.grid.depth= depthfield.data`


parcels/field.py
```diff
@@ -270,7 +274,12 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         if 'depth' in dimensions:
             with NetcdfFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method) as filebuffer:
                 filebuffer.name = filebuffer.parse_name(variable[1])
-                depth = filebuffer.read_depth
+                if dimensions['depth'][:6] == 'Field.':
+                    depth = filebuffer.read_depth_dim
+                    kwargs['depth_field'] = dimensions['depth'][6:]
+                else:
+                    depth = filebuffer.read_depth
+                print('Temporary print rather than error: Time varying depth data cannot be read in netcdf files yet')
```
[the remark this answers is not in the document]
erikvansebille: Also, would this require a tutorial to explain to users how to use this new 4D S-grid functionality?

parcels/field.py
```diff
@@ -270,7 +274,12 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         if 'depth' in dimensions:
             with NetcdfFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method) as filebuffer:
                 filebuffer.name = filebuffer.parse_name(variable[1])
-                depth = filebuffer.read_depth
+                if dimensions['depth'][:6] == 'Field.':
```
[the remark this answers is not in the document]
erikvansebille: Much better with the 'no_set_yet' flag!

delandmeterp:
Some small issue:
Since we work with `dimension['depth'] = 'not_yet_set'`, we can't compare dimensions anymore to check if fields are on the same grid. So the quick check performed in the fieldset does not work anymore and dimensions need to be loaded for every field with `not_yet_set` flag (this wasn't occurring with dimension `Field.XXX`... So we have some cost here, but we  gained a lot in clarity)

cjongedijk:
I seem not authorised to merge the master into the origin/branch but locally I have merged since the update for the memory leak.  There is still the same memory problem as discussed in #703 #711 and #668 and even with the master as it was this morning, I still seem to have it (in 1h I ran out of my RAM). It seemed solved and tested for other users so I was wondering @erikvansebille or @CKehl, could you merge the master in this branch as there are no conflicts (yet) and then if we're all on the same page we could hopefully fix where the memory issue still seems to persist.

erikvansebille:
Hi @cjongedijk. I have just merged `master` into this PR. To be honest, I'm not really sure what the status of this PR is; it was mostly a discussion between you and @delandmeterp, right?

We have done (and are still doing) quite some optimisations of memory usage, so it may be that this updated version fixes your RAM issues. See also http://oceanparcels.org/faq.html#performance

Let us know how it goes!

CKehl:
Ho - so, I also encounter still some problems with long-running simulations, but not due to the memory but because there are file access conflicts, which I intent to resolve soon. Can you confirm (via the procedure in http://oceanparcels.org/faq.html#performance) that your error actually comes from memory overflow ?

CKehl:
Also: I am right now encountering some issues with 4D data myself when compiling the MPI documentation notebook for 4D simulations, so we're working very much on the same issue, I believe.

cjongedijk:
@erikvansebille thanks for the merge. @CKehl are you using this branch to simulate with 4D data? I haven't ran into any problems yet but @delandmeterp and me worked on it indeed but nothing has changd since I have not had any issues and could not really do anything or check any performance differences to the 3D fields with memory running out so fast on my longer sims. I should have flagged earlier that I cannot merge myself but I have locally been merging all the time and since I was understanding being the only user of this branch didnt bother pushing back to the repository. I use 4D fields for all of my simulations, however only the ones where the fields are large I run into memory problems, not for ones that have small fields and run for a longer time. I have monitored my memory over two simulations I did this week, one that has run (and is still running) with small fields and one that ran out of memory in one hour with large fields (similar pset sizes). I will post here in a minute how it looks and we can continue from there.

cjongedijk:
so this is how it looks like for me. I randomly exported the memory during the last few days sorry for the intermittent signal compared to the graphs @CKehl showed in the other threads , working on an automated memory tracker for python processes only but haven't done that yet. approximately 2 days in my slowly aggregating first (light) simulation, I had started one that gives me lots of troubles. I swapped after about 1 hour and then I killed the simulation. I had the long running one on purpose in the background because I saw there was a memory drop earlier whilst simultaneously running sims. This seems to indeed the case around 48 h into the sim where I kill the trouble simulation and the other one continues on a lower memory load.


memory_usage

cjongedijk:
with 'one field size' I mean the size of the netcdf file containing all variables/field information for one time step.

cjongedijk:
updated: after the weekend my 'light' simulation linearly increased the memory uptake as expected, and with netdata I tracked yesterday where my local system was only doing the parcels run with similar memory behaviour as in #703  [https://github.com/OceanParcels/parcels/issues/668#issuecomment-576667920](url) tomorrow morning I'll have to kill it to free up space for other projects. 

Fig 1:
memory_python3
Fig 2: netdata output of my total RAM used for 24h.
netdata_commited_memory_snapshot

Also, following the conversation in #740 today is that testing for _fromnemo specifically or also generally for _fromNetCDF?  It seems like that data format of NEMO (sigma grids 3d+time, C grid) generally is very similar to mine.

parcels/field.py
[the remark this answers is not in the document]
cjongedijk: happy to help creating one. I have plenty of (light) simulations that could be cleaned up and used for it.

CKehl:
Hi Cleo - sorry to come back to you that late. Indeed this memory issue was keeping me up now the last 2 weeks, cause when I fixed it initially, I ran all tests just with 2D+time data. In 3D+time, I ran into massive issues because I wasn't converting the field_chunksize things properly for those cases - they are more 'involved'. Also, there was a separate issue popping up very often when running Erik's Galapagos case - conflicting file access, because with the lazy loading now actually implemented properly, the locking mechanism needs to be managed. As you already saw in #740 , those issue should be resolved now. I tested #740 by running the full galapagos case (the backward simulation) as it is in the repository (https://github.com/OceanParcels/GalapagosBasinPlastic) successfully on the cluster in a job system without crashes, using auto-chunking. Thus, if we can rebase this branch (update it to the current master), then it would be worth testing again. I am also running the python notebook (https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_MPI.ipynb) now on a cluster with the 3D data, regenerating the images. It would be very kind if you were running you simulation with the updates again. Conversly, I'll now browse through your changes in this branch, make a review, and get some inspiration from your changes were I may still need to adapt things.

CKehl:
BTW: if you wanna see how to define `field_chunksize` for 3D fields, have a look on #731 , the file `parcels/examples/documentation_MPI.ipynb`. There, in the second part of the file, you have the function `set_cmems_fieldset_3D(cs)`, where `cs=field_chunksize`. That is for how to do that for general `Field[set].from_netcdf()` access - for NEMO and the others, that step is done automatically (though only tested in detail with NEMO, via the Galapagos case).

parcels/field.py
```diff
@@ -490,15 +501,26 @@ def cell_areas(self):
...
-            raise FieldOutOfBoundError(0, 0, z, field=self)
-        depth_index = grid.depth <= z
-        if z >= grid.depth[-1]:
-            zi = len(grid.depth) - 2
+        if grid.depth[-1] > grid.depth[0]:
+            if z < grid.depth[0]:
+                raise FieldOutOfBoundSurfaceError(0, 0, z, field=self)
+            elif z > grid.depth[-1]:
+                raise FieldOutOfBoundError(0, 0, z, field=self)
+            depth_index = grid.depth <= z
```
CKehl: minor formulation feedback: this function does not return just 1 index, but a mask field for depth indices, so it may be (for comprehension) better to call the variable `depth_indices`. Just a suggestion.

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -157,6 +190,34 @@ def test_pop(mode, chunk_mode):
...
+@pytest.mark.parametrize('mode', ['jit'])
+@pytest.mark.parametrize('chunk_mode', [False, 'auto', 'specific'])
+def test_swash(mode, chunk_mode):
+    if chunk_mode == 'auto':
+        dask.config.set({'array.chunk-size': '32KiB'})
+    else:
+        dask.config.set({'array.chunk-size': '128MiB'})
+    field_set = fieldset_from_swash(chunk_mode)
+    npart = 20
+    # lonp = [i for i in np.arange(start=9.1, stop=11.3, step=0.1)[0:20]]
```
erikvansebille, line 202: Stray comment?

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -157,6 +190,34 @@ def test_pop(mode, chunk_mode):
...
+    lonp = [i for i in 9.5 + (-0.2 + np.random.rand(npart) * 2.0 * 0.2)]
+    # latp = [i for i in 12.7+(-0.25+np.random.rand(npart)*2.0*0.25)]
```
erikvansebille, line 204: another stray debug comment?

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -78,6 +78,32 @@ def fieldset_from_pop_1arcs(chunk_mode):
...
+    dimensions = {'U': {'lon': 'x', 'lat': 'y', 'depth': 'not_yet_set', 'time': 't'},
+                  'V': {'lon': 'x', 'lat': 'y', 'depth': 'not_yet_set', 'time': 't'},
+                  'W': {'lon': 'x', 'lat': 'y', 'depth': 'not_yet_set', 'time': 't'},
+                  'depth': {'lon': 'x', 'lat': 'y', 'depth': 'not_yet_set', 'time': 't'},
+                  'depth_u': {'lon': 'x', 'lat': 'y', 'depth': 'not_yet_set', 'time': 't'}}
+    chs = False
+    if chunk_mode == 'auto':
+        chs = 'auto'
+    elif chunk_mode == 'specific':
+        # chs = {'x': 4, 'j': 4, 'z': 7, 'z_u': 6, 't': 1}
```
erikvansebille, line 98: stray debug comment?

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -78,6 +78,32 @@ def fieldset_from_pop_1arcs(chunk_mode):
     return fieldset
 
 
+def fieldset_from_swash(chunk_mode):
+    filenames = path.join(path.join(path.dirname(__file__), 'SWASH_data'), 'field_*.nc')
+    # filenames = []
```
CKehl, line 83: to-be-removed comment

PR #661, opened 2019-10-02
Creating Particleset.from_particlefile() method

### LINKED ISSUE #640: create ParticleSet.from_particlefile() method for restart functionality
Some users have asked for a 'restart' function to Parcels. Perhaps the easiest way to do this would be to write a `ParticleSet.from_particlefile()` method, that then either takes a netcdf file or a npy temporary file and creates a ParticleSet from the last (by default?) positions in that file

Should be relatively easy to create, and will make many users happy

erikvansebille:
This PR adds a method for creating a `ParticleSet` from a netcdf file as written by `ParticleFile`. This is useful for e.g. restarting an experiment

This PR fixes #640


tests/test_particle_sets.py
```diff
@@ -54,6 +54,27 @@ class MyParticle(ptype[mode]):
...
+def test_pset_create_fromparticlefile(fieldset, mode, tmpdir):
+    filename = tmpdir.join("pset_fromparticlefile.nc")
+    lon = np.linspace(0, 1, 10, dtype=np.float32)
+    lat = np.linspace(1, 0, 10, dtype=np.float32)
+    pset = ParticleSet(fieldset, lon=lon, lat=lat, pclass=ptype[mode])
+    pfile = pset.ParticleFile(filename, outputdt=1)
+
+    def DeleteLast(particle, fieldset, time):
+        if particle.lon == 1.:
+            particle.delete()
```
delandmeterp: So you wanted to check also deleted particles are written, right?

parcels/particleset.py
```diff
@@ -278,6 +279,37 @@ def from_field(cls, fieldset, pclass, start_field, size, mode='monte_carlo', dep
 
         return cls(fieldset=fieldset, pclass=pclass, lon=lon, lat=lat, depth=depth, time=time, lonlatdepth_dtype=lonlatdepth_dtype, repeatdt=repeatdt)
 
+    @classmethod
+    def from_particlefile(cls, fieldset, pclass, filename, repeatdt=None, lonlatdepth_dtype=None):
+        """Initialise the ParticleSet from a ParticleFile
+
+        :param fieldset: :mod:`parcels.fieldset.FieldSet` object from which to sample velocity
+        :param pclass: mod:`parcels.particle.JITParticle` or :mod:`parcels.particle.ScipyParticle`
```
delandmeterp: I think we need a bit more explanation in the docstring. What does the method exactly do?
Like, it looks in the code you take the last time step of the data, so:
* if all particles are started at t=0 and some are deleted at t=3 and other at t=4, we only keep the one at t=4, right?
but if particles are started at t=[0,1,2,3] and are all deleted after 5 days of runtime, you will load all particles.
* Should we only keep particles with `time=max(pfile.variables['time']` (or min if `pfile.variables['time'[0,1] < pfile.variables['time'[0,0]`) ?
* Should we mention that particles not started yet won't be exported and then won't be reloaded, such that `from_particlefile` is not the exact restart of the simulation as it was before quitting (I don't think it should, but important to say it)
* Mentioning that p.id are updated and not the one of the original particles
* Should we add a `time` param such that we load the particles at some time of the simulation? That might be cool actually (with default=None, when we load the last time step)

parcels/examples/example_globcurrent.py
```diff
@@ -225,3 +225,22 @@ def DeleteParticle(particle, fieldset, time):
                   dt=delta(minutes=5))
 
     assert np.allclose([pset0[-1].lon, pset0[-1].lat], [pset1[-1].lon, pset1[-1].lat])
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('dt', [-300, 300])
+def test_globcurrent_pset_fromfile(mode, dt, tmpdir):
+    filename = tmpdir.join("pset_fromparticlefile.nc")
```
delandmeterp: About this `tmpdir.join`, should we make this more systematic in the example_* files? After running `py.test example_*` on my laptop, I end with many *.nc files

erikvansebille: Hmm, that's strange because the whole point of the `tmpdir` is that `pytest` handles this to be a temporary directory (http://doc.pytest.org/en/latest/tmpdir.html). That works for me

tests/test_particle_sets.py
erikvansebille: Yes correct

parcels/particleset.py
erikvansebille: All very good points! You thought this through much more than I did. I will keep working on this implementation then

parcels/examples/example_globcurrent.py
delandmeterp: Yes sure, that's the good approach. What I mean is that other examples don't use this tmpdir. We could do it for all example_* files !

erikvansebille: Sure, feel free to go through and check them. Should probably be a different PR though?

delandmeterp: Yes, I would also say a different PR

PR #663, opened 2019-10-06
Implementing script support for 2d grid interpolation

### LINKED ISSUE #643: ``FieldSet.from_pop`` not working for surface data in SciPy mode.
When loading POP data from one layer (e.g. surface) to perform simulations in 2D, ``FieldSet.from_pop()`` does not work in SciPy mode, while it works for JITParticles. 

In the end, when working in 2D, a B-grid should reduce to an A-grid, so ``FieldSet.from_netcdf()`` does the trick here, but there is no reason why ``FieldSet.from_pop()`` should break.

erikvansebille:
This fixes #643

### DIFF
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -492,6 +492,17 @@ def test_fieldset_from_xarray(maxlatind):
     assert np.allclose(pset[0].lon, 4.5) and np.allclose(pset[0].lat, 10)
 
 
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+def test_fieldset_frompop(mode):
+    filenames = path.join(path.join(path.dirname(__file__), 'test_data'), 'POPtestdata_time.nc')
+    variables = {'U': 'U', 'V': 'V', 'W': 'W', 'T': 'T'}
+    dimensions = {'lon': 'lon', 'lat': 'lat', 'time': 'time'}
+
+    fieldset = FieldSet.from_pop(filenames, variables, dimensions, mesh='flat')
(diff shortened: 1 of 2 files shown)


tests/test_fieldset.py
```diff
@@ -492,6 +492,17 @@ def generate_dataset(xdim, ydim, zdim=1, tdim=1):
     assert np.allclose(pset[0].lon, 4.5) and np.allclose(pset[0].lat, 10)
 
 
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+def test_fieldset_frompop(mode):
+    filenames = path.join(path.join(path.dirname(__file__), 'test_data'), 'POPtestdata_time.nc')
+    variables = {'U': 'U', 'V': 'V', 'W': 'W', 'T': 'T'}
```
delandmeterp: Could you just remove the `W` field?

erikvansebille: Well, whether or not there is a `W` field has nothing to do with 2D or 3D interpolation. That is controlled by the `depth` dimensions

delandmeterp: Yes, it's just that we usually don't put `W` field in 2D. But you are right that it works.
So basically we can choose before testing more (like it is: assessing that `W` does not affect anything) or having something more similar to a real application (that is also easier to read as an example).

delandmeterp: So yes, maybe better to keep it

erikvansebille: So merge?

Issue #664, opened 2019-10-06
Remove py27 from Continuous Integration workflow


erikvansebille:
The (imminent) release of Parcels v2.1.0 will be the last to officially support py27, following the depreciation of python in the wider community at the end of the year (https://pythonclock.org/)

This means that we can remove py27 testing from our CI workflow (Travis and Appveyor).

PR #666, opened 2019-10-09
Fix deprecation warning in field.py

angus-g:
Using `collections.Iterable` instead of `collections.abc.Iterable` was raising the following warning:

    DeprecationWarning: Using or importing the ABCs from 'collections' instead of from
    'collections.abc' is deprecated, and in 3.8 it will stop working


angus-g:
I'm assuming this will fail on Python 2, but there is #664

PR #671, opened 2019-10-15
Add CI with github actions

jelletreep:
I'm Jelle Treep, research engineer at ITS, this draft pull request is created together with Raoul. 
It adds a ci workflow folder to run tests across multiple platforms and python 3 in one github actions job.

The yml file is created using github actions documentation and has not been tested yet.
Python 2 will be added when needed.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -1,52 +1,40 @@
 name: test-suite
-on: [push pull-request]
+on: [push, pull-request]
 jobs:
-  PytestLinux:
-    name: Github Action for testing on Linux
-    runs-on: ubuntu-latest
+  test:
+    name: Github Action for testing across platforms
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [osx, windows-2016, ubuntu-latest]
     env:
       PY_VERSION=3
-      OS_NAME=linux
     steps:
     - uses: actions/checkout@master
-    - name: Install Linux
-      shell: bash
+    - name: Test linux
+      if: matrix.os == 'ubuntu-latest'
       env: 
+        OS_NAME=linux
         MINICONDA_NAME=Miniconda3-latest-Linux-x86_64.sh
-      run: ./.github/scripts/linux_install.sh
-    - name: Github Action for testing on Linux
+      run: |
+        ./.github/scripts/linux_install.sh
+        ./.github/scripts/linux_test.sh
       shell: bash
-      run: ./.github/scripts/linux_test.sh
-  PytestOSX:
-    name: Github Action for testing on OSX
-    runs-on: ubuntu-latest
-    env:
-      PY_VERSION=3
-      OS_NAME=linux
-    steps:
-    - uses: actions/checkout@master
-    - name: Install OSX
-      shell: bash
+    - name: Test OSX
+      if: matrix.os == 'osx'
       env: 
+        OS_NAME=osx
         MINICONDA_NAME=Miniconda3-latest-MacOSX-x86_64.sh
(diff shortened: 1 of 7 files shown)


erikvansebille:
Thanks @jelletreep. So can you also help us what I do need to do on our side to turn this feature on?

jelletreep:
> 
> 
> Thanks @jelletreep. So can you also help us what I do need to do on our side to turn this feature on?

We will try to test it first with a colleague that has a Beta account with github actions, will let you know when I know more

J535D165:
@erikvansebille Any idea why AppVeyor doesn't trigger in this PR? This seems to be independent of the implementation proposed in this PR.

RoelBrouwer:
@jelletreep I have checked the syntax of the workflow. Changed a few small things, but it should work now. I did not check the scripts in /scripts themselves. I believe they can be tested independent of the workflow itself.

A few suggestions:
- The install_miniconda.ps1 script fails, I believe it expects the environmentvariables PYTHON_VERSION, PYTHON_ARCH and PYTHON to be provided, but they are not. I think the scripts for windows have to be looked at still.
- You might want to turn the "Fail-fast" feature off. By default, all jobs abort as soon as one of them fails. I am not sure if that is what you want here. https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
- You are currently using the matrix-strategy, running the same job for three different operating systems. However, you define three completely independent steps (one for each OS), testing for the OS in use. As they are almost completely separate, you could just define three separate jobs (without the matrix-strategy), which saves a bit of clutter in each jobs output (as it does list the skipped steps). But that is a choice.

@erikvansebille You shouldn't have to do anything. As soon as this PR is merged, GitHub should detect the workflow in .github/workflows, and start running it with every push and every PR.

jelletreep:
@RoelBrouwer Thank you Roel! Will look into the first two suggestions next week. 

Regarding the third suggestion. Are there any downsides of separating the work in separate jobs? 
Perhaps it makes sense to keep the matrix-strategy in place if a different version of Python should be included, or would it be best to define 3 jobs with a matrix of python versions?

erikvansebille:
Thanks for all this work @jelletreep. Let me know when this PR comes out of draft status and is ready for merging with `master`.

jelletreep:
@erikvansebille the pull request is ready for review and merging. It implements CI using Github Actions and runs tests on Linux, Windows and MacOS in parallel. Tests run successfully in my branch in 26 minutes (see 
https://github.com/jelletreep/parcels/actions/ ). 

At the moment the CI workflow is triggered only when a commit is pushed to the master branch. This can be changed to all branches or other events in line 2-5 of .github/workflows/ci-workflow.yml . If necessary, I can change it on request. Documentation can be found here: https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows

When this CI workflow satisfies and Github Actions comes out of Beta (13 november), the Travis and Appveyor CI workflows can be removed. I will prepare a separate pull request for this.

.github/workflows/ci-workflow.yml
```diff
@@ -0,0 +1,51 @@
+name: test-suite
+on: 
+  push:
+    branches:
+    - master
```
erikvansebille: I think it's quite important we also do this other branches. And can we do these Actions on external PRs too?

jelletreep: Yes this is fixed now for all branches and pull requests: 

"When you create a pull request from a forked repository to the base repository, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository." : so this will trigger the CI workflow at the base repository

Issue #665, opened 2019-10-07
Particle creation efficiency


angus-g:
I realise #263 sort of touched on this issue, but I think there's still a lot to be gained here. I made a pretty minimal benchmark for doing nothing but creating `JITParticle` objects from a list:

```python
import numpy as np
import parcels

n = 10000
l = np.zeros(n)
fs = parcels.FieldSet.from_data(
    {"U": np.array([[0.0]]), "V": np.array([[0.0]])},
    {"lon": np.array([0.0]), "lat": np.array([0.0])},
)
ps = parcels.ParticleSet.from_list(fs, parcels.JITParticle, l, l)
```

For 10,000 particles, this takes 2.5s (0.25ms/particle). This does seem to scale linearly: 100,000 particle takes 25s. In my opinion however, this is quite slow.

As some context, parcels uses the _array of structures_ method of storing particles (https://en.wikipedia.org/wiki/AoS_and_SoA), which is the most intuitive and natural, but also triggers two big pain points in Python: creating objects, and direct loops. I've been leveraging parcels for some simulations with high particle counts, with frequent initialisation of particle sets. For this, I transposed the particle storage to SoA (https://github.com/angus-g/parcels/commit/5d85fb5efd4d4d568279309f26bf3d837c0f3509). Running the same benchmark on my branch takes 2.6s for 10,000,000 particles (1000x speedup; this would take something like 40 minutes normally).

I don't think this should be taken as advocation for switching the particle storage scheme, rather as motivation for seeking out different available methods. One thought that comes to mind is to use Cython rather than JIT particles: the interpolation routines can all be written in Cython-flavoured Python, particles can be AoS of fast extension types, and the need to generate C code on-the-fly may be removed.

### Initial field sampling

The other big performance killer during particle creation is specifying a variable on a derived particle class whose initial value is sampled from a field. Because this happens on the Python side, it is very expensive. My workaround has been to initialise the variable value to zero, then run a sample-only kernel for zero time. This has the desired effect, utilising the much more efficient interpolation routines, and is scalable to high particle counts. Perhaps this mechanism could be formalised (at least for JIT particles).

### Other possible improvements

[...]

PR #678, opened 2019-10-20
Convert ParticleSet to structure of arrays

angus-g:
This means initialisation of particles is very cheap from Python, and shouldn't be any worse from the C kernel. This gives a massive speedup. See #665 for other discussion.

As part of this change, the semantics of accessing properties on a `ParticleSet` has changed: `pset[idx].property` becomes `pset.property[idx]`. It would be possible to rearrange `__getitem__` to return some kind of accessor, like the `ParticleAccessor` to make this backward-compatible, but I'm not sure of the efficiency implications of such a change. Speaking of which, I created `ParticleAccessor` before implementing `__getattr__` on `ParticleSet`. It might be worth ripping this back out since it's somewhat unneeded.

A few of the tests are (randomly) failing in JIT mode with segfaults. It looks like maybe something isn't being initialised correctly but it's being a little tricky to debug. I thought I'd just put the bulk of the work up here first.

There are a couple of other remaining tasks:
* get `ParticleFile` working again. I hacked this in another branch at a6684db, but that was before the writing to temporary dictionaries, so it'd probably be worth just responding to the few errors that pop up, rather than trying to merge this other change across. I'll do this after dealing with the JIT segfaults.
* test all tutorials/examples (I imagine things like particle plotting probably won't work)


erikvansebille:
Thanks so much for all this development, @angus-g! I'd really like to try give it a spin in a real-life setup to investigate how much faster Parcels becomes!

As you've seen, I did quite some development over the last few weeks. I've today finally fixed the ParticleFile, so that works too

As I can see, there are two open issues:
- [ ] `print(ParticleSet)` is now very ugly
- [ ] there are Segmentation faults. I've now found why they happen: because `particle.xi` is not set to an array the size of the number of `Grids` (we need a `particle.xi` for each grid separately). For some reason, https://github.com/angus-g/parcels/blob/soa-storage/parcels/particle.py#L239-L254 is not called anymore?

The second open issue obviously is much more problematic than the first one. But we're almost there!

angus-g:
Thanks for rolling with this! In response to your second issue, you're correct that individual particles aren't initialised any more: the class is used as a structure to hold information about the required variables for which to create arrays. A lot of the work has been shifted to the particle set instead. I suppose this introduces a level of indirection that may make it more difficult to reason about what's going on.

This seems like the sort of thing that needs to be solved with more metaprogramming, so that we can treat particles as individual (for initialisation and printing), whilst having them stored in the more efficient format.

parcels/particleset.py
```diff
@@ -26,6 +29,51 @@
 __all__ = ['ParticleSet']
 
 
+class ParticleAccessor(object):
```
CKehl: That a really need implementation to make the structural change transparent for users.

.github/workflows/ci-workflow.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         shell: bash
   test_windows:
     name: Testing on Windows
-    runs-on: windows-2016
+    runs-on: windows-2019
```
CKehl: May this be the reason for the test failure on windows ?

parcels/codegenerator.py
```diff
@@ -864,42 +865,43 @@ def __init__(self, fieldset, ptype=None):
...
-            if v.dtype == np.uint64:
-                vdecl.append(c.Pointer(c.POD(np.void, v.name)))
-            else:
-                vdecl.append(c.POD(v.dtype, v.name))
-
+        # Generate type definition for particle struct type
+        vdeclp = [c.Pointer(c.POD(v.dtype, v.name)) for v in self.ptype.variables]
+        ccode += [str(c.Typedef(c.GenerableStruct("", vdeclp, declname=pname)))]
+        # Generate type definition for single particle type
+        vdecl = [c.POD(v.dtype, v.name) for v in self.ptype.variables if v.dtype != np.uint64]
```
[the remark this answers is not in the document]
erikvansebille: Good catch! That might actually also be related the failing CI on Windows. which is also about `uint64` (although then of course it's strange that it does work on macos and linux)

parcels/codegenerator.py
```diff
@@ -864,42 +865,43 @@ def __init__(self, fieldset, ptype=None):
...
-        # Generate type definition for particle type
-        vdecl = []
-        for v in self.ptype.variables:
-            if v.dtype == np.uint64:
-                vdecl.append(c.Pointer(c.POD(np.void, v.name)))
-            else:
-                vdecl.append(c.POD(v.dtype, v.name))
-
+        # Generate type definition for particle struct type
+        vdeclp = [c.Pointer(c.POD(v.dtype, v.name)) for v in self.ptype.variables]
```
CKehl: why is this not `vdeclp = [c.Pointer(c.POD(np.void, v.name)) for v in self.ptype.variables]` ? May that be the issue for Windows failing the JIT tests ? Because on Windows the test that fail are exclusively the JIT tests, it would be a start to look for the error in the codegenerator.

parcels/particleset.py
```diff
@@ -147,37 +194,94 @@ def convert_to_array(var):
...
+                    raise RuntimeError('Particle class does not have Variable %s' % kwvar)
+                self.particle_data[kwvar][:] = kwval
+                initialised.add(kwvar)
+
+            # initialise the rest to their default values
+            for v in self.ptype.variables:
+                if v.name in initialised:
+                    continue
+
+                if isinstance(v.initial, Field):
```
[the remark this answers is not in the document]
erikvansebille: Yes, the idea is that a user can initialise a Particle with the (interpolated) value of the Field at its location. 

However, this is very slow, as the interpolation is done in scipy mode (there ia also a warning if users do this). For backward compatibility, we keep this feature?

.github/workflows/ci-workflow.yml
erikvansebille: I changed this because it was also failing on `windows-2016`. I hoped that `windows-2019` would fix it, but alas...

parcels/codegenerator.py
erikvansebille: I've just checked and changing this line to `vdeclp = [c.Pointer(c.POD(np.void, v.name)) for v in self.ptype.variables]` actually causes all linux JIT tests to fail; so is not the solution

CKehl: okay, I understand. Unfortunate - it's also annoying that the Windows log doesn't really provide a lot of information why all that failed ... @angus-g : do you have a Windows machine running this PR at your workplace ? We here only have MacOS and Linux machines for testing, and if you could provide us with a more detailed error log, we can probably fix it. Otherwise I need to install Windows on some VirtualBox machine here or so, and try to get an error log that way.

parcels/particleset.py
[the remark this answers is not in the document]
CKehl: Well, if it runs and it doesn't break anything (which apparently seems all fine for now), we can surely keep it. It's appropriate to spawn a warning for that, indeed.

parcels/codegenerator.py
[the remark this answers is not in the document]
CKehl: I just recognized that the variables are now named differently, so they don't override one another. That clears up my confusion.

This, this looks a bit confusing, because now, wouldn't we have two fields (instead of one) for each variable, thus being either wasteful or ambiguous for the c-kernel ?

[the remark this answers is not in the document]
erikvansebille: Good point, something to explore

In any case, removing the `if v.dtype != np.uint64` part of the statement did not change anything in the unit tests (see commit 67bf24b)

parcels/codegenerator.py
angus-g: It looks like cgen doesn't think the Windows CI is running 64-bit. Reproducing locally, I get the exception in cgen's `dtype_to_ctype`, which will only map `uint64` if `is_64_bit_platform()`. This runs:

```python
import struct as _struct

def is_64_bit_platform():
    return _struct.calcsize('l') == 8
```

Which I think is an incorrect way to test this! `platform.architecture()` gives `('64bit', 'WindowsPE')` as we'd expect.

I'll raise this upstream with cgen.

erikvansebille: Thanks for exploring this, @angus-g! Any way we could work around the `cgen` Issue for now, while we wait for them to fix this and release a new version?

parcels/codegenerator.py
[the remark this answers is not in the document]
CKehl: so, thanks to @angus-g feedback from cgen, I'm thinking again: would it inappropriate to apply this differentiation also for `uint32` ? most of the other variables are anyway still floating-point metrics, so that wouldn't really confuse it. I'd try that out.

parcels/codegenerator.py
angus-g: As a stop-gap, we could either change the cgen dependency to a different conda channel that incorporates the fix in `environment_py3_win.yml`;

or monkey-patch cgen by modifying `cgen.dtype_to_ctype` when we import it:

```python
import cgen as c

def patched_dtype_to_ctype(dtype):
  ...

c.dtype_to_ctype = patched_dtype_to_ctype
```

Neither solution is particularly clean, but at least it's possible that we'd have meaningful Windows CI

erikvansebille:
Even though we are now using [the fixed version](https://github.com/inducer/cgen/pull/26) of `cgen` (thanks @angus-g!),  the windows CI still breaks. Surprisingly, this seems to be related to printing to stdout. The error we are now getting is
```
./.github/scripts/windows_test.bat : WARNING: Casting field data to np.float32
At D:\a\_temp\cc668354-a74d-4222-8293-0beb2f6f0a93.ps1:4 char:1
+ ./.github/scripts/windows_test.bat
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (WARNING: Castin...a to np.float32:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```
Note that this `WARNING` is the first message printed to stdout. With https://github.com/OceanParcels/parcels/pull/678/commits/d835667589616af060fd9990292aaa7650f225c9, the same error appeared immediately after the `INFO Compiled JITParticleAdvectionRK4` message. 

I don't know enough about Windows why this now breaks, but the really strange thing is that it doesn't happen on `master`, so does seem related to `cgen` again?

angus-g:
That's an odd error, I'm not reproducing that locally. Particularly, all the JIT advection tests are passing for me, which suggests that the fixed `cgen` is working.

I'm only hitting the following error where compiling with `-fPIC` raises a warning on Windows:

```
tests/test_kernel_execution.py::test_execution_keep_cfiles_and_nocompilation_warnings[False] FAILED

================================== FAILURES ===================================
________ test_execution_keep_cfiles_and_nocompilation_warnings[False] _________

fieldset = <parcels.fieldset.FieldSet object at 0x000001EA2709F940>
delete_cfiles = False

    def test_execution_keep_cfiles_and_nocompilation_warnings(fieldset, delete_cfiles):
        pset = ParticleSet(fieldset, pclass=JITParticle, lon=[0.], lat=[0.])
        pset.execute(pset.Kernel(AdvectionRK4, delete_cfiles=delete_cfiles), endtime=1., dt=1.)
        cfile = pset.kernel.src_file
        logfile = pset.kernel.log_file
        del pset.kernel
        if delete_cfiles:
            assert not path.exists(cfile)
        else:
            assert path.exists(cfile)
            with open(logfile) as f:
>               assert 'warning' not in f.read(), 'Compilation WARNING in log file'
E               AssertionError: Compilation WARNING in log file
E               assert 'warning' not in 'C:\\Users\\Angus\\App...c4c_0.c -shared -m64\n'
E                 'warning' is contained here:
E                   C:\Users\Angus\AppData\Local\Temp\parcels-tmp\5fc35a3ed3c047645697155520b62c4c_0.c:1:0: warning: -fPIC ignored for target (all code is position independent)
E                 ?                                                                                         +++++++
E                    #include "parcels.h"
E                    ^
E                   Compiling: gcc -Wall -fPIC -IC:\Miniconda36-x64\envs\parcels\lib\site-packages\parcels-0.1.dev1+g0b49473-py3.6.egg\parcels\include -g -O3 -m64 -o C:\Users\Angus\AppData\Local\Temp\parcels-tmp\5fc35a3ed3c047645697155520b62c4c_0.dll C:\Users\Angus\AppData\Local\Temp\parcels-tmp\5fc35a3ed3c047645697155520b62c4c_0.c -shared -m64

tests\test_kernel_execution.py:280: AssertionError
```

angus-g:
Good call on the revert; it looks like anything written to stderr is considered an error in 2019? Weird!

angus-g:
Just pushed through a little change in the recovery kernel logic: we can handle all the particles with the `Repeat` state at once, which again removes a Python loop over all particles (e.g. in the case of the first execution step with deferred load, where all particles will be `Repeat`). The recovery kernels are still executed in a Python loop, which could be slow if there are many particles with errors.

I'm able to initialise and sample about 7 million particles on 0.1 deg global surface velocities, though the actual advection is letting me down for speed now.

CKehl:
@angus-g : I like the work you are recently doing on optimizing Parcels, and I would appreciate some talk via Slack, Mail or Skype to coordinate our optimization efforts a bit. I am preparing some branches at the moment for clean benchmarking of the source code and, afterwards, iterative hypothesis testing on some (code) optimization procedures and algorithmic changes. That benchmarking is also meant as an optimization guideline. Please contact me if you're in for a conversation, also on your goals and development- or research intents. Cheers, @CKehl

angus-g:
@CKehl no problem, I’m happy to discuss via Slack, preferably

CKehl:
How can I find you on Slack ? Alternatively, you can find me by searching for my full name.

parcels/kernel.py
```diff
@@ -379,22 +379,22 @@ def remove_deleted(pset):
...
+                elif particles.state in recovery_map:
+                    recovery_kernel = recovery_map[particles.state]
+                    particles.state = ErrorCode.Success
+                    recovery_kernel(particles, self.fieldset, particles.time)
+                    if particles.state == ErrorCode.Success:
+                        particles.state = ErrorCode.Evaluate
                 else:
                     logger.warning_once('Deleting particle because of bug in #749 and #737')
-                    p.delete()
+                    particles.delete()
```
CKehl: ah, now I see why in sao my nice state-access functions don't work - they need to be now member functions of the ParticleSet, not the Particle itself. With **delete**, that switch from `Particle` member function to `ParticleSet` member function worked cause there where mutual callers in those classes. We could pretty simply switch back to the member functions rather than directly setting the state, by just introducing the functions `isComputed()`, `reset_state()` and `succeeded()` to the ParticleSet. But it's just a suggestion - if you think it's fine to externally allow access to the state variable, then that's ok.

parcels/kernel.py
```diff
@@ -405,7 +405,7 @@ def remove_deleted(pset):
             else:
                 self.execute_python(pset, endtime, dt)
 
-            error_particles = [p for p in pset.particles if p.state not in [ErrorCode.Success, ErrorCode.Evaluate]]
+            error_particles = np.isin(pset.particle_data['state'], [ErrorCode.Success, ErrorCode.Evaluate], invert=True)
```
CKehl: NICE - that should speed up computation of large particles quite a bit

parcels/particle.py
```diff
@@ -198,15 +201,6 @@ def __repr__(self):
...
 
-    def succeeded(self):
-        self.state = ErrorCode.Success
-
-    def isComputed(self):
-        return self.state == ErrorCode.Success
-
-    def reset_state(self):
-        self.state = ErrorCode.Evaluate
-
```
CKehl: Again, depending on your view of access protection for the `state` variable, we either leave it out as such, or we also introduce those functions in the ParticleSet. Choice is up to you.

parcels/particleset.py
```diff
@@ -31,6 +34,51 @@
...
+    def update_next_dt(self, next_dt=None):
+        if next_dt is None:
+            if not np.isnan(self._next_dt):
+                self.dt, self._next_dt = self._next_dt, np.nan
+        else:
+            self._next_dt = next_dt
+
+    def delete(self):
+        self.state = ErrorCode.Delete
+
```
CKehl: Here would be the place for `isComputed()`, `reset_state()` and `succeeded()`

parcels/kernel.py
erikvansebille: I understand yes. Well, perhaps you can create a separate PR to implement this?

erikvansebille: I've now implemented a general `Particle.set_state(res)` function

angus-g:
Thanks @erikvansebille and @CKehl!

PR #688, opened 2019-11-19
Adding kernel for calculation of polyTEOS10_bsq

erikvansebille:
Adding kernel for calculation of polyTEOS10_bsq, based on Appendix A2 of https://www.sciencedirect.com/science/article/pii/S1463500315000566

Note that this requires `fieldset.abs_salinity` and `fieldset.cons_temperature` `Fields` in the `fieldset` and a `particle.density` `Variable` in the `ParticleSet`

### DIFF
--- a/tests/test_kernel_language.py
+++ b/tests/test_kernel_language.py
@@ -1,4 +1,5 @@
 from parcels import FieldSet, ParticleSet, ScipyParticle, JITParticle, Kernel, Variable
+from parcels.kernels.seawaterdensity import polyTEOS10_bsq
 from parcels import random as parcels_random
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


erikvansebille:
@pdnooteboom, I've now created a `seawaterdensity.py` file in the `kernels` directory. Perhaps we can also add your [UNESCO density kernel](https://github.com/OceanParcels/UNESCOdensity_kernel/blob/master/UNESCO_desnity_kernel.py ) to that file, so that it ships with Parcels?

tests/test_kernel_language.py
```diff
@@ -300,3 +301,70 @@ def modif_dt(particle, fieldset, time):
...
+        density = Variable('density', dtype=np.float32)
+
+    pset = ParticleSet(fieldset, pclass=DensParticle, lon=5, lat=5, depth=1000)
+
+    pset.execute(polyTEOS10_bsq, runtime=0, dt=0)
+    assert np.allclose(pset[0].density, 1022.85377)
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('pressure', ['0', '10'])
```
erikvansebille: What if you use `@pytest.mark.parametrize('pressure', [0, 10])`? Then you shouldn't have to convert between int and float?

PR #690, opened 2019-11-21
Adding timedelta64 calendar support

erikvansebille:
Adding support for `np.timedelta64` calendars in netcdf files


VeckoTheGecko:
Is there somewhere I can read up more on the datetime types in NetCDF files? I'm surprised that `timedelta` would be supported here?

JamiePringle:
netCDF does not support dates or datetime or anything like that -- here is the documentation for supported types: https://docs.unidata.ucar.edu/nug/current/md_types.html 

However, xarray, after parsing the attributes of the data, will by default convert the variables to date formats if it thinks that is appropriate. You can turn the date conversion on or off when you read the data set in. For example, see the `decode_times` and `decode_timedelta` kwargs in `xarray.open_dataset`.

Jamie

PR #710, opened 2020-01-07
Updating link for the Github Actions badge

erikvansebille:


CKehl:
The modification that has been made concerning the image addressing has nothing to do with the CI AppVeyor failure. AppVeyor fails cause it's lacking some Visual Studio project to build - which I don't have an idea right now to fix. I would put this in as a new issue.

erikvansebille:
Thanks for the review, @CKehl! The appveyor shouldn't even be triggered; since merging #682, appveyor and travis aren't used anymore. No idea why it was triggered now..

PR #715, opened 2020-01-18
delete_cfiles argument in Kernel initialisation

erikvansebille:
Adding an argument to `Kernel.__init__` to keep the C files after `Kernel` object is deleted

This aids debugging, as C-files are not removed anymore at end of python run and so can be more easily investigated if something goes wrong in JIT mode

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -61,6 +61,7 @@ class Kernel(object):
 
     :arg fieldset: FieldSet object providing the field information
     :arg ptype: PType object for the kernel particle
+    :param delete_cfiles: Boolean whether to delete the C-files after compilation in JIT mode (default is True)
 
     Note: A Kernel is either created from a compiled <function ...> object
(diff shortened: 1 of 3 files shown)


parcels/kernel.py
```diff
@@ -69,10 +70,11 @@ class Kernel(object):
     """
 
     def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
-                 funccode=None, py_ast=None, funcvars=None, c_include=""):
+                 funccode=None, py_ast=None, funcvars=None, c_include="", delete_cfiles=False):
```
CKehl: delete_cfiles should be True

erikvansebille:
To be clear, this is mostly a debugging feature; it is during debugging sometimes really annoying that C-files are deleted at the end of the simulation. 
But for normal use, we indeed want these files to be deleted. Hence `delete_cfiles` default is `True`

PR #718, opened 2020-01-27
adding a new branch to fix a bug in the timestamp

pdnooteboom:
When using timestamps, parcels should not use the 'assert(len(filenames[k]) == len(timestamps))' for the 'lon', 'lat' and 'depth.' if filenames is a dictionary, because those are not the length of the timestamps.


parcels/field.py
```diff
@@ -230,7 +230,8 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                 assert len(filenames) == len(timestamps), 'Outer dimension of timestamps should correspond to number of files.'
             elif isinstance(filenames, dict):
                 for k in filenames.keys():
-                    assert(len(filenames[k]) == len(timestamps)), 'Outer dimension of timestamps should correspond to number of files.'
+                    if(k not in ['lat','lon','depth']):
```
erikvansebille: No need for parentheses around `if` clause

erikvansebille: And what about `time`? Should that also be exempt here?

Issue #711, opened 2020-01-08
Issue with memory exceeding limit


claudiofgcardoso:
Hello all,

I am having a similar issue as #703. When using the beta 2.0.0 version of Parcels I had no problem running a 2D simulation with 10 years of daily Mercator (nested grid with 0.083 and 0.25º) files and 15 particles being released daily.  Now, using parcels 2.1.2 (no MPI) on a single node of 120Gb it raises the following error: 

"(...)
 26% (49852800.0 of 189043200.0) |#      | Elapsed Time: 0:27:24 ETA:   1:52:44
 26% (49939200.0 of 189043200.0) |#      | Elapsed Time: 0:27:28 ETA:   2:00:27
/var/log/slurm/spool_slurmd/job3571814/slurm_script: line 16: 13617 Killed                  python run_MAC_NEMO.py -stype backward -pnumber 2 > run_MAC_NEMO_backward.log
Wed Jan  8 16:53:03 CET 2020
slurmstepd: error: Exceeded step memory limit at some point."

First I thought it was related with the fact I was now running a simulation with 300 particles released at a 5 day interval in a HPC. When I decreased the number of particles to 12 the issue persisted, always when the simulation was at ~26%. So I guess the problem isn't related with the number of particles. 

I also ran the original sim. (15 particles in a normal linux laptop with parcels 2.1.2) and the process stops at 8% because of lack of memory.  deferred_load is set as default (i.e. True), so I don't understand what is going on..

[MAC_NEMO.txt](https://github.com/OceanParcels/parcels/files/4036168/MAC_NEMO.txt)
[run_MAC_NEMO.txt](https://github.com/OceanParcels/parcels/files/4036169/run_MAC_NEMO.txt)

Any help and suggestion is greatly appreciated!  
Cláudio

claudiofgcardoso:
Forgot to mention that after all I tried several thing:
1) created a new python enviromnent with Parcels v2.1.1 and the problem persisted;
2) removed parcels v2.1.1 from the newly created enviroment and installed the v2.0.0, problem persisted;
3) then I removed the newly created enviromnent and created a new one, but this time with parcels v2.0.0 and it worked! 

So maybe the problem is related with a dependency...?

erikvansebille:
Thank you for letting us know, @claudiofgcardoso. This indeed appears similar to #703 and perhaps also #668? 

[...]

Issue #703, opened 2019-12-18
memory problem loading long list of daily model output


sjvg:
Hello,

I'm fairly new to parcels, so apologies if such error has been dealt before, but I could not see it amongst the previous issues raised.
I have a problem with parcels when loading a large number of daily output of an ORCA12 run (arrays of size (4032,3059, 50)), when doing a simple 2D advection experiment (7000 particles, 20 timesteps).
The experiment works fine when loading 30 daily u,v fields, but crashes when I choose to load 100 daily u,v fields instead. I thought that the option deferred_load in=True, in: 
fieldset = FieldSet.from_netcdf(filenames, variables, dimensions,deferred_load=True)
would avoid this memory problem.
Do you know why this happens? 
For info, I use the normal parcels python environment (no MPI) on a single node of 64Gb.

Many thanks in advance for your help.

erikvansebille:
Hi @sjvg, thanks for logging this issue. I'm not sure what is happening here, not do I have any idea how to fix it. For us to tackle this, we would need to be able to at least reproduce the Issue on our systems. We do us ORCA12 fields here too, so perhaps we could try if your code also breaks on our system? 

Could you share a minimal example of a breaking code?

sjvg:
Hi Erik, thanks for you reply, please see code attached. All I change is the number of files fed to fieldset in :  'data': data_path + 'U00??.nc'. 
When I load 30 files, no problem, when I load 100, I get:
INFO: Compiled JITParticleAdvectionRK4periodicBC ==> /tmp/parcels-985/bd35fb50f54210ceb82df350dbb549fb_0.so
INFO: Temporary output files are stored in /home/PARCELS/WORK/out-IPPOUKYX.
INFO: You can use "parcels_convert_npydir_to_netcdf /home/PARCELS/WORK/out-IPPOUKYX" to convert these to a NetCDF file during the run.
N/A% (0 of 1728000.0) |                                                                                                                                                     | Elapsed Time: 0:00:00 ETA:  --:--:--
 10% (172800.0 of 1728000.0) |##############                                                                                                                                | Elapsed Time: 0:01:27 ETA:   0:13:11
 15% (259200.0 of 1728000.0) |#####################                                                                                                                         | Elapsed Time: 0:02:55 ETA:   0:24:47
 20% (345600.0 of 1728000.0) |############################                                                                                                                  | Elapsed Time: 0:04:20 ETA:   0:22:36
[...]

Issue #668, opened 2019-10-09
Memory issue due to flow field data


VictorOnink:
I've just updated python to version 3.7, and therefore I reinstalled parcels with the most up to date version (v2.1 that was just released). However, with this new version of parcels I find that my runs crash due to a memory error. Specifically, the error message reads:

`MemoryError: Unable to allocate array with shape (1, 3251, 4500) and data type float32`

The shape of the array is the same as that of one time step of the HYCOM surface circulation data that I have been using, which makes me suspect that the issue is related to releasing the files from memory once the simulation has passed them. While running a simulation I have also followed the working memory being used by the script, which increases until it is so high that the program crashes. Furthermore, when I run the script using only one flow field data file, then it runs without issue.

I've come across the same issue with parcels installed both on a linux and a windows machine, while it didn't come up when i installed parcels v2.0, so it likely lies somewhere in the more recent additions to the parcels code. It came up already last week, so Philippe already has the code and datafiles that I used that led to the issue on my machines.

delandmeterp:
Hi Victor, could you provide the exact file you are running?
We've run your file last week with `totalKernal=AdvectionRK4+diffusion+Beach
`, and it was going fine on Macos. I've run it again now, and on gemini as well, with similar fine result (see below)




![linux_memory](https://user-images.githubusercontent.com/31093031/66493241-00cea180-eab6-11e9-8349-8d33826eebbe.png)
![macos_memory](https://user-images.githubusercontent.com/31093031/66493242-01673800-eab6-11e9-81a1-194c1fa7e487.png)

VictorOnink:
The file I used is the exact same as the one last week, with the only difference being that the paths to the files were changed due to it being run on a different computer. But that is odd then that the error doesn't come up on your servers. I've now been using v2.0 and that works fine without any errors, but then I'm not sure what the issue is with v2.1. 

Since posting the issue I've been working with my code so quite a few of the kernels have changed somewhat. I'll try reinstalling v2.1 at some later point and I'll see if a similar error comes up again. It might just be an issue specific to my computer or due to some part of a kernel.

CKehl:
[...]

PR #719, opened 2020-01-27
Fix chunking mem leak

CKehl:
fixing the recent memory leaks.

Tested locally extensively. Ran and fixed PEP8 problems. Getting locally errors running tests with other kinds of datasets (help welcome).

Related Issues:
https://github.com/OceanParcels/parcels/issues/703
https://github.com/OceanParcels/parcels/issues/711
https://github.com/OceanParcels/parcels/issues/668


parcels/grid.py
```diff
@@ -199,17 +201,24 @@ def computeTimeChunk(self, f, time, signdt):
...
                     self.ti -= 1
                 if self.ti >= len(self.time_full) - 2:
                     self.ti = len(self.time_full) - 3
+
                 self.time = self.time_full[self.ti:self.ti+3]
                 self.tdim = 3
-                self.update_status = 'first_updated'
+                if prev_time_indices is None or len(prev_time_indices) != 3 or len(prev_time_indices) != len(self.time):
+                    self.update_status = 'first_updated'
+                elif functools.reduce(lambda i, j: i and j, map(lambda m, k: m == k, self.time, prev_time_indices), True) and len(prev_time_indices) == len(self.time):
```
erikvansebille: `lambda` functions are not very easy to read (we only have one other in Parcels); I thought that in python3 they were slowly being phased out and replaced by e.g. lost comprehension?

parcels/scripts/get_examples.py
```diff
@@ -128,7 +128,7 @@ def main(target_path=None):
         args = parser.parse_args()
         target_path = args.target_path
 
-    if os.path.exists(target_path):
+    if os.path.exists(os.path.join(target_path, "MovingEddies_data")):
```
erikvansebille: Why only check for one of the paths? is that intentional?

parcels/examples/example_performanceProfiling.py
```diff
@@ -0,0 +1,288 @@
+from parcels import FieldSet, ParticleSet, JITParticle, AdvectionRK4
```
erikvansebille: Very nice and elaborate function, but I'm not sure this should go into  the `examples` directory, which is intended for new users to get a feel for how parcels can be used. This might shock them, ;-) 

Should we create e.g. a `performance` directory? Or move this function to a separate repository?

parcels/grid.py
CKehl: this lambda function was, after google browse, the best and quicked option to check if the indices in both python arrays (prev_time_indices, self.time) are the same at the same position - also the quickest, cause all other versions depend on a `for i in range(len(self.time))` loop, which should never be done in performance-critical applications. I could rewrite all those simple loops with `itertools.islice()` to increase performance, but at this point it seemed the simplest and quickest option.

Generally, I do not like lambda expressions either cause they sometimes hide function calls as being variables. It's just that in this case, it makes sense due to performance.

parcels/scripts/get_examples.py
CKehl: what was before checked if the given head path exists. Hence, on my machine and on GEMINI, I was not able to say `parcels_get_examples examples/` because `examples` already existed as a folder. This way, we only check if the actual example data (e.g. the folder of *MovingEddies_data*) have already been pushed into the folder.

Yes, we can be more elaborate, checking for all example data, and then pick up aborted downloads. That seemed in this case just a bit too much - I just wanted to be able to make the tests run the way they are descriped by the linux pipeline.

parcels/examples/example_performanceProfiling.py
CKehl: Sure, we can - that's up to you. @delandmeterp also added his performance analysis on the chunksize as a jupyter notebook to the examples as `documentation_MPI,ipynb`, so I thought this may be a good place to add it. I'm open for suggestions what to do.

parcels/grid.py
erikvansebille: OK, is fine then

parcels/scripts/get_examples.py
erikvansebille: No, is fine like this then

CKehl:
> Do we also need to provide info for how to` setup` up the `dask.yaml` file, e.g. in the `documentation_MPI,ipynb` and/or in the installation instructions at oceanparcels.org?

Added issue for this (https://github.com/OceanParcels/parcels/issues/720)

parcels/examples/example_performanceProfiling.py
delandmeterp: https://github.com/OceanParcels/parcels_efficiency ?

erikvansebille: The disadvantage of a separate repository is that it will be hard to find there and may become obscure. 

I suggest to move this function to a new directory `performance`, directly in the root (so not within the `parcels` sub-dir)

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+
+        init_chunk_dict = {}
+        if isinstance(self.field_chunksize, dict):
+            init_chunk_dict = self.field_chunksize
+        elif isinstance(self.field_chunksize, tuple):
+            spatStart = 0
+            if 'time' in self.dimensions and self.dimensions['time'] in self.dataset.dims:
+                if(len(self.dimensions) > 2):
+                    init_chunk_dict[self.dimensions['time']] = self.field_chunksize[0]
+                    spatStart += 1
```
erikvansebille: what does `spatStart` mean? From its name, I can't guess its use. Expand the name?

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+                else:
+                    init_chunk_dict[self.dimensions['time']] = 1
+            else:
+                raise RuntimeWarning("Did not find {} in NetCDF dims - Please specify field_chunksize as dictionary for NetCDF dimension names.".format(self.dimensions['depth']))
```
erikvansebille: Do we need a bit more guidance _how_ to specify this `field_chuncksize`? E.g. link to a tutorial or FAQ (we can do that later, once we created the tutorial)?

erikvansebille: Should be `format(self.dimensions['time'])` in this clause (depth is dealt with later)?

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+            if 'depth' in self.dimensions and self.dimensions['depth'] in self.dataset.dims:
+                if(len(self.dimensions) > 3):
+                    init_chunk_dict[self.dimensions['depth']] = self.field_chunksize[1]
+                    spatStart += 1
+                else:
+                    init_chunk_dict[self.dimensions['depth']] = 1
+            else:
+                raise RuntimeWarning("Did not find {} in NetCDF dims - Please specify field_chunksize as dictionary for NetCDF dimension names.".format(self.dimensions['depth']))
```
erikvansebille: Make this a function/method so that we can more easily repeat it for each dimension?

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+                init_chunk_dict[self.dimensions['lat']] = self.field_chunksize[spatStart]
+            else:
+                raise RuntimeWarning("Did not find {} in NetCDF dims. Please specifiy field_chunksize as dictionary for NetCDF dimension names.".format(self.dimensions['lat']))
+            if 'lon' in self.dimensions and self.dimensions['lon'] in self.dataset.dims:
+                init_chunk_dict[self.dimensions['lon']] = self.field_chunksize[spatStart+1]
+            else:
+                raise RuntimeWarning("Did not find {} in NetCDF dims. Please specifiy field_chunksize as dictionary for NetCDF dimension names.".format(self.dimensions['lon']))
+        elif self.field_chunksize == 'auto':
+            av_mem = psutil.virtual_memory().available
+            chunk_cap = av_mem * (1/8) * (1/3)
```
erikvansebille: Perhaps give a (very short) explanation of why the multiplication factors 1/3 and 1/8?

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+            if 'array.chunk-size' in da_conf.config.keys():
```
erikvansebille: Do we need to throw a warning/info message to notify users that they can improve performance by creating a dask yaml file? Perhaps by adding a link to the website (e.g. http://oceanparcels.org/faq.html#performance; to be created)

parcels/field.py
```diff
@@ -1545,18 +1553,76 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
+                chunk_cap = da_utils.parse_bytes(da_conf.config.get('array.chunk-size'))
+            if ('lat' in self.dimensions and self.dimensions['lat'] in self.dataset.dims) and ('lon' in self.dimensions and self.dimensions['lon'] in self.dataset.dims):
```
erikvansebille: We seem to use this `(dim in self.dimensions and self.dimensions[dim] in self.dataset.dims)` clause a lot. Make it a method/function?

parcels/particleset.py
```diff
@@ -528,6 +528,10 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
             if abs(time-next_movie) < tol:
                 self.show(field=movie_background_field, show_time=time, animation=True)
                 next_movie += moviedt * np.sign(dt)
+            # ==== insert post-process here to also allow for memory clean-up via external func ==== #
+            if postIterationFunctions is not None:
+                for extFunc in postIterationFunctions:
+                    extFunc()
```
erikvansebille: This is a new feature that's now buried in this larger PR; would it make sense to separate it out? And then also add a unit test so that we test its working?

More fundamentally, I'm not sure this feature is even necessary. In principle, a user could do the same with a for loop around the `pset.execute()` and setting `runtime` and `dt` the same value. Then, the execute would only do one kernel call each time. Admittedly, this adds a bit of overhead on the start of `pset.execute()`, but that's probably fairly limited.

So aren't we over-engineering by adding this feature?

parcels/field.py
CKehl: yes - I'll rename it to 'spatial_start_index'

parcels/field.py
CKehl: I can make a little example there for each case. And when we updated the documentation (https://github.com/OceanParcels/parcels/issues/720), then we should also update these messages.

parcels/field.py
CKehl: agreed

parcels/field.py
CKehl: Okay, so estimating this is anyway quite difficult, but: 1/3 => we always have 3 timesteps in memory to take care of.; 1/8 = makes sure you can at least fit 8 different fields into memory. Here, psutil.virtual_memory().available refers to the overall physically available space (psutil is a bit misleading there), so if one has 8 GB, that would return ~333MB, which is an appropriate upper bound, having in mind that we also need to fit particles in that memory. On clusters, this number is absurdly large - not really a boundary.

On the other hand, I want to remind you that this is just for initialisation in cases where no other information on the data is available. Also, it will be auto-rechunked anyway by dask on actual data access, internally.

parcels/field.py
CKehl: yes, I can add that.

parcels/field.py
CKehl: yes, that will be done.

parcels/particleset.py
CKehl: a) put it in there cause otherwise the whole performance measuring script doesn't work, as we can't obtain performance metrics otherwise at each iteration
b) yes, agreed - it may be a bit of its own PR, but again: I need that function for the example script anyway.
c) it's not useless - see #723 
d) the feature is necessary, cause in the for loop, you also then measure the setup time for the ParticleSet.__init__ (considerable), and the first-kernel start-up including the JIT compilation of the kernels - in essence: the measurements are deeply flawed, at would not survive CS paper peer review. This way, the first measurement taken (the blue spike at the start of the graphs) includes the whole setup procedure (JIT compilation etc.), and the other measurements are really the per-iteration measurements without overhead.

parcels/field.py
CKehl: yes, typo fixed, thanks.

parcels/field.py
CKehl: acutally, I reworked this part for now being 'chunk_index'

parcels/field.py
CKehl: I made the error message now being generated by a function. Unfortunately, reworking the whole code section into a simple loop doesn't work because indices for time or depth can be optional, but when actually specifying a `field_chunksize`, there should be at least values for *lat* and *lon* (or: their equivalents). Hence, those dimensions are actually treated differently.

CKehl: I also replaced `raise Warning(...)` with `logger.warning_once`, because apparently, raising a warning results in pytest failure. Learning experience +1

Issue #724, opened 2020-01-30
Random.VonMises C function


pierrick-giffard:
To simulate that active particles (tuna, sea turtles, ...) go in a given direction +/- a random angle, a function equivalent to the python random.vonmisesvariate can be useful. I implemented it in C based on the python source code ([](https://github.com/python/cpython/blob/master/Lib/random.py)). It is reproducing the expected behaviour. 
Below I plotted the result of 10000 calls to the function with a mean value of pi and a concentration parameter of 4. 
![image](https://user-images.githubusercontent.com/56306211/73471401-948bec80-4389-11ea-84de-cb1e43879dfe.png)

Here is the source code, it is just needed to add it to parcels/include/random.h.
[C_VonMises.txt](https://github.com/OceanParcels/parcels/files/4135204/C_VonMises.txt)

PR #725, opened 2020-01-31
Adding vonmisesvariate function

erikvansebille:
This PR adds a `vonmisesvariate` function to the parcels library of random functions. It follows @pierrick-giffard suggestion in #724 

@pierrick-giffard, could you check if this implementation works for you? And could you also propose another `assert` statement in the `test_diffusion.py` function to check if implementation of kappa is correct?


pierrick-giffard:
It's working for me, thank you for the implementation!

PR #727, opened 2020-02-03
Support setting a particle state to skip advection

angus-g:
Instead of deleting a particle under a certain condition (like an error), it may be useful to keep the particle in the output files, but cease any computation on that particle. For this, a particle state can be set to `Skip`.

I've also renumbered the particle states, though maybe it would make sense to make skip = 3 and error = 4, with the various sub-errors numbered 4x (or even something else entirely)?


erikvansebille:
Thanks for this PR, @angus-g. This looks good code, and I'm happy with the renumbering indeed. But is it by design that, even if the kernels are not executed on a particle, it will still be outputted?

Could you provide a unit-test in `test_kernel_execution.py` (where most error codes are being tested) that checks that this works?

angus-g:
It is indeed the intention (in my case) that the particles are kept to still be outputted. Otherwise I tihnk

erikvansebille:
> I'll push through a unit test shortly!

Have you already found time to create a simple unit test for this new functionality, @angus-g?

erikvansebille:
Hi @angus-g, you may have seen that we've just rebased your other PR (#678), @CKehl and I are keen to merge that into `master` early next week!

What do you want with this PR? Note that the recent #672 has implemented significant rewrite of codegenerator, rebasing might be a considerable effort. Are you still working on these unit tests? 

Or, if you don't intent to work on this more, shall we close this PR for now?

angus-g:
I'm happy to close this for now. If I find I need it again I'll reopen a new PR against the updated code. Thanks!

PR #730, opened 2020-02-05
Add a unit test that checks if there are warnings in the compilation log

erikvansebille:
Note that this breaks at the moment because at the moment there are two compilation warnings. This will need to be fixed before we can merge this PR. But after that, introduction of warnings will raise this unit test

### DIFF
--- a/tests/test_kernel_execution.py
+++ b/tests/test_kernel_execution.py
@@ -266,12 +266,15 @@ def test_errorcode_repeat(fieldset, mode):
 
 
 @pytest.mark.parametrize('delete_cfiles', [True, False])
-def test_execution_keep_cfiles(fieldset, delete_cfiles):
+def test_execution_keep_cfiles_and_nocompilation_warnings(fieldset, delete_cfiles):
     pset = ParticleSet(fieldset, pclass=JITParticle, lon=[0.], lat=[0.])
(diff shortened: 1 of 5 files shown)


tests/test_kernel_execution.py
```diff
@@ -266,12 +266,15 @@ def simpleKernel(particle, fieldset, time):
 
 
 @pytest.mark.parametrize('delete_cfiles', [True, False])
-def test_execution_keep_cfiles(fieldset, delete_cfiles):
+def test_execution_keep_cfiles_and_nocompilation_warnings(fieldset, delete_cfiles):
     pset = ParticleSet(fieldset, pclass=JITParticle, lon=[0.], lat=[0.])
-    pset.execute(pset.Kernel(DoNothing, delete_cfiles=delete_cfiles), endtime=1., dt=1.)
+    pset.execute(pset.Kernel(AdvectionRK4, delete_cfiles=delete_cfiles), endtime=1., dt=1.)
```
CKehl: the test still fails, and I can imagine that is because `pset.Kernel(..., endtime=1., dt=1.)` - the main func actually does not progress to the next step. Does `pset.Kernel(..., endtime=1.1, dt=1.)` make a difference for the online CI OSX test ?

PR #731, opened 2020-02-05
Updating memory performance doc

CKehl:
This issue does 3 things:

- Update the `documentation_MPI.ipynb` notebook with more information and comments on memory consumption
- Integrate the referral to the new FAQ page of Parcels on Memory and Performance to the related runtime warnings
- **[important]** Fix a bug with file locking that became an issue after actually doing lazy loading with dask/xarray


CKehl:
PS: the actual figures are auto-generated by the binder/runner thingy when evaluated on the server - so: they will update then later, I assume. Still, as I have the figures locally, I did comment on them in text.

parcels/examples/documentation_MPI.ipynb
```diff
@@ -59,14 +59,14 @@
     "\n",
     "However, in tests we have experienced that this may not necessarily be the most efficient chunking. Therefore, Parcels provides control over the chunk-size via the `field_chunksize` keyword in `Field` creation, which requires a tuple that sets the typical size of chunks for each dimension.\n",
     "\n",
-    "It is strongly encouraged to explore what the best value for chunksize is for your experiment, which will depend  on the `FieldSet`, the `ParticleSet` and the type of simulation (2D versus 3D). As a guidance, we have found that chunksizes in the zonal and meridional direction of approximately `500` are typically most effective."
+    "It is strongly encouraged to explore what the best value for chunksize is for your experiment, which will depend  on the `FieldSet`, the `ParticleSet` and the type of simulation (2D versus 3D). As a guidance, we have found that chunksizes in the zonal and meridional direction of approximately around *128-512* are typically most effective. The binning relates to the size of the model and its data size, so power-of-two values are advantageous but not required."
```
erikvansebille: change *128-512* to *128 to 512* for better readability?

erikvansebille:
> PS: the actual figures are auto-generated by the binder/runner thingy when evaluated on the server - so: they will update then later, I assume. Still, as I have the figures locally, I did comment on them in text.

How does that work? Normally, we also load the figures in the notebook already

parcels/examples/documentation_MPI.ipynb
CKehl: yes, I can do that.

CKehl:
Well, so if I look into the code, there are no 'show image' statements for the graphs, just sections of markdown text and python code. Secondly, when I pulled the new versions from master this morning, they already had the new figures in them - and the notebook on the website also already has new (re-computed) figures. So, those figures come from some CI procedure actually computing or evaluating the code, if I'm not mistaken. Otherwise, I'd need to see how to get the figures in there.

CKehl:
@erikvansebille I have a problem with `documentation_MPI.ipynb`. This file has been expanded in jupyter at some point with printing also 3D-temporal simulations (benchmarking that), but the loading and computation of those things was done in interactive mode and the code has been removed, so I cannot re-run that notebook to create the figures because I lack the code to run the 3D+time simulation. Can you provide me with that sub-routine ?

parcels/examples/documentation_MPI.ipynb
CKehl: still need to do that before merge

erikvansebille: I've just updated it

PR #740, opened 2020-02-10
Bugfixes for 4D data access using field chunking

CKehl:
In issue #719 , we have adapted the data access in the NetcdfFileBuffer to actually make use of the lazy field evaluation to prevent loading all data per time step at once. We have thoroughly tested the procedure, but only for 2D+time (3D) data. For 3D+time (4D) data, this procedure was still buggy.

In this pull request, we resolve those issues with 4D data access. For testing, we used the Galapagos backward advection, as shown in van Sebille et al. 2019 "Basin-scale sources and pathways of microplastic that ends up in the Galápagos Archipelago" [code here](https://github.com/OceanParcels/GalapagosBasinPlastic), which is now running through successfully. Furthermore, we tested the process on the 3D NEMO data. in issue #731 , we will later use the fix to re-generate the MPI doc with the global reanalysis data for 3D fields. This run requires the bugfixes made here.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -273,7 +273,7 @@ class Field(object):
             else:
                 raise RuntimeError('interp_method is a dictionary but %s is not in it' % variable[0])
 
-        with NetcdfFileBuffer(lonlat_filename, dimensions, indices, netcdf_engine, field_chunksize=False) as filebuffer:
+        with NetcdfFileBuffer(lonlat_filename, dimensions, indices, netcdf_engine, field_chunksize=False, lock_file=False) as filebuffer:
             lon, lat = filebuffer.read_lonlat
             indices = filebuffer.indices
             # Check if parcels_mesh has been explicitly set in file
@@ -281,7 +281,7 @@ class Field(object):
                 mesh = filebuffer.dataset.attrs['parcels_mesh']
 
         if 'depth' in dimensions:
-            with NetcdfFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method, field_chunksize=False) as filebuffer:
+            with NetcdfFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method, field_chunksize=False, lock_file=False) as filebuffer:
                 filebuffer.name = filebuffer.parse_name(variable[1])
                 depth = filebuffer.read_depth
                 data_full_zdim = filebuffer.data_full_zdim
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -309,8 +309,7 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                 timeslices = []
                 dataFiles = []
                 for fname in data_filenames:
-                    with NetcdfFileBuffer(fname, dimensions, indices, netcdf_engine,
-                                          field_chunksize=False) as filebuffer:
+                    with NetcdfFileBuffer(fname, dimensions, indices, netcdf_engine, field_chunksize=False) as filebuffer:
```
erikvansebille: Specify here if `lock_file=False` or `lock_file=True`, to make choice explicit?

parcels/field.py
```diff
@@ -1556,55 +1555,72 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
...
 
     def __enter__(self):
         if self.netcdf_engine == 'xarray':
             self.dataset = self.filename
             return self
+
         init_chunk_dict = self._get_initial_chunk_dictionary()
         try:
-            self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine=self.netcdf_engine, chunks=init_chunk_dict)
+            # unfortunately we need to do if-else here, cause the lock-parameter is either False or a Lock-object (we we rather want to have auto-managed)
```
erikvansebille: Double 'we'  in the comment here?

parcels/field.py
```diff
@@ -1785,19 +1852,23 @@ def data_access(self):
...
                     self.chunk_mapping = {}
                     chunkIndex = 0
                     for chunkDim in data.numblocks[1:]:
                         self.chunk_mapping[chunkIndex] = chunkDim
                         chunkIndex += 1
+                    self._chunkmap_to_chunksize()
                     self.rechunk_callback_fields()
-                elif self.field_chunksize != 'auto':
+                    self.chunking_finalized = True
+                elif self.field_chunksize != 'auto' and not self.chunking_finalized:    # why ? why here ? why at every single data access, instead of fixing that size when you open the file ?!
```
erikvansebille: This is a somewhat desperate-sounding comment. As if someone who didn't write the code added it here? Could you expand?

CKehl: oops - old left-over from my private repo - will be removed (what is standing there actually now fixes what I was mourning over at that point).

parcels/field.py
CKehl: yes, grammar mistake - will be corrected. I left the comment in on purpose to make it understandable why `lock=False` is okay but `lock=True` is not, in that case.

parcels/field.py
CKehl: we could make `lock_file=True` in this case (which is the default, which is why I did not specify that). Here, we need to lock because we actually access the time dimension of the NetCDF - and actual data access requires the lock.

Issue #741, opened 2020-02-10
"compute_on_defer" not working with dask array


claudiofgcardoso:
Hello all, 

I realized that 'fieldset.U.data' is now a dask array:
`dask.array<where, shape=(3, 28, 205, 223), dtype=float32, chunksize=(1, 28, 205, 223), chunktype=numpy.ndarray>`

This makes the simplest operation using 'compute_on_defer' not working:
```python
def compute(fieldset):
    for tind in fieldset.W[0].loaded_time_indices:
        fieldset.W[0].data[tind,0]=0
        fieldset.W[1].data[tind, 0] = fieldset.W[1].data[tind, 0].clip(min=0)
```
Result:
`NotImplementedError: Item assignment with <class 'tuple'> not supported`

Any ideas of how to go around this?

erikvansebille:
Thanks for reporting, @claudiofgcardoso. You're right that with the new dask development, `compute_on_defer` may not work as expected anymore because mixing `numpy` methods in the `compute` function with `dask` Fields may lead to the entire array being converted to numpy

However, the output from your function seems to hint at something else being wrong. Can you investigate a bit _which_ of the two lines in your `compute` functions break? And what if you do 'simpler' functions? This kind of info is really useful to us for debugging!

claudiofgcardoso:
I changed the `compute` function to something much simple, such as: 
```python
def compute(fieldset):
    fieldset.W[0].data[:, 0, :, :] = 0.
    fieldset.W[1].data[:, 0, :, :] = 0.
```
The outcome was the same: 
``` 
fieldset.computeTimeChunk(1,1)
Traceback (most recent call last):

  File "<ipython-input-5-a7923779232e>", line 1, in <module>
    fieldset.computeTimeChunk(1,1)

  File "/home/ccardoso/anaconda3/lib/python3.7/site-packages/parcels/fieldset.py", line 853, in computeTimeChunk
    self.compute_on_defer(self)

  File "<ipython-input-3-1f2061b49552>", line 19, in compute
    fieldset.W[0].data[:, 0, :, :] = 0.

  File "/home/ccardoso/anaconda3/lib/python3.7/site-packages/dask/array/core.py", line 1480, in __setitem__
    "Item assignment with %s not supported" % type(key)

NotImplementedError: Item assignment with <class 'tuple'> not supported 
```

[...]

PR #742, opened 2020-02-10
Changing compute_on_defer unit test to use dask

erikvansebille:
This PR addresses how `FieldSet.compute_on_defer` can best be combined with `dask` arrays, following the bug report at #741

### DIFF
--- a/parcels/tools/error.py
+++ b/parcels/tools/error.py
@@ -2,7 +2,7 @@
 
 
 __all__ = ['ErrorCode', 'FieldSamplingError', 'FieldOutOfBoundError', 'TimeExtrapolationError',
-           'KernelError', 'OutOfBoundsError', 'ThroughSurfaceError',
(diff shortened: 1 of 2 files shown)


lucvandenbulcke:
Hello,
I'm sorry to write in this old PR already merged. I don't understand exactly whether this PR solved issue #741 ? Using a Parcels version more recent than this merge, I still have the same error that I cannot set `fieldset.W.data[:,0,:,:]=0.` when it is a dask array.
Is the workaround still what @erikvansebille recommended in his May 13 answer in issue #741 , i.e. to use the `ErrorCode.ThroughSurface` error ?
Thanks !

erikvansebille:
If your files are not too big, then the easiest solution to implement  is to run with `field_chunksize=False` - see bullet point 2 of https://github.com/OceanParcels/parcels/issues/741#issuecomment-627763495. But the other bullets are also options

lucvandenbulcke:
Indeed this works !
Thank you very much again

Issue #746, opened 2020-02-13
Extreme slowdown at specific execution timesteps


reint-fischer:
Hi all,

I am releasing particles in an open 3D domain with velocities only in the [x,z]-plane.

I am running into a sudden extreme increase in execution time at specific moments for specific combinations of dt and outputdt. At different starttimes, fieldsets, and both in forwards in time and backwards in time execution parcels takes extremely long at the same specific timestep. 

Decreasing dt seems to make this moment come earlier, while decreasing outputdt seems to do the opposite. However, some values for dt work significantly better, i.e. no difference between 0.005 and 0.01 but large difference between 0.005 and 0.004.

A number of combinations I have found:
DT [s]                      OutputDT [s]                     Timestep [s] where parcels starts taking very long
0.002                      0.1                                     0.6
0.002                      0.08                                   0.64
0.002                      0.01                                   5.42
0.004                      0.1                                     0.9999
0.004                      0.08                                   1.28
0.005                      0.1                                     44.8
0.005                      0.08                                   44.8
0.01                        0.1                                     44.8
0.01                        0.08                                   44.8

Since the bug is irrespective of starttime or fieldset, I do not think anything is happening to the particles. My memory looks fine. I get no errors. The problem is exactly reproducible.

[ExtremeSlowdown.txt](https://github.com/OceanParcels/parcels/files/4197095/ExtremeSlowdown.txt)

Any ideas what this might be?

reint-fischer:
Hi, as a quick fix for the bug, we've tried increasing the `__tol` variable. This seems to increase the time it takes before the extreme slowdown occurs and perhaps even speed up the continuation. It doesn't get rid of the problem altogether however. 

Now, I have found that for `__tol` == 1e-4, outputdt = 0.1 and dt = 0.001 I can get Parcels to execute for 44 seconds, with a slowdown at 21.3 seconds. After the slowdown, it seems something changes in the outputdt, as you can see in these figures:
![bug1zoom](https://user-images.githubusercontent.com/33000433/75148762-b2d2d700-5700-11ea-8d4b-f56a212e4d5f.png)
![bug1](https://user-images.githubusercontent.com/33000433/75148763-b36b6d80-5700-11ea-9277-aa9deb2f9cce.png)

Up until 21.3 seconds, each timestep takes 0.1 seconds, so the line is nicely 1:1. After the slowdown, the dt becomes variable. You can see this in this figure: 
[...]

PR #747, opened 2020-02-13
increasing __tol value in JIT particle loop

erikvansebille:
If dt is very small (~<0.001 s), Parcels can get stuck in an infinite loop (see #746). 

Commit fff3c97 fixes that by increasing the tolerance `__tol` in the JIT particles loop. 
This, however, is not a very clean solution. Perhaps a more robust solution can be developed


erikvansebille:
Closing this PR, as #762 is a more robust implementation/fix

PR #750, opened 2020-02-18
'KeyError: 2' hotfix

CKehl:
only recover the particle errors that are listed in the combined recovery map, delete all particles that are unsuccessful and for whose errors there is no recovery specified.

Nature: hotfix

The error (until now tracked) only occurs when your simulation dynamically adds AND deletes particles during the execute() function. The actual errors source for this is still investigated, thus a better fix will be submitted separately in the future. This given fix is tested.

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -353,10 +353,12 @@ class Kernel(object):
             for p in error_particles:
                 if p.state == ErrorCode.Repeat:
                     p.state = ErrorCode.Success
-                else:
+                elif p.state in recovery_map:
                     recovery_kernel = recovery_map[p.state]
                     p.state = ErrorCode.Success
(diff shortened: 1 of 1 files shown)


parcels/kernel.py
```diff
@@ -353,10 +353,12 @@ def remove_deleted(pset):
             for p in error_particles:
                 if p.state == ErrorCode.Repeat:
                     p.state = ErrorCode.Success
-                else:
+                elif p.state in recovery_map:
                     recovery_kernel = recovery_map[p.state]
                     p.state = ErrorCode.Success
                     recovery_kernel(p, self.fieldset, p.time)
+                else:
```
erikvansebille: Add here in comments that this is a `HOTFIX for #750`?

CKehl: agreed

PR #751, opened 2020-02-18
KeyError 2 - discussion and bugfix

CKehl:
Reason and bugfix ideas for 'KeyError: 2' kernel abortion


CKehl:
Several experts, adopters, users and students recently experienced a problem that appears as follows (or similar):

```
  File "/git_code/private_parcels/performance/benchmark_perlin.py", line 362, in <module>
    pset.execute(kernels, runtime=delta(days=time_in_days), dt=delta(minutes=dt_minutes), output_file=output_file, recovery={ErrorCode.ErrorOutOfBounds: delete_func}, postIterationCallbacks=postProcessFuncs, callbackdt=delta(hours=12))
  File "/var/scratch/bionic/lib/python3.7/site-packages/parcels-2.1.3.dev81+g86a2147.d20200218-py3.7.egg/parcels/particleset.py", line 550, in execute
    self.kernel.execute(self, endtime=time, dt=dt, recovery=recovery, output_file=output_file)
recovery_kernel = recovery_map[p.state]; KeyError: 2
```

This error key maps to one of the error codes defined in `parcels/tools/error.py` and according to line 12, it is `ErrorCode.Delete`. This code is given to all particles that are deleted - either internally by the main loop when a recovery map deletes out-of-bound particles, or externally by dynamic removed procedures, such as aging, beaching, surfacing, and so forth. Now, the internal *quasi-state machine* for handling out-of-bound particles and other interpolation errors that are fixed in the provided recovery map of `pset.execute()` are not the source of this error.

In fact, we were only able to reproduce the error in scenarios where particles are dynamically added and deleted (e.g. releasing continuous particle streams via `repeatdt=<some time>` and deleting particle by aging).

The reason for the error can be found in `Kernel.execute()` in `parcels/kernel.py` (currently: line 314 to 374):

```
    def execute(self, pset, endtime, dt, recovery=None, output_file=None):
        """Execute this Kernel over a ParticleSet for several timesteps"""

        def remove_deleted(pset):
            """Utility to remove all particles that signalled deletion"""
            indices = [i for i, p in enumerate(pset.particles)
                       if p.state in [ErrorCode.Delete]]
            if len(indices) > 0 and output_file is not None:
                output_file.write(pset[indices], endtime, deleted_only=True)
            pset.remove(indices)

        if recovery is None:
            recovery = {}
        elif ErrorCode.ErrorOutOfBounds in recovery and ErrorCode.ErrorThroughSurface not in recovery:
            recovery[ErrorCode.ErrorThroughSurface] = recovery[ErrorCode.ErrorOutOfBounds]
        recovery_map = recovery_base_map.copy()
        recovery_map.update(recovery)

        # SOME FIELD TREATMENT - UNIMPORTANT

        # EXECUTE KERNEL #
        if self.ptype.uses_jit:
            self.execute_jit(pset, endtime, dt)
        else:
            self.execute_python(pset, endtime, dt)

        # REMOVE DELETED PARTICLES #
        remove_deleted(pset)

        # COLLECT ALL ERROR PARTICLE NOT FLAGGED FOR DELETION #
        error_particles = [p for p in pset.particles
                           if p.state != ErrorCode.Success]

        # WHILE THERE ARE STILL KERNEL ERRORS #
        while len(error_particles) > 0:
            # RECOVER ERROR PARTICLES FOR WHICH A RECOVERY ENTRY EXISTS #
            for p in error_particles:
                if p.state == ErrorCode.Repeat:
                    p.state = ErrorCode.Success
                elif p.state in recovery_map:               # hotfix for #749, #737 and related issues
                    recovery_kernel = recovery_map[p.state]
                    p.state = ErrorCode.Success
                    recovery_kernel(p, self.fieldset, p.time)
                else:
                    logger.warning_once('Deleting particle because of bug in #749 and #737')
                    p.delete()

            # REMOVE DELETED PARTICLES (e.g. where recovery => particle.delete() ) #
            remove_deleted(pset)

            # RE-RUN THE KERNEL WITHOUT ADVANCING endtime #
            if self.ptype.uses_jit:
                self.execute_jit(pset, endtime, dt)
            else:
                self.execute_python(pset, endtime, dt)

            # COLLECT ERROR PARTICLES #
            error_particles = [p for p in pset.particles
                               if p.state != ErrorCode.Success]
```

Now, the top-outlined kernel algorithm has a disadvantage as follows: if your (accumulated) kernel has an aging function similar as such

```
def Age(particle, fieldset, time):
    particle.age = particle.age + math.fabs(particle.dt)
    if particle.age > threshold:
        particle.delete()
```

then, in update cycles where out-of-bound errors and the like occur, the `Age` function is called multiple times for the same time integration step. Now, because particle uses exact integration, `particle.dt` is in this case not always what is defined within `pset.execute()`, but can be smaller integral steps. As a result, in some out-of-bound error runs, the situation may occur where a particles in the original `pset.execute_jit()` or `pset.execute_python()` (both cases tested) do not exceed the `threshold` in the `Age` function but, because of floating-point accuracy limits, DO exceed the threshold in the *recovery re-runs*, they are flagged as *to-be-deleted* even though they (pedantically-speaking) should not, because those recovery runs are not meant to do a time integration. It is a result of floating point limits with `particle.dt`.

A rather obvious-and-dirty fix is to re-call `remove_deleted(pset)` at the beginning of every recovery run, I personally do not support this approach because (a) we shouldn't re-run kernels for particle recovery in any case; (b) particle deletion is a costly operation, in terms of memory and time (see the implementation for `numpy.delete()` form the [official repository](https://github.com/numpy/numpy/blob/v1.17.0/numpy/lib/function_base.py#L4225-L4429) ) that should not be executed if not needed; (c) finding all deleted particles in a large particle set is an exceedingly time-consuming (meaning: performance-consuming) operation, which should not be done if it can be circumvented.

There is a way the user can fix it. An alternative aging particle and kernel that is error-resilient looks like this:
```
class AgeParticle_JIT(JITParticle):
    age = Variable('age', dtype=np.float64)
    agetime = Variable('agetime', dtype=np.float64)
    def __init__(self, *args, **kwargs):
        type(self).age.initial=0.
        type(self).agetime.initial=-1.0
        super(AgeParticle_JIT, self).__init__(*args, **kwargs)

def Age(particle, fieldset, time):
    if math.fabs(time-particle.agetime) > math.fabs(1.e10 * particle.dt):
        particle.age = particle.age + math.fabs(particle.dt)
        particle.agetime = time

        if particle.age > (60.0*60.0*24.0*62.0):
            particle.delete()
```
Here, the important bit is ` if math.fabs(time-particle.agetime) > math.fabs(1.e10 * particle.dt)`, in order to prevent actions in those recovery runs. This is working reliably, but I admit that it's limited to fairly simple kernels and particles.

Ideally, we can (together, as a community - suggestions welcome) find an appropriate fix to prevent the re-execution of kernels on particles that finished their time step integration successfully.

CKehl:
resolved by #762

Issue #752, opened 2020-02-19
Exception with version 2.1.4 in Python 3.8


albertotb:
If I create an environment with Python 3.8

     conda create -n py38_parcels -c conda-forge python>=3.8 parcels jupyter cartopy ffmpeg

activate it

    conda activate py38_parcels

and run the examples

    python example_peninsula.py --fieldset 100 100

I get the following error

```
INFO: Generating FieldSet output with basename: peninsula
WARNING: Particle initialisation from field can be very slow as it is computed in scipy mode.
Peninsula: Advecting 20 particles for 1 day, 0:00:00
Traceback (most recent call last):
  File "example_peninsula.py", line 206, in <module>
    peninsula_example(fieldset, outfile, args.particles, mode=args.mode,
  File "example_peninsula.py", line 118, in peninsula_example
    pset.execute(k_adv + k_p, runtime=time, dt=dt, output_file=out)
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/kernel.py", line 387, in __add__
    return self.merge(kernel)
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/kernel.py", line 379, in merge
    return Kernel(self.fieldset, self.ptype, pyfunc=None,
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/kernel.py", line 117, in __init__
    exec(compile(py_mod, "<ast>", "exec"), user_ctx)
TypeError: required field "type_ignores" missing from Module
Exception ignored in: <function ParticleFile.__del__ at 0x7fb778d9c940>
Traceback (most recent call last):
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/particlefile.py", line 195, in __del__
    self.close()
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/particlefile.py", line 200, in close
    self.export()
  File "/home/atorres/miniconda3/envs/py38_parcels/lib/python3.8/site-packages/parcels/particlefile.py", line 350, in export
    raise RuntimeError("No npy files found in %s" % self.tempwritedir_base)
RuntimeError: No npy files found in out-EMKLQHPA
Exception ignored in: <function Pool.__del__ at 0x7fb781ec7040>
Traceback (most recent call last):
[...]

PR #753, opened 2020-02-19
Bug fixes for parcels in python3.8

erikvansebille:
As reported in #752, parcels does not currently work with python3.8. To address this, we need to implement at least the following fixes
 
- [x] Implementing change in `ast.Module` call in py3.8 (see e.g. https://github.com/pallets/werkzeug/issues/1551)
- [x]  Proper parsing of comment strings in kernels
- [x] Fix exception in `function Pool.__del__` at end of parcels run


environment_py3_osx.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - defaults
   - conda-forge
 dependencies:
-  - python>=3.4
+  - python=3.8
```
CKehl: do you want to leave it in ? I mean, you tested it works now for 3.8, but shouldn't >=3.4 be the correct dependency ?

erikvansebille: Yes, this can be reverted now. That means we don't check for python3.8 explicitly, but at least now we know that there are no more errors with the present code

PR #760, opened 2020-02-25
Speed up field initialization

CKehl:
Help for @pdnooteboom for testing plankton advection with small NEMO or POP files by 'tuning' the field init a bit


CKehl:
Fixes are of general nature, thus generally applicable (can be merged to master without trouble)

Issue #759, opened 2020-02-25
s-coordinates


ana-machado:
Hello,

I am doing some tests in order to use parcels with roms outputs. To certify if everything was running as expected, I made two tests:
-keep u,v, and set w=0 (teste2_3rsteps): the files continue moving in the vertical, I was expecting them to stay at the deployment depth and being advected only horizontally.
-set u,v,w=0 (teste3_3rsteps), as expected the particles don't move.
The nc files are in this link, if you think they're useful.
https://meocloud.pt/link/496136f3-00e6-4131-a993-88ebbf77eec8/roms_outputs_3tSteps/
Am I doing something wrong, or is this a bug?

Best regards,
Ana

![teste2_3rsteps](https://user-images.githubusercontent.com/34549357/75203597-6fab4f00-5766-11ea-81ac-9132b9b7bcc6.png)
![teste3_3rsteps](https://user-images.githubusercontent.com/34549357/75203600-720da900-5766-11ea-8594-59cc574d988b.png)

erikvansebille:
Hi @ana-machado, can you also send the scripts that you used to create these simulations?

ana-machado:
Hello. Of course. I added the grid file, the script to run parcels, and the one to plot the previous figures in the same shared folder:
https://meocloud.pt/link/496136f3-00e6-4131-a993-88ebbf77eec8/roms_outputs_3tSteps/
Thank you very much
Ana

erikvansebille:
Hi @ana-machado. I have been looking at your code, but had to change quite a bit in your script, for example because the `out3.nc` file that you provided seems to contain only 0s... I thus now use the `out2.nc` file

Anyway, the main reason it seems why your particles were also moving up and down is because you used the `AdvectionRK4_3D` kernel, instead of the `AdvectionRK4` kernel. As soon as you provide vertical velocity (`W`), then your particles will move up and down...

The following code works for me
```python
from parcels import FieldSet, ParticleSet, JITParticle, AdvectionRK4, ErrorCode
from datetime import timedelta as delta

data_file = 'out2.nc'
outFileName = 'teste3_3tsteps.nc'

def DeleteParticle(particle, fieldset, time):
    particle.delete()

filenames = {'U': data_file, 'V': data_file, 'W': data_file}

variables = {'U': 'u', 'V': 'v', 'W': 'w'}

[...]

Issue #749, opened 2020-02-17
memory error when deferred_load = False in fieldset


ignasivalles:
Hi! I had similar problem than #658. Following @delandmeterp suggestions I set `deferred_load=False` . A short simulation was working for parcels version 2.1.2 but not for version 2.1.4, where fieldset crashes with `MemoryError`.  I tried to use `field_chunksize` to `auto`, `False` and a number without success.

So my question is if there is somehow a better way to solve the calendar issue without scarifying the `deferred_load` option. 

Thank you!

erikvansebille:
Have you tried using the `timestamps` keyword argument, as explained in https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_timestamps.ipynb? Does that fix this issue?

ignasivalles:
Thanks Erik!, It works but I had to split my netcdf files into single time-step files to make it work. 
But  now some another errors appeared again (#712) : `recovery_kernel = recovery_map[p.state]; KeyError: 2`. 

First I added recovery kernel as you suggested in #737, only for `ErrorCode.Error`:

```
def DeleteParticle(particle, fieldset, time):
    particle.delete()
pset.execute(..., recovery={ErrorCode.Error: DeleteParticle}))
```
Then I had a `ThroughSurfaceError` with  `Through-surface sampling by particle at (-7.472253, -4.896896, 21.116211)`. Note that depth value is 21.11m which is obviously not at the surface.

Then I set `recovery={ErrorCode.Error: DeleteParticle, ErrorCode.ErrorOutOfBounds: DeleteParticle})` and the same recovery kernel error appears.

This is always happening in the first time-step of my simulation.

CKehl:
we're working on the KezError bug right now - we keep you informed when it's solved.

Issue #736, opened 2020-02-06
Chunking broken in v2.1.3 for NEMO data?


willirath:
With v2.1.3, the only `field_chunks` value working with `from_nemo` seems to be `False` which seems to lead to chunks spanning the whole data arrays.

erikvansebille:
Thanks for reporting, @willirath! Indeed, this is bug in `FieldSet.from_nemo()`. I've just pushed a fix to #738. Can you check if this works for you?

willirath:
Doesn't work for me. I've tried the following cases:

Not setting `field_chunks` or setting it to `field_chunks='auto'` results in `IndexError: list assignment index out of range`.

```python-traceback
INFO: Compiled SampleParticleAdvectionRK4_3Dtemperature_sampling ==> /tmp/parcels-1001/3497c45c845653ab64d45d49cb094fac_0.so
INFO: Unable to locate chunking hints from dask, thus estimating the max. chunk size heuristically. Please consider defining the 'chunk-size' for 'array' in your local dask configuration file (see http://oceanparcels.org/faq.html#field_chunking_config and https://docs.dask.org).
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-27-b9ad495d7e42> in <module>
      4      dt=timedelta(minutes=10),
      5      output_file=outputfile,
----> 6      recovery=recovery_cases
      7 )

/opt/tljh/user/envs/py3_lagrange_dev/lib/python3.6/site-packages/parcels/particleset.py in execute(self, pyfunc, endtime, runtime, dt, moviedt, recovery, output_file, movie_background_field, verbose_progress, postIterationFunctions)
    517             else:
    518                 time = max(next_prelease, next_input, next_output, next_movie, endtime)
--> 519             self.kernel.execute(self, endtime=time, dt=dt, recovery=recovery, output_file=output_file)
    520             if abs(time-next_prelease) < tol:
    521                 pset_new = ParticleSet(fieldset=self.fieldset, time=time, lon=self.repeatlon,

/opt/tljh/user/envs/py3_lagrange_dev/lib/python3.6/site-packages/parcels/kernel.py in execute(self, pset, endtime, dt, recovery, output_file)
    339         # Execute the kernel over the particle set
    340         if self.ptype.uses_jit:
--> 341             self.execute_jit(pset, endtime, dt)
    342         else:
    343             self.execute_python(pset, endtime, dt)

[...]

Issue #722, opened 2020-01-28
Create a test to check and verify memory behaviour and containment with different Field() structures


CKehl:
Many issues recently experienced severe bugs in terms of memory leakage, which required significant fixes. In order to avoid, prevent or mitigate future memory leaks, we should develop and include a test that checks the memory behaviour and containment of the following things:

- [ ] different pre-defined chunk sizes: auto, False, tuple-definition, dict-definition

- [ ] check usage of Dask parameters: field_chunksize='auto' *with* and *without* `dask.yaml`

- [ ] deferred loading vs. direct loading

- [ ] time extrapolation: None, time_extrapolate, time_periodic

- [ ] MPI vs. non-MPI (if possible)

- [ ] time-forward vs. time-backward computing

erikvansebille:
Very good idea! This can go into a new `performance` folder on the parcels repo? We could then automate testing/analysis on our own HPC systems, without burdening the github actions CI (which wouldn't have access to the large netcdf files anyways). Would that work?

PR #764, opened 2020-02-27
Fix auto field chunking

CKehl:


CKehl:
last commit fixes errors such as:
`AttributeError: 'DeferredArray' object has no attribute 'shape'`

CKehl:
related to:
#711 #612 #759 

further tests / re-runs after merge required for:
#722 #736 #749

parcels/fieldset.py
```diff
@@ -778,16 +778,23 @@ def computeTimeChunk(self, time, dt):
                 continue
             g = f.grid
             if g.update_status == 'first_updated':  # First load of data
+                if f.data is not None and not isinstance(f.data, DeferredArray):
+                    if not isinstance(f.data, list):
+                        f.data = None
+                    else:
+                        for i in range(len(f.data)):
+                            del f.data[i, :]
```
erikvansebille: Why does this need to be done for each `i` separately?

parcels/particleset.py
```diff
@@ -420,7 +420,8 @@ def execute(self, pyfunc=AdvectionRK4, endtime=None, runtime=None, dt=1.,
         :param movie_background_field: field plotted as background in the movie if moviedt is set.
                                        'vector' shows the velocity as a vector field.
         :param verbose_progress: Boolean for providing a progress bar for the kernel execution loop.
-        :param postIterationFunctions: Array of functions that are to be called after each iteration (post-process, non-Kernel)
+        :param postIterationCallbacks: Array of functions that are to be called after each iteration (post-process, non-Kernel)
```
erikvansebille: Add in the docstring that these two functions are most useful for optimisation purposes?

tests/test_fieldset.py
```diff
@@ -346,6 +349,88 @@ def test_vector_fields(mode, swapUV):
...
+    if with_GC:
+        assert np.allclose(mem_steps_np[7:], perflog.memory_steps[-1], rtol=0.01)
+    if cs is not False or with_GC:
+        assert np.alltrue((mem_steps_np-mem_0) < 4712832)   # represents 4 x [U|V] * sizeof(field data)
+    assert not mem_exhausted
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('time_periodic', [4*86400.0, False])
+@pytest.mark.parametrize('cs', [False, 'auto', {'x': 32, 'y': 32}, {'time_counter': 1, 'x': 32, 'y': 32}, (32, 32), (1, 32, 32)])
```
erikvansebille: Also test here for simply `32`? Because I heard from some users that they expected that they could put an integer for `field_chunksize`

tests/test_fieldset.py
```diff
@@ -346,6 +349,88 @@ def test_vector_fields(mode, swapUV):
         assert abs(pset[0].lat - .5) < 1e-9
 
 
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('time_periodic', [4*86400.0, False])
+@pytest.mark.parametrize('cs', [False, 'auto', (1, 32, 32)])
```
erikvansebille: Call this variable `field_chunksize` instead of `cs` for readability?

parcels/fieldset.py
CKehl: that is because, if called as `del f.data[:]`, one cannot address the sub-elements anymore - they would need to be reallocated as `f.data = [None, ] * 3`. That's not necessary, because we only want to remove the larger field data within the python list, not the list itself. That is why it's written the way it is, removing each sub-element individually. It's als `range(len(f.data))` to avoid problems when only having a period of two (or one) discrete time periods. We will see if this is a performance bottleneck now, but it's a short loop and only occurs after (re-)initialization, so I don't expect a large hit on performance in common use cases.

erikvansebille: OK that's clear. makes sense then

parcels/particleset.py
CKehl: if you think we should add it there, we can. The function is flexible, as we use if for performance measurements as well as garbage collection - which I assume scientists with large-scale case studies or long-running simulations may want to make use of.

tests/test_fieldset.py
CKehl: No, we don't facilitate THAT right now, cause it's ambiguous - which data dimension does this apply to ? All of them ? If can be done, if requested, but it's also not great in performance cause the used dimensions are anisotropic - meaning: either the laterial dimensions are much larger than the vertical, or vice versa (for cross-section/sinking studies).

If you wish that to happen, we can integrate it, though I'd rather prefer that working separately in an issue - this one is really just for testing and fixing what should already work.

tests/test_fieldset.py
CKehl: yes, we can do that

parcels/particleset.py
erikvansebille: Well, I'd like to make sure that typical users don't expect this to be something they _have_ to use. So even the word '(Optional)' in front would help?

tests/test_fieldset.py
erikvansebille: Well, then better to throw an error if `field_chunksize` is not a tuple

parcels/particleset.py
CKehl: k, let's do that

PR #777, opened 2020-03-27
Implementing AnalyticalAdvection Kernel

erikvansebille:
This PR implements the Analytical Advection scheme of e.g. [TRACMASS](tracmass.org) in Parcels. 

Implemented:
- [x] Support for non-uniform flow fields
- [x] Support for curvilinear grids (spherical meshes)
- [x] Support for time-varying fields
- [x] Support for backward-in-time integration
- [x] Support for 3D (also in depth) advection; probably via a separate `AnalyticalAdvection3D` Kernel
- [x] Tutorial notebook

Note the following constraints in this implementation:
1. The velocity field should be defined on a C-grid (this will always be the case for Analytical Advection)
2. The AdvectionAnalytical kernel only works for `Scipy` Particles.
3. Since Analytical Advection does not use timestepping, the `dt` parameter in `pset.execute()` should be set to `np.inf`. For backward-in-time simulations, it should be set to `-np.inf`.
4. For time-varying fields, only the 'intermediate timesteps' scheme ([section 2.3 of Döös et al 2017](https://www.geosci-model-dev.net/10/1733/2017/gmd-10-1733-2017.pdf)) is implemented. While there is also a way to also analytically solve the time-evolving fields ([section 2.4 of Döös et al 2017](https://www.geosci-model-dev.net/10/1733/2017/gmd-10-1733-2017.pdf)), this is not yet implemented in Parcels.


parcels/field.py
```diff
@@ -721,6 +721,7 @@ def search_indices_curvilinear(self, x, y, z, xi, yi, ti=-1, time=-1, search2D=F
                          [1, -1, 1, -1]])
         maxIterSearch = 1e6
         it = 0
+        tol = 1.e-10
```
CKehl, line 724: from a computational perspective, I'm not a big advocate of tolerance values (as it suggests an algorithm is numerically unstable), but: if that is how it works in this scenario, then it works and that's fine.

parcels/kernel.py
```diff
@@ -300,9 +318,6 @@ def execute_python(self, pset, endtime, dt):
                     if res is ErrorCode.Success and particles.state != state_prev:
                         res = particles.state
 
-                    if res == ErrorCode.Success and not np.isclose(particles.dt, pdt_prekernels):
-                        res = ErrorCode.Repeat
-
```
CKehl, line 305: hm, clearing this line kind'a disallows kernel repetition. The line's intent is actually that, if a particle is not advected by dt in full (e.g if the write-out timer to file or the plotting timer is between `particles.time` and `particles.time+particles.dt`), then the computation is continued until a full time step has been performed. I see how that clashes with the analytical advection, but clearing out this line regardless of analytical or iterative optimization looks like a potential error source.

parcels/kernels/advection.py
```diff
@@ -100,3 +101,159 @@ def AdvectionRK45(particle, fieldset, time):
...
+            U0 = U0 * (1 - tau) + tau * direction * fieldset.U.data[ti+1, yi+1, xi] * c4 * dz
+            U1 = U1 * (1 - tau) + tau * direction * fieldset.U.data[ti+1, yi+1, xi+1] * c2 * dz
+            V0 = V0 * (1 - tau) + tau * direction * fieldset.V.data[ti+1, yi, xi+1] * c1 * dz
+            V1 = V1 * (1 - tau) + tau * direction * fieldset.V.data[ti+1, yi+1, xi+1] * c3 * dz
+
+    def compute_ds(F0, F1, r, direction, tol):
+        up = F0 * (1-r) + F1 * r
+        r_target = 1. if direction * up >= 0. else 0.
+        B = F0 - F1
+        delta = - F0 - B * 0.
```
CKehl, line 204: ehm, that statement I don't fully understand (or: the intent of it) - isn't that just `delta = -F0` ? Unless, of course, we are dealing with values close to the numerical accuracy, where `delta` could switch signs (iff F0 ~ +0, B ~ -0 and abs(B) > abs(F0))

parcels/kernels/advection.py
```diff
@@ -100,3 +101,159 @@ def AdvectionRK45(particle, fieldset, time):
...
+        elif B == 0:
+            ds = -(r_target - r) / delta
+        elif F_r1 * F_r0 <= 0:
+            ds = float('inf')
+        else:
+            ds = - 1. / B * math.log(F_r1 / F_r0)
+
+        if abs(ds) < tol:
+            ds = float('inf')
+        return ds, B, delta
```
CKehl, line 224: This is computationally dicy, judging convergence on user-provided tolerances (though admittedly 1e-10 is quite low), comparing floating-point limited-accuracy values on their equality or inequality to `.0`, and that in Python where - due to the typeless variables - all those math operations obey their internal tolerance cap. Another question: what happens if `delta`, `B`, `F1` or `F0` is actually once so close to absolute zero that they are actually `np.nan` ? That is, in such a geometrically sensitive scheme, a distinct possibility.

I mean: if it works, it works, and that's great and that's the fact I rely on. 
Judging from a theoretical and computational perspective, it doesn't look computationally robust and is prone to floating point-accuracy bound issues at some case. That would be particularly true if the cell sizes are very small (metre - and sub-metre range). 
But again, it may also just be a 'theoretical' caveat ...

tests/test_advection.py
```diff
@@ -364,3 +365,58 @@ def test_decaying_eddy(fieldset_decaying, mode, method, rtol, npart=1):
...
+        failed = True
+    assert failed
+
+
+@pytest.mark.parametrize('mode', ['scipy'])  # JIT not implemented
+@pytest.mark.parametrize('u', [1, -0.2, -0.3, 0])
+@pytest.mark.parametrize('v', [1, -0.3, 0, -1])
+@pytest.mark.parametrize('w', [None, 1, -0.3, 0, -1])
+@pytest.mark.parametrize('direction', [1, -1])
+def test_uniformanalytical(mode, u, v, w, direction, tmpdir):
```
CKehl, line 391: for readability and simplicity, could you rename `test_uniformanalytical` to `test_uniform_analytical` ? That would be nice.

tests/test_advection.py
```diff
@@ -364,3 +365,58 @@ def test_decaying_eddy(fieldset_decaying, mode, method, rtol, npart=1):
...
+    lon = np.arange(0, 15, dtype=np.float32)
+    lat = np.arange(0, 15, dtype=np.float32)
```
CKehl, line 393: with respect to the issue of computational accuracy, I think it would be more convincing if the lateral grid was small (the whole grid being perhaps 1m wide and tall, resolved in steps of 0.01m horizontally and vertically), so the kernel is also tested on interpolation values and this within the bounds where floating-point accuracy could result in failure.

just a suggestion.

parcels/kernels/advection.py
erikvansebille: Note that all these computations are done in the 'unit cell', where the length of each grid edge is 1. This is like in Figure 3b of @delandmeterp's paper ([see here](https://www.geosci-model-dev.net/12/3571/2019/gmd-12-3571-2019.html)). The conversion is done with the `px` and `py` at the beginning and end of the Kernel

So we don't need to worry about how this tolerance would work for very small grids (in fact, the new double gyre example in the tutorial notebook is on a 1x2 m grid).

parcels/kernels/advection.py
erikvansebille: I left the `B * 0.` in to stay close to the notation of Döös et al (2017); [their equation (17)](https://www.geosci-model-dev.net/10/1733/2017/gmd-10-1733-2017.pdf) where in our unit cell by definition r<sub>i-1</sub>=0.

But I agree it could only cause numerical errors, so will remove

parcels/kernel.py
erikvansebille: OK, I changed to 
```python
if not analytical and res == ErrorCode.Success and not np.isclose(particles.dt, pdt_prekernels):
    res = ErrorCode.Repeat
```

parcels/field.py
erikvansebille: This `tol` is what we also use in JIT interpolation, see e.g. https://github.com/OceanParcels/parcels/blob/master/parcels/include/index_search.h#L287

CKehl: agreed.

parcels/kernels/advection.py
```diff
@@ -100,3 +101,161 @@ def AdvectionRK45(particle, fieldset, time):
...
+            ds = float('inf')
+        elif B == 0:
+            ds = -(r_target - r) / delta
+        elif F_r1 * F_r0 < tol:
+            ds = float('inf')
+        else:
+            ds = - 1. / B * math.log(F_r1 / F_r0)
+
+        if abs(ds) < tol:
+            ds = float('inf')
```
CKehl, line 225: yes, that is numerically more robust that 'equal-zero' operations. Thanks for the change.

parcels/kernels/advection.py
```diff
@@ -100,3 +101,161 @@ def AdvectionRK45(particle, fieldset, time):
...
+        W1 = direction * fieldset.W.data[ti, zi+1, yi+1, xi+1] * dxdy
+        if withTime:
+            W0 = W0 * (1 - tau) + tau * direction * fieldset.W.data[ti+1, zi, yi + 1, xi + 1] * dxdy
+            W1 = W1 * (1 - tau) + tau * direction * fieldset.W.data[ti+1, zi + 1, yi + 1, xi + 1] * dxdy
+        ds_z, B_z, delta_z = compute_ds(W0, W1, zeta, direction, tol)
+    else:
+        ds_z = float('inf')
+
+    # take the minimum travel time
+    s_min = min(abs(ds_x), abs(ds_y), abs(ds_z), abs(ds_t / (dxdy * dz)))
```
CKehl, line 241: elegant formulation, with the jacobian

Issue #779, opened 2020-03-30
Option to stop the simulation on demand


miquelmassot:
First of all, thanks for this great software piece!

We are using OceanParcels for simulating underwater objects and we would like to have the ability to stop the simulation when a certain condition is reached. 

For example: how could the argo float example be stopped after drifting 10 km or after returning to the surface for the second time? Is there any other way to stop the simulation than running over the entire simulation time?

erikvansebille:
Very good idea, @miquelmassot; thanks for the suggestion!

I have now implemented an `ErrorCode.StopExecution` in PR #780. If you change to that branch, you can then add the following to your Kernel
```python
if particle.driftdistance > 10e3:
    return ErrorCode.StopExecution
```
This should do what you suggested (assuming of course that you keep track of `driftdistance` for your particle)

miquelmassot:
Thanks @erikvansebille! I will give it a try.

PR #780, opened 2020-03-31
Adding ErrorCode.StopExecution

erikvansebille:
Adding an ErrorCode that break the pset.execute() loop, as requested in #779. 

To use, add the following to a Kernel
```python
return ErrorCode.StopExecution
```


parcels/codegenerator.py
```diff
@@ -165,7 +165,8 @@ def __getattr__(self, attr):
 class ErrorCodeNode(IntrinsicNode):
     symbol_map = {'Success': 'SUCCESS', 'Evaluate': 'EVALUATE', 'Repeat': 'REPEAT', 'Delete': 'DELETE',
                   'Error': 'ERROR', 'ErrorInterpolation': 'ERROR_INTERPOLATION',
-                  'ErrorOutOfBounds': 'ERROR_OUT_OF_BOUNDS', 'ErrorThroughSurface': 'ERROR_THROUGH_SURFACE'}
+                  'ErrorOutOfBounds': 'ERROR_OUT_OF_BOUNDS', 'ErrorThroughSurface': 'ERROR_THROUGH_SURFACE',
+                  'StopExecution': 'STOP_EXECUTION'}
```
CKehl, line 169: this may not be the best (though it is the easiest) place to add that error. I think it would be better to have all the general, pipeline-related flags before 'Error' - even though it means shifting the error codes. I though understand that 'Success' is necessarily at position 0 because that is the standard return code of C for successful executions. Still, I would semantically not just append this flag at the end.

parcels/kernel.py
```diff
@@ -384,6 +384,8 @@ def remove_deleted(pset):
             # Apply recovery kernel
             for p in np.where(error_particles)[0]:
                 particles.set_index(p)
+                if particles.state == ErrorCode.StopExecution:
+                    return
```
CKehl, line 388: yes, simple and easy. That means that the manual stop of the execution still somehow needs to be attached to a particle, and it's limited kernel instruction set, right ? Still, it does match what #779 discusses and requests, and for that, this solution is nice.

PR #782, opened 2020-04-01
Fixing a bug with recent Structure-of-arrays PR and multiple grids

erikvansebille:
Since the recent #678, `ParticleSet.particle_data.xi` (and same for `yi`, `zi` and `ti`) is a vector of length of the `ParticleSet` instead of the length of the `Gridset`. This means that if multiple `Grids` are present, Parcels currently breaks.  

A solution could be to make `ParticleSet.particle_data.xi` an array of size `(Particleset.size, fieldset.gridset.size)`. However, I had a look but couldn't see where to make these changes. 

@angus-g, since you so wonderfully drafted most of the initial #678, could you also have a look at how to incorporate this change?


angus-g:
Hi @erikvansebille, I think your suggestion of changing the size of the index arrays is pretty sensible! I've made that change and pushed to my fork at https://github.com/angus-g/parcels/tree/soa_multigrid_indexing

I haven't run the full test suite, but this is failing the `test_pset_from_field` test because the index arrays are never assigned. I'm a little bit confused as to why this wasn't failing before?

erikvansebille:
Thanks @angus-g, for the quick response! Can you merge your changes into this PR, so that I can easily check them out locally and work on them too?

angus-g:
Unfortunately not, since I don't have commit access!

erikvansebille:
Thanks for committing this fix, @angus-g!

I think one of the reasons that some of the unit tests break is because of the code below:
https://github.com/OceanParcels/parcels/blob/soa_multigrid_indexing/parcels/particleset.py#L720-L726
In particular, if this is changed to
```python
        for i in range(self.size):
            p.set_index(i)
            # try:  # breaks if either p.xi, p.yi, p.zi, p.ti do not exist (in scipy) or field not in fieldset
            #     if p.ti[field.igrid] < 0:  # xi, yi, zi, ti, not initialised
            #         raise('error')
            #     xi = p.xi[field.igrid]
            #     yi = p.yi[field.igrid]
            # except:
            _, _, _, xi, yi, _ = field.search_indices(p.lon, p.lat, p.depth, 0, 0, search2D=True)
            density[yi, xi] += particle_val[i]
```
then the unit tests pass. Note however that removing the commented code altogether is not ideal, because it means that calculating `ParticleSet.density()` is much slower (as  field.search_indices is very slow). So ideally the commented lines are fixed. Do you see how, @angus-g?

angus-g:
I've just pushed a possible idea for this, which is a bit ugly. If we make a JIT kernel that samples the field, the indices get set correctly.

erikvansebille:
> If we make a JIT kernel that samples the field, the indices get set correctly.

Nice idea; I actually like it quite a lot!

parcels/particleset.py
```diff
@@ -700,7 +708,23 @@ def density(self, field=None, particle_val=None, relative=False, area_scale=Fals
...
+        """.format(field_name)
+
+        k = Kernel(
+            self.fieldset,
+            self.ptype,
+            funcname="search_kernel",
+            funcvars=["particle", "fieldset", "time", "x"],
+            funccode=f_str,
+        )
+        self.execute(k, runtime=0)
```
CKehl, line 726: Neat, to define the search kernel in C dynamically here.

parcels/particleset.py
```diff
@@ -490,7 +498,7 @@ def remove_indices(self, indices):
     def remove_booleanvector(self, indices):
         """Method to remove particles from the ParticleSet, based on an array of booleans"""
         for d in self.particle_data:
-            self.particle_data[d] = self.particle_data[d][~indices]
+            self.particle_data[d] = self.particle_data[d][..., ~indices]
```
CKehl, line 501: uncommon notation (thought this only works with ``...`` being at the end of a command) - nice that it works as simplification!

parcels/particleset.py
```diff
@@ -295,7 +298,12 @@ def ctypes_struct(self):
         class CParticles(Structure):
             _fields_ = [(v.name, POINTER(np.ctypeslib.as_ctypes_type(v.dtype))) for v in self.ptype.variables]
 
-        cdata = [np.ctypeslib.as_ctypes(self.particle_data[v.name]) for v in self.ptype.variables]
+        def cdata_for(v):
+            cdata_array = np.ctypeslib.as_ctypes(self.particle_data[v.name])
+            # cast to bare pointer
+            return cast(cdata_array, POINTER(np.ctypeslib.as_ctypes_type(v.dtype)))
```
CKehl, line 304: so, question for my understanding: are now all particle cdata pointers-of-base types ?

parcels/kernel.py
```diff
@@ -218,6 +218,9 @@ def load_lib(self):
 
     def execute_jit(self, pset, endtime, dt):
         """Invokes JIT engine to perform the core update loop"""
+        if len(pset) > 0 and pset.particle_data['xi'].ndim == 2:
+            assert pset.fieldset.gridset.size == pset.particle_data['xi'].shape[0], \
+                'FieldSet has different number of grids than Particle.xi. Have you added Fields after creating the ParticleSet?'
```
CKehl, line 223: yes, great - I also thought of re-introducing that check from earlier version. Good idea.

parcels/particleset.py
angus-g: That's correct

erikvansebille:
OK, so is this ready for merge when we get green ticks on the CI?

parcels/particleset.py
CKehl: k - sad that it makes thing that bit more complex to keep in mind, but I agree with your reasoning: this way seems the most applicable solution.

tests/test_fieldset_sampling.py
```diff
@@ -403,27 +405,50 @@ def test_sampling_out_of_bounds_time(mode, allow_time_extrapolation, k_sample_p,
...
+              lon=np.linspace(0., 1., xdim*10, dtype=np.float32),
+              lat=np.linspace(0., 1., ydim*10, dtype=np.float32))
+    fieldset.add_field(P)
+
+    fail = False
+    try:
+        pset.execute(AdvectionRK4, runtime=10, dt=1)
+    except:
+        fail = True
+    assert fail
```
CKehl, line 451: well, yes, that's a better way of writing the try-exception test than.I take that as inspiration.

PR #787, opened 2020-04-06
Expanding CI to run all unit tests on all platforms

erikvansebille:
This PR updates the Github CI commands so that all unit tests (both `tests/*.py`, `examples/*.py`, and `examples/*.ipynb`) are run on all three platforms (linux, macos and windows), for full unit test coverage

### DIFF
--- a/parcels/scripts/plottrajectoriesfile.py
+++ b/parcels/scripts/plottrajectoriesfile.py
@@ -13,7 +13,7 @@ from parcels.plotting import plotfield
 try:
     if sys.platform == 'darwin' and sys.version_info[0] == 3:
         import matplotlib
-        matplotlib.use("TkAgg")
+        matplotlib.use("Agg")
     import matplotlib.animation as animation
     from matplotlib import rc
 except:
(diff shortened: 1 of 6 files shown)


parcels/scripts/plottrajectoriesfile.py
```diff
@@ -13,7 +13,7 @@
 try:
     if sys.platform == 'darwin' and sys.version_info[0] == 3:
         import matplotlib
-        matplotlib.use("TkAgg")
+        matplotlib.use("Agg")
```
CKehl, line 16: hm - weird that TkAgg doesn't work - it's (for me for auto-generating the notebooks) actually the only thing that works consistently. Anyway, if this works - nice!

tests/test_kernel_execution.py
```diff
@@ -270,6 +271,7 @@ def simpleKernel(particle, fieldset, time):
 
 
 @pytest.mark.parametrize('delete_cfiles', [True, False])
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="skipping windows test as windows compiler generates warning")
```
CKehl, line 274: interesting compiler pragma - I'll make a note of this for myself.

Issue #791, opened 2020-04-08
Exception thrown by kernel finaliser


angus-g:
The `Kernel` class defines a `__del__` finaliser, which is meant to unload the JIT module and delete its files. This seems to work in a lot of cases, but can throw the following exception:

```
Exception ignored in: <bound method Kernel.__del__ of <parcels.kernel.Kernel object at 0x7fc0b0875c50>>
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/parcels/kernel.py", line 172, in __del__
TypeError: 'NoneType' object is not callable
```

From the [CPython documentation](https://docs.python.org/3/reference/datamodel.html#object.__del__):

> `__del__()` can be executed during interpreter shutdown. As a consequence, the global variables it needs to access (including other modules) may already have been deleted or set to `None`. Python guarantees that globals whose name begins with a single underscore are deleted from their module before other globals are deleted; if no other references to such globals exist, this may help in assuring that imported modules are still available at the time when the `__del__()` method is called

In this particular case, `Kernel.__del__()` calls functions from the `_ctypes` module, which is deleted first, according to the snippet above. Maybe it would be possible to qualified-import `_ctypes` to something that doesn't begin with an underscore, but I'm not really sure of the right solution here.

erikvansebille:
Fixed in #794

PR #792, opened 2020-04-08
Qualified importing _ctypes

erikvansebille:
To avoid error on `Kernel.__del__`, see also #791

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -1,4 +1,4 @@
-import _ctypes
+import _ctypes as ctp
 import inspect
 import math  # noqa
 import random  # noqa
@@ -169,7 +169,7 @@ class Kernel(object):
         # This is not really necessary, as these programs are not that large, but with the new random
         # naming scheme which is required on Windows OS'es to deal with updates to a Parcels' kernel.
         if self._lib is not None:
-            _ctypes.FreeLibrary(self._lib._handle) if platform == 'win32' else _ctypes.dlclose(self._lib._handle)
+            ctp.FreeLibrary(self._lib._handle) if platform == 'win32' else ctp.dlclose(self._lib._handle)
             del self._lib
             self._lib = None
             if path.isfile(self.lib_file) and self.delete_cfiles:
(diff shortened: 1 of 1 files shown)


angus-g:
Sorry to be a bother on this, but this doesn't seem to have fixed the error on my travis build! Any thoughts?

angus-g:
Maybe as an alternative:

```python
import weakref
...
class Kernel(object):
...
    # old __del__:
    def _cleanup(self):
        if self._lib is not None:
            print("cleaning up, ctp:", ctp)
            ctp.FreeLibrary(self._lib._handle) if platform == 'win32' else ctp.dlclose(self._lib._handle)
            del self._lib
            self._lib = None
            if path.isfile(self.lib_file) and self.delete_cfiles:
                [remove(s) for s in [self.src_file, self.lib_file, self.log_file]]

    def compile(self, compiler):
        """ Writes kernel code to file and compiles it."""
        with open(self.src_file, 'w') as f:
            f.write(self.ccode)
        compiler.compile(self.src_file, self.lib_file, self.log_file)
        logger.info("Compiled %s ==> %s" % (self.name, self.lib_file))

        weakref.finalize(self, self._cleanup)
```

From the docs for `weakref.finalize`:
> Return a callable finalizer object which will be called when obj is garbage collected. Unlike an ordinary weak reference, a finalizer will always survive until the reference object is collected, greatly simplifying lifecycle management.

> A finalizer will never invoke its callback during the later part of the interpreter shutdown when module globals are liable to have been replaced by None.

But perhaps the finaliser should be registered on the `Kernel` object, and it can be marked dead with `finalizer.detach()` if `Kernel.remove_lib()` is called?

Thoughts?

erikvansebille:
Does that solution with `finalizer` work on your Travis? If so, then it might be a good solution yes!

Note that we could also use a similar approach in `ParticleFile.__del__`, which can experience the same problem with unlinked objects that cannot be deleted

angus-g:
Just tested, and yes, using the finalizer seems to clean up correctly (no files left in `/tmp/parcels-1000/...`) and not throw errors, both on my test container and Travis. I think it might be a more robust solution for `ParticleFile` too: I think if you explicitly call the finalizer object, it will be dead and won't be called again, so it implicitly has the `convert_at_end` logic.

I can create a new PR for my `Kernel` changes?

erikvansebille:
> Just tested, and yes, using the finalizer seems to clean up correctly (no files left in `/tmp/parcels-1000/...`) and not throw errors, both on my test container and Travis. I think it might be a more robust solution for `ParticleFile` too: I think if you explicitly call the finalizer object, it will be dead and won't be called again, so it implicitly has the `convert_at_end` logic.

OK, great. That's a good solution then

> I can create a new PR for my `Kernel` changes?

Yes please

erikvansebille:
Fixed in #794

PR #794, opened 2020-04-08
Use weakref.finalize in Kernel

angus-g:
As discussed in #792, this seems like a more robust solution to cleaning up objects when they get deleted. Documentation for `weakref.finalize` is at https://docs.python.org/3/library/weakref.html#weakref.finalize

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -15,6 +15,7 @@ from os import path
 from os import remove
 from sys import platform
 from sys import version_info
+from weakref import finalize
 
 import numpy as np
 import numpy.ctypeslib as npct
@@ -73,6 +74,8 @@ class Kernel(object):
         self.ptype = ptype
         self._lib = None
         self.delete_cfiles = delete_cfiles
+        self._cleanup_files = None
+        self._cleanup_lib = None
 
         # Derive meta information from pyfunc, if not given
         self.funcname = funcname or pyfunc.__name__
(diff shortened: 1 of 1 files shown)


angus-g:
I had to do a few tweaks to this, because the finaliser can't have any references to the `Kernel` itself, or it'll never be garbage-collected until interpreter shutdown! It's a bit more verbose, but hopefully robust in this form.

angus-g:
Do you mean the current behaviour is as you posted, but ideally wouldn't raise an error? I can do that and tack that onto this PR

erikvansebille:
Yes exactly. We now get this error, but the code shouldn't give an error altogether. Sorry for the confusion

angus-g:
Gotcha!

angus-g:
I had a further think about applying the same approach to `ParticleFile`, but I think it could be a bit more complicated. Because the finaliser can't refer to the object or any of its methods, the export logic would have to become a (static) class method, and the variables it needs like `self.var_names` would have to be stored on the finaliser directly.

It seems like the context manager approach would be more appropriate in this case. It would look something like the following: define `__enter__` and `__exit__` methods as

```python
class ParticleFile(object):
    ...
    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        if exc_type is not None:
            return False # reraise any exceptions
        if self.convert_at_end:
            self.close()
    ...
````
        
and the call would turn into something like:

```python
filepath = "pfile_closeondelete.nc"
pset = ParticleSet(fieldset(), pclass=JITParticle, lon=0, lat=0)
with pset.ParticleFile(filepath, outputdt=1) as pfile:
    pset.execute(AdvectionRK4, runtime=2, dt=1, output_file=pfile)
```

erikvansebille:
OK thanks for investigating the `ParticleFile` finaliser too. In that case, i suggest we merge this solution to the `Kernel` finalizer, and leave the  `ParticleFile` finaliser for now. Note that a simple workaround is to issue `pfile.close()` at the end of the script.

Issue #803, opened 2020-04-14
Field sampling near coast


pierrick-giffard:
I noticed that when particles are close to coast (I mean in the same cell that landmask, see first figure), field sampling is computed based on interpolation of oceanic true values but also based on zero values (because of land). Here for my example with temperature it results in particulary low temperatures near coast (second figure). A way to fix this might be to interpolate only with oceanic points available or with nearest neighbour ?

![Traj_temperature](https://user-images.githubusercontent.com/56306211/79207330-e1d41200-7e40-11ea-876a-4823d7db68a4.PNG)

![Temperature](https://user-images.githubusercontent.com/56306211/79207608-3081ac00-7e41-11ea-86c1-3c7cd2078a58.PNG)

erikvansebille:
Thanks for reporting this Issue, @pierrick-giffard. This problem of interpolating near land indeed pops up quite frequently in my own research too,  and I don't have a perfect solution. I do have a few thoughts, though

1) The simplest 'solution' is to set the `interp_method` of your temperature field to `nearest`. You will then be able to remove all the locations where the particle temperature is zero in post-processing. However, this is not ideal because you don't go to the root of the cause (there will still be 'zeros', they're just easier to identify now) and you also have to use an interpolation method that you don't want to use * ...

2) Write a new interpolation method that 'ignores' zeros of land values. However, I don't think it's trivial what this interpolation method would look like, let alone implement it into [parcels.h](https://github.com/OceanParcels/parcels/blob/master/parcels/include/parcels.h).

3) Change your `Field` data and change the zeros to e.g. the closest non-zero value. The tricky thing here is that you're probably using deferred-loading of the `FieldSet`, so you have to do this 'on the fly'. While that is in principle possible in Parcels (using the `FieldSet.compute_on_defer` method that is used e.g. [here](https://github.com/OceanParcels/parcels/blob/master/tests/test_fieldset.py#L665-L673)), this is not for the faint-at-heart. It would also likely come at a considerable efficiency loss

So in short, there is no 'perfect' solution, I think. Unless you have an equation in mind you would want to use for the interpolation, then we could try option 2)

_* note that nearest interpolation is the default tracer interpolation for C-grids by the way, see [@delandmeterp's paper](https://www.geosci-model-dev.net/12/3571/2019/gmd-12-3571-2019.html) for an explanation why_

pierrick-giffard:
[...]

PR #815, opened 2020-04-22
Interpolation near coast

pierrick-giffard:


pierrick-giffard:
Yhanks @angus-g for your corrections. I'll make the 3D function as soon as possible.

erikvansebille:
Thanks for writing these C-functions, @pierrick-giffard 

I've now incorporated them into parcels. They can be set with `interp_method=linear_invdist_land_tracer`. I've also made a very rudimentary unit test to check if the method works as expected. Things to do now

- [ ] Develop a better/more sensitive unit test
- [ ] Add warnings/errors when `interp_method=linear_invdist_land_tracer` is used on velocities (?)
- [ ] Add documentation (both in docstring and perhaps in notebook tutorial)
- [ ] Implement SciPy equivalent to interpolation method (?)

CKehl:
looks conceptually similar to the [Natural Neighbour Interpolation](https://en.wikipedia.org/wiki/Natural_neighbor_interpolation)

parcels/include/parcels.h
```diff
@@ -73,6 +73,65 @@ static inline ErrorCode spatial_interpolation_bilinear(double xsi, double eta, f
...
 
+/* Bilinear interpolation routine for 2D grid for tracers with inverse distance weighting near land*/
+static inline ErrorCode spatial_interpolation_bilinear_invdist_land(double xsi, double eta, float data[2][2], float *value)
+{
+  int k, l, nb_land = 0, land[2][2] = {{0}};
+  float weight[2][2] = {{0.}}, w_sum = 0.;
+  // count the number of surrounding land points (assume land is where the value is close to zero)
+  for (int i = 0; i < 2; i++) {
+    for (int j = 0; j < 2; j++) {
+      if (fabs(data[i][j]) < 1e-14) {
```
CKehl, line 84: is this 'new' tolerance threshold required ? Also, `data` is a float, so `1e-14` is beyond the numerical resolution of `data`. How about using either 
```
if(fabs(data[i][j] < FLT_EPSILON) { ... }
```
or the pre-build function:
```
if(is_zero_flt(data[i][j])) { ... }
```
?

parcels/include/parcels.h
```diff
@@ -73,6 +73,65 @@ static inline ErrorCode spatial_interpolation_bilinear(double xsi, double eta, f
...
+  case 4:  // only land
+    return ERROR_INTERPOLATION;
+  default:
+    break;
+  }
+  // interpolate with 1 or 2 land points
+  for (int i = 0; i < 2; i++) {
+    for (int j = 0; j < 2; j++) {
+      float distance = sqrt(pow((xsi - j), 2) + pow((eta - i), 2));
+      if (fabs(distance) < 1e-14) {
```
CKehl, line 111: Here, again, I'd prefer using `if(is_zero_flt(distance)) { ... }` to avoid extra tolerance thresholds.

parcels/include/parcels.h
```diff
@@ -73,6 +73,65 @@ static inline ErrorCode spatial_interpolation_bilinear(double xsi, double eta, f
   return SUCCESS;
 }
 
+/* Bilinear interpolation routine for 2D grid for tracers with inverse distance weighting near land*/
+static inline ErrorCode spatial_interpolation_bilinear_invdist_land(double xsi, double eta, float data[2][2], float *value)
```
CKehl, line 77: Looks very good to me - also the inverse distance-weighting scheme for the interpolation is a nice idea in this sort of scenario.

parcels/include/parcels.h
```diff
@@ -90,6 +149,72 @@ static inline ErrorCode spatial_interpolation_trilinear(double xsi, double eta,
...
+/* Trilinear interpolation routine for 3D grid for tracers with inverse distance weighting near land*/
+static inline ErrorCode spatial_interpolation_trilinear_invdist_land(double xsi, double eta, double zeta, float data[2][2][2], float *value)
+{
+  int l, m, n, nb_land = 0, land[2][2][2] = {{{0}}};
+  float weight[2][2][2] = {{{0.}}}, w_sum = 0.;
+  // count the number of surrounding land points (assume land is where the value is close to zero)
+  for (int i = 0; i < 2; i++) {
+    for (int j = 0; j < 2; j++) {
+      for (int k = 0; k < 2; k++) {  
+        if (fabs(data[i][j][k]) < 1e-14) {
```
CKehl, line 161: comment as in 2D - see l. 84

parcels/include/parcels.h
```diff
@@ -90,6 +149,72 @@ static inline ErrorCode spatial_interpolation_trilinear(double xsi, double eta,
...
+    return ERROR_INTERPOLATION;
+  default:
+    break;
+  }
+  // interpolate with 1 to 6 land points
+  for (int i = 0; i < 2; i++) {
+    for (int j = 0; j < 2; j++) {
+        for (int k = 0; k < 2; k++) {  
+          float distance = sqrt(pow((zeta - i), 2) + pow((eta - j), 2) + pow((xsi - k), 2));
+          if (fabs(distance) < 1e-14) {
```
CKehl, line 191: comment as in 2D - see l. 111

angus-g:
> the square-root on the distance being computed 4*N per run is a bit expensive

Is the square-root needed? Couldn't we get away with working with the squared distance?

pierrick-giffard:
One note about the square root. I realized that my figure done here https://github.com/OceanParcels/parcels/issues/803#issuecomment-617260630 was computed without the square root (it means working with squared distance). I just tried with and without square root. The results are pretty close, and working with the squared distance seemed slightly better. In MITAS and MITASOVA 1998 it is written inverse distance weighting is typically used with square distance. Though, I propose to remove the square root.

parcels/include/parcels.h
```diff
@@ -73,6 +73,65 @@ static inline ErrorCode spatial_interpolation_bilinear(double xsi, double eta, f
...
+    }
+  }
+  switch (nb_land) {
+  case 0:  // no land, use usual routine
+    return spatial_interpolation_bilinear(xsi, eta, data, value);
+  case 3:  // single non-land point
+    *value = data[k][l];
+    return SUCCESS;
+  case 4:  // only land
+    return 0.;
```
erikvansebille, line 103: This `return 0` is not the cleanest way to do this. In parcels, the way you would normally handle these `ERROR_INTERPOLATION` is with a recovery kernel. So I would suggest to keep `return ERROR_INTERPOLATION`

(by the way, also note that since these errorcodes are simple enums, `return 0` is the same as `return SUCCESS`; see https://github.com/OceanParcels/parcels/blob/master/parcels/include/index_search.h#L19-L22)

PR #817, opened 2020-04-22
Support for Fields with longitude or latitude of length 1

### LINKED ISSUE #784: Support for Parcels in 1D simulations
Parcels now requires Fields with at least two grid points in longitude and two grid points in latitude. For many idealised experiments (e.g. one-dimensional vertical simulations), this is overly cumbersome. 

It would be good to remove this constraint and support grids with only one longitude and/or latitude.

erikvansebille:
Until now, Parcels requires longitude and latitudes vectors to have a length of at least 2. For some cases, however, this is not ideal. For example for Fields in the x-z plane (i.e. with no latitude), or for Fields that are constant in space.

With this PR, users can also create Fields from data that have only 1 longitude or latitude value. Parcels will then automatically extrapolate in this dimensions, so users should never get an `ErrorOutOfBounds` in e.g. Longitude if the `xdim==1`

Note that this does not work with `dask` chunking or C-grids, but since the usecase is really only for `Field.from_data()` or `FieldSet.from_data()` for now, that should not be an issue for most uses

As an example, adding a contant diffusivity `Field` to an existing `fieldset` is now simply

```python
fieldset.add_field(Field('Kh_zonal', 100, lon=0, lat=0, mesh=fieldset.U.grid.mesh))
fieldset.add_field(Field('Kh_meridional', 100, lon=0, lat=0, mesh=fieldset.U.grid.mesh))
```

This fixes #784


parcels/field.py
```diff
@@ -636,61 +640,66 @@ def reconnect_bnd_indices(self, xi, yi, xdim, ydim, sphere_mesh):
 
     def search_indices_rectilinear(self, x, y, z, ti=-1, time=-1, search2D=False):
```
CKehl, line 641: Okay, if I understand the function correctly, it transforms the continuous grid-bound indices x and y in barymetric coordinates (xsi, eta, zeta) .. that was helpful for understanding.

parcels/include/parcels.h
```diff
@@ -208,9 +208,19 @@ static inline ErrorCode getCell2D(CField *f, int xi, int yi, int ti, float cell_
...
+        for (xii=0; xii<2; ++xii){
+          if (xdim == 1)
+            xiid = 0;
+          else
+            xiid = xii;
+          if (ydim == 1)
+            yiid = 0;
+          else
+            yiid = yii;
+          cell_data[tii][yii][xii] = data[ilocal[0]+yiid][ilocal[1]+xiid];
```
CKehl, line 221: these if-statements for such a low-level, many-call function may inhibit performance as it cripples CPU branch prediction. A better formulation to achieve the same:
```
      int xiid = ((xdim==1) ? 1 : 0);
      int yiid = ((ydim==1) ? 1 : 0);
      for (yii=0; yii<2; ++yii)
        for (xii=0; xii<2; ++xii)
          cell_data[tii][yii][xii] = data[ilocal[0]+(yii*yiid)][ilocal[1]+(xii*xiid)];
```
from a machine instruction perspective, this is faster.

parcels/include/parcels.h
```diff
@@ -303,10 +313,25 @@ static inline ErrorCode getCell3D(CField *f, int xi, int yi, int zi, int ti, flo
...
+              xiid = xii;
+            if (ydim == 1)
+              yiid = 0;
+            else
+              yiid = yii;
+            if (zdim == 1)
+              ziid = 0;
+            else
+              ziid = zii;
+            cell_data[tii][zii][yii][xii] = data[ilocal[0]+ziid][ilocal[1]+yiid][ilocal[2]+xiid];
```
CKehl, line 331: analogous to comment l. 211 - 221

parcels/include/parcels.h
erikvansebille: Thanks, your suggestion indeed is much cleaner and nicer! My C is a bit rusty ;-)

CKehl: You're welcome. Also good that you spotted my mistake in the conditional statement

PR #823, opened 2020-04-24
Improved Diffusion

daanreijnders:
Diffusion in Parcels used to be implemented in two ways: 1) a simple random perturbation when the diffusivity is constant everywhere 2) an Euler-Maruyama implementation without the advective part (including the 'boost' out of areas of low diffusivity) called `SpatiallyVaryingBrownianMotion2D`. 

This pull request adds the Milstein advection-diffusion scheme of order 1 (see Kloeden and Platen, 2010), which is much more accurate than the Euler-Maruyama scheme, especially when the diffusivity field contains sharp gradients (Gräwe et al., 2012). The Milstein-1 scheme comes with a small computational cost from an extra correction term, but can resolve a similar accuracy as the Euler-Maruyama scheme with a larger time steps such that it Milstein-1 scheme is much more efficient and preferable over Euler-Maruyama.

In this pull request, `diffusion.py` is replaced with `advectiondiffusion.py`, containing kernels that solve advection and diffusion at the same time. This is important, since the chaining of seperate advection and diffusion kernels causes the velocity for advection to be sampled at a different location that the diffusivity for the diffusion step. Therefore, the `SpatiallyVaryingBrownianMotion2D` Euler-Maruyama (EM) scheme is renamed `AdvectionDiffusionEM` and includes Euler-forward advection. If fourth-order Runge-Kutta advection is desired, this is available in the `AdvectionRK4DiffusionEM` kernel. Note that the error in the diffusion is of a lower order than that stemming from RK4 advection. Still, RK4 can be useful in cases where the diffusivity is low. Similarly, for the Milstein-1 scheme (M1), the `AdvectionDiffusionM1` and `AdvectionRK4DiffusionM1` kernels are available.

## Other changes
 - `BrownianMotion2D` is called `DiffusionUniformKh`, stressing its applicability only when Kh is uniform.
 - A constant `dres` needs to be added to the fieldset. `dres` is the spatial resolution at which the central difference is taken to compute the local gradient in diffusivity (and joint terms in Milstein-1). This value used to be hard-coded in the kernel, but the use of a constant makes the user mindful that care must be taken in choosing a value.
 - An reference is added in the docstring indicating the higher efficiency of Milstein-1 over Euler-Maruyama, linking to Gräwe et al. (2012).
 - The docstring contains an explanation on the requirements for the Wiener increment `dW`. If convergence of particle paths is necessary, the user can adapt the kernel manually to use Gaussian random variables (e.g. using `random.normalvariate` with a standard deviation of `particle.dt`).

## References
Gräwe, U., Deleersnijder, E., Shah, S. H. A. M., & Heemink, A. W. (2012). Why the Euler scheme in particle tracking is not enough: The shallow-sea pycnocline test case. Ocean Dynamics, 62(4), 501–514. https://doi.org/10.1007/s10236-012-0523-y

Kloeden, P. E., & Platen, E. (2010). Numerical solution of stochastic differential equations. Springer.


daanreijnders:
# To be added
 - A tutorial notebook explaining when to use which diffusivity kernels, including some caveats.
 - A test case for the advection part of the advection-diffusion kernels

parcels/kernels/advectiondiffusion.py
```diff
@@ -0,0 +1,230 @@
...
+__all__ = ['DiffusionUniformKh', 'AdvectionDiffusionEuler', 'AdvectionDiffusionMilstein1',
+           'AdvectionRK4DiffusionEuler', 'AdvectionRK4DiffusionMilstein1']
+
+
+def DiffusionUniformKh(particle, fieldset, time):
+    """Kernel for diffusion where diffusivity (Kh) is assumed uniform.
+    Assumes that fieldset has fields Kh_zonal and Kh_meridional
+
+    This kernel neglects gradients in the diffusivity field and is
+    therefore more efficient in cases with constant diffusivity.
```
erikvansebille, line 16:
```suggestion
    therefore more efficient in cases with uniform diffusivity.
```

parcels/kernels/advectiondiffusion.py
```diff
@@ -0,0 +1,230 @@
...
+
+    bx = math.sqrt(2 * fieldset.Kh_zonal[time, particle.depth, particle.lat, particle.lon])
+    by = math.sqrt(2 * fieldset.Kh_meridional[time, particle.depth, particle.lat, particle.lon])
+
+    particle.lon += bx * dWx
+    particle.lat += by * dWy
+
+
+def AdvectionDiffusionEuler(particle, fieldset, time):
+    """Kernel for simple advection-diffusion, solved using the Euler-Maruyama
```
erikvansebille, line 42: But confusing that Kernel name is `AdvectionDiffusionEuler`; wouldn't `AdvectionDiffusionEulerMaruyama` be clearer?

parcels/kernels/advectiondiffusion.py
```diff
@@ -0,0 +1,230 @@
...
+    byp1 = math.sqrt(2 * Kyp1)
+    bym1 = math.sqrt(2 * Kym1)
+    dbdy = (byp1 - bym1) / (2 * fieldset.dres)
+
+    ay = fieldset.V[time, particle.depth, particle.lat, particle.lon] + dKdy
+    by = math.sqrt(2 * fieldset.Kh_meridional[time, particle.depth, particle.lat, particle.lon])
+
+    # Particle positions are updated only after evaluating all terms.
+    particle.lon += ax * particle.dt + bx * dWx + \
+        0.5 * bx * dbdx * (dWx**2 - particle.dt)
```
erikvansebille, line 128: Does `**2` work in JIT? Normally, we use `math.pow(.., 2)` in Kernels.

parcels/kernels/advectiondiffusion.py
```diff
@@ -0,0 +1,230 @@
...
+    Kym1 = fieldset.Kh_meridional[time, particle.depth, particle.lat - fieldset.dres, particle.lon]
+    dKdy = (Kyp1 - Kym1) / (2 * fieldset.dres)
+    by = math.sqrt(2 * fieldset.Kh_meridional[time, particle.depth, particle.lat, particle.lon])
+
+    # Particle positions are updated only after evaluating all terms.
+    particle.lon += ((u1 + 2 * u2 + 2 * u3 + u4) / 6. + dKdx) * particle.dt + bx * dWx
+    particle.lat += ((v1 + 2 * v2 + 2 * v3 + v4) / 6. + dKdy) * particle.dt + by * dWy
+
+
+def AdvectionRK4DiffusionMilstein1(particle, fieldset, time):
```
erikvansebille, line 179: Would it make sense to also add 3D version of these Kernels? Or is that overkill?

parcels/kernels/advectiondiffusion.py
daanreijnders: In the literature it's referred to both as Euler and as Euler-Maruyama. I think the Maruyama part stresses that we're dealing with SDEs, but I figured Euler in the kernel name is nice and short...

parcels/kernels/advectiondiffusion.py
daanreijnders: Certainly no overkill. Just only did 2D now since the previous kernels were also just 2D. Since it'd take some time (also to develop testing), I'll do so in a new PR.

parcels/kernels/advectiondiffusion.py
daanreijnders: Yes, seems to give no trouble. Particle trajectories are the same in JIT and Scipy for my test runs.

parcels/kernels/advectiondiffusion.py
erikvansebille: OK, let' leave as is then

daanreijnders: I'll rename it to EM

mathildejutras:
Hi. Is there a plan to have the diffusion schemes work in 3D, i.e. with 3D velocity fields?

daanreijnders:
> Hi. Is there a plan to have the diffusion schemes work in 3D, i.e. with 3D velocity fields?

Yes, but on the longer term. See #849 and #858 for related discussions. Thanks to the possibility of creating custom kernels, you can for now adapt the existing kernels to work in 3D, choosing a different `dres` for the vertical direction. See the [documentation for the current structure of the advection-diffusion kernels](https://oceanparcels.org/gh-pages/html/#parcels-kernels-advectiondiffusion-module) and see this [tutorial about creating custom kernels](http://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_Argofloats.ipynb).

mathildejutras:
Great thank you very much for the information.

PR #831, opened 2020-05-06
Fixing a bug with length-1 dimensions

erikvansebille:
Fixing a bug when Field created with length-1 longitude and latitude, but with full depth dimension

Thanks to @dlobelle for finding this bug

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -474,16 +474,15 @@ class Field(object):
             logger.warning_once("Casting field data to np.float32")
             data = data.astype(np.float32)
         lib = np if isinstance(data, np.ndarray) else da
-        if data.size == 1:
-            data = lib.tile(data, [self.grid.tdim, self.grid.zdim, self.grid.ydim, self.grid.xdim])
         if transpose:
             data = lib.transpose(data)
         if self.grid.lat_flipped:
             data = lib.flip(data, axis=-2)
 
-        if len(data.shape) == 1:
-            if self.grid.zdim > 1 and self.grid.xdim == 1 and self.grid.ydim == 1:
-                data = data[:, None, None]  # if vertical section
+        if self.grid.xdim == 1:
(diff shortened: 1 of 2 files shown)


parcels/field.py
```diff
@@ -481,6 +481,9 @@ def reshape(self, data, transpose=False):
         if self.grid.lat_flipped:
             data = lib.flip(data, axis=-2)
 
+        if len(data.shape) == 1:
+            if self.grid.zdim > 1 and self.grid.xdim == 1 and self.grid.ydim == 1:
+                data = data[:, None, None]  # if vertical section
```
CKehl, line 486: why is the `data = data[:, None, None]` instead of `data = data[:, 1, 1]` ? Doesn't the non-notation cause issues when start computing the field on execute (throwing an index error in Field::computeTimeChunk) ?

erikvansebille: It's instead of `np.newaxis`, see e.g. https://stackoverflow.com/questions/26166471/3d-tiling-of-a-numpy-array

CKehl: Okay, I see. well, if that's pythonically fine, then I'm okay. It looks a but visually disturbing to me, but that's probably just different perspectives of C++ and Python coders. Test comply, so the fix is good to go.

erikvansebille:
Actually, it's not a very clean implementation. I just recoded with `np.expand_dims()`, see https://github.com/OceanParcels/parcels/pull/831/commits/c3a61e8d13c24bd52dbca20836b87c716017a49a. Is this better, @CKehl?

CKehl:
perhaps better change back to the previous code - it looked discomforting, but if that worked and this explicit dimension extension doesn't, then changing it would be obvious

erikvansebille:
I needed to also check if the data is not already 4D. Fixed it this morning and now it seems to work. I'll merge

PR #832, opened 2020-05-08
Fixing another bug in length-1 dimensions

erikvansebille:
This PR fixes another small bug in the length1-dimensions, found by @daanreijnders

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -479,6 +479,7 @@ class Field(object):
         if self.grid.lat_flipped:
             data = lib.flip(data, axis=-2)
 
+        data = np.squeeze(data)  # First remove all length-1 dimensions in data, so that we can add them below
         if self.grid.xdim == 1 and len(data.shape) < 4:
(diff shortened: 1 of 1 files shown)


parcels/field.py
```diff
@@ -479,6 +479,7 @@ def reshape(self, data, transpose=False):
         if self.grid.lat_flipped:
             data = lib.flip(data, axis=-2)
 
+        data = np.squeeze(data)  # First remove all length-1 dimensions in data, so that we can add them below
```
CKehl, line 482: nope - I disagree. we should use `lib.squeeze` instead - or use `lib.reshape` and determine which dimension to clear manually. Using `np.squeeze` with initialize the reading operation that we do not want to enact at that point (at least when using dask)

erikvansebille: Oops, good catch! You're absolutely right that we need to do this with `lib`! I've now updated the PR, and for safety also embedded the squeeze only if xdim or ydim is length 1; otherwise we're in the 'normal' scenario of full fields

Issue #846, opened 2020-05-28
Particle set from particle file doesn't include last values of custom variables


stellema:
Is it possible to include the final values of custom particle variables when using `ParticleSet.from_particlefile`? I noticed that only default particle variables (e.g. time, depth, lat) are carried over to the `ParticleSet`, while custom variables go back to their initial value.
For example, in the [Calculating distance travelled](https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/parcels_tutorial.ipynb#Calculating-distance-travelled) example at the end of `parcels_tutorial.ipynb`, the output of pset is:
```
P[0](lon=27.875189, lat=-33.055775, depth=0.000000, distance=13.197482, time=518400.000000)
P[1](lon=22.721094, lat=-36.037327, depth=0.000000, distance=641.251831, time=518400.000000)
P[2](lon=24.074530, lat=-35.525574, depth=0.000000, distance=543.712524, time=518400.000000)
P[3](lon=28.883598, lat=-34.525509, depth=0.000000, distance=183.668457, time=518400.000000)
P[4](lon=30.882704, lat=-34.146477, depth=0.000000, distance=172.781235, time=518400.000000)`
```

However, the custom "distance" variable goes back to the initial value of 0. in the newly created `ParticleSet`:

`ParticleSet.from_particlefile(fieldset, filename="GlobCurrentParticles_Dist.nc", pclass=DistParticle, restart=True)`
```
P[5](lon=27.875189, lat=-33.055775, depth=0.000000, distance=0.000000, time=518400.000000)
P[6](lon=22.721094, lat=-36.037327, depth=0.000000, distance=0.000000, time=518400.000000)
P[7](lon=24.074530, lat=-35.525574, depth=0.000000, distance=0.000000, time=518400.000000)
P[8](lon=28.883598, lat=-34.525509, depth=0.000000, distance=0.000000, time=518400.000000)
P[9](lon=30.882704, lat=-34.146477, depth=0.000000, distance=0.000000, time=518400.000000)
```

erikvansebille:
Good point, I'll implement asap! Should not be too difficult; stay tuned

erikvansebille:
I've now implemented it in #847. Could you check if this work for you @stellema?

stellema:
Thanks! It works when all variables are written to the output file. 
[...]

PR #847, opened 2020-05-29
Adding support for Custom Variables in ParticleSet.from_particlefile()

erikvansebille:
This is an implementation of the request at #846

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -428,6 +428,11 @@ class ParticleSet(object):
                and np.float64 if the interpolation method is 'cgrid_velocity'
         """
 
+        if repeatdt is not None:
+            logger.warning('Note that the `repeatdt` argument is not retained from %s, and that '
+                           'setting a new repeatdt will start particles from the _new_ particle '
+                           'locations.' % filename)
+
         pfile = xr.open_dataset(str(filename), decode_cf=True)
         pfile_vars = [v for v in pfile.data_vars]
 
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -408,7 +408,7 @@ class ParticleSet(object):
         return cls(fieldset=fieldset, pclass=pclass, lon=lon, lat=lat, depth=depth, time=time, lonlatdepth_dtype=lonlatdepth_dtype, repeatdt=repeatdt)
 
     @classmethod
-    def from_particlefile(cls, fieldset, pclass, filename, restart=True, restarttime=None, repeatdt=None, lonlatdepth_dtype=None):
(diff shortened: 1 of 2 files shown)


erikvansebille:
Good point about the `repeatdt`, @ckehl. I have now added a check to the unit test. 

It is important to point out that users should (again) set repeatdt when they create the new `pset`, because this information is not stored in the ParticleFile. This also means, if they do set the `repeatdt` again for the new `pset`, the new particles will be repeatedly released from the _new_ locations. I have now also added a warning message to clarify this

parcels/particleset.py
```diff
@@ -428,36 +428,47 @@ def from_particlefile(cls, fieldset, pclass, filename, restart=True, restarttime
                and np.float64 if the interpolation method is 'cgrid_velocity'
         """
 
+        if repeatdt is not None:
+            logger.warning('Note that the `repeatdt` argument is not retained from %s, and that '
+                           'setting a new repeatdt will start particles from the _new_ particle '
+                           'locations.' % filename)
```
CKehl, line 434: yes, nice feedback as reminder.

tests/test_particle_sets.py
```diff
@@ -67,8 +67,10 @@ def test_pset_create_fromparticlefile(fieldset, mode, restart, tmpdir):
 
     class TestParticle(ptype[mode]):
         p = Variable('p', np.float32, initial=0.33)
+        p2 = Variable('p2', np.float32, initial=1, to_write=False)
+        p3 = Variable('p3', np.float32, to_write='once')
```
CKehl, line 71: Adding this in the TestParticle is good, but down the test function there's no check if `p2` is not written and `p3` is only there once.

parcels/particleset.py
```diff
@@ -156,7 +156,7 @@ def convert_to_array(var):
         for kwvar in kwargs:
             kwargs[kwvar] = convert_to_array(kwargs[kwvar])
             assert lon.size == kwargs[kwvar].size, (
-                '%s and positions (lon, lat, depth) don''t have the same lengths.' % kwargs[kwvar])
+                '%s and positions (lon, lat, depth) don''t have the same lengths.' % kwvar)
```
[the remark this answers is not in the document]
erikvansebille: Because `kwargs[kwvar]` is the vector of data, and `kwvar` is the name of the variable. The latter is more useful for debugging purposes.

Previously, the error would be something like`[0 0 0 0 ] and positions (lon, lat, depth) don't have the same lengths`. 
Now it is something like `p3 and positions (lon, lat, depth) don't have the same lengths.`

PR #852, opened 2020-06-03
Throwing error for custom Variables with same name as built-in Variables

erikvansebille:
Generating an error when users create a custom variable that has the same name as a built-in one

This fixes the Issue mentioned in https://github.com/OceanParcels/parcels/issues/846#issuecomment-638101903

### DIFF
--- a/parcels/particle.py
+++ b/parcels/particle.py
@@ -24,9 +24,6 @@ class Variable(object):
     def __init__(self, name, dtype=np.float32, initial=0, to_write=True, _custom=True):
         if name == 'z':
             raise NotImplementedError("Custom Variable name 'z' is not allowed, as it is used for depth in ParticleFile")
-        if _custom and name in ['lon', 'lat', 'depth', 'time', 'xi', 'yi', 'zi', 'ti', 'id',
-                                'fileid', 'dt', 'state', '_next_dt']:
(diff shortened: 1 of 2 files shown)


HugoDENISFR:
I am not sure to understand well the issue with the name of the variable. 

The file I created is initially empty so there is no variable yet. 
`init_pos = netCDF4.Dataset('parcels_examples/Data_NEMO/indian_drifters.nc','w',format="NETCDF4")`

I tried to create a variable 'time' using : 
`i_time` = init_pos.createVariable('time', 'f8', dimensions='time')`
because I thought (from what I have seen in the documentation) that it is a required arguments for ParticleSet. 

If I change the name of the time variable, how is ParticleSet.from_particlefile supposed to handle the data file ?

erikvansebille:
Thanks @CKehl, you're absolutely right that a loop over existing Variable names would be the cleaner way to do this. I had a look at how to do that, but couldn't work it out, which is why I went for this solution. The problem is that [`Variable.__init__`](https://github.com/OceanParcels/parcels/blob/master/parcels/particle.py#L24) does not have information about the other Variables in the ParticleClass. I thought I could do it in [`Variable.__set__`](https://github.com/OceanParcels/parcels/blob/master/parcels/particle.py#L40), but that does not seem to be called. Do you have any idea where to get info to the other Variables?

CKehl:
we can browse ideas separately for a later fix - for this case, the code is fine and sufficient.

Issue #768, opened 2020-03-11
Silent chunking failure for dimensions not in name_maps


angus-g:
The automatic chunking mechanism (e.g. passing a tuple of chunk sizes) determines dimension names for each variable independently, which has no problems assuming automatic chunking is supported.

For files with dimensions that aren't in `parcels.field.NetcdfFileBuffer._name_maps`, we have to specify chunks as a dictionary. However, because of staggering, different variables have different dimension names. This chunking dictionary is passed directly to xarray, which throws an error if extra dimensions are present.

## Example

For example, we have B-grid data which is all chunked in the same way:
```python
chunks = {
  "st_ocean": 7, "sw_ocean": 7,
  "xt_ocean": 400, "xu_ocean": 400,
  "yt_ocean": 300, "yu_ocean": 300,
}
```

This gives the following error:

```
ValueError: some chunks keys are not dimensions on this object: {'yt_ocean', 'xt_ocean', 'sw_ocean'}
```

At this point, I reduced the chunks to only the subset of dimensions that appear on the first field that's processed:
```python
chunks = {
  "st_ocean": 7,
  "xu_ocean": 400,
  "yu_ocean": 300,
}
```

However, because the dimension names are still not expected, the chunksize/chunkmap mechanism silently breaks. The relevant part of the traceback is:

```
  File "/nix/store/9racrka4shwcahxrbzk14w6lpzbp3nrs-python3.7-parcels/lib/python3.7/site-packages/parcels/field.py", line 1976, in data_access
    data = data.rechunk(self.chunk_mapping)
```

Which leads to a division by zero in dask: `self.chunk_mapping = {0: 0, 1: 0}`.

## Thoughts
One way to get around this would be just to add the extra dimension names to the name map, but this doesn't scale to the general case. It does seem that the whole chunking mechanism is a bit complicated and magic.

CKehl:
well, Angus, so this problem is connected to the connection between netcdf file and the Grid and the Field that are using this file. So, I would opt e.g. for the first example
```
chunks = {
  "st_ocean": 7, "sw_ocean": 7,
  "xt_ocean": 400, "xu_ocean": 400,
  "yt_ocean": 300, "yu_ocean": 300,
}
```
because the function `_get_initial_chunk_dictionary` in the NetcdfFileBuffer will try then to map those provided dimensions to the largest subset of parameters for a variable (which by assumption is the actual data variable) and apply those numbers to it (line 1808 to 1839 in field.py).

[...]

PR #853, opened 2020-06-03
Fixing dynamic addition of NetCDF dimensions to the NetCDF parsing for dask

CKehl:
As discussed in #768 , chunking *novel* field data is tricky because the dimension names (the dimensions of the NetCDF file) need to be known in advance (i.e. before file opening), as otherwise the data need to be loaded fully in memory (which is exactly what dynamic loading in `dask` forces to avoid).
Now, if you need chunking, you can define the dimension names (and their connection to Parcels' dimensions `lon`, `lat`, `depth` and `time`) with the `name_maps` parameter during FieldSet creation. This parameter `name_maps` is forwarded to all related NetcdfFileBuffer instances in the background.

Tests of this can be found in the [parcels_contributions repository](https://github.com/OceanParcels/parcels_contributions), for the OFAM dataset.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1765,7 +1765,6 @@ class NetcdfFileBuffer(object):
             #         if value not in self._name_maps[key]:
             #             self._name_maps[key].append(value)
 
-
     def __enter__(self):
         if self.netcdf_engine == 'xarray':
             self.dataset = self.filename
@@ -1902,7 +1901,7 @@ class NetcdfFileBuffer(object):
         # ==== check if the chunksize reading is successful. if not, load the file ONCE really into memory and ==== #
         # ==== deduce the chunking from the array dims.                                                         ==== #
         try:
-            if len(init_chunk_dict)<3:
+            if len(init_chunk_dict) < 3:
                 raise AttributeError("Too few known chunk dimension arguments.")
             self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine=self.netcdf_engine, chunks=init_chunk_dict, lock=False)
         except:
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -74,8 +74,8 @@ class Field(object):
     :param time_periodic: To loop periodically over the time component of the Field. It is set to either False or the length of the period (either float in seconds or datetime.timedelta object).
            The last value of the time series can be provided (which is the same as the initial one) or not (Default: False)
            This flag overrides the allow_time_interpolation and sets it to False
-    :param name_maps (opt.): gives a name map to the NetCDFFileBuffer that declared a mapping between chunksize name, NetCDF dimension and Parcels dimension;
-           required only if unknown OCM field is loaded and chunking is used by 'field_chunksize' (which is the default)
+    :param chunkdims_name_map (opt.): gives a name map to the NetCDFFileBuffer that declared a mapping between chunksize name, NetCDF dimension and Parcels dimension;
+           required only if currently incompatible OCM field is loaded and chunking is used by 'field_chunksize' (which is the default)
     """
 
     def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=None, mesh='flat', timestamps=None,
@@ -173,7 +173,7 @@ class Field(object):
         self.loaded_time_indices = []
         self.creation_log = kwargs.pop('creation_log', '')
         self.field_chunksize = kwargs.pop('field_chunksize', None)
-        self.netcdf_name_maps = kwargs.pop('name_maps', None)
+        self.netcdf_chunkdims_name_map = kwargs.pop('chunkdims_name_map', None)
         self.grid.depth_field = kwargs.pop('depth_field', None)
 
         if self.grid.depth_field == 'not_yet_set':
@@ -1294,7 +1294,7 @@ class Field(object):
                                       data_full_zdim=self.data_full_zdim,
                                       field_chunksize=self.field_chunksize,
                                       rechunk_callback_fields=self.chunk_setup,
-                                      name_maps=self.netcdf_name_maps)
+                                      chunkdims_name_map=self.netcdf_chunkdims_name_map)
         filebuffer.__enter__()
         time_data = filebuffer.time
         time_data = g.time_origin.reltime(time_data)
@@ -1757,13 +1757,13 @@ class NetcdfFileBuffer(object):
         self.rechunk_callback_fields = rechunk_callback_fields
         self.chunking_finalized = False
         self.lock_file = lock_file
-        if "name_maps" in kwargs.keys() and kwargs["name_maps"] is not None and isinstance(kwargs["name_maps"], dict):
-            self._name_maps = kwargs["name_maps"]
+        if "chunkdims_name_map" in kwargs.keys() and kwargs["chunkdims_name_map"] is not None and isinstance(kwargs["chunkdims_name_map"], dict):
+            # self._name_maps = kwargs["chunkdims_name_map"]
             # ==== code for merging instead of replacing the map - choose which may be most applicable ==== #
-            # for key, dim_name_arr in kwargs["name_maps"].items():
-            #     for value in dim_name_arr:
-            #         if value not in self._name_maps[key]:
-            #             self._name_maps[key].append(value)
+            for key, dim_name_arr in kwargs["chunkdims_name_map"].items():
+                for value in dim_name_arr:
+                    if value not in self._name_maps[key]:
+                        self._name_maps[key].append(value)
 
     def __enter__(self):
         if self.netcdf_engine == 'xarray':
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -74,6 +74,8 @@ class Field(object):
     :param time_periodic: To loop periodically over the time component of the Field. It is set to either False or the length of the period (either float in seconds or datetime.timedelta object).
            The last value of the time series can be provided (which is the same as the initial one) or not (Default: False)
            This flag overrides the allow_time_interpolation and sets it to False
+    :param name_maps (opt.): gives a name map to the NetCDFFileBuffer that declared a mapping between chunksize name, NetCDF dimension and Parcels dimension;
+           required only if unknown OCM field is loaded and chunking is used by 'field_chunksize' (which is the default)
     """
 
     def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=None, mesh='flat', timestamps=None,
@@ -171,6 +173,7 @@ class Field(object):
         self.loaded_time_indices = []
         self.creation_log = kwargs.pop('creation_log', '')
         self.field_chunksize = kwargs.pop('field_chunksize', None)
+        self.netcdf_name_maps = kwargs.pop('name_maps', None)
         self.grid.depth_field = kwargs.pop('depth_field', None)
 
         if self.grid.depth_field == 'not_yet_set':
@@ -1290,7 +1293,8 @@ class Field(object):
                                       interp_method=self.interp_method,
                                       data_full_zdim=self.data_full_zdim,
                                       field_chunksize=self.field_chunksize,
-                                      rechunk_callback_fields=self.chunk_setup)
+                                      rechunk_callback_fields=self.chunk_setup,
+                                      name_maps=self.netcdf_name_maps)
         filebuffer.__enter__()
         time_data = filebuffer.time
         time_data = g.time_origin.reltime(time_data)
@@ -1738,7 +1742,7 @@ class NetcdfFileBuffer(object):
 
     """ Class that encapsulates and manages deferred access to file data. """
     def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
-                 interp_method='linear', data_full_zdim=None, field_chunksize='auto', rechunk_callback_fields=None, lock_file=True):
+                 interp_method='linear', data_full_zdim=None, field_chunksize='auto', rechunk_callback_fields=None, lock_file=True, **kwargs):
         self.filename = filename
         self.dimensions = dimensions  # Dict with dimension keys for file data
         self.indices = indices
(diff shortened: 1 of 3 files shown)


angus-g:
Cool, this makes things a bit nicer than having to change the `NetcdfFileBuffer` class directly. There's still a limitation that crops up because of the linear scan through the name map, however. Consider a dataset with staggered velocity components:

```
netcdf test {
dimensions:
        T = 4 ;
        Z = 2 ;
        Y = 200 ;
        Xp1 = 200 ;
        Yp1 = 200 ;
        X = 200 ;
variables:
        double UVEL(T, Z, Y, Xp1) ;
        double VVEL(T, Z, Yp1, X) ;
        int64 T(T) ;
                T:units = "days since 1900-01-01" ;
        int64 Z(Z) ;
        int64 X(X) ;
        double Xp1(Xp1) ;
        int64 Y(Y) ;
        double Yp1(Yp1) ;
}
```

We create a fieldset, supposing that we want to chunk into quarters spatially:

```python
fs = parcels.FieldSet.from_netcdf(
    "test.nc",
    {"U": "UVEL", "V": "VVEL"},
    {"lon": "Xp1", "lat": "Yp1", "depth": "Z", "time": "T"},
    field_chunksize={"X": 50, "Xp1": 50, "Y": 50, "Yp1": 50, "Z": 1, "T": 1},
    name_maps={"lon": ["X", "Xp1"], "lat": ["Y", "Yp1"], "depth": "Z", "time": "T"},
)
```

However, this maps to an initial chunk dictionary of `{'X': 50, 'Y': 50, 'Z': 1, 'T': 1}`, so when the dataset gets loaded, the data isn't properly chunked as we would expect:

```
Data variables:
    UVEL     (T, Z, Y, Xp1) float64 dask.array<chunksize=(1, 1, 50, 200), meta=np.ndarray>
    VVEL     (T, Z, Yp1, X) float64 dask.array<chunksize=(1, 1, 200, 50), meta=np.ndarray>
```

I think the same would occur in any case where there are multiple lon/lat dimensions within a single file. If I understand correctly, the `NetcdfFileBuffer` is created on a per-field basis, so perhaps in `NetcdfFileBuffer._get_initial_chunk_dictionary` where the dataset is opened, it could be restricted to the DataArray corresponding to the field -- that way dimensions from other variables in the same file would be excluded:

```
>>> d = xr.open_dataset("test.nc").UVEL
>>> d.dims
('T', 'Z', 'Y', 'Xp1')
```

Anyway, this a little tangential to the PR itself, which looks fine if you consider only its scope.

parcels/field.py
```diff
@@ -74,6 +74,8 @@ class Field(object):
     :param time_periodic: To loop periodically over the time component of the Field. It is set to either False or the length of the period (either float in seconds or datetime.timedelta object).
            The last value of the time series can be provided (which is the same as the initial one) or not (Default: False)
            This flag overrides the allow_time_interpolation and sets it to False
+    :param name_maps (opt.): gives a name map to the NetCDFFileBuffer that declared a mapping between chunksize name, NetCDF dimension and Parcels dimension;
```
erikvansebille, line 77: I suggest a more explicit (and non-plural) name for this argument, like `chunking_dims_name_map`?

parcels/field.py
```diff
@@ -74,6 +74,8 @@ class Field(object):
...
+           required only if unknown OCM field is loaded and chunking is used by 'field_chunksize' (which is the default)
```
erikvansebille, line 78: 'unkown' is a bit strange here. Suggests 'unknown' to the user, but you mean within parcels. Better use 'non-compatible' or something?

parcels/field.py
```diff
@@ -171,6 +173,7 @@ def __init__(self, name, data, lon=None, lat=None, depth=None, time=None, grid=N
         self.loaded_time_indices = []
         self.creation_log = kwargs.pop('creation_log', '')
         self.field_chunksize = kwargs.pop('field_chunksize', None)
+        self.netcdf_name_maps = kwargs.pop('name_maps', None)
```
erikvansebille, line 176: Again, add the word 'chunking' to this variable name? E.g. `self.chunking_dims_name_map`?

parcels/field.py
```diff
@@ -1753,6 +1757,13 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
         self.rechunk_callback_fields = rechunk_callback_fields
         self.chunking_finalized = False
         self.lock_file = lock_file
+        if "name_maps" in kwargs.keys() and kwargs["name_maps"] is not None and isinstance(kwargs["name_maps"], dict):
+            self._name_maps = kwargs["name_maps"]
+            # ==== code for merging instead of replacing the map - choose which may be most applicable ==== #
```
erikvansebille, line 1762: I'd say merging is better, also for when users have some pre-built and some non-standard chunking dimensions

parcels/field.py
```diff
@@ -1880,9 +1898,11 @@ def _get_initial_chunk_dictionary(self):
                 init_chunk_dict[depthname] = max(1, depthvalue)
         # ==== closing check-opened requested dataset ==== #
         self.dataset.close()
-        # ==== check if the chunksize reading is successfull. if not, load the file ONCE really into memory and ==== #
+        # ==== check if the chunksize reading is successful. if not, load the file ONCE really into memory and ==== #
         # ==== deduce the chunking from the array dims.                                                         ==== #
         try:
+            if len(init_chunk_dict) < 3:
```
erikvansebille, line 1904: Why is this the case? What if a field only has two dimensions (e.g. bathymetry)?

CKehl:
Hey Angus,

well, it is a problem with this idea to 'repurpose' one data file for the mesh grid of all other data files - this can lead to very, very weird circumstances (I know,it's kind'a standard in Parcels to do that, but data-access wise, it's sub-optimal somehow). One thing that you can always do: instead of giving it the named dictionary, give it the 3- or 4D tuple ´(time, depth, lat, lon) => (1, 1, 50, 50)´. This works at least for velocity fields (U,V, W), but this chunking is not ideal for other field data and **will** fail for fieldsets that mix 2D, 3D and 4D dimensions or different grids. In those circumstances, one needs the named dictionary. Apart from that, I don't understand while you get the flawed dictionary 

> ```
> fs = parcels.FieldSet.from_netcdf(
>     "test.nc",
>     {"U": "UVEL", "V": "VVEL"},
>     {"lon": "Xp1", "lat": "Yp1", "depth": "Z", "time": "T"},
>     field_chunksize={"X": 50, "Xp1": 50, "Y": 50, "Yp1": 50, "Z": 1, "T": 1},
>     name_maps={"lon": ["X", "Xp1"], "lat": ["Y", "Yp1"], "depth": "Z", "time": "T"},
> )
> ```

looks good to me ... so I don't get the reason for the result of:

> ```
> Data variables:
>     UVEL     (T, Z, Y, Xp1) float64 dask.array<chunksize=(1, 1, 50, 200), meta=np.ndarray>
>     VVEL     (T, Z, Yp1, X) float64 dask.array<chunksize=(1, 1, 200, 50), meta=np.ndarray>
> ```

parcels/field.py
CKehl: agreed.

parcels/field.py
CKehl: agreed.

parcels/field.py
CKehl: agreed.

parcels/field.py
CKehl: agreed.

parcels/field.py
CKehl: if a field only has 2 dimensions by design, we don't do chunking anyway. So if parcels requests chunking and is not able to detect 3 or more working chunking dimensions, something has gone wrong.

angus-g:
I think the same thing would happen even if you had e.g. B-grid data in the same file: velocities on q-points and tracers on t-points, then you'd have two X and Y dimensions in the file (corresponding to the different staggering). So while the redefining grids bit is a bit weird, it's not causing the problem in this case.

As to why it actually happens, I think it comes down to this loop: https://github.com/OceanParcels/parcels/blob/b6e7ec96ce9fae1c18db32158f62af145d4a9181/parcels/field.py#L1952-L1956:

In the example we're using, and when searching for the "lon" dimension, it will always find `X` in the file, regardless of which field is actually being loaded. So for `UVEL`, it will incorrectly put `X` in the chunking dictionary, so it won't get the chunking correct.

My suggestion was to restrict from the full Dataset corresponding to a given file in `NetcdfFileBuffer` to only the DataArray representing the active field. This way the dimension data of the other variables in the file don't get in the way of this process (but maybe there are further reaching implications)

parcels/field.py
```diff
@@ -1757,13 +1757,13 @@ def __init__(self, filename, dimensions, indices, netcdf_engine, timestamp=None,
         self.rechunk_callback_fields = rechunk_callback_fields
         self.chunking_finalized = False
         self.lock_file = lock_file
-        if "name_maps" in kwargs.keys() and kwargs["name_maps"] is not None and isinstance(kwargs["name_maps"], dict):
-            self._name_maps = kwargs["name_maps"]
+        if "chunkdims_name_map" in kwargs.keys() and kwargs["chunkdims_name_map"] is not None and isinstance(kwargs["chunkdims_name_map"], dict):
+            # self._name_maps = kwargs["chunkdims_name_map"]
```
erikvansebille, line 1761: Remove this an line below?

CKehl:
@angus-g ,
I had another look on your script and the basic problem is that, when using `from_netcdf`, your data are both supposed to use the exact same dimension names for the grid. That is not the case (U uses `Xp1` and `Y`, V uses `X` and `Yp1`). that is why - when looking at the `parcels_contributions` repository, you basically create 1 grid that both fields share (which is mandatory if they live in the same grid), and only that their `data` access differs.
What I see in your example is that the same file should chunk *all* dimensions given in the `field_chunksize`, irrespective if they are addressed in Parcels `dimensions` dictionary or not. I need to think about that again if we can do it. One conceivable issue is that for each field access, we would then attempt to chunk more data than we need. Chunking is an expensive operation depending on the number of dimensions, so say instead of only chunking `T, Z, Xp1, Yp1` we chunk `T, Z, Zp1, X, Xp1, Y, Yp1` even though we don't use 3 of those dimensions, it will take unnecessary long. Also, when doing auto-chunking, the actual chunk size would vary quite a bit between both case, possibly leading to index errors.
It would take more time to think about it. For now, the best thing to do is to use one mesh grid file that indexes lat, lon, depth and time, and then files for the different data one wants to access.
Cheers,
Christian

angus-g:
Sorry to bring this back up, but I'm not sure I agree with your conclusion here. My point relates more to xarray/netcdf behaviour than parcels itself.

If you have a single file with data of different staggerings, the file could have up to two lat dimensions, and two lon dimensions (`X`, `Xp1` and `Y`, `Yp1`, per my example). Let's suppose this is natural B-grid data, so we're not "tricking" parcels by overriding the grid. Velocity `U` is defined on (`Xp1`, `Yp1`) and tracer `T` is defined on (`X`, `Y`).

When the file is opened in parcels, `NetcdfFileBuffer`, through `_get_initial_chunk_dictionary()` calls `_is_dimension_in_dataset('lon')` (and lat, etc.) if `field_chunksize` is a dictionary. The code snippet I highlighted here https://github.com/OceanParcels/parcels/pull/853#issuecomment-638782050 does not work correctly for this kind of staggered file:

Taking the first "lon" element from the name map, we have `X`. This value is of course in the file (indeed it has both `X` and `Xp1`), and therefore the dataset. Thus, `_is_dimension_in_dataset('lon')` will always return the `X` dimension from this file, and will never return `Xp1`.

This means that when the dataset is opened with chunking in `__enter__`, for both `U` *and* `T`, the chunk dictionary looks like `{..., "X": 50, "Y": 50}`. In particular, this means that `U` is not chunked in either of its lateral dimensions.

Also:
> One conceivable issue is that for each field access, we would then attempt to chunk more data than we need. Chunking is an expensive operation depending on the number of dimensions, so say instead of only chunking T, Z, Xp1, Yp1 we chunk T, Z, Zp1, X, Xp1, Y, Yp1 even though we don't use 3 of those dimensions, it will take unnecessary long.

If I understand correctly, this is because parcels' chunking is tied to the grid?

> For now, the best thing to do is to use one mesh grid file that indexes lat, lon, depth and time, and then files for the different data one wants to access.

I'm not sure I understand how this fixes the problem? The dataset that backs the `NetcdfFileBuffer` would still be incorrectly chunked, wouldn't it?

CKehl:
Thanks for your feedback. I'm still reading through your comment. I take the comments step by step. 

Regarding the commented snippet [in 853](https://github.com/OceanParcels/parcels/pull/853#issuecomment-638782050): the loop so far is correct, it's its application which is not in your interest. I start to understand that this has something to do with tracer vs. grid definition. The issue here is: each parcels dimension has an equal priority and all of them are accessed with the same class of NetCDF reader. Therefore, as long as all dimensions are declared on the same fieldset and on the same grid, we can't just define that a certain field dimension (e.g. U, V) is the 'head field' and that all other dimensions are sub-ordinate.

This could be done, but as it requires rewrites of grids, fields and file access classes, it's not done in one issue. If this really is a strict issue right now, then I suggest using the index mechanism to manually define smaller, halo-overlapped regions for the computation.

On the subject of chunking taking a long time: it's not entirely because it's related to grids - chunking always takes a long time (in the order of seconds), because in order to generate the virtual access layout the file needs to be opened and accessed (though not the data-part of the file). 
The issue that relates to the grid: fields and grids are connected, and both the grids and the fields need to hold parts of the chunking information. In particular: the grid makes the conversion between the global lon-lat-depth-time index of a particle to the actually loaded chunks and their local address space. As this requires the interpolation, it needs to reside in the grid. On the other hand, the field needs to know which indices need to be loaded, and needs to have a local field that tells which data chunks are loaded or not. That cannot reside in the grid because the field data are not loaded simultaneously - fields are asynchronous to one another, hence they need to manage their own data space. As both the lon-lat-time-depth conversion to the chunk structure and the overview of loaded chunks are interconnected, all fields connected to the same 'shared grid' need to share one and the same chunking structure - they need to match exactly for all fields and dimensions.

On the last comment: we probably misunderstood each other - what I said is that the best thing would be to use one mesh that is consistent (no mixing of interpolations or dimensions or such) and then address the data in separate files, so as to avoid naming confusion and putting too many dimensions in one file. From what I read from your comment, it does not seem to be a practical solution.

Issue #618, opened 2019-07-18
Disentangling Parcels


willirath:
@delandmeterp In Paris, we briefly talked about disentangling the structure of Parcels into essentially a bunch of microservices.

In this Gist https://gist.github.com/willirath/6943b136ad85ad83078f201316f9170b I've played with Dask actors to have a spatial domain (like a fieldset) on one worker, the random number generator on another worker and the particles on the front end.

(Note that this is more of a thought experiment to better understand the structure of the problem.)

delandmeterp:
Hi Willi,

Sorry for the very late reply, I was on holiday and came back last week.
Thanks! It's nice to see we can really reduce the Particle tracker to minimal **independent** modules. 

You already know that what I don't really like is the use of `scipy.interpolate.interp2d()`, which does not work for us as it should. But in that same spirit, when I was in scipy I've started to write a vectorial interpolator, using the library of [fbriol](https://github.com/pangeo-data/pangeo/issues/629#issuecomment-509243031) to locate the particles.
Combining those pieces, we could have a simple and very modulable particle tracker.

We should list what is needed for a minimal program, how to cut it in modules as small as possible, and how to interconnect those modules.

PR #880, opened 2020-07-03
Support for creating a ParticleSet with empty FieldSet

### LINKED ISSUE #867: KernelGenerator with empty FieldSet
The Kernel generator now needs a FieldSet, which sometimes hampers code development. @CKehl has already fixed this in #862, these changes need to be brought over to a separate PR.

In order to make it work, @CKehl needed to switch the parameter order from `KernelGenerator(fieldset, pyfunc)` to
`KernelGenerator(pyfunc, fieldset)` (next to some other adaptations). Hence after manually merging the two files, we may need to adapt some `KernelGenerator` calls in the `ParticleSet`, the `Kernel` and/or the `RNG` to the new order.

This will also help fix #618. Thanks to @CKehl for bringing this up

erikvansebille:
This PR support creating a `ParticlseSet` without a `FieldSet`, or with `fieldset=None`.

This fixes #867 and addresses #618, to deconvolute Parcels class dependencies

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -12,6 +12,7 @@ from parcels.field import NestedField
 from parcels.field import SummedField
 from parcels.field import VectorField
 from parcels.grid import Grid
+from parcels.particle import JITParticle
 from parcels.tools.loggers import logger
 
 
@@ -218,7 +219,7 @@ class IntrinsicTransformer(ast.NodeTransformer):
     names, such as 'particle' or 'fieldset', inserts placeholder objects
     and propagates attribute access."""
 
-    def __init__(self, fieldset, ptype):
+    def __init__(self, fieldset=None, ptype=JITParticle):
         self.fieldset = fieldset
         self.ptype = ptype
 
@@ -237,7 +238,7 @@ class IntrinsicTransformer(ast.NodeTransformer):
 
     def visit_Name(self, node):
         """Inject IntrinsicNode objects into the tree according to keyword"""
-        if node.id == 'fieldset':
+        if node.id == 'fieldset' and self.fieldset is not None:
             node = FieldSetNode(self.fieldset, ccode='fset')
         elif node.id == 'particle':
             node = ParticleNode(self.ptype, ccode='particles')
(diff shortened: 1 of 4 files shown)


parcels/codegenerator.py
```diff
@@ -218,7 +219,7 @@ class IntrinsicTransformer(ast.NodeTransformer):
     names, such as 'particle' or 'fieldset', inserts placeholder objects
     and propagates attribute access."""
 
-    def __init__(self, fieldset, ptype):
+    def __init__(self, fieldset=None, ptype=JITParticle):
```
CKehl, line 222: okay, so you decided to keep this parameter order, and then also rather just define 'JITParticle' as ptype standard. Yeah, fine - relieves you from a lot of tiny adaptions in the other files, I agree.

parcels/particleset.py
```diff
@@ -112,9 +113,13 @@ class ParticleSet(object):
     Other Variables can be initialised using further arguments (e.g. v=... for a Variable named 'v')
     """
 
-    def __init__(self, fieldset, pclass=JITParticle, lon=None, lat=None, depth=None, time=None, repeatdt=None, lonlatdepth_dtype=None, pid_orig=None, **kwargs):
+    def __init__(self, fieldset=None, pclass=JITParticle, lon=None, lat=None, depth=None, time=None, repeatdt=None, lonlatdepth_dtype=None, pid_orig=None, **kwargs):
         self.fieldset = fieldset
-        self.fieldset.check_complete()
+        if self.fieldset is None:
+            logger.warning_once("No FieldSet provided in ParticleSet generation. "
+                                "This breaks most Parcels functionality")
```
CKehl, line 120: good to set a warning. The formulation is very user-centric, but that's probably how it should be.

parcels/particleset.py
```diff
@@ -227,7 +235,8 @@ def _convert_to_reltime(time):
         initialised = set()
         for v in self.ptype.variables:
             if v.name in ['xi', 'yi', 'zi', 'ti']:
-                self.particle_data[v.name] = np.empty((len(lon), fieldset.gridset.size), dtype=v.dtype)
+                ngrid = fieldset.gridset.size if fieldset is not None else 1
```
CKehl, line 238: wouldn't that be `0` if no fieldset is given ?

parcels/particleset.py
```diff
@@ -227,7 +235,8 @@ def _convert_to_reltime(time):
...
+                self.particle_data[v.name] = np.empty((len(lon), ngrid), dtype=v.dtype)
```
CKehl, line 239: ah, okay, now I see why 'ngrid' should not be 0 ... the variable naming semantic here is a bit misleading then, but I understand the purpose.

PR #884, opened 2020-07-08
List of Particle Arrays structure for ParticleSet - DO NOT MERGE - DO NOT CLOSE

CKehl:


CKehl:
the auto-testing probably won't work before a rebase (which is NOT to be done).
If you want to check the test output, install flake8 locally to check the linter and run pytest to check the actual tests.

parcels/particleset.py
```diff
@@ -336,19 +495,38 @@ def lonlatdepth_dtype_from_field_interp_method(field):
...
         return self.size
 
     def __getitem__(self, key):
-        return self.particles[key]
+        bracket_index = 0
+        start_index = 0
+        end_index = self._pid_mapping_bounds[bracket_index][2]
+        # while start_index<key:
+        # while end_index >= key:
+        while key >= end_index:
```
erikvansebille, line 517: This could be a bisect search on the `end_index` for speedier (log(N)) search

parcels/kernel.py
```diff
@@ -250,11 +254,18 @@ def execute_jit(self, pset, endtime, dt):
 
         fargs = [byref(f.ctypes_struct) for f in self.field_args.values()]
         fargs += [c_double(f) for f in self.const_args.values()]
-        particle_data = pset._particle_data.ctypes.data_as(c_void_p)
-        self._function(c_int(len(pset)), particle_data,
-                       c_double(endtime),
-                       c_double(dt),
-                       *fargs)
+
+        for subfield_c in pset.particles_c:
```
erikvansebille, line 258: It's a bit confusing you call this `subfield`, because a `Field` is a specific object in Parcels, and I don't think that these subfields have anything to do with `Fields`

PR #886, opened 2020-07-22
Bugfix: Using correct library to create empty DeferredArray and error for dask.expand_dims()

erikvansebille:
Fixing a small bug where DeferredArrays were always created as dask, even if field_chunksize was set to False or None

This then later caused problems in field.reshape() because `dask` does not have a method `expand_dims`. We therefore now throw a `NotImplementedError`

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -922,7 +922,8 @@ class FieldSet(object):
                         for i in range(len(f.data)):
                             del f.data[i, :]
 
-                data = da.empty((g.tdim, g.zdim, g.ydim-2*g.meridional_halo, g.xdim-2*g.zonal_halo), dtype=np.float32)
+                lib = np if f.field_chunksize in [False, None] else da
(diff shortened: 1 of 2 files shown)


parcels/fieldset.py
```diff
@@ -922,7 +922,8 @@ def computeTimeChunk(self, time, dt):
                         for i in range(len(f.data)):
                             del f.data[i, :]
 
-                data = da.empty((g.tdim, g.zdim, g.ydim-2*g.meridional_halo, g.xdim-2*g.zonal_halo), dtype=np.float32)
+                lib = np if f.field_chunksize in [False, None] else da
+                data = lib.empty((g.tdim, g.zdim, g.ydim-2*g.meridional_halo, g.xdim-2*g.zonal_halo), dtype=np.float32)
```
CKehl, line 926: nice that you saw that one! indeed this is better.

PR #887, opened 2020-07-23
Support deferred loading from xarray datasets

angus-g:
This allows the use of the deferred/chunked loading mechanism on xarray datasets. In particular, this may be useful for dask-backed datasets which are not loaded directly from files (e.g. some transformation/chunking/etc. is performed on the data after it's loaded).

In its current form, this will break the usual deferred arrays, which assume `field.data` only contains 3 elements along the first axis — using xarray datasets, this is now the full time dimension and must be indexed accordingly. I'm not quite familiar enough to suggest an elegant way of solving this. The obvious method would be to have a `grid.block_ti` which is either `grid.ti` if the fieldset came from an xarray dataset, and `0` otherwise. This feels a little too special-casey, and I don't know how to actually make this distinction.

There would also need to be changes to `fieldset.computeTimeChunk` to shuffle the loaded chunks along in time when the grid time index is updated. In my own code, I'm handling this with the `compute_on_defer` functionality, but duplicating the existing code inside the `g.update_status == 'updated'` block that is specific to the loaded chunks.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -436,7 +436,7 @@ class Field(object):
 
     @classmethod
     def from_xarray(cls, da, name, dimensions, mesh='spherical', allow_time_extrapolation=None,
-                    time_periodic=False, **kwargs):
+                    time_periodic=False, deferred_load=False, **kwargs):
         """Create field from xarray Variable
 
         :param da: Xarray DataArray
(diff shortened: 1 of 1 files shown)


CKehl:
see #905 and @erikvansebille 's recent efforts to include xarray in the file-like loading process, and refactor field file interfaces themselves. ;-)

angus-g:
Sounds like a better place to focus efforts!

Issue #891, opened 2020-07-30
ParticleFile options - removing default variables and time-independent variables?


Plagioclase:
I have two questions about ParticleFile options - firstly, is there a non-hacky way to set `to_write=False` for one of the default JITParticle variables (e.g. depth)? Secondly, is there by any chance a way to have `write_ondelete=True` for only certain variables (or alternatively have time-independent variables)? I have a variable where I am only interested in the final state, but still would like the time-varying lat/lon information. Not problem if not, but it would save some space.

erikvansebille:
Re your first question: I've just created a PR (#892) that implements a method to set the write status of a Variable. Until that is merged into master, you can 'hack' your way around it with
```python
for v in pset.ptype.variables:
            if v.name == 'depth':
                v.to_write = False
```

Re your second question on setting `to_write=False` for only some Variables; that is not possible yet and not as easy to implement. We'll keep it on the backburner for a future feature

Plagioclase:
Thank you Erik, I think this method would be a very useful addition. However, for some reason, that code snippet isn't working for me. It successfully changes `to_write = False` for the depth variable, but Parcels is still writing it to the output netcdf. The status of `to_write` is consistent for all of the other variables. I've tried doing this for other variables using this method and it isn't working either. Could it be because I'm using a custom particle class?

erikvansebille:
That's because you also need to make a change in particlefile.py: https://github.com/OceanParcels/parcels/blob/particleset_set_variable_write_status/parcels/particlefile.py#L174
Otherwise, the file will still be created with a z variable

Plagioclase:
Perfect, thank you!

PR #892, opened 2020-07-30
Set Variable write status after creation of ParticleSet

erikvansebille:
This PR implements and tests a method to set the write-status of a Variable when ParticleSet has been created

Useful to change e.g. the write status of default Variables like 'depth'

This addresses #891

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -819,6 +819,10 @@ def search_kernel(particle, fieldset, time):
         :param var: Name of the variable (string)
         :param status: Write status of the variable (True, False or 'once')
         """
+        var_changed = False
         for v in self.ptype.variables:
             if v.name == var:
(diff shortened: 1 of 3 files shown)


parcels/particleset.py
```diff
@@ -813,3 +813,16 @@ def ParticleFile(self, *args, **kwargs):
...
+        :param var: Name of the variable (string)
+        :param status: Write status of the variable (True, False or 'once')
+        """
+        var_changed = False
+        for v in self.ptype.variables:
+            if v.name == var:
+                v.to_write = write_status
+                var_changed = True
+        if not var_changed:
+            raise SyntaxError('Could not change the write status of %s, because it is not a Variable name' % var)
```
CKehl, line 828: yes, that looks like a good 'z'/'depth' problem treatment

PR #894, opened 2020-07-30
Include UML class diagram of ParticleSet in docs folder

CKehl:
Inclusion, discussion and refactoring of UML class diagram for upcoming ParticleSet refactoring


erikvansebille:
Thanks @CKehl! What's the easiest way to view and/or comment on this xlm file?

erikvansebille:
> Thanks @CKehl! What's the easiest way to view and/or comment on this xlm file?

After some digging around for the correct website, I found that https://app.diagrams.net/#HOceanParcels%2Fparcels%2Finclude_pset_class_hierarchy_in_docs%2Fdocs%2FParticleSet_UML.drawio is the way to visualise this file

Any idea how to best comment and/or give feedback on this diagram?

CKehl:
you can use https://draw.io, connect it to Github, and then you should be able to view the file.
Alternative: go to draw.io, download the desktop app, install it, checkout the issue and then you can work with the file locally.

Comment: well, you can use a non-block shape (perhaps a cloud ?) to attach your questions. If it's a longer question, we need to talk, I think. UML software modelling and CI was never something to-be-combined ...

erikvansebille:
Thanks @CKehl. I've now been able to see the diagram; very nice!

A few comments (easier here as a list, rather than as clouds):
1. Is `Collection.ncount` the same as what is now `ParticleSet.size`? If so, should we rename `ncount` to `size`?
2. I don't see anything on MPI and moving Particles between processors. Should that also be a (set of) method(s) in Collection?
3. Should we rename Collection to something more specific to Parcels? E.g. `ParticleCollection`
4. Same for Iterator: e.g. `ParticleIterator`?
5. The `print` (`__repr__`) methods are missing for Particle and ParticleSet?

CKehl:
k, I'll answer the questions one by one.

1. Is Collection.ncount the same as what is now ParticleSet.size? If so, should we rename ncount to size?
Comment: I vote against it. 'Size' is rather ambiguous. You know this from the 'Particle' itself. Does it describe the byte size, or the number of elements (e.g. Particles, variables, ...) ? Thus, usually a variable (or: property) describing the number of entities is called 'ncount', and returned by the 'length()' function. 'size' usually describes the byte size in memory, which is returned by 'size()' or 'sizeOf()'.

CKehl:
2. I don't see anything on MPI and moving Particles between processors. Should that also be a (set of) method(s) in Collection?
Comment: MPI parallelisation is quite a specific technique, which rather has to do with the layout and processing and management of the variables, not the variables specifically. UML class diagrams usually has just the purpose to model the data structure and the relations and inheritance structure between classes. The main purpose of it is to devise a structure that increases inheritance and minimizes code duplication. So MPI is an important topic - but I think right now it might be too early to already tackle that problem. Also, in view of our code specifically, most of the MPI data splitting is part of the init() function of each specific ParticleSet. What I did miss in the diagram is to mention the 'init()' function in each specific ParticleSet.

CKehl:
3. Should we rename Collection to something more specific to Parcels? E.g. `ParticleCollection`
Comment: yes, we can do that. I just wanted to make the relation to classic software classes, common to nearly all object-oriented computer languages, obvious and explicit. Helps for example also to check if I missed some essential functions - just having to look up the member function list of Collections in Python, Java or C++.

CKehl:
4. Same for Iterator: e.g. `ParticleIterator`?
Comment: yes, reasoning is again as in (3)

CKehl:
5. The `print` (`__repr__`) methods are missing for Particle and ParticleSet?
Comment: `__repre__` (Python-specific) or `toString` (generic) are missing in there. I think this function has a base implementation in the `Collections` class, just printing the class name and `ncount`, and a specific implementation in each `ParticleSet` class, which prints the core properties of each structure. But that's just my thinking.

RoelBrouwer:
A few early obeservations/questions:

1. Not really a crucial point, but in the current diagram almost all relations are diamond-shaped, which signifies an aggregation, I believe. Most of these are meant to represent inheritance/implementation, right?
2. Are Parcels Particles, Collection and nD_Cluster meant to be groupings of functions that should be implemented in the BaseParticleSet and its subclasses, or do you actually intend to have these as separate classes?
3. Collection contains a really large number of ways of adding/removing particles, not all of which have a natural translation in all ParticleSet implementations. I am not sure how pushing an element works in a tree, for example. Do we need/want all of them?
4. Related to that, it is something to think about how all these manipulation options work if we want to keep our datastructure somewhat ordered to facilitate quick access to neighbouring particles. I know that that is not what this is about yet, and that it will probably only be implemented for one or some of the ParticleSet implementations.

CKehl:
to (1): yes, inheritance
to (2): partially. 'Parcels Particles' is - as your saw correctly - just a collection of functions, so those would not get a dedicated class but would be integrated in `BaseParticleSet`. In order to separate and better orient the functions, classes like `Collection` and `Iterator` (or their name derivatives) would stand alone. Python >3.6 has multi-class inheritance, so it would be implemented then the way it is modeled in the diagram. `nD Cluster` is a borderline case. We can put it ALL in `BaseParticleSet` - but if we later find out that this functionality is helpful for other parts, we need to disentangle it again. In the `nD Cluster` case, I personally have a preference for it being a separate class, but would not object too much in integrating it in the `BaseParticleSet`.

CKehl:
to (3): in an ordered set - such as a `Set`, `OrderedList` or a `Tree` - pushing is the same as inserting. yes - if you strictly say 'pushing means it's the last element of the collection', then that would be wrong, but because an ordered set defines that 'last element' implicitly, tree-push is the same as tree-insert is the same as tree-add of an element. Remember: those classes - collection & iterator at least, nD Cluster probably too - are 'abstract classes'. You define which functions need to be there, but the specific implementation can decide that e.g. `add()` is the major function, and if `push` or `insert` are semantically the same in the context of that data structure, then we can link them to call `add()`. The alternative: just having an `add()` function, and not being able to control where in the data structure an element should be added. Or semantic bloopers like `remove` returning an element ... possibly would be good to speak that through in Teams.

CKehl:
to (4): yes - see the very expanded list of thoughts on trees in the Slack group ;-) Thing is: it CAN be done - it's just not viable if you demand not to code the tree yourself. There are many tree implementations out, but most are targeting nD data, with N>=64, where tree updates become none-sensical. With our case of N<=4, it may be worth writing or own multiway kD tree - again, discussion see Slack group.

CKehl:
I now updated the class diagram to the state of the the class structure we have now. Ready for review and merge.

erikvansebille (approved):

PR #895, opened 2020-07-30
Fixing bug in github's macOS CI

erikvansebille:
Fixing a bug where the github CI on macOS (and only on that platform) breaks on test_mpirun, with an error like

```
=================================== FAILURES ===================================
_________________________________ test_mpi_run _________________________________

tmpdir = local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_mpi_run0')

   def test_mpi_run(tmpdir):
       if MPI:
           repeatdt = 200*86400
     
             system('mpirun -np 2 python %s -p 4 -o %s -r %d' % (stommel_file, outputMPI, repeatdt))
             system('python %s -p 4 -o %s -r %d' % (stommel_file, outputNoMPI, repeatdt))
     
 >           ncfile1 = Dataset(outputMPI, 'r', 'NETCDF4')
 
 tests/test_mpirun.py:21: 
 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 netCDF4/_netCDF4.pyx:2321: in netCDF4._netCDF4.Dataset.__init__
     ???
 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 
 >   ???
 E   FileNotFoundError: [Errno 2] No such file or directory: b'/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_mpi_run0/StommelMPI.nc'
```

### DIFF
--- a/tests/test_mpirun.py
+++ b/tests/test_mpirun.py
@@ -1,13 +1,15 @@
 from os import path, system
 from netCDF4 import Dataset
 import numpy as np
-import pytest  # noqa
+import pytest
+import sys
 try:
     from mpi4py import MPI
 except:
     MPI = None
 
 
+@pytest.mark.skipif(sys.platform.startswith("darwin"), reason="skipping macOS test as problem with file in pytest")
 def test_mpi_run(tmpdir):
     if MPI:
         repeatdt = 200*86400
--- a/tests/test_mpirun.py
+++ b/tests/test_mpirun.py
@@ -1,6 +1,7 @@
 from os import path, system
 from netCDF4 import Dataset
 import numpy as np
+import pytest  # noqa
 try:
     from mpi4py import MPI
 except:


erikvansebille:
> My question: why now ? why never before ?

Typically, this is because of an update in one of the packages (my hunch would be pytest, in this case). Since we don't pin most of the packages in our requirements, we every now and then face these bugs

CKehl:
> > My question: why now ? why never before ?
> 
> Typically, this is because of an update in one of the packages (my hunch would be pytest, in this case). Since we don't pin most of the packages in our requirements, we every now and then face these bugs

I see - well, it interesting to know it.

PR #898, opened 2020-08-04
Support sampling fields without applying conversion

angus-g:
Velocity fields on a spherical grid will automatically be converted to degrees when sampled within a kernel. This makes sense for advection, but we may want an interpolated sampling, without unit conversion. By calling `field.eval(..., applyConversion=False)` directly, we can get the desired behaviour. This works in both scipy and JIT modes.

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -54,18 +54,25 @@ class FieldNode(IntrinsicNode):
         if isinstance(getattr(self.obj, attr), Grid):
             return GridNode(getattr(self.obj, attr),
                             ccode="%s->%s" % (self.ccode, attr))
+        elif attr == "eval":
+            return FieldEvalCallNode(self)
         else:
             raise NotImplementedError('Access to Field attributes are not (yet) implemented in JIT mode')
 
-    def __getitem__(self, attr):
-        return FieldEvalNode(self.obj, attr)
+
+class FieldEvalCallNode(IntrinsicNode):
+    def __init__(self, field):
+        self.field = field
+        self.obj = field.obj
+        self.ccode = ""
 
 
 class FieldEvalNode(IntrinsicNode):
-    def __init__(self, field, args, var, var2=None):
+    def __init__(self, field, args, var, convert=True):
         self.field = field
         self.args = args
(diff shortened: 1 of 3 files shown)


parcels/codegenerator.py
```diff
@@ -352,9 +359,28 @@ def visit_Assign(self, node):
...
             node = IntrinsicNode(node, "return DELETE")
+
+        if isinstance(node.func, FieldEvalCallNode):
+            # get a temporary value to assign result to
+            tmp = self.get_tmp()
+            # whether to convert
+            convert = True
+            if "applyConversion" in node.keywords:
+                k = node.keywords["applyConversion"]
+                if isinstance(k, ast.NameConstant):
```
angus-g, line 375: This could be handled more indirectly, adding an `if` statement to the `conv_stat` below, but I've just handled the static case for now.

parcels/codegenerator.py
```diff
@@ -54,18 +54,25 @@ def __getattr__(self, attr):
         if isinstance(getattr(self.obj, attr), Grid):
             return GridNode(getattr(self.obj, attr),
                             ccode="%s->%s" % (self.ccode, attr))
+        elif attr == "eval":
+            return FieldEvalCallNode(self)
```
erikvansebille, line 58: Why did you make this change, using an `elif` instead of the `__getitem__`? Is this why the CI breaks?

erikvansebille:
Thanks for the PR, @angus-g. A nice feature indeed! We need to make sure the github CI passes before we can merge, though

parcels/codegenerator.py
angus-g: Whoops! I didn't mean to get rid of `__getitem__`, that was part of my testing. As to the change itself, because we're looking at `field.eval`, it will call `__getattr__` on the `FieldNode` (as part of `self.visit(node.func)` in `visit_Call`, I think).

However, it did seem that `__getitem__` here is never called, and is incorrect: `FieldEvalNode` has a 3-argument constructor, so we can't call it as `FieldEvalNode(self.obj, attr)` without raising a `TypeError` anyway? Because of that, I would still advocate for removing it.

PR #890, opened 2020-07-30
Updates to field.py and fieldset.py to include from_mom5 B-grid method

hrsdawson:
The vertical dimension of MOM5 fields differs from other models in that the vertical velocity is defined at the base of the grid cell.  To overcome this the vertical fields should be indexed differently and the top layer of W has been repeated. This relates to issue #860.

PR #900, opened 2020-08-04
Support MITgcm grid indexing

### LINKED ISSUE #897: Reading files from MITgcm c-grid output
OceanParcels currently supports reading input from c-grid data using `from_c_grid_dataset`. The c-grid indices are defined using the convention used by NEMO, as is shown [HERE](https://www.nemo-ocean.eu/doc/img360.png). However, the MITgcm c-grid definition is slightly different, where the U- and V- points are to the north and west of the f-points, instead of to the south and east. Therefore, using `from_c_grid_dataset` with the MITgcm output results in different (wrong) pathways, see for an example [HERE](https://github.com/SLYpma/GalapagosAnalysis/blob/master/documentation/20.08_CgridTest/Test_Cgrid.ipynb).

At the moment, the U and V fields need to be corrected by hand before the simulation by shifting (by adding a row for zonal velocities and a column for meridional velocities to shift the gridpoints) to correctly be interpreted by parcels. Is is possible to instead tell parcels to take for the f-point (j,i) the u-point (j-1,i) and the v-point (j,i-1)?

Moreover, to obtain netCDF output from MITgcm, native output data (.data/.meta) can be converted to netCDF using a package like `xmitgcm`, or data can be written directly to netCDF using the `mnc` module in MITgcm. In the latter case, velocity output is of the shape `UVEL[nx+1, ny]` and `VVEL[nx, ny+1]` (`nx` and `ny` being the number of grid points). Instead, Parcels assumes c-grid data to be of the shape `[nx+1, ny+1]`. Again, adding a dummy row/column will shift the data so that it can be read correctly by Parcels, but making Parcels aware of the MITgcm output conventions would be much more user-friendly to people working with MITgcm data.

erikvansebille:
This PR implements native grid indexing for MITgcm-style models (where u is south and v is west of the f-points, instead of north and east). This would then fix #897


erikvansebille:
MITgcm-style grid indexing is now supported in JIT and Scipy for 2D-grids. Next steps:

- [ ] Support for 3D grids
- [ ] Create `FieldSet.from_mitgcm()` method to automatically set the gridindexingtype correctly
- [ ] Add information to documentation (tutorials and docstring)

daanreijnders:
I've added a unit test case where U and V have unequal shapes (U[nx+1, ny], V[nx, ny+1]), which is default for netCDF output obtained with the MITgcm `mnc` package.

angus-g:
I wonder if this same approach could be extended/generalised to support different vertical indexing (c.f. #890)

erikvansebille:
> I wonder if this same approach could be extended/generalised to support different vertical indexing (c.f. #890)

Good idea. So if I understand correctly, MOM5 indexing is the same as NEMO for the horizontal directions, but at the bottom of the cell in the vertical. That would now be quite trivial to implement, following the method here. 

@angus-g or @hrsdawson, can you draft a unit test like the `test_mitgridindexing_3D()` one in test_grids.py? 

Is this for all MOM versions, or only MOM5? (i.e., should we call the `gridindexingtype='mom'`, or `gridindexingtype='mom5'`)

adele157:
Erik, that’s correct about the MOM5 indexing. It would also be relevant for earlier MOM versions, including OFAM. MOM6 is a switch to c-grid, but I’m not sure if the vertical indexing changes or not (@angus-g?).

angus-g:
In MOM6, all interfaces are present. `k=0` corresponds to the top interface, and the top layer — so I guess it's the same as the NEMO convention?

hrsdawson:
Hi @erikvansebille, I've drafted a unit test for the mom5 indexing which is sitting here: https://github.com/hrsdawson/parcels/blob/mom5_test_gridindexing/tests/test_grids.py#L749. 

I am entirely unfamiliar with writing unit tests, so apologies in advance if this is not what you're after. I've taken the `test_mitgridindexing_3D()` example and modified the vertical indexing for W when `gridindexingtype` is mom5 (W placed at bottom of grid cell, i.e. `depth[k] + dz`). It currently fails when running `pytest` because the mom5 grid indexing is not yet implemented.

daanreijnders:
With the latest commits, `from_mitgcm` should be working. I've added a unit test with MITgcm data in a zonally re-entrant domain, which uses a periodic halo, improving code coverage. 

Let's keep the scope for this PR to MITgcm. @hrsdawson Could you add the unit test you wrote for MOM5 to #890? 

After tests & review, this PR should be ready for merging.

erikvansebille:
Thanks @daanreijnders! I'll review soon. But should we not first also add some further info to https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_indexing.ipynb about the indexing of MITgcm?

daanreijnders:
> Thanks @daanreijnders! I'll review soon. But should we not first also add some further info to https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_indexing.ipynb about the indexing of MITgcm?

I've done this in https://github.com/OceanParcels/parcels/pull/900/commits/713995606f6c85efac44ca5fe49c78f5455f7ef1, including an updated diagram, highlighting the differences between MITgcm-style and NEMO-style C-grids (see also https://github.com/OceanParcels/oceanparcels_website/commit/7d3c08828e1b0be3658ad6b1b1c31ace8826b840).

parcels/examples/documentation_indexing.ipynb
```diff
@@ -11,17 +11,35 @@
...
+    "\n",
+    "**MITgcm** uses a different indexing: `U[j, i]` is located between the cell corners `[j, i]` and `[j+1, i]`, and  `V[j, i]` is located between cell corners `[j, i]` and `[j, i+1]`. If you use [`xmitgcm`](https://xmitgcm.readthedocs.io/en/latest/) to write your data to a NetCDF file, `U` and `V` will have the same dimensions as your grid (in the above case 4x4 rather than 5x5 as in NEMO), meaning that the last column of `U` and the last row of `V` are omitted. In MITgcm, these either correspond to a solid boundary, or in the case of a periodic domain, to the first column and row respectively. In the latter case, you can use the `FieldSet.add_periodic_halo` method to make sure particles can be correctly interpolated in the last zonal/meridional cells.\n",
```
erikvansebille, line 32:
```suggestion
    "**MITgcm** uses a different indexing: `U[j, i]` is located between the cell corners `[j, i]` and `[j+1, i]`, and  `V[j, i]` is located between cell corners `[j, i]` and `[j, i+1]`. If you use [`xmitgcm`](https://xmitgcm.readthedocs.io/en/latest/) to write your data to a NetCDF file, `U` and `V` will have the same dimensions as your grid (in the above case 4x4 rather than 5x5 as in NEMO), meaning that the last column of `U` and the last row of `V` are omitted. In MITgcm, these either correspond to a solid boundary, or in the case of a periodic domain, to the first column and row respectively. In the latter case, and assuming your model domain is periodic, you can use the `FieldSet.add_periodic_halo` method to make sure particles can be correctly interpolated in the last zonal/meridional cells.\n",
```

erikvansebille: (Adding the phrase 'and assuming your model domain is periodic' to the last line)

parcels/examples/documentation_indexing.ipynb
```diff
@@ -11,17 +11,35 @@
...
+    "\n",
+    "Parcels can take care of loading C-grid data for you, through the general `FieldSet.from_c_grid_dataset` method (which takes a `gridindexingtype` argument), and the model-specific methods `FieldSet.from_nemo`, `FieldSet.from_pop`, and `FieldSet.from_mitgcm`.\n",
```
erikvansebille, line 34: POP is a B-Grid, so should we removed from the list here

parcels/examples/documentation_indexing.ipynb
```diff
@@ -11,17 +11,35 @@
...
+    "\n",
+    "The only information that Parcels needs is the location of the 4 corners of the cell, which are called the $f$-node. Those are the ones provided by `U.grid` (and are equal to the ones in `V.grid`). Parcels does not need the exact location of the `U` and `V` nodes, because in C-grids, `U` and `V` nodes are always located in the same position relative to the $f$-node.\n",
+    "\n",
+    "Importantly, the interpolation of velocities on Arakawa C-grids is done in barycentric coordinates: $\\xi$, $\\eta$ and $\\zeta$ instead of longitude, latitude and depth. These coordinates are always between 0 and 1, measured from the corner of the grid cell. This is more accurate than simple linear interpolation\n",
```
erikvansebille, line 38: Add 'on curvilinear grids' to the end of this sentence

parcels/examples/documentation_indexing.ipynb
```diff
@@ -184,33 +202,20 @@
...
-    "\n",
-    "For 3D C grids and B grids, the idea is the same. It is important to follow the indexing notation, which is defined in Parcels and in [Delandmeter and Van Sebille (2019)](https://www.geosci-model-dev.net/12/3571/2019/gmd-12-3571-2019.html). \n",
+    "### 3D C- and B-grids\n",
     "\n",
-    "Note that Parcels can take care of loading these B and C grid data for you, through the general `FieldSet.from_c_grid_dataset` and `FieldSet.from_b_grid_dataset` methods, and the model-specific methods `FieldSet.from_nemo` and `FieldSet.from_pop`."
+    "For 3D C-grids and B-grids, the idea is the same. It is important to follow the indexing notation, which is defined in Parcels and in [Delandmeter and Van Sebille (2019)](https://www.geosci-model-dev.net/12/3571/2019/gmd-12-3571-2019.html). In 3D C-grids, the vertical (`W`) velocities at the top and bottom.  Note that in the vertical, the bottom velocity is often omitted, since a no-flux boundary conditions implies a zero vertical velocity at the ocean floor. That means that the vertical dimension of `W` often corresponds to the amount of vertical levels."
```
erikvansebille, line 218:
```suggestion
    "For 3D C-grids and B-grids, the idea is the same. It is important to follow the indexing notation, which is defined in Parcels and in [Delandmeter and Van Sebille (2019)](https://www.geosci-model-dev.net/12/3571/2019/gmd-12-3571-2019.html). In 3D C-grids, the vertical (`W`) velocities are at the top and bottom.  Note that in the vertical, the bottom velocity is often omitted, since a no-flux boundary conditions implies a zero vertical velocity at the ocean floor. That means that the vertical dimension of `W` often corresponds to the amount of vertical levels."
```

parcels/examples/example_mitgcm.py
```diff
@@ -0,0 +1,67 @@
...
+        "U": {"lon": "XG", "lat": "YG", "time": "time"},
+        "V": {"lon": "XG", "lat": "YG", "time": "time"},
+    }
+    fieldset = FieldSet.from_mitgcm(filenames, variables, dimensions, mesh="flat")
+
+    fieldset.add_periodic_halo(zonal=True)
+
+    def periodicBC(particle, fieldset, time):
+        if particle.lon < 0:
+            particle.lon += 1000000
```
erikvansebille, line 37: Would it be nicer to use a FieldSet constant for this value of 1000000? I.e. 
```python
fieldset.add_constant('domain_width', 1000000)
```
And then use `fieldset.domainwidth` in this kernel here?

parcels/field.py
```diff
@@ -1339,11 +1352,13 @@ def __init__(self, name, U, V, W=None):
...
         self.vector_type = '3D' if W else '2D'
+        self.gridindexingtype = U.gridindexingtype
         if self.U.interp_method == 'cgrid_velocity':
             assert self.V.interp_method == 'cgrid_velocity', (
                 'Interpolation methods of U and V are not the same.')
-            assert self.U.grid is self.V.grid, (
-                'Grids of U and V are not the same.')
+            if U.gridindexingtype == 'nemo':
+                assert self.U.grid is self.V.grid, (
+                    'Grids of U and V are not the same.')
```
erikvansebille, line 1361: Since we chose not to support uneven grids for mitgcm after all, this change is not needed. If-statement can be removed

parcels/include/parcels.h
```diff
@@ -690,15 +702,27 @@ static inline StatusCode temporal_interpolationUV_c_grid(type_coord x, type_coor
...
-      status = getCell2D(U, xi[igrid], yi[igrid], ti[igrid], data2D_U, 0); CHECKSTATUS(status);
-      status = getCell2D(V, xi[igrid], yi[igrid], ti[igrid], data2D_V, 0); CHECKSTATUS(status);
+      if (gridindexingtype == NEMO) {
+          status = getCell2D(U, xi[igrid], yi[igrid], ti[igrid], data2D_U, 0); CHECKSTATUS(status);
+          status = getCell2D(V, xi[igrid], yi[igrid], ti[igrid], data2D_V, 0); CHECKSTATUS(status);
+      }
+      else if (gridindexingtype == MITGCM) {
+          status = getCell2D(U, xi[igrid], yi[igrid]-1, ti[igrid], data2D_U, 0); CHECKSTATUS(status);
+          status = getCell2D(V, xi[igrid]-1, yi[igrid], ti[igrid], data2D_V, 0); CHECKSTATUS(status);
+      }
```
erikvansebille, line 712: This block is indented with 4 spaces, where we use 2-space indentation in the rest of the C-code (pedantic detail, I know...)

erikvansebille (commented):
Nice work on this PR! Apart from a few minor comments, I think it's ready to go. @daanreijnders, can you implement these last changes?

tests/test_grids.py
```diff
@@ -601,3 +601,128 @@ class MyParticle(ptype[mode]):
...
+                r, phi = calc_r_phi(lon[i], lat[j] + isign * dy / 2)
+                U[j, i] = omega * r * np.cos(phi)
+        return U, V, R
+
+    U, V, R = calculate_UVR(lat, lon, dx, dy, omega)
+    data = {'U': U, 'V': V, 'R': R}
+    dimensions = {'lon': lon, 'lat': lat}
+    fieldset = FieldSet.from_data(data, dimensions, mesh='flat', gridindexingtype=gridindexingtype)
+    fieldset.U.interp_method = 'cgrid_velocity'
+    fieldset.V.interp_method = 'cgrid_velocity'
```
CKehl, line 641: interesting - I still didn't figure out how the field loading works in that case, but I guess that your fixes really just had an influence on the indexing and on nothing else, so nothing had to be done with the FileBuffer .. that would be quite cool

daanreijnders: In this case, it's indeed only indexing. When using MITgcm, there's multiple ways to output netCDF data. By default, output is in its native binary format. This can be converted to netCDF with the [`xmitgcm`](https://github.com/MITgcm/xmitgcm) package. Alternatively, output can be written to netCDF directly with the `mnc` or the `NF90IO` packages. `xmitgcm` outputs a velocity field with similar shapes as NEMO (UVEL[nx+1, ny+1]). Here U is located at the inflow coordinates instead, and this was easy to implement (credits to @erikvansebille and @SLYpma). Using `mnc` or `NF90IO` is where the trouble starts: dimensions will in that case correspond to the exact number of U or V faces (so UVEL[nx+1, ny], VVEL[nx, ny+1]). Turned out to pretty complicated to rewrite Parcels to accept fields with uneven dimensions in a C-grid, so for now those are not supported.

Issue #889, opened 2020-07-29
Correct cell not found as particle crosses nested model border


DEHewitt:
Hi,

I am attempting to nest a small scale oceanographic model, which is configured in the Regional Oceanographic Modelling System (NEMO grid; described here: https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2019JC015227) within BRAN 3p5, configured using the Ocean Forecasting Australia Model (which is a configuration of the GFDL Modular Ocean Model; details here: https://wp.csiro.au/bluelink/global/bran/). 

I am using version 2.1.4 of Parcels and attempting to run these simulations on the HPC at UNSW (Katana). I have successfully run small scale simulations (i.e. 1 particle released every 4 days from 4 locations) - that is the output looks how I expect it to - however, associated with the output is an error/warning log with the following message:

```
Correct cell not found for (154.050926, -25.534158) after 1000000 iterations
Debug info: old particle indices: (yi, xi) 314 45
            new particle indices: (yi, xi) 314 226
            Mesh 2d shape:  316 271
            Relative particle position:  (xsi, eta) -6.1320155892265049e-01 1.0127507928492394e+00
```

When I attempt to scale up to full scale simulations (i.e. 100 particles released daily from 10 locations) the jobs consistently time out with the same error/warning message. 

These coordinates roughly match the boundary where particles pass from one model into the next.

The following is the code used to run the simulations:

```
from parcels import FieldSet, Field, AdvectionRK4, ParticleSet, JITParticle, Variable, BrownianMotion2D, random, NestedField
from parcels import ErrorCode
import numpy as np
from glob import glob
from datetime import timedelta as delta
from datetime import datetime as datetime
import math
import os
from operator import attrgetter

out_dir = '/home/z5278054/Nesting'

npart = 1  # number of particles to be released
repeatdt = delta(days=4)  # release from the same set of locations every X day

# Forward: 9
# lon_array = []
# lat_array = []

# Backwards: 13
lon_array = [153.790]#, 153.799, 153.736, 153.172, 153.048, 152.946, 152.775, 152.312, 152.045, 151.577]
lat_array = [-28.165]#, -28.890, -29.432, -30.864, -31.645, -31.899, -32.193, -32.719, -32.917, -33.578]

lon = np.repeat(lon_array,npart)
lat = np.repeat(lat_array,npart)

array_ref = 0 # which season to simulate (0 = 1994/95, 22 = 2015/16)

[...]

PR #904, opened 2020-08-05
Fixing Dask ZeroDivisionError - DO NOT MERGE

erikvansebille:
Some users experience a dask `ZeroDivisonError` in the latest master branch of Parcels. See e.g. https://github.com/OceanParcels/parcels/issues/889#issuecomment-666869317. 

By running `git-bisect`, we were able to pinpoint this bug to https://github.com/OceanParcels/parcels/commit/cbf9e287e3e5bcefd6c60821fddaf30425cb094c.

While removing the line `data = data.rechunk(self.chunk_mapping)` from field.py fixes this bug, it causes other unit tests to fail. Clearly, some further work is needed on the dask chunking

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -2200,8 +2200,8 @@ class NetcdfFileBuffer(object):
                             self.rechunk_callback_fields()
                             self.chunking_finalized = True
                     else:
-                        # ==== I think this can be "pass" too ==== #
-                        data = data.rechunk(self.chunk_mapping)
+                        # # ==== I think this can be "pass" too ==== #
+                        # data = data.rechunk(self.chunk_mapping)
                         self.chunking_finalized = True
             else:
                 da_data = da.from_array(data, chunks=self.field_chunksize)


angus-g:
In my opinion, this change is the right way to go, but triggers an error due to incorrect behaviour elsewhere. The test requests `field_chunksize=(1, 2, 2)`. Since the depth chunking is only singular, it gets dropped from the `chunk_mapping` dictionary:

```python
            self.chunk_mapping = {}
            if(isinstance(self.field_chunksize, tuple)):
                j = 0
                for i in range(len(self.field_chunksize)):
                    if self.field_chunksize[i] <= 1:
                        continue
                    self.chunk_mapping[j] = self.field_chunksize[i]
                    j += 1
                self.field_chunksize = tuple([self.chunk_mapping[i] for i in range(len(self.chunk_mapping))])
```

So after this, you end up with `chunk_mapping = {0: 2, 1: 2}`. Before the `data.rechunk()` call prior to its removal in this PR, data looks like:

```
dask.array<getitem, shape=(4, 3, 3), dtype=float32, chunksize=(1, 2, 2), chunktype=numpy.ndarray>
```

but afterwards, `chunksize=(2, 2, 2)`. In other words, it _incorrectly_ rechunks the depth dimension to size-2 chunks.

erikvansebille:
Thanks @angus-g, yes we saw that too. We are now working on a more comprehensive refactoring of the NetcdfFileBuffer class (#905), and will properly fix this bug in that PR. This PR is only here so that users who encounter the ZeroDivision Error can in the meantime find a quick fix. We have no intention to merge it into master

erikvansebille:
This is now fixed in #905, so closing this PR

Issue #872, opened 2020-06-24
Disentangling Dask between Grids and Fields


erikvansebille:
When looking on recent discussions, it would be good to access NetCDF files that contain the mesh grid separately from the `data` NetCDF files, to make it possible to chunk one and the same file differently. 

In the current form, all fields' chunk sizes of the same shared grid need to be the same, as well as the chunk sizes for the mesh of a grid (which is loaded with the first field of a fieldset?). Disentangling that would make it possibly easier in the post-summer times to make this lazy-loading more adaptable.

Thanks to @CKehl for bringing this up

erikvansebille:
This is now fixed in #932

PR #905, opened 2020-08-05
Refactoring NetcdfFileBuffer class

### LINKED ISSUE #721: re-organize NetcdfFileBuffer into FileBuffer package
Currently, the NetcdfFileBuffer is part of the Field package. This has disadvantages:

- lack of overview (the class is not a field and located out-of-place and the end of the package)
- it invites programmers to mix file functionality and field functionality
- after introducing dask, Fields operate only on virtual data arrays (i.e. lazy allocation in Dask), whereas FileBuffers organise and orchestrate the transfer from physical disk data to dask arrays, which should be obvious from code organisation
- FileBuffer objects exclusively operate in Python, whereas Fields are an interface between Python and the C-backend data structure, which should be obvious from code organisation

Due to those reasons, the NetcdfFileBuffer shall be externalized in a package, which also provides options to write other FileBuffers (e.g. HDF5) instead of integrating all functionality in one class.

Hand in hand, code structure between the different computeTimeChunk() functions can be reviewed, improved, and possibly given more distinct names.

erikvansebille:
This PR is an attempt to significantly refactor the NetcdfFileBuffer class, and implements the following changes:
- [x] moving NetcdfFileBuffer class from field.py to its own file (field.py was getting really bloated)
- separate FileBuffer into classes for:
  - [ ] xarray datasets
  - [x] netcdf: plain numpy, unbuffered, full-file loading
  - [x] netcdf: deferred loading (3-time step procedure) with numpy
  - [x] netcdf: dynamic loading (combined with deferred, as now) with dask
- [ ] disentangling dask between grids and fields (#872)
- [ ] provide more user control on name maps (#768)
- [x] further enhancing the unit tests to e.g. also fix the bug in #904 

This PR fixes #721


parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -121,6 +121,23 @@ def fieldset_from_ofam(chunk_mode):
     return FieldSet.from_netcdf(filenames, variables, dimensions, allow_time_extrapolation=True, field_chunksize=chs, chunkdims_name_map=name_map)
 
 
+def fieldset_from_mitgcm_regrid(chunk_mode):
```
CKehl, line 124: very nice - I like that

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -300,6 +317,28 @@ def test_ofam_3D(mode, chunk_mode):
...
+        dask.config.set({'array.chunk-size': '1024KiB'})
+    else:
+        dask.config.set({'array.chunk-size': '128MiB'})
+    field_set = fieldset_from_mitgcm_regrid(chunk_mode)
+    lons, lats = 5e5, 5e5
+
+    pset = ParticleSet.from_list(fieldset=field_set, pclass=ptype[mode], lon=lons, lat=lats)
+    pset.execute(AdvectionRK4, runtime=delta(days=1), dt=delta(minutes=5))
+    # MITgcm_regridded sample file dimensions: T=5, Xp1=401, lon=201
+    assert (len(field_set.U.grid.load_chunk) == len(field_set.V.grid.load_chunk))
```
CKehl, line 335: if that is enough for you for the test, then it's okay. I think with those chunking tests, we should really check that, if we request a specific chunking set, it is in the end also formatted the way it was requested (hence the specific dblocks, vblocks, ublocks - should rather be called zblocks (for 'depth'), yblocks (for 'lat') and xblocks (for 'lon')).

parcels/field.py
```diff
@@ -284,15 +285,20 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
             else:
                 raise RuntimeError('interp_method is a dictionary but %s is not in it' % variable[0])
 
-        with NetcdfFileBuffer(lonlat_filename, dimensions, indices, netcdf_engine, field_chunksize=False, lock_file=False) as filebuffer:
+        if netcdf_engine == 'xarray':
+            FieldFileBuffer = XarrayFileBuffer
+        else:
+            FieldFileBuffer = NetcdfFileBuffer
```
CKehl, line 291: Sorry but: please don't overwrite package- or class names with other classes. This creates massive confusion later, cause most programmers will assume `FieldFileBuffer` to be the abstract super-class, not one of its main classes.

you can put that in a variable, e.g.:
```
self._field_fb_class = FieldFileBuffer
self._field_fb_class = NetcdfFileBuffer if netcdf_engine is not None else self._field_fb_class
self._field_fb_class = self._field_fb_class if netcdf_engine != 'xarray'

with self._field_fb_class( ... )
```

parcels/field.py
```diff
@@ -284,15 +285,20 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
...
+        with FieldFileBuffer(lonlat_filename, dimensions, indices, netcdf_engine) as filebuffer:
             lon, lat = filebuffer.read_lonlat
             indices = filebuffer.indices
             # Check if parcels_mesh has been explicitly set in file
             if 'parcels_mesh' in filebuffer.dataset.attrs:
                 mesh = filebuffer.dataset.attrs['parcels_mesh']
 
         if 'depth' in dimensions:
-            with NetcdfFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method, field_chunksize=False, lock_file=False) as filebuffer:
+            with FieldFileBuffer(depth_filename, dimensions, indices, netcdf_engine, interp_method=interp_method) as filebuffer:
```
CKehl, line 301: this will be interesting how it's going to be solves. My suggestions here, explicit: Have 2 'field fielbuffers'! One that is used in the init-process to gather the file metadata - ideally a Netcdf buffer that is non-dask, just a regular-old netcdf-reader. The second 'field filebuffer' then would be for the actually data read access during field computations, and that one - if used in conjecture with Dask, needs to be file-locked (see Dask & xarray documentation on chunking).

parcels/field.py
```diff
@@ -426,6 +432,9 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         kwargs['indices'] = indices
         kwargs['time_periodic'] = time_periodic
         kwargs['netcdf_engine'] = netcdf_engine
+        if 'field_chunksize' in kwargs.keys() and kwargs['field_chunksize'] not in [False, None]:
+            FieldFileBuffer = DaskFileBuffer
```
CKehl, line 436: see comment from above l. 291

parcels/fieldfilebuffer.py
```diff
@@ -14,20 +14,232 @@
...
+                self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine=self.netcdf_engine, lock=False)
+            self.dataset['decoded'] = True
+        except:
+            logger.warning_once("File %s could not be decoded properly by xarray (version %s).\n         "
+                                "It will be opened with no decoding. Filling values might be wrongly parsed."
+                                % (self.filename, xr.__version__))
+            if self.lock_file:
+                self.dataset = xr.open_dataset(str(self.filename), decode_cf=False, engine=self.netcdf_engine)
+            else:
+                self.dataset = xr.open_dataset(str(self.filename), decode_cf=False, engine=self.netcdf_engine, lock=False)
```
CKehl, line 68: the locking only makes sense when one uses Dask, so in the 'regular' NetCDF file buffer, the locking can (or: shall) go. Conversely, when using a DaskFileBuffer, the locking needs to be done - hence the 'except' branch would throw the error then.

parcels/field.py
erikvansebille: Fixed in ff39e0e6fa6bc54ffe14de423bea09b48d556faf

parcels/field.py
erikvansebille: Good idea. I've renamed to `_grid_fb_class` in ff39e0e6fa6bc54ffe14de423bea09b48d556faf, as it mostly deals with reading the Grid (and could even go into grid.py??)

parcels/fieldfilebuffer.py
erikvansebille: Fixed in cc794eec3d2a2bf1b479dccf6d8884c45e987f9d

parcels/fieldfilebuffer.py
```diff
@@ -0,0 +1,580 @@
...
+            # (which we would rather want to have being auto-managed).
+            # If 'lock' is not specified, the Lock-object is auto-created and managed bz xarray internally.
+            self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine=self.netcdf_engine)
+            self.dataset['decoded'] = True
+        except:
+            logger.warning_once("File %s could not be decoded properly by xarray (version %s).\n         "
+                                "It will be opened with no decoding. Filling values might be wrongly parsed."
+                                % (self.filename, xr.__version__))
+            self.dataset = xr.open_dataset(str(self.filename), decode_cf=False, engine=self.netcdf_engine)
+            self.dataset['decoded'] = False
```
CKehl, line 47: nice - simplifies that again

parcels/fieldfilebuffer.py
```diff
@@ -0,0 +1,580 @@
...
+                data = data[self.indices['depth'], self.indices['lat'], self.indices['lon']]
+            else:
+                data = data[ti, self.indices['lat'], self.indices['lon']]
+        else:
+            if self._check_extend_depth(data, 1):
+                data = self._extend_depth_dimension(data, ti)
+            else:
+                data = data[ti, self.indices['depth'], self.indices['lat'], self.indices['lon']]
+
+        return data
```
CKehl, line 184: nice helper function!

parcels/fieldfilebuffer.py
```diff
@@ -0,0 +1,580 @@
...
+
+    @property
+    def data(self):
+        return self.data_access()
```
CKehl, line 188: thanks for keeping that! A 'lil background explanation: This way, I can also put 'benchmark decarators' on the actual data function (or timing decarators or anything) without 'removing' the property decarator. That is why I initially did that split-up.

parcels/fieldfilebuffer.py
```diff
@@ -0,0 +1,580 @@
...
+        convert_xarray_time_units(time_da, self.dimensions['time'])
+        time = np.array([time_da[self.dimensions['time']]]) if len(time_da.shape) == 0 else np.array(time_da[self.dimensions['time']])
+        if isinstance(time[0], datetime.datetime):
+            raise NotImplementedError('Parcels currently only parses dates ranging from 1678 AD to 2262 AD, which are stored by xarray as np.datetime64. If you need a wider date range, please open an Issue on the parcels github page.')
+        return time
+
+
+class DeferredNetcdfFileBuffer(NetcdfFileBuffer):
+    def __init__(self, *args, **kwargs):
+        super(DeferredNetcdfFileBuffer, self).__init__(*args, **kwargs)
```
CKehl, line 217: Little hint on that class: Basically the difference between the NetcdfFileBuffer and the Deferred... is that the NetcdfFileBuffer could load THE WHOLE file in one go into a common buffer in one go (-> 1 file access), while the deferred loader basically just loads 3 timesteps each time (-> N-2 file accesses).

parcels/examples/example_nemo_curvilinear.py
```diff
@@ -29,7 +29,7 @@ def run_nemo_curvilinear(mode, outfile, advtype='RK4'):
                        'data': data_path + 'V_purely_zonal-ORCA025_grid_V.nc4'}}
     variables = {'U': 'U', 'V': 'V'}
     dimensions = {'lon': 'glamf', 'lat': 'gphif'}
-    field_chunksize = {'y': 2, 'x': 2}
+    field_chunksize = {'y': 256, 'x': 512}
```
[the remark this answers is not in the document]
erikvansebille: Because otherwise running the example would become _extremely_ slow, timing out the 6-hour max of github CI. 

Note that previously these fields were not actually loaded with dask, as they have less than three time snapshots. So this would be loaded in netcdf, and the `field_chunksize` was irrelevant. 

In the new implementation, we also have a non-deferred dask fieldfilebuffer (which could be useful for e.g. super-fine-resolution 0.01 degree global bathymetry Fields), so the `field_chunksize` is now used. And that blew out the CI ;-)

Makes sense?

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -300,6 +317,27 @@ def test_ofam_3D(mode, chunk_mode):
...
+    lons, lats = 5e5, 5e5
+
+    pset = ParticleSet.from_list(fieldset=field_set, pclass=ptype[mode], lon=lons, lat=lats)
+    pset.execute(AdvectionRK4, runtime=delta(days=1), dt=delta(minutes=5))
+    # MITgcm_regridded sample file dimensions: time=10, XG=400, YG=200
+    assert (len(field_set.U.grid.load_chunk) == len(field_set.V.grid.load_chunk))
+    if chunk_mode in [False, 'auto']:
+        assert (len(field_set.U.grid.load_chunk) == 1)
+    elif chunk_mode == 'specific':
+        assert (len(field_set.U.grid.load_chunk) == (1 * int(math.ceil(400.0/50.0)) * int(math.ceil(200.0/100.0))))
```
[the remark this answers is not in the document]
erikvansebille: Yes, I'm confident these numbers are correct. I checked them during development, and the CI now passes

parcels/fieldfilebuffer.py
erikvansebille: Exactly; for now there's no difference in the implementation between the two yet, but creating the class now will allow us to more easily develop this functionality in following development

parcels/examples/example_nemo_curvilinear.py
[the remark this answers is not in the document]
CKehl: Makes total sense - yes, with chunks of 2x2, if the field is more than 100x100 cells, managing the individual chunks becomes the main bottleneck in Dask. Absolutely understandable reason.

parcels/field.py
```diff
@@ -343,63 +341,58 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
...
-                field_time[0] = 0
-            time_origin = TimeConverter(field_time[0])
-            field_time = time_origin.reltime(field_time)
-            if not np.all((field_time[1:]-field_time[:-1]) > 0):
-                id_not_ordered = np.where(field_time[1:] < field_time[:-1])[0][0]
-                raise AssertionError('Please make sure your netCDF files are ordered in time. First pair of non-ordered files: %s, %s'
-                                     % (dataFiles[id_not_ordered], dataFiles[id_not_ordered+1]))
-            if 'field_chunksize' in kwargs.keys() and grid.master_chunksize is None:
-                grid.master_chunksize = kwargs['field_chunksize']
+            _, _, _, dataFiles = collect_timeslices()
```
CKehl, line 361: okay, this looks nice - but functions definitions inside function definitions are calls for maintenance disasters. I approve the general idea of slimming and streamlining it - I don't like the particular way it is done.

parcels/field.py
```diff
@@ -317,9 +317,7 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
         if len(data_filenames) > 1 and 'time' not in dimensions and timestamps is None:
             raise RuntimeError('Multiple files given but no time dimension specified')
 
-        if grid is None:
-            # Concatenate time variable to determine overall dimension
-            # across multiple files
+        def collect_timeslices():
```
CKehl, line 320: see comment new-line 361.

parcels/fieldfilebuffer.py
```diff
@@ -0,0 +1,565 @@
...
+
+
+class DaskFileBuffer(NetcdfFileBuffer):
```
CKehl, line 220: not further reviewing this in-depth, as this will be adapted in another PR later - just looking if all required functions are there

CKehl (approved):

parcels/field.py
```diff
@@ -205,6 +205,38 @@ def get_dim_filenames(cls, filenames, dim):
         else:
             return filenames
 
+    @classmethod
```
CKehl, line 208: well, the question here really is: do we need any class-internal or object-internal functions ? having this be a class method is already MUCH nicer than some function-in-function construct, but the way I see it, it could also be a static method. @erikvansebille it's up to you - you can leave it this way, which is fine, or you can make it a static method. Either way you choose, I approve it.

CKehl (approved):

erikvansebille:
Since this is a major refactoring (mostly splitting off the NetcdfFileBuffer class into its own file) that is difficult to maintain as an open PR, I'm merging this PR now and will then open a new one that addresses the outstanding points of this

PR #906, opened 2020-08-05
Move convert_pset_to_dict from ParticleFile to ParticleSet

RoelBrouwer:
This pull request contains the minimal changes needed to move ParticleSet-implementation specific methods from the ParticleFile class to the ParticleSet class. This means moving the `convert_pset_to_dict` method and its supporting private function `_to_write_particles`.

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -35,6 +35,16 @@ if MPI:
 __all__ = ['ParticleSet']
 
 
+def _to_write_particles(pd, time):
+    """We don't want to write a particle that is not started yet.
+    Particle will be written if particle.time is between time-dt/2 and time+dt/2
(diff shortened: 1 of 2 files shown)


parcels/particleset.py
```diff
@@ -503,6 +513,58 @@ def lonlatdepth_dtype_from_field_interp_method(field):
                 return np.float64
         return np.float32
 
+    def convert_pset_to_dict(self, pfile, time, deleted_only=False):
```
CKehl, line 516: a simple rename to 'to_dict(self, ...)' I think would be nice.

RoelBrouwer: Ok! Changed in 0cfbf92.

PR #908, opened 2020-08-11
Kernel support for math.nan and NotEq operator

erikvansebille:
Somewhat surprisingly, we did not yet support the NotEq (`!=`) operator in Kernels. With this PR, that is now fixed. 

Also, there was no easy way to include a nan into a Kernel. Now, Kernels support `math.nan`

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -152,7 +152,7 @@ class ConstNode(IntrinsicNode):
 
 
 class MathNode(IntrinsicNode):
-    symbol_map = {'pi': 'M_PI', 'e': 'M_E'}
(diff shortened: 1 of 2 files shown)


parcels/codegenerator.py
```diff
@@ -152,7 +152,7 @@ def __getitem__(self, attr):
 
 
 class MathNode(IntrinsicNode):
-    symbol_map = {'pi': 'M_PI', 'e': 'M_E'}
+    symbol_map = {'pi': 'M_PI', 'e': 'M_E', 'nan': 'NAN'}
```
CKehl, line 155: agreed - I didn't even know that the NAN symbol exists in math.h, but it appears to be standard.

PR #911, opened 2020-08-31
Refactor C code for structured grid temporal interpolation

angus-g:
There's a lot of code duplication in the structured grid temporal function which hides the actual intent.

This was also hiding a behavioural difference where interpolation on a 2D field with `CGRID_VELOCITY` would cause an error only if the interpolation time exactly coincides with the time grid. The new behaviour is to not raise the error, regardless of the interpolation time.

I've factored out the logic to handle interpolation at 1/2 time points into `tii`, and wrapped it, along with the separate handling of 2D and 3D interpolation into a single macro called `INTERP`.

### DIFF
--- a/parcels/include/parcels.h
+++ b/parcels/include/parcels.h
@@ -222,7 +222,8 @@ static inline StatusCode spatial_interpolation_nearest2D(double xsi, double eta,
 }
 
 /* C grid interpolation routine for tracers on 2D grid */
-static inline StatusCode spatial_interpolation_tracer_c_grid_2D(float data[2][2], float *value)
+static inline StatusCode spatial_interpolation_tracer_c_grid_2D(double _xsi, double _eta,
+								float data[2][2], float *value)
 {
   *value = data[1][1];
   return SUCCESS;
@@ -241,7 +242,8 @@ static inline StatusCode spatial_interpolation_nearest3D(double xsi, double eta,
 }
 
 /* C grid interpolation routine for tracers on 3D grid */
-static inline StatusCode spatial_interpolation_tracer_c_grid_3D(float data[2][2][2], float *value)
+static inline StatusCode spatial_interpolation_tracer_c_grid_3D(double _xsi, double _eta, double _zeta,
+								float data[2][2][2], float *value)
 {
   *value = data[0][1][1];
   return SUCCESS;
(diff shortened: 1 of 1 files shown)


parcels/include/parcels.h
```diff
@@ -459,131 +461,82 @@ static inline StatusCode temporal_interpolation_structured_grid(type_coord x, ty
...
+  } while (0)
+
+  if ((interp_method == LINEAR) || (interp_method == CGRID_VELOCITY) ||
+      (interp_method == BGRID_VELOCITY) || (interp_method == BGRID_W_VELOCITY)) {
+    // adjust the normalised coordinate for flux-based interpolation methods
+    if ((interp_method == CGRID_VELOCITY) || (interp_method == BGRID_W_VELOCITY)) {
+      if (gridindexingtype == NEMO) {
+	// velocity is on the northeast of a tracer cell
+	xsi = 1;
+	eta = 1;
```
erikvansebille, line 513: Indentation not correct?

parcels/include/parcels.h
```diff
@@ -459,131 +461,82 @@ static inline StatusCode temporal_interpolation_structured_grid(type_coord x, ty
...
+      } else if (gridindexingtype == MITGCM) {
+	// velocity is on the southwest of a tracer cell
+	xsi = 0;
+	eta = 0;
+      }
```
erikvansebille, line 518: Again, indentation not correct?

parcels/include/parcels.h
```diff
@@ -459,131 +461,82 @@ static inline StatusCode temporal_interpolation_structured_grid(type_coord x, ty
...
-    else {
-        return ERROR;    
-    }
-    return SUCCESS;
+    status = getCell3D(f, xi[igrid], yi[igrid], zi[igrid], ti[igrid], data3D, tii == 1); CHECKSTATUS(status);
+  }
+
+  // define a helper macro that will select the appropriate interpolation method
+  // depending on whether we need 2D or 3D
+#define INTERP(fn_2d, fn_3d)						\
```
erikvansebille, line 491: Why the back-slashes here and in the lines below? Do they do anything?

erikvansebille (commented):
Thanks for this refactoring @angus-g; really helpful! And good that you caught the bug in cgrid-velocity :-)

A few minor review questions below

parcels/include/parcels.h
angus-g: C macros terminate at the end of the line, so we need continuation characters -- the whole backslashed block is the macro definition.

erikvansebille: Ah I see. Sorry for my ignorance on C; it's been a while since I've actively used it

parcels/include/parcels.h
angus-g: Yep, my editor is being silly in C mode (clearly I don't use it much either!)

Issue #873, opened 2020-06-24
Renaming parcels random library


erikvansebille:
The name of the Parcels random library often shadows the built-in random libraries, which can cause problems on compilation (both in JIT and Scipy). Giving the library a more explicit name would make future development easier

Thanks to @CKehl for bringing this up

erikvansebille:
This is fixed in #912

PR #912, opened 2020-09-02
Refactoring random to ParcelsRandom class

erikvansebille:
This implements #873. Note that this renaming of the random function is not backward-compatibel; users can choose two ways to adapt their code

1. Change 
    ```diff
    - from parcels import random
    + from parcels import ParcelsRandom as random
    ```

2. Change 
    ```diff
    - from parcels import random
    + from parcels import ParcelsRandom
    ``` 
    and rename all calls to `random` functions to `ParcelsRandom`

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -255,7 +255,7 @@ class IntrinsicTransformer(ast.NodeTransformer):
             node = StatusCodeNode(math, ccode='')
         elif node.id == 'math':
             node = MathNode(math, ccode='')
-        elif node.id == 'random':
+        elif node.id == 'ParcelsRandom':
             node = RandomNode(math, ccode='')
         elif node.id == 'print':
             node = PrintNode()
--- a/parcels/rng.py
+++ b/parcels/rng.py
@@ -13,7 +13,7 @@ from parcels.tools.loggers import logger
 __all__ = ['seed', 'random', 'uniform', 'randint', 'normalvariate', 'expovariate', 'vonmisesvariate']
 
 
-class Random(object):
+class RandomC(object):
     stmt_import = """#include "parcels.h"\n\n"""
     fnct_seed = """
 extern void pcls_seed(int seed){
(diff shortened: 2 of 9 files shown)


daanreijnders (approved):
Looks all good to me 👍

CKehl:
On my list - scheduled for Thursday morning

parcels/codegenerator.py
```diff
@@ -255,7 +255,7 @@ def visit_Name(self, node):
             node = StatusCodeNode(math, ccode='')
         elif node.id == 'math':
             node = MathNode(math, ccode='')
-        elif node.id == 'random':
+        elif node.id == 'ParcelsRandom':
```
CKehl, line 258: interesting that this change was required - I can't remember that I needed to do that. Well spotted!

parcels/rng.py
```diff
@@ -13,7 +13,7 @@
 __all__ = ['seed', 'random', 'uniform', 'randint', 'normalvariate', 'expovariate', 'vonmisesvariate']
 
 
-class Random(object):
+class RandomC(object):
```
CKehl, line 16: like the name - encapsulates the reason of why that is needed

tests/test_kernel_language.py
```diff
@@ -265,7 +265,7 @@ def kernel(particle, fieldset, time):
 
 
 def random_series(npart, rngfunc, rngargs, mode):
-    random = parcels_random if mode == 'jit' else py_random
+    random = ParcelsRandom if mode == 'jit' else py_random
     random.seed(1234)
     func = getattr(random, rngfunc)
     series = [func(*rngargs) for _ in range(npart)]
```
CKehl, line 271: yes, exactly - these parts are the ones that become quite difficult of all the different libraries are named 'random'.

CKehl (approved):
Yes - the changes look all very good. I think this now makes it more robust and reasy to switch RNG libraries and avoid scipy-jit deviations.

PR #913, opened 2020-09-02
ParticleSet Class Hierarchy 

RoelBrouwer:
Implementing the design from the UML class diagram as discussed in #894


parcels/particleset/baseparticleset.py
```diff
@@ -0,0 +1,38 @@
...
+            self.p.set_index(self._index)
+            result = self.p
+            self._index += 1
+            return result
+        # End of Iteration
+        raise StopIteration
+
+
+class BaseParticleAccessor(ABC):
+    """Interface."""
```
CKehl, line 32: Well, related to my previous comment: miscommunication of ideas lead to this weird construct now where we try to derive abstract classes from specific classes. Hence, in my head, it would rather look like
```
class ParticleSetIterator(ABC):
    """Interface"""
    def ___init___(self):
        _head = None
        _tail = None
        _current = None

    def __next__(self):
        """... <fill purpose here> ...."""
        return None

    @property
    def head(self):
        return self._head

    @property
    def tail(self):
        return self._tail

    @property
    def current(self):
        return self._current

class BaseParticleAccessor(ParticleIterator):
    """Generic base class for accessing a Particle"""
    def __init__(self):
        super(BaseParticleAccessor, self).__init__()

    @property
    def current_cptr(self):
        p = self.current()
        if isinstance(p, Particle):
            return p.cptr()
        else:
            return -1

<in e.g. particlesets/trees/accessor.py>
from parcels.particlesets.baseparticleset import BaseParticleAccessor

class TreeParticleAccessor(BaseParticleAccessor):
    """Implementation of the ParticleAccessor for particles organized in a tree data structure"""
    def __init__(self):
        <fill in stuff here for what you need to parse a tree>

    def __next__(self):
        <parse your tree to fast-access the 'next' element>
        return p

    @property
    def current(self):
        <ideally your 'next' method stores some information that manipulates the '_current' variable>
        <perform some validity checks>
        return self._current
        
```

remember that in that latter implementation class, you do not need to override all functions - just the ones that needed to be left undefined in the base class as they depend on the underlying data structure. Also, ideally ALL functions in the base class make sense for ALL implementations - but if not so, the base class should per default return None, and is not overriden in the implementations.

CKehl (commented):
Some comments remain - I guess we kind-of misunderstood or misaligned some intentions here. Perhaps it would be better to first again talk and fix the diagram before running further in code. That said, depends on how you'd like to consider the feedback. As the UML diagram is a baseline for your work now, I'm hesitant to just refit that diagram. But it would be good - for keeping the overview - to keep both the code and its structure in UML aligned and up-to-date.

RoelBrouwer:
Quick summary of what we discussed:
- The ParticleSetIterator as it currently exists will be removed. Its functionality will be merged into the ParticleAccessor. 
- The ParticleIterator interface from the UML diagram will not be a separate class, but will be directly included in the BaseParticleAccessor.

parcels/particleset/baseparticleset.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __next__(self):
+        ''''Returns the next value from ParticleSet object's lists '''
+        if self._index < self.max_len:
+            self.p.set_index(self._index)
+            result = self.p
+            self._index += 1
+            return result
+        # End of Iteration
+        raise StopIteration
```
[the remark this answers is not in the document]
RoelBrouwer: So, it is/was clear that `Collection` etc. were meant as abstract/interfaces. The confusion was caused by the current implementation of the `ParticleSetIterator` as a concrete class, which encapsulates part of the functionality of the `ParticleAccessor`. We will do away with this concrete  `ParticleSetIterator` class, as discussed, and include its functionality in the accessor.

parcels/particleset/baseparticleset.py
RoelBrouwer: The current commit was only meant to show really the very basic outline of which files/classes we would end up using and how they should relate to each other. I deliberately did not implement any interfaces (i.e. abstract methods) yet, but we were mostly on the same page there. As discussed, I will merge the `ParticleSetIterator` interface into the `BaseParticleAccessor` abstract class, and not have it as a separate class.

parcels/particlesets/particlesetsoa.py
```diff
@@ -45,9 +47,20 @@ def _to_write_particles(pd, time):
             & (np.isfinite(pd['time'])))
 
 
-class ParticleAccessor(object):
+class ParticleAccessorSOA(BaseParticleAccessor):
     def __init__(self, pset):
         self.pset = pset
+        self.max_len = pset.size
+        self._index = 0
+        self._iterindex = 0
```
RoelBrouwer, line 55: Working on this I begin to see a bit better why the Accessor and Iterator were implemented as separate objects and I am no longer convinced that it is definitely best to merge the two.

CKehl:
> Quick summary of what we discussed:
> 
>     * The ParticleSetIterator as it currently exists will be removed. Its functionality will be merged into the ParticleAccessor.
> 
>     * The ParticleIterator interface from the UML diagram will not be a separate class, but will be directly included in the BaseParticleAccessor.

* We try to keep the UML updated to reflect this class structure for ease of discussions

parcels/particlesets/particlesetsoa.py
CKehl: hm, okay - let's perhaps talk about that either Friday or Wednesday - looking forward to your explanation!

parcels/particlesets/baseparticleset.py
```diff
@@ -23,6 +23,13 @@ class NDCluster(ABC):
 # == class BaseParticleSetNDCluster):
 # ==     def __init__(self, collection_type): # this is NOT an abstract method then
 # ==         assert inspect.isclass(collection_type, ParticleCollection)
+# == Response RB: Yes, I think I agree with option (a). This disentangles the inheritance-structure quite a bit. It
+# ==              would mean that collections and particlesets form their own separate hierarchies (so no ParticleSet
+# ==              is a subclass of a collection), bounded together by a "has-a"-relation, meaning that a ParticleSet
+# ==              always 'carries' an instance of a (Particle)Collection. However, this has an impact on the data-access:
+# ==              Currently you call e.g. `p = pset.data_accessor().set_index(3)`, in the old inheritance-idea this would
+# ==              have become something like `p = pset.get_by_index(3)`, but with this suggestion that would change to
+# ==              `p = pset.collection.get_by_index(3)`.
```
CKehl, line 32: well, the actual implication is as such: if we do (a), then we need to replicate quite some data-access functions into the collection as forward-methods of particle set into its collection. So there will be some wrapper functions that some (but not me) would consider 'code replication'. I wouldn't be too nutty on that.

Option (b) is a little work-around some people would call a 'hack'. Still this would just be a minute formality, and considering the many 'hacks' that parcels already does for making the C-interface with arbitrary particles work, I'd say (b) is still quite a viable option.

Note to you @RoelBrouwer : it would be good for you to consider that we are probably (i) not gonna get around some implementation tricks to make this work, due to the C-embedding and the strict 'all MPI processes are workers' paradigm for the MPI simplicity (-> see my tricky workaround on ID management in the ID generator for lists) and that (ii) we are also not fixing all the 'hacks' in parcels. Target is: make it work, make it usable, make it adaptive, make it more manageable than now - not: make it UML-MVC conform (cause that is not gonna happen).

I'm okay with (a) as well as (b), but I do have a preference for (b) cause it would auto-enforce the implementation-via-inheritance of the required abstract functions in the specific ParticleSet implementation, where it counts.

parcels/particlesets/collections.py
```diff
@@ -619,7 +619,7 @@ def clear(self):
         pass
 
 
-class ParticleCollection(ABC, Collection):
+class ParticleCollection(Collection):
```
CKehl: why leaving out ABC ? ParticleCollection is still itself also an abstract class - and Python >=3.5 does actually facilitate multi-class inheritance (one of the actually cool features of newer Python versions).

parcels/particlesets/iterators.py
```diff
@@ -63,6 +72,12 @@ def update_next_dt(self, next_dt=None):
     def delete(self):
         # == OBJECTION CK: the actual operation, which is the particle's state, shall be done by the particle. So, == #
         # == this function should just forward the delete-call to the particle in question.                        == #
+        # == RESPONSE RB: There might be a problem with that in some cases. As we discussed, the ParticleAccessor  == #
+        # == basically acts like a shell around the particle (data), providing uniform access regardless of the    == #
+        # == underlying datastructure. The SOA approach, for example, does not use a particle class on which       == #
+        # == functions can be defined. So it makes sense, I think, to implement any function that requires more    == #
+        # == logic than just setting a property on the Accessor level. The state-thing is not really a good        == #
+        # == example, although the delete-alias may be useful over just treating it as a property.                 == #
```
CKehl, line 80: The issue is that if we, e.g., change state codes, or need to differentiate and check properties of a particle for execution the right functions, we wanna do that once, on the particle level, and then have this affecting all ParticleSet implementations. I do not wanna go over each specific particle set implementation when changing a particular particle behaviour that is common to all kinds of particle sets.

parcels/particlesets/iterators.py
```diff
@@ -74,6 +89,12 @@ def set_state(self, state):
 
         # == OBJECTION CK: the actual operation, which is the particle's state, shall be done by the particle. So, == #
         # == this function should just forward the delete-call to the particle in question.                        == #
+        # == RESPONSE RB: There might be a problem with that in some cases. As we discussed, the ParticleAccessor  == #
+        # == basically acts like a shell around the particle (data), providing uniform access regardless of the    == #
+        # == underlying datastructure. The SOA approach, for example, does not use a particle class on which       == #
+        # == functions can be defined. So it makes sense, I think, to implement any function that requires more    == #
+        # == logic than just setting a property on the Accessor level. The state-thing is not really a good        == #
+        # == example, although the delete-alias may be useful over just treating it as a property.                 == #
```
CKehl, line 97: see comment above.

CKehl:
General comment: I will now start taking over the pieces of the ParticleSetSOA init procedure over to the ParticleCollectionSOA. So, we are then gonna work on the same file (soa.py), but on different classes within it, minimising merge conflicts.

parcels/particlesets/collections.py
RoelBrouwer: It does facilitate multi-class inheritance, but not in this way. `Collection` already inherits from `ABC`. If `ParticleCollection` inherits from both, Python will not know what to do: `Cannot create a consistent method resolution order (MRO) for bases ABC, Collection`.
Besides, `ParticleCollection` does not explicitly need to inherit from `ABC` anyway. As long as not all abstract methods are instantiated, any subclass of Collection will be abstract itself. So leaving `ABC` out here has no negative consequences.

parcels/particlesets/iterators.py
RoelBrouwer: I sort of see your point. There is however no way around some code duplication in this way: as whatever you implement on the particle level has to be recreated somewhere else for datastructures that do not use particle objects (like the SoA approach). So then you end up with a bit of a halfway-solution, where for most datastructures the Accessor just passes the function call along, while for at least one it actually implements something by itself.

parcels/particlesets/baseparticleset.py
CKehl: we shall leave this discussion open as we may return to this later ...

parcels/particlesets/iterators.py
CKehl: agreed

parcels/kernel.py
```diff
@@ -411,11 +405,13 @@ def remove_deleted(pset):
         remove_deleted(pset)
 
         # Identify particles that threw errors
-        error_particles = np.isin(pset.particle_data['state'], [StateCode.Success, StateCode.Evaluate], invert=True)
+        # NOTE: this relies on the SOA implementation, it does not
+        # generalize. This needs to be reworked.
+        error_particles = np.isin(pset.collection._data['state'], [StateCode.Success, StateCode.Evaluate], invert=True)
```
CKehl: okay, this should rather be a list comprehension - just as it was prior to SoA and as it is in all branches. Reasoning here is that this execute function should later be a generic super-class function, where only the `execute_[ji|python]` actually change.

Can be in here for now as-is, but it will be subject to one of the earliest follow-up changes.

parcels/kernel.py
```diff
@@ -378,10 +378,14 @@ def execute(self, pset, endtime, dt, recovery=None, output_file=None, execute_on
...
         def remove_deleted(pset):
             """Utility to remove all particles that signalled deletion"""
-            indices = pset.particle_data['state'] == OperationCode.Delete
-            if np.count_nonzero(indices) > 0 and output_file is not None:
-                output_file.write(pset, endtime, deleted_only=indices)
-            pset.remove_booleanvector(indices)
+            # Indices marked for deletion.
+            bool_indices = np.array([
+                p.state == OperationCode.Delete for p in pset])
+            indices = np.where(bool_indices)[0]
```
CKehl, line 384: these are 2 separate loopings which effect performance negatively - why change that ?!

CKehl (changes requested):
FULL REJECT of the changes - they hamper performance, they make the whole code once more less-generalisable, and - most importantly - they are right now unnecessary to do.

CKehl:
> FULL REJECT of the changes - they hamper performance, they make the whole code once more less-generalisable, and - most importantly - they are right now unnecessary to do.

Proposal: if you really want to make that drastic change, we may wanna have a talk first ;-)

parcels/kernel.py
```diff
@@ -405,28 +409,35 @@ def remove_deleted(pset):
...
+                    particle.set_state(StateCode.Success)
+                    recovery_kernel(particle, self.fieldset, particle.time)
+                    if particle.state == StateCode.Success:
+                        particle.set_state(StateCode.Evaluate)
                 else:
-                    logger.warning_once('Deleting particle because of bug in #749 and #737')
-                    particles.delete()
+                    logger.warning_once(
+                        'Deleting particle because of bug in #749 and #737')
+                    particle.delete()
```
CKehl, line 434: I actually recently fixed that one in my benchmarking branch. It's a simple fix, so I may just add that right here.

The issue is: if the kernel sets the particle state already to 'Delete', this if-branch is executed and hence emits the error - although everything behaves just-as-planned.

CKehl (changes requested):
marked the requested changes in more detail

parcels/kernel.py
```diff
@@ -405,28 +409,35 @@ def remove_deleted(pset):
         remove_deleted(pset)
 
         # Identify particles that threw errors
-        # NOTE: this relies on the SOA implementation, it does not
-        # generalize. This needs to be reworked.
-        error_particles = np.isin(pset.collection._data['state'], [StateCode.Success, StateCode.Evaluate], invert=True)
-        while np.any(error_particles):
+        while True:
```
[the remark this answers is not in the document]
qubixes: How about something along the lines of 

`while n_tries < max_tries:`

That way you can catch infinite loops if they somehow occur.

parcels/kernel.py
```diff
@@ -405,28 +409,35 @@ def remove_deleted(pset):
...
+            n_error = 0  # Number of particles with a different code.
             # Apply recovery kernel
-            for p in np.where(error_particles)[0]:
-                particles = pset.collection.get_single_by_index(p)
-                if particles.state == OperationCode.StopExecution:
+            # Effectively loop over all particles that don't have sucess of
+            # evaluate codes.
+            for particle in pset:
```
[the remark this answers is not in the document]
qubixes: Yes, I understand. This will get fixed.

parcels/kernel.py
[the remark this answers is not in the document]
qubixes: Nvm, I already changed it. Still, a check on the number of tries probably wouldn't hurt?

parcels/particlesets/soa.py
```diff
@@ -799,7 +799,7 @@ def cstruct(self):
         'cstruct' returns the ctypes mapping of the particle data. This depends on the specific structure in question.
         """
         class CParticles(Structure):
-            _variable_ = [(v.name, POINTER(np.ctypeslib.as_ctypes_type(v.dtype))) for v in self._ptype.variables]
+            _fields_ = [(v.name, POINTER(np.ctypeslib.as_ctypes_type(v.dtype))) for v in self._ptype.variables]
```
RoelBrouwer, line 802: Changed this back to fields to make the current implementation work. But this is something to look into in the future w.r.t. appropriate naming (as discussed) @CKehl

CKehl (commented):
re-introducing the accessors is nice, but in the end, those functions actually need to modify the particle states themselves - or, in an SoA-case, the states of the related variables array. Otherwise, those modifications won't be visible in the kernel in the next run, which may cause trouble.

parcels/kernel.py
```diff
@@ -234,8 +234,8 @@ def load_lib(self):
 
     def execute_jit(self, pset, endtime, dt):
         """Invokes JIT engine to perform the core update loop"""
-        if len(pset) > 0 and pset.particle_data['xi'].ndim == 2 and pset.fieldset is not None:
-            assert pset.fieldset.gridset.size == pset.particle_data['xi'].shape[1], \
+        if len(pset) > 0 and pset.collection._data['xi'].ndim == 2 and pset.fieldset is not None:
+            assert pset.fieldset.gridset.size == pset.collection._data['xi'].shape[1], \
```
CKehl, line 238: just as I saw before, if you actually wanna change the state of the 'data' of a collection, a proper setter-getter would be appropriate for the data array, otherwise we're just breaking the member variable protection 'convention'

parcels/kernel.py
```diff
@@ -378,7 +378,7 @@ def execute(self, pset, endtime, dt, recovery=None, output_file=None, execute_on
 
         def remove_deleted(pset):
             """Utility to remove all particles that signalled deletion"""
-            indices = pset.particle_data['state'] == OperationCode.Delete
+            indices = pset.collection._data['state'] == OperationCode.Delete
```
CKehl: also here, better have a setter/getter for the data array, parametrized by the name of the variable.

parcels/particlesets/particlesetsoa.py
```diff
@@ -740,15 +740,15 @@ def search_kernel(particle, fieldset, time):
...
+            self._collection.ptype,
             funcname="search_kernel",
             funcvars=["particle", "fieldset", "time", "x"],
             funccode=f_str,
         )
         self.execute(pyfunc=k, runtime=0)
 
         if isinstance(particle_val, str):
-            particle_val = self.collection._data[particle_val]
+            particle_val = self._collection._data[particle_val]
```
CKehl, line 751: also here please setter/getter for `_data`

parcels/particlesets/soa.py
CKehl: noted #TODO

CKehl: ah, if you're extra-awesome, you could add a 
```
#TODO adapt naming of _fields_ in CParticles (code generator) into _variables_
```
at that code-step, which would then highlight at least in PyCharm

RoelBrouwer:
> re-introducing the accessors is nice, but in the end, those functions actually need to modify the particle states themselves - or, in an SoA-case, the states of the related variables array. Otherwise, those modifications won't be visible in the kernel in the next run, which may cause trouble.

Maybe I am misunderstanding your point, but the functions that are defined on the Accessor (in the BaseParticleAccessor class) do exactly this, right? They do depend on the correct implementation of __setattr__ and __getattr__ in the subclass for this, however. In the SoA implementation, these read from/write to the arrays in the ParticleCollection. So this should work.

parcels/kernel.py
RoelBrouwer: Absolutely, every line of code that looks like this means that section needs to be reworked. The purpose of changing these lines in this way, is to make them work again (for now at least). That is step 1. Step 2 is to gradually replace all code that depends on the specific implementation of the datastructure by code that does not, like Raoul has been doing in the execute function below.

parcels/kernel.py
RoelBrouwer: This line has already been replaced with an alternative, that does not use the _data array directly, by Raoul.

parcels/kernel.py
qubixes: @CKehl For this part of the code I made a particle_field_check() as a member of the particle set. Does that make sense to you? 

Is there a reason why this check is performed in execute_jit(), but not in execute_python()?

parcels/particlesets/particlesetsoa.py
```diff
@@ -0,0 +1,675 @@
...
+                                             subset=error_indices)
+
+    def __getattr__(self, name):
+        # Comment CK: this either a member function of the accessor or the collection - not the PSet itself
+        if name in self._collection._data:
+            return getattr(self._collection, name)
+        elif name in self.__dict__:
+            return self.__dict__[name]
+        else:
+            return False
```
RoelBrouwer, line 309: This is a point we have to discuss a bit more. We talked about it previously @CKehl. Anyway, it is not just the tests that use notation like `pset.lon`, there are user-use-cases as well (e.g. plotting trajectories). So it is probably worth keeping this notation in place (at least for now), even though it does not have an obvious efficient implementation in every imaginable datastructure.
I suggest we discuss this this afternoon, but I wanted to have this comment in place to refer to for the discussion :)
If we keep it this function should probably move up to the BaseParticleSet and the implementation of __getattr__ should be required by the Collection interface.

parcels/kernel.py
erikvansebille: I can answer that question. It's a legacy choice because we used to only have particle.xi in JIT mode. However, since #940 we also support particle.xi in Scipy mode, so I guess that this check also needs to be done in `execute_python()`

CKehl:
> > re-introducing the accessors is nice, but in the end, those functions actually need to modify the particle states themselves - or, in an SoA-case, the states of the related variables array. Otherwise, those modifications won't be visible in the kernel in the next run, which may cause trouble.
> 
> Maybe I am misunderstanding your point, but the functions that are defined on the Accessor (in the BaseParticleAccessor class) do exactly this, right? They do depend on the correct implementation of **setattr** and **getattr** in the subclass for this, however. In the SoA implementation, these read from/write to the arrays in the ParticleCollection. So this should work.

hm, yes, you're right in that argument. Well, we do have ParticleSet tests that change particle properties via the ParticleSet (before kernel execution) and change values in the kernel - if those test all succeed, the implementation is apparently correct.

parcels/particlesets/particlesetsoa.py
CKehl: I think we resolved it in the DevTalk today.

parcels/particlesets/baseparticleset.py
```diff
@@ -271,3 +246,257 @@ def ParticleFile(self, *args, **kwargs):
         """Wrapper method to initialise a :class:`parcels.particlefile.ParticleFile`
         object from the ParticleSet"""
         pass
+
+    def _set_particle_vector(self, name, value, indices=None):
```
CKehl, line 250: still not really generalisable - I'm gonna make a counter-proposal here.

parcels/particlesets/baseparticleset.py
```diff
@@ -271,3 +246,257 @@ def ParticleFile(self, *args, **kwargs):
...
+        if indices is None:
+            bool_indices = np.full(len(self), True)
+        else:
+            bool_indices = np.full(len(self), False)
+            bool_indices[indices] = True
+        for i, p in enumerate(self):
+            if bool_indices[i]:
+                setattr(p, name, value)
+
+    def _get_particle_vector(self, name, indices=None):
```
CKehl, line 271: still not really generalisable - I'm gonna make a counter-proposal here.

qubixes: That's not necessary. At the moment it is not used anywhere. I'll comment it out at least in my next commit.

parcels/particlesets/baseparticleset.py
qubixes: Same for this one. It will either be removed or rewritten.

parcels/particlesets/particlesetsoa.py
```diff
@@ -0,0 +1,528 @@
...
+                       return all indices that do not equal (one of)
+                       `compare_values`.
+        :return: Numpy array of indices that satisfy the test.
+        """
+        compare_values = np.array([compare_values, ]) if type(compare_values) not in [list, dict, np.ndarray] else compare_values
+        return np.where(np.isin(self._collection.data[variable_name], compare_values, invert=invert))[0]
+
+    def indexed_subset(self, indices):
+        return ParticleCollectionIteratorSOA(self._collection,
+                                             subset=indices)
```
RoelBrouwer, line 188: I believe this function is not currently used internally (correct me if I am wrong). Do we want to leave this in as a way (for the user) to interact with a subset of the particles?

parcels/particlesets/particlesetsoa.py
```diff
@@ -0,0 +1,528 @@
...
+    def remove_indices(self, indices):
+        """Method to remove particles from the ParticleSet, based on their `indices`"""
+        if type(indices) in [int, np.int32, np.intp]:
+            self._collection.remove_single_by_index(indices)
+        else:
+            self._collection.remove_multi_by_indices(indices)
+
+    def remove_booleanvector(self, indices):
+        """Method to remove particles from the ParticleSet, based on an array of booleans"""
+        self.remove_indices(np.where(indices)[0])
```
RoelBrouwer, line 424: I removed the code comments, but will restate what they said here: the functions need to be replaced / removed in the future.

tests/test_particle_collections.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.mark.parametrize('pt', ['soa'])
+def test_pset_get_single_by_index(fieldset, pt, npart=10):
+    pset = psettype[pt](fieldset, lon=np.linspace(0, 1, npart), lat=np.zeros(npart), pclass=JITParticle)
+    assert np.all(np.diff([pset._collection.get_single_by_index(i).lon for i in range(npart)]))
+
+
+@pytest.mark.parametrize('pt', ['soa'])
+def test_pset_get_single_by_ID(fieldset, pt, npart=10):
+    pset = psettype[pt](fieldset, lon=np.linspace(0, 1, npart), lat=np.zeros(npart), pclass=JITParticle)
+    ids = pset._collection._data['id']
```
RoelBrouwer, line 47: This way of getting the IDs works for the SOA implementation, but does not generalize to future datastructure implementations. Previously (currently in master) you used to be able to just call `pset.id` (like in the test below this one) to get the array with all particle IDs. This should still work. However, it will not be efficient for all (future) datastructures, but that is not really an issue for tests.

tests/test_particle_collections.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.mark.parametrize('pt', ['soa'])
+def test_pset_iteration_backward(fieldset, pt, npart=10):
+    pset = psettype[pt](fieldset, lon=np.linspace(0, 1, npart), lat=np.zeros(npart), pclass=JITParticle)
+    assert np.all(np.diff(np.array([p.id for p in reversed(pset)])) < 0)
+
+
+@pytest.mark.parametrize('pt', ['soa'])
+def test_pset_get_single_by_index(fieldset, pt, npart=10):
+    pset = psettype[pt](fieldset, lon=np.linspace(0, 1, npart), lat=np.zeros(npart), pclass=JITParticle)
+    assert np.all(np.diff([pset._collection.get_single_by_index(i).lon for i in range(npart)]))
```
RoelBrouwer, line 41: This comment also applies for the test(s) below.
I think these tests are fine, they test if we get 10 different particles. Which implies we encounter all ten particles at some point when retrieving them by index.
It would be nice to have a test that actually checks if the particles have the value we expect (so using e.g. `np.isclose()`). For the SOA implementation at least we should be able to predict what index / ID will be assigned to a particle, so we could actually do that. However, I am not sure we would also be able to do this for datastructures that do not internally work with indices (linked lists, trees). Then again, we might choose not to implement the get_by_index function for those at all...
For IDs it does generalize, I think. A particle that is added later, will have a higher ID. So sorting the list of IDs should allow for the prediction of the actual values for each particle.

To expand the coverage of the test even further, you could consider calling `pset.collection.get(i)`. This should forward the function call to the appropriate get function based on the type of the argument. So if `i` is a 32-bit integer, it will use `get_single_by_index` in the background (64-bit integer for IDs), and we get a test of the forwarding "for free". But this is just an idea.

In general I would say when referring to the collection, it would be better to use `pset.collection` rather than `pset._collection`. Both work, as a property is defined within the baseparticleset for the collection. But using `pset.collection` would set the right example, just in case somebody looks to the tests for inspiration.

CKehl:
I just read the diff-compare [`cab3c18..0f6755b`](https://github.com/OceanParcels/parcels/compare/cab3c18..0f6755b) - looks all fine for now, so I give an approve on the current state of the PR. The new test file looks also good.

Looks good.

tests/test_mpirun.py
```diff
@@ -10,16 +10,17 @@
...
-        repeatdt = 200*86400
         stommel_file = path.join(path.dirname(__file__), '..', 'parcels',
                                  'examples', 'example_stommel.py')
         outputMPI = tmpdir.join('StommelMPI.nc')
         outputNoMPI = tmpdir.join('StommelNoMPI.nc')
 
-        system('mpirun -np 2 python %s -p 4 -o %s -r %d' % (stommel_file, outputMPI, repeatdt))
-        system('python %s -p 4 -o %s -r %d' % (stommel_file, outputNoMPI, repeatdt))
+        system('mpirun -np 2 python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputMPI, repeatdt, maxage))
+        system('python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputNoMPI, repeatdt, maxage))
```
CKehl, line 23: well, with respect to what I showed with pyccel, this isn't great testing. The issue here is that it doesn't really check the correct processing of that MPI call. `system` just starts the python program an returns the return value of the `__main__` function. Hard to trace back the error here. Even more so, if using the `system(...)` system call, then we should at least evaluate the return value of the function and assert that that value is 0.

Alternatively, we could let the mpi-command be run by the ci-workflow, as *pyccel* is doing it.

If uou generally agree, but rather like to move it to a new issue, then let's do that - equally valid, I could understand that argumentation.

parcels/examples/example_stommel.py
```diff
@@ -71,8 +71,14 @@ def UpdateP(particle, fieldset, time):
     particle.p = fieldset.P[time, particle.depth, particle.lat, particle.lon]
 
 
+def AgeP(particle, fieldset, time):
+    particle.age += particle.dt
+    if particle.age > fieldset.maxage:
+        particle.delete()
```
CKehl, line 77: yes, testing the dynamic particle removal is a good idea.

parcels/examples/example_stommel.py
```diff
@@ -88,6 +94,7 @@ def stommel_example(npart=1, mode='jit', verbose=False, method=AdvectionRK4, gri
     class MyParticle(ParticleClass):
         p = Variable('p', dtype=np.float32, initial=0.)
         p_start = Variable('p_start', dtype=np.float32, initial=fieldset.P)
+        age = Variable('age', dtype=np.float32, initial=0.)
```
CKehl, line 97: `float32` should be sufficient for this case study. Generally, `age` shall be the 64-bit float as is the `time` variable of a particle.

CKehl (approved):

tests/test_mpirun.py
erikvansebille: Yes @ckehl, I think that more careful MPI testing should be an Issue/development on its own, potentially together with a refactoring of the MPI functionality in Parcels?

PR #916, opened 2020-09-07
Fixing a bug when Kernels with multiple docstrings could not be generated

### LINKED ISSUE #910: Docstring not removed from concatenated kernel
I'm seeing this on Python 3.8.5, running the test suite. The `test_mitgridindexing[mitgcm-jit]` (in `test_grids.py`) fails during kernel compilation:
```
/tmp/parcels-1000/4ecf0962eceb88a6505bd7495efc5b9f_0.c: In function ‘UpdateRAdvectionRK4’:
/tmp/parcels-1000/4ecf0962eceb88a6505bd7495efc5b9f_0.c:95:3: error: unknown type name ‘Advection’
   95 |   Advection of particles using fourth-order Runge-Kutta integration.
      |   ^~~~~~~~~
/tmp/parcels-1000/4ecf0962eceb88a6505bd7495efc5b9f_0.c:95:16: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘particles’
   95 |   Advection of particles using fourth-order Runge-Kutta integration.
      |                ^~~~~~~~~
```

The relevant generated kernel code is:
```c
static inline StatusCode UpdateRAdvectionRK4(MyParticlep *particles, int pnum, double time, CField *R, CField *U, CField *V)
{
  float parcels_tmpvar0, parcels_tmpvar1, parcels_tmpvar2, parcels_tmpvar3, parcels_tmpvar4, parcels_tmpvar5, parcels_tmpvar6, parcels_tmpvar7, parcels_tmpvar8;
  type_coord u1, v1, lon1, lat1, u2, v2, lon2, lat2, u3, v3, lon3, lat3, u4, v4;
  StatusCode err;
  {
    err = temporal_interpolation(particles->lon[pnum], particles->lat[pnum], particles->depth[pnum], time, R, &particles->xi[pnum*ngrid], &particles->yi[pnum*ngrid], &particles->zi[pnum*ngrid], &particles->ti[pnum*ngrid], &parcels_tmpvar0, LINEAR, MITGCM);
    parcels_tmpvar0 *= 1.0;
    CHECKSTATUS(err);
  }
  particles->radius[pnum] = parcels_tmpvar0;
  Advection of particles using fourth-order Runge-Kutta integration.

    Function needs to be converted to Kernel object before execution;
```

As you can see, the docstring from the `AdvectionRK4` function is included directly in the C code. Because the combination of kernels simply combines all the body code within a single function, the mechanism to strip the docstring doesn't remove the docstring embedded in the middle of the function.

This comes from the kernel merge with `AdvectionRK4` in the second position:
```python
    pset.execute(pset.Kernel(UpdateR) + AdvectionRK4,
                 runtime=delta(hours=14), dt=delta(minutes=5))
```

erikvansebille:
As reported in #910, the ast fix for python3.8 in #753 only appeared to work when the docstring was on the top of the concatenated kernel.

Now, we assume that all strings are docstrings or comments by default and can be removed. The only exception is the string `'parcels_customed_Cfunc_pointer_args'`, which is needed when c kernels are included (see `tests/test_kernel_language.py` -> `test_c_kernel`)

This fixes #910

### DIFF
--- a/parcels/codegenerator.py
+++ b/parcels/codegenerator.py
@@ -426,9 +426,6 @@ class KernelGenerator(ast.NodeVisitor):
         # Untangle Pythonic tuple-assignment statements
         py_ast = TupleSplitter().visit(py_ast)
 
-        # Store the docstring so that it can be removed in visit_Str (#753)
-        self.docstr = ast.get_docstring(py_ast)
-
         # Generate C-code for all nodes in the Python AST
         self.visit(py_ast)
         self.ccode = py_ast.ccode
@@ -909,15 +906,10 @@ class KernelGenerator(ast.NodeVisitor):
         node.ccode = c.Statement('printf("%s\\n", %s)' % (stat, vars))
 
     def visit_Str(self, node):
-
-        def _isdocstr(node):
-            # Check if node is docstr. Comparison only on text, not whitespace etc
-            return [c for c in node.s if c.isalpha()] == [c for c in self.docstr if c.isalpha()]
-
-        if self.docstr is not None and _isdocstr(node):
-            node.ccode = ''
-        else:
+        if node.s == 'parcels_customed_Cfunc_pointer_args':
             node.ccode = node.s
+        else:
+            node.ccode = ''
 
 
 class LoopGenerator(object):


angus-g (approved):
Thanks Erik, this does the trick on my Python 3.8 install!

Maybe the CI should be testing against a few Python versions, especially given there are quite a few changes along the Python 3 series?

CKehl (approved):

CKehl:
seems like a reasonable fix - I agree that we don't need to include comments and docstrings in the compilation. I think it would be good to see what the Parcels users make out of it - if there are any complaints for those comment-inclusions in the jit-compilations actually missing.

PR #919, opened 2020-09-09
Removing max version on pytest from environment

erikvansebille:
Trying to update the environment, so that it does not use python3.7 but instead uses the latest version on miniconda (currently 3.8.4). See also @angus-g's comment on #916


.github/workflows/ci-workflow.yml
```diff
@@ -27,23 +27,23 @@ jobs:
           ./.github/scripts/linux_install.sh
           ./.github/scripts/osx_test.sh
         shell: bash
-        timeout-minutes: 45
   test_linux:
-    name: Testing on Linux
+  test_linux_py36:
+    name: Testing on Linux: Python 3.6
     runs-on: ubuntu-latest
```
CKehl, line 33: so, this is interesting, and I am not a yaml expert so, for me, the question is now:

```
test_linux:
test_linux_py36:
    name: Testing on Linux: Python 3.6
[...]
```
Does that mean that all run parameters written for 'test_linux_py36' also apply to 'test_linux', or is test-linux non-function and left unparameterized ?

erikvansebille: Ah, that might actually be why the CI is not triggered... Removing the `test_linux` now

erikvansebille: Actually, it were the `:` in the name descriptions. Found out through using https://github.com/adrienverge/yamllint

PR #921, opened 2020-09-15
support specifying central longitude in 2d & 2dmovie trajectory plotting

dklink:
Hi there!  I'm using Parcels for investigating flows in the North Pacific, and while the default plotting is almost perfect, I thought it could benefit from being able to specify a central longitude.  This is a pretty essential feature if you're interested in anything in the Pacific.  Some of your plotting already supports this, I just connected the functionality to plotTrajectoriesFile, and made a minor adjustment to create_parcelsfig_axis to generate the correct xlabels when utilizing its central_longitude parameter.

I would write a test, but I'm not really sure how to do that, since "correctness" kind of depends a graphical output.  Anyways, please let me know what else I need to do to make this mergeable, if you're interested in it.  Thanks for such a great library!

P.S. method signature of plotTrajectoriesFile is modified, so [this](https://oceanparcels.org/gh-pages/html/#module-scripts.plottrajectoriesfile) documentation page should change; I modified the docstring of the method, not sure if that automatically modifies the documentation.  If not, let me know where I can do that.

### DIFF
--- a/parcels/plotting.py
+++ b/parcels/plotting.py
@@ -278,7 +278,10 @@ def create_parcelsfig_axis(spherical, land=True, projection=None, central_longit
         projection = cartopy.crs.PlateCarree(central_longitude) if projection is None else projection
         fig, ax = plt.subplots(1, 1, subplot_kw={'projection': projection})
         try:  # gridlines not supported for all projections
-            gl = ax.gridlines(crs=projection, draw_labels=True)
+            if isinstance(projection, cartopy.crs.PlateCarree) and central_longitude != 0:
+                gl = ax.gridlines(crs=cartopy.crs.PlateCarree(), draw_labels=True)  # central_lon=0 necessary for correct xlabels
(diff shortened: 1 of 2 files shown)


erikvansebille:
Thanks for proposing this change in the plotting function, @dklink! We very much appreciate your contribution.

Although it is not our intention to develop very advanced and fully customisable plotting functionality (that would be a project on its own; rather we suggest users to create their own plotting routines for more control), this is a nice feature that is useful for a lot of users working in the Pacific Ocean indeed

As for the docstring, we will not update it on http://oceanparcels.org/gh-pages/html/ just yet, because we want that documentation to mirror the version of the conda release. So as soon as we create a next release of Parcels, we will also update the documentation. OK?

dklink:
Awesome!  I totally agree with the design decision of not developing complex plotting features.  Glad you are interested in the change anyways!

Issue #922, opened 2020-09-17
Change to vertical indexing for W field in MOM5 B-grid output


hrsdawson:
Hi @erikvansebille and @daanreijnders. As discussed, the vertical velocity in MOM5 is defined at the base of the tracer cell, shown in the sketch below. Parcels currently expects the vertical velocity to be defined at the top surface of the grid cell, meaning that `w` is applied at the upper face rather than the bottom face of the grid cell when using MOM5 output in Parcels. Could the vertical indexing be modified for this case so that `w` is applied correctly to the base of the grid cell? This would be a neater approach than PR #890. Modifying the vertical indexing for `w` in this way assists with, though does not fix, the issue of stuck particles along coastal boundaries. 

[Indexing_sketch.pdf](https://github.com/OceanParcels/parcels/files/5236795/Indexing_sketch.pdf)

It does however raise the question of what `w` values to use at the surface boundary. I am currently applying `w` from the bottom of the top z layer (using the stationary z layer approximation) as discussed in issue #860. This causes particles to move through the surface which then need to be mapped to a submerge kernel. The other option is to set `w` to zero at the surface of the top z layer which would prevent the through-surface issue. I'm not sure which option is more appropriate - @Adele157 do you have any more thoughts on this?  

I have included a unit test for the vertical indexing below (apologies, couldn't get it to format properly). This is based on the MITgcm grid unit test with indexing modified in the vertical rather than the horizontal. I hope it works. I will send some time snapshots of the hydrodynamic data to you shortly. 

It would also be great if this [tutorial ](https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_indexing.ipynb) could be updated to include some more information on B-grids, such as MOM5 and this vertical indexing issue. I think this could be beneficial to other/future users. 

```python
@pytest.mark.parametrize('mode', ['scipy', 'jit'])
@pytest.mark.parametrize('gridindexingtype', ['mom5', 'nemo'])
@pytest.mark.parametrize('withtime', [False, True])`
def test_mom5gridindexing_3D(mode, gridindexingtype, withtime):
    xdim = zdim = 201
    ydim = 2
    a = c = 20000  # domain size
    b = 2
    lon = np.linspace(-a / 2, a / 2, xdim, dtype=np.float32)
    lat = np.linspace(-b / 2, b / 2, ydim, dtype=np.float32)
[...]

PR #923, opened 2020-09-18
Implementing gridindexing for MOM5 B-grids

erikvansebille:
Following code suggestion by @hrsdawson in #922, this PR will implement grid indexing for MOM5 B-grids.


tests/test_grids.py
```diff
@@ -675,8 +675,10 @@ def test_mitgridindexing_3D(mode, gridindexingtype, withtime):
         dimensions = {"lon": lon, "lat": lat, "depth": depth}
         dsize = (depth.size, lat.size, lon.size)
 
-    index_signs = {'nemo': -1, 'mitgcm': 1}
-    isign = index_signs[gridindexingtype]
+    hindex_signs = {'nemo': -1, 'mitgcm': 1, 'mom5': -1}
```
daanreijnders, line 678: I've been trying to wrap my head around it for a while, but also not yet sure why this passes. I'll look at it again next week, with a clear mind.

If MOM5s indexing is different than that in POP, we should have a unit test for both. Once we have the sample data from @hrsdawson, we'll also better be able to compare differences in both model outputs.

erikvansebille: OK, I think I know why changing `vindex_signs` does not make any difference to the particle trajectory. It's because the `W` field for such a radial flow does not vary with depth, only with longitude. So no matter how we change `depth[k] + vsign * dz` in calculating `r` and `phi`, `W` will stay the same.

So the question then is: is this a non-useful idealised unit test for this mom-grid example? Or am I thinking wrongly about how to calculate the grid levels for `U` and `W`? @daanreijnders and @hrsdawson, what are your thoughts?

daanreijnders: Why does W only change with longitude only change with longitude and not with depth? Particles are advected clockwise on a circle in the XZ-plane, right? W then changes with depth to provide the up- and downward motion to keep it on the circle.

daanreijnders: This should be resolved with https://github.com/OceanParcels/parcels/pull/923/commits/062ce93a91daf0b3c78560a30ac57cc57987726e. The problem was indeed that W in its current definition had no vertical gradient, so both indexing types (`mom5` and `pop`-like B-grids) would sample the same velocities. The solution was to rotate the coordinate system (swap U and W), also in the kernel.

CKehl (approved):

CKehl:
Apart from the adapted assertion message, it all looks okay to my eyes.

PR #931, opened 2020-10-08
Throw IOError when ParticleFile.tempwritedir already exists

### LINKED ISSUE #925: Strange netcdf behaviour when running multiple instances of parcels
I've come across some strange behaviour when running multiple oceanparcels processes at the same time (not using parcel's parallel capability, I mean running multiple parcels scripts on one machine at the same time). One of two things happens - the first is that there's no error, but the netcdf file generated at the end contains extra information. For instance, when running two processes at once, instead of the time axis being 

`0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100`

it might be 

`0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 90, 100`

The amount of 'extra data' seems to increase with the number of processes I'm running simultaneously (e.g. when running 15 at once I've had files that are 3x larger than they should be), so it looks like the process that generates the netcdf file is getting mixed up between the different processes. 

The second option is that the process crashes on the `output_file.export()` line, with the error message `No such file or directory: 'out-XXXXXXX/0/0.npy`. I've mainly seen this version when running the different instances as foreground processes.

I've tested this on two machines, and with the processes being run in the foreground or as background nohup processes, with the same issues in all cases. This issue occurs even when the different scripts are independent, i.e. using different files for the velocity field (and obviously writing to different netcdf files). All of these scripts work with no issues if only one process is running. It looks like the final step that converts the numpy files into netcdf is getting mixed up between output from different processes if more than one instance of parcels is running. Is there any way of getting around this (apart from running each file in its own directory)?

erikvansebille:
This fixes #925, by throwing an error if multiple processes want to write to the same temporary output directory

### DIFF
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -96,8 +96,10 @@ class ParticleFile(object):
             mpi_rank = 0
         self.tempwritedir = os.path.join(self.tempwritedir_base, "%d" % mpi_rank)
 
-        if pset_info is None:  # otherwise arrive here from convert_npydir_to_netcdf
-            self.delete_tempwritedir()
+        if not os.path.exists(self.tempwritedir):
+            os.makedirs(self.tempwritedir)
+        elif pset_info is None:
+            raise IOError("output directory %s already exists. Please remove first." % self.tempwritedir)
 
     def open_netcdf_file(self, data_shape):
         """Initialise NetCDF4.Dataset for trajectory output.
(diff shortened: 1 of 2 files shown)


CKehl (approved):
The changes are fine. Don't think I agree that the initial issue qualifies as a bug, but: the PR seems to break no tests, so that's okay.

PR #932, opened 2020-10-12
Updating Dask name maps

erikvansebille:
This PR is a follow-up of #905, and is meant to fix the last outstanding issues in that PR:

- [x] disentangling dask between grids and fields (#872)
- [x] provide more user control on name maps (#768)


erikvansebille:
After discussion with @CKehl, we came up with the following road-map/to-do-list

One core problem is that in NetCDF files, different Fields with the same dimensions can have different chunking sizes. In Parcels, however, these Fields will share the same Grid. Now, with auto-chunking, the chunking of the NetCFD files is used, and this creates problems. 
The simplest solution is:
- [ ] If using `field_chunksize=auto`, then assure that each `Field` has its own `Grid` (i.e. # Fields == # Grids). 

The drawback of this is memory overhead (both in the GridSet and in the Particle.xi etc). 
That is why we also propose
- [ ] Change the default `field_chunksize` from `auto` to `None`
- [ ] Implement/improve the `DeferredNetcdfFileBuffer` class, and use this class if `field_chunksize=None` and there are more than four timesteps in the NetCDF file

erikvansebille:
Note that forcing each `Field` to have its own `Grid` when `field_chunksize=auto` will not play nice with B-grid and C-grid interpolation, where U, V and W are expected to be on the same Grid. 

For this, I suggest to 
- [ ] change the assert so that we don't check if the Grids are the same objects/instances, but rather whether each of the velocity Grids has the same lon/lat/depth dimensions.

CKehl:
- [ ] rename `grid.master_chunksize` and `field.field_chunksize` and `filebuffer.field_chunksize` to: `grid.chunksize`, `field.chunksize` and `filebuffer.chunksize` (remedy the confusion in the terminology)

CKehl:
When integrating now the new dict-structure of the chunksize, I encountered that there are some 'abomination'-case possible, that we need to restrict, forbid or treat, e.g.

```
field_chunksize = {'U': False,
                   'V': None,
                   'W': 'auto'}
```

or

```
field_chunksize = {'U': {'xu_ocean': ('lon', 128), 'yu_ocean': ('lat', 128), 'zt_ocean': ('depth', 3)},
                   'V': (1, 5, 10, 10),
                   'W': 'auto'}
```

or

```
field_chunksize = {'U': {'xu_ocean': ('lon', 128), 'yu_ocean': ('lat', 128), 'zt_ocean': 'auto'},
                   'V': {'xv_ocean': ('lon', 128), 'yv_ocean': ('lat', 128), 'zt_ocean': 'auto'},
                   'W': {'timer': 'auto', 'xt_ocean': ('lon', 128), 'yt_ocean': ('lat', 128), 'zw_ocean': 'auto'}}
```

The  combination possibilities of '(semi-)valid' is so large that treating all combinations is tricky, so we may need to constrain what we allow and what we prohibit.

erikvansebille:
Regarding all the possible corner cases: I suggest we make it easy and clean and _only_ support explicit chunk-sizes in chunk_namemaps. IN other words, we have three options for `field_chunksize`:
1. `None` for all Fields (default, but with info message if total FieldSet is very large, than maybe chunking is a good idea)
2. `auto` for all Fields
3. explicit dict with _only_ integers for all Fields

The idea is that if users go through the trouble of specifying the chunksizes explicitly, it's little effort to do that for all Fields

CKehl:
@erikvansebille good, I like your definition. Then all the other cases should raise an ideally-comprehensible error message. Otherwise I already foresee practitioners to try those unintended combinations -> Murphy's Law

CKehl:
- [ ] provide (programmer-)comprehensible doctstrings to those member functions to make the learning curve and future adaptations easier

tests/test_fieldset.py
```diff
@@ -287,13 +288,16 @@ def test_fieldset_samegrids_from_file(tmpdir, filename='test_subsets'):
...
-    assert fieldset.U.grid == fieldset.V.grid
-    assert fieldset.U.grid.master_chunksize == fieldset.V.grid.master_chunksize
-    assert fieldset.U.field_chunksize == fieldset.V.field_chunksize
+    if field_chunksize == 'auto':
+        assert fieldset.gridset.size == 2
+        assert fieldset.U.grid != fieldset.V.grid
+    else:
+        assert fieldset.gridset.size == 1
+        assert fieldset.U.grid == fieldset.V.grid
+        assert fieldset.U.grid.master_chunksize == fieldset.V.grid.master_chunksize
```
CKehl, line 299: isn't this completely verbose ? if the grids are the same, it means it's the same object, so this assertion is entirely redundant.

tests/test_fieldset.py
```diff
@@ -328,14 +333,17 @@ def test_fieldset_diffgrids_from_file(tmpdir, filename='test_subsets'):
...
 
-    fieldset = FieldSet.from_netcdf(files, variables, dimensions, timestamps=timestamps, allow_time_extrapolation=True, field_chunksize=chs)
-    assert fieldset.gridset.size == 2
-    assert fieldset.U.grid != fieldset.V.grid
+    fieldset = FieldSet.from_netcdf(files, variables, dimensions, timestamps=timestamps, allow_time_extrapolation=True, field_chunksize=field_chunksize)
+    if field_chunksize == 'auto':
+        assert fieldset.gridset.size == 2
+    else:
+        assert fieldset.gridset.size == 2
+        assert fieldset.U.grid != fieldset.V.grid
```
CKehl, line 342: in this case, both cases are exactly the same. We have two variables, so if we have two grids the grids automatically need to be different, isn't that so ?

tests/test_fieldset.py
```diff
@@ -352,11 +360,14 @@ def test_fieldset_diffgrids_from_file_data(tmpdir, filename='test_subsets'):
...
-    fieldset_file = FieldSet.from_netcdf(files, variables, dimensions, timestamps=timestamps, allow_time_extrapolation=True, field_chunksize=chs)
+    fieldset_file = FieldSet.from_netcdf(files, variables, dimensions, timestamps=timestamps, allow_time_extrapolation=True, field_chunksize=field_chunksize)
 
     fieldset_file.add_field(field_data, "B")
     assert len(fieldset_file.get_fields()) == 3
-    assert fieldset_file.gridset.size == 2
+    if field_chunksize == 'auto':
+        assert fieldset_file.gridset.size == 3
+    else:
+        assert fieldset_file.gridset.size == 2
```
CKehl, line 370: That's a nice check - works really well for distinguishing the scenarios

parcels/fieldset.py
```diff
@@ -444,25 +464,13 @@ def from_netcdf(cls, filenames, variables, dimensions, indices=None, fieldtype=N
...
-                            res = False
-                            if (isinstance(chunksize, tuple) and isinstance(grid.master_chunksize, tuple)) or (isinstance(chunksize, dict) and isinstance(grid.master_chunksize, dict)):
-                                res |= functools.reduce(lambda i, j: i and j, map(lambda m, k: m == k, chunksize, grid.master_chunksize), True)
-                            else:
-                                res |= (chunksize == grid.master_chunksize)
-                            if res:
-                                grid_chunksize = grid.master_chunksize
-                                logger.warning_once("Trying to initialize a shared grid with different chunking sizes - action prohibited. Replacing requested field_chunksize with grid's master chunksize.")
-                            else:
-                                raise ValueError("Conflict between grids of the same fieldset chunksize and requested field chunksize as well as the chunked name dimensions - Please apply the same chunksize to all fields in a shared grid!")
```
CKehl, line 458: Now, I am not entirely happy with this solution out of performance reasons. We did agree that, if the chunksize is 'auto' or the chunksizes are different, the fields get different grids. Now, as we compare full sub-dictionaries in line 450, it leads us to a situation where, even if U an V essentially chunk their fields the same sizes, if the NetCDF dimensions and their chunking only differs by the name (see MITgcm -> YC/YG and XC/XG), then they will get different grids. Even if, numerically, they could get the same grid. It's not satisfying from an efficieny viewpoint. 

What this restriction does do is lead to trimmed, less code, indeed.

erikvansebille: Good point, @CKehl. I have now pushed a suggestion (plus unit test) that checks if the values (not the dimension names) are the same for the different chunking dictionary entries. Would that be a solution?

CKehl: i'll look at it - but probably not today. Today I need to wrap up something else.

CKehl:
well, now we have to take over the changes in `from_netcdf` to `add_field` in `fieldset.py`

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -224,17 +240,21 @@ def test_pop(mode, chunk_mode):
...
     assert (len(field_set.U.grid.load_chunk) == len(field_set.W.grid.load_chunk))
     if chunk_mode is False:
+        assert field_set.gridset.size == 1
         assert (len(field_set.U.grid.load_chunk) == 1)
     elif chunk_mode == 'auto':
+        assert field_set.gridset.size == 3  # because three different grids in 'auto' mode
+        # assert (len(field_set.U.grid.load_chunk) != 1)
         assert (len(field_set.U.grid.load_chunk) == 1)
     elif chunk_mode == 'specific':
+        assert field_set.gridset.size == 1
```
CKehl, line 250: good point checking that!

parcels/fieldfilebuffer.py
```diff
@@ -188,36 +188,47 @@ def __init__(self, *args, **kwargs):
...
-                  'lat': ['lat', 'nav_lat', 'y', 'latitude', 'la', 'lt', 'j', 'YC', 'YG'],
-                  'depth': ['depth', 'depthu', 'depthv', 'depthw', 'depths', 'deptht', 'depthx', 'depthy', 'depthz',
-                            'z', 'z_u', 'z_v', 'z_w', 'd', 'k', 'w_dep', 'w_deps', 'Z', 'Zp1', 'Zl', 'Zu', 'level'],
-                  'time': ['time', 'time_count', 'time_counter', 'timer_count', 't']}
+    _static_name_maps = {'time': ['time', 'time_count', 'time_counter', 'timer_count', 't'],
+                         'depth': ['depth', 'depthu', 'depthv', 'depthw', 'depths', 'deptht', 'depthx', 'depthy',
+                                   'depthz', 'z', 'z_u', 'z_v', 'z_w', 'd', 'k', 'w_dep', 'w_deps', 'Z', 'Zp1',
+                                   'Zl', 'Zu', 'level'],
+                         'lat': ['lat', 'nav_lat', 'y', 'latitude', 'la', 'lt', 'j', 'YC', 'YG'],
+                         'lon': ['lon', 'nav_lon', 'x', 'longitude', 'lo', 'ln', 'i', 'XC', 'XG']}
```
erikvansebille, line 196: Do we still need this now that we're asking users to provide the mapping explicitly?

parcels/field.py
```diff
@@ -375,9 +375,8 @@ def from_netcdf(cls, filenames, variable, dimensions, indices=None, grid=None,
                                                         dimensions, indices, netcdf_engine)
             kwargs['dataFiles'] = dataFiles
 
-        field_chunksize = kwargs.get('field_chunksize', None)
-        if field_chunksize and grid.master_chunksize is None:
-            grid.master_chunksize = field_chunksize
+        chunksize = kwargs.get('chunksize', None)
+        grid.chunksize = chunksize
```
CKehl, line 379: yes, nice - this way, with Field.chunksize and Grid.chunksize are clear -> less confusing

parcels/fieldfilebuffer.py
```diff
@@ -188,36 +188,47 @@ def __init__(self, *args, **kwargs):
...
+        This function enters the physical file (equivalent to a 'with open(...)' statement) and returns a file object.
+        In Dask, with dynamic loading, this is the point where we have access to the header-information of the file.
+        Hence, this function initializes the dynamic loading by parsing the chunksize-argument and maps the requested
+        chunksizes onto the variables found in the file. For auto-chunking, educated guesses are made (e.g. with the
+        dask configuration file in the background) to determine the ideal chunk sizes. This is also the point
+        where - due to the chunking, the file is 'locked', meaning that it cannot be simultaneously accessed by
+        another process. This is significant in a cluster setup.
+        """
+        if self.chunksize not in [False, None, 'auto'] and type(self.chunksize) is not dict:
+            raise AttributeError("'chunksize' is of wrong type. Parameter is expected to be a list, tuple or dict per data dimension, or be False, None or 'auto'.")
```
erikvansebille, line 226: Are list and tuple still allowed? Would it not be even simpler to only support dicts?

erikvansebille (commented):
I've now gone through the new fieldfilebuffer:DaskFileBuffer and it looks good and much cleaner. 

I wonder whether we could even further clean this up by being more stringent on the input format of the chunking dictionary?

parcels/fieldset.py
```diff
@@ -383,7 +320,10 @@ def from_netcdf(cls, filenames, variables, dimensions, indices=None, fieldtype=N
                'linear_invdist_land_tracer', 'cgrid_velocity', 'cgrid_tracer' and 'bgrid_velocity'
         :param gridindexingtype: The type of gridindexing. Either 'nemo' (default) or 'mitgcm' are supported.
                See also the Grid indexing documentation on oceanparcels.org
-        :param field_chunksize: size of the chunks in dask loading
+        :param chunksize: size of the chunks in dask loading. Default is None (no chunking). Can be None or False (no chunking),
+               'auto' (chunking is done in the background, but results in one grid per field individually), a four-value
+               tuple (time_chunksize, z_chunksize, y_chunksize, x_chunksize) that only works on A-Grids, or a dict in the format
+               '{parcels_varname: {netcdf_dimname : (parcels_dimname, chunksize_as_int)}, ...}', where 'parcels_dimname' is one of ('time', 'depth', 'lat', 'lon')
```
CKehl: Nice description - I killed the note on the tuple, cause that is firmly removed now.

parcels/gridset.py
```diff
@@ -17,39 +15,27 @@ def add_grid(self, field):
...
-                    logger.warning_once("Trying to initialize a shared grid with different chunking sizes - action prohibited. Replacing requested field_chunksize with grid's master chunksize.")
-                else:
-                    raise ValueError("Conflict between grids of the same gridset: major grid chunksize and requested sibling-grid chunksize as well as their chunk-dimension names are not equal - Please apply the same chunksize to all fields in a shared grid!")
-                break
+
+            if (g.chunksize != grid.chunksize) and (grid.chunksize not in [False, None]):
+                for dim in grid.chunksize:
+                    if grid.chunksize[dim][1] != g.chunksize[dim][1]:
+                        sameGrid &= False
+                        break
```
CKehl, line 37: nice and short

CKehl (approved):

parcels/fieldfilebuffer.py
CKehl: oh, tuples and lists are not supported anymore.

parcels/fieldfilebuffer.py
CKehl: we need this - for auto-chunking and for fall-back fixes in case user-definition and file definitions don't match. See l. 573 ff.

CKehl:
> After discussion with @CKehl, we came up with the following road-map/to-do-list
> 
> One core problem is that in NetCDF files, different Fields with the same dimensions can have different chunking sizes. In Parcels, however, these Fields will share the same Grid. Now, with auto-chunking, the chunking of the NetCFD files is used, and this creates problems.
> The simplest solution is:
> 
>     * [x]  If using `field_chunksize=auto`, then assure that each `Field` has its own `Grid` (i.e. # Fields == # Grids).
> 
> 
> The drawback of this is memory overhead (both in the GridSet and in the Particle.xi etc).
> That is why we also propose
> 
>     * [x]  Change the default `field_chunksize` from `auto` to `None`
> 
>     * [ ]  Implement/improve the `DeferredNetcdfFileBuffer` class, and use this class if `field_chunksize=None` and there are more than four timesteps in the NetCDF file

I think we can accept the latter point too - unless the code (i.e. the FieldSet- and Field-code) still opens a `DaskFileBuffer` when the chunksize is `False` or `None`. In that case, we may wanna change that to use the `DeferredNetcdfFileBuffer` class.

parcels/field.py
erikvansebille: Well, Ideally I would have wanted to get rid of `grid.chunksize` and only have `field.chunksize`, but then we can't check for duplicate grids in https://github.com/OceanParcels/parcels/blob/fixing_dask_namemaps/parcels/gridset.py#L33-L37. If you have a way to check the chunking there without need for the `grid.chunksize`, then we can get rid of the line `grid.chunksize = chunksize`

CKehl: no no - misunderstanding. I just said that this change the way it is now, that is fine.

erikvansebille: Well, but I would still prefer to get rid of `grid.chunksize`, just can't figure out how. My question to you @CKehl was: can you think of a way to get rid of `grid.chunksize` and still do the check in https://github.com/OceanParcels/parcels/blob/fixing_dask_namemaps/parcels/gridset.py#L33-L37?

erikvansebille:
> > After discussion with @CKehl, we came up with the following road-map/to-do-list
> > One core problem is that in NetCDF files, different Fields with the same dimensions can have different chunking sizes. In Parcels, however, these Fields will share the same Grid. Now, with auto-chunking, the chunking of the NetCFD files is used, and this creates problems.
> > The simplest solution is:
> > ```
> > * [x]  If using `field_chunksize=auto`, then assure that each `Field` has its own `Grid` (i.e. # Fields == # Grids).
> > ```
> > 
> > 
> > The drawback of this is memory overhead (both in the GridSet and in the Particle.xi etc).
> > That is why we also propose
> > ```
> > * [x]  Change the default `field_chunksize` from `auto` to `None`
> > 
> > * [ ]  Implement/improve the `DeferredNetcdfFileBuffer` class, and use this class if `field_chunksize=None` and there are more than four timesteps in the NetCDF file
> > ```
> 
> I think we can accept the latter point too - unless the code (i.e. the FieldSet- and Field-code) still opens a `DaskFileBuffer` when the chunksize is `False` or `None`. In that case, we may wanna change that to use the `DeferredNetcdfFileBuffer` class.

OK, I've just updated the tick!

parcels/field.py
CKehl: well, I needed to add that to prevent adding fields (via `add_fields(...)`) to the grid even though the grid was initialised with a different chunksize already. I'll look into that.

CKehl: so, the short answer is: because there is no way to get the field(s) of a grid fromout the `Grid` or `GridSet` itself, we cannot get rid of `Grid.chunksize` without also loosing the failsafe-check.

CKehl:
hm, just saw that a test from one of the other examples now failed - disappointing. Need to look into it. @erikvansebille perhaps you can have a look on ´examples/text_nemo_curvilinear.py´ and correlate it with the NEMO tests from ´examples/example_dask_chunk_OCMs.py´ ?

CKehl:
PS: I think the test fails for a reason - showing that the ´test_nemo_curvilinear.py´ was not correctly chunking the field before, it seems.

erikvansebille:
> PS: I think the test fails for a reason - showing that the ´test_nemo_curvilinear.py´ was not correctly chunking the field before, it seems.

OK fixed it in https://github.com/OceanParcels/parcels/pull/932/commits/bb2545abff86460e04c2ec69a4dc76fa7e501f42. It was an old-syntaxed chunking dictionary that now threw an error. Good that that was found

parcels/tools/statuscodes.py
```diff
@@ -26,6 +26,16 @@ class ErrorCode(object):
...
+class DaskChunkingError(RuntimeError):
+    """
+    Error indicating to the user that something with setting up Dask and chunked fieldsets went wrong.
+    """
+
+    def __init__(self, src_class_type, message):
+        msg = "[{}]: {}".format(str(src_class_type), message)
+        super(DaskChunkingError, self).__init__(msg)
+
+
```
erikvansebille, line 38: Nice implementation of a new ErrorCode, @CKehl!

PR #935, opened 2020-10-13
Fixing bug when concatenating kernels with ParcelsRandom functions in scipy

erikvansebille:
As reported by @daanreijnders, scipy breaks with an `Error: name 'ParcelsRandom' is not defined` when concatenating two kernels in which at least one call ParcelsRandom. This does not occur in JIT, only in Scipy. 

This PR will fix this bug

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -37,6 +37,7 @@ from parcels.field import VectorField
 from parcels.grid import GridCode
 from parcels.kernels.advection import AdvectionRK4_3D
 from parcels.kernels.advection import AdvectionAnalytical
+import parcels.rng as ParcelsRandom  # noqa
 from parcels.tools.statuscodes import StateCode, OperationCode, ErrorCode
 from parcels.tools.statuscodes import recovery_map as recovery_base_map
 from parcels.tools.loggers import logger
@@ -116,6 +117,7 @@ class Kernel(object):
             try:
                 user_ctx = stack[-1][0].f_globals
                 user_ctx['math'] = globals()['math']
+                user_ctx['ParcelsRandom'] = globals()['ParcelsRandom']
                 user_ctx['random'] = globals()['random']
                 user_ctx['StateCode'] = globals()['StateCode']
                 user_ctx['OperationCode'] = globals()['OperationCode']
--- a/tests/test_kernel_language.py
+++ b/tests/test_kernel_language.py
@@ -4,7 +4,6 @@ from parcels.kernels.EOSseawaterproperties import PressureFromLatDepth, PtempFro
 from parcels import ParcelsRandom
 import numpy as np
 import pytest
-import random as py_random
 from os import path
 import sys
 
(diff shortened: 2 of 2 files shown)


parcels/kernel.py
```diff
@@ -116,6 +117,7 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None,
             try:
                 user_ctx = stack[-1][0].f_globals
                 user_ctx['math'] = globals()['math']
+                user_ctx['ParcelsRandom'] = globals()['ParcelsRandom']
                 user_ctx['random'] = globals()['random']
```
CKehl, line 121: why is this then still required ? What kind of test breaks when removing this line ? to my understanding, those stack-parses are only required because of JIT, JIT should always use ParcelsRandom, so from the current viewpoint I don't see why this is still here. Maybe I just check out the PR and see it myself, for curiosity

tests/test_kernel_language.py
```diff
@@ -265,11 +264,10 @@ def kernel(particle, fieldset, time):
...
 
 def random_series(npart, rngfunc, rngargs, mode):
-    random = ParcelsRandom if mode == 'jit' else py_random
-    random.seed(1234)
-    func = getattr(random, rngfunc)
+    ParcelsRandom.seed(1234)
+    func = getattr(ParcelsRandom, rngfunc)
     series = [func(*rngargs) for _ in range(npart)]
-    random.seed(1234)  # Reset the RNG seed
+    ParcelsRandom.seed(1234)  # Reset the RNG seed
```
CKehl, line 270: why has this now changed ? This would kind-of defy the purpose of making sure one can - if wanted - use the normal Python RNG, but then just in SciPy mode.

tests/test_kernel_language.py
```diff
@@ -287,13 +285,31 @@ class TestParticle(ptype[mode]):
...
+
+    def RandomKernel(particle, fieldset, time):
+        particle.p += ParcelsRandom.uniform(0, 1)
+
+    def AddOne(particle, fieldset, time):
+        particle.p += 1.
+
+    kernels = pset.Kernel(RandomKernel)+pset.Kernel(AddOne) if concat else RandomKernel
+    pset.execute(kernels, dt=0)
+    assert pset.p > 1 if concat else pset.p < 1
```
CKehl, line 310: that's a nice test making sure that concatenated kernels just the same RNG

CKehl (commented):
Conditional approve - the implemented functionality is okay, but there are 2 things in the review that I am unsure. The removal of the Python-internal RNG for SciPy mode simulations makes little sense to me (possibly need to review the random module again), and conversely leaving in the random package in the stack-parse then makes even less sense to me. But I may just be wrong myself ... I'll check out the PR.

parcels/kernel.py
erikvansebille: I could indeed remove `user_ctx['random'] = globals()['random']`, but then users won't be able to concat python 'random' functions in scipy mode (note that this all applies to scipy kernel concatenation). For backward compatibility, I decided to keep it

tests/test_kernel_language.py
erikvansebille: Well, I could leave it the old way, but in fact I'd support also using the ParcelsRandom in scipy mode. The old version of this unit test also still worked with the bugfix of course, but I think this new one is tidier

tests/test_kernel_language.py
erikvansebille: Thanks!

CKehl (approved):

PR #936, opened 2020-10-14
Support POP interpolation on bottom layer through custom interpolation

erikvansebille:
Before, interpolation in the bottom layer of the POP B-Grid was supported by adding a layer of zeros to the field, upon `FieldSet` creation. However, this could be slow and code was not very clean

With this PR, we implement support for the bottom layer interpolation through custom extrapolation, like we did with MOM5 data in #923 (there the extrapolation was at the surface, as the W-layers were _below_ U and V, in POP they are _above_)


parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -226,9 +226,8 @@ def test_pop(mode, chunk_mode):
     if chunk_mode is False:
         assert (len(field_set.U.grid.load_chunk) == 1)
     elif chunk_mode == 'auto':
-        assert (len(field_set.U.grid.load_chunk) != 1)
+        assert (len(field_set.U.grid.load_chunk) == 1)
```
erikvansebille, line 229: To be honest; not sure why this unit test had to be changed now...

parcels/include/parcels.h
```diff
@@ -262,6 +274,13 @@ static inline StatusCode spatial_interpolation_tracer_c_grid_3D(double _xsi, dou
   return SUCCESS;
 }
 
+static inline StatusCode spatial_interpolation_tracer_c_grid_bottom(double _xsi, double _eta, double _zeta,
```
daanreijnders, line 277: This naming doesn't make sense if it's used for B-grids, right?

parcels/fieldset.py
```diff
@@ -636,6 +636,8 @@ def from_pop(cls, filenames, variables, dimensions, indices=None, mesh='spherica
                       W nodes are at the centre of the horizontal interfaces.
                       They are interpolated linearly (as a function of z) in the cell.
                       T node is at the cell centre, and constant per cell.
+                      Note that Parcels assumes that the length of the depth dimension (at the W-points)
+                      is one larger than the size of the velocity and tracer fields in the depth dimension.
```
daanreijnders, line 640: Is this how pop-data is usually provided?

daanreijnders (commented):
I'm not yet sure about whether "the dimension of length of the depth dimension (at the W-points) is [/should be] one larger than the size of the velocity and tracer fields in the depth dimension." Is this common for B-grid data? Or should the user specify the bathymetry through this manually?

parcels/field.py
```diff
@@ -945,6 +949,9 @@ def interpolator3D(self, ti, z, y, x, time):
                 xsi*(1-eta) * data[yi, xi+1] + \
                 xsi*eta * data[yi+1, xi+1] + \
                 (1-xsi)*eta * data[yi+1, xi]
+            if self.gridindexingtype == 'pop' and zi >= self.grid.zdim-2:
+                # Since POP is indexed at cell top, allow linear interpolation of W to zero in upper cell)
```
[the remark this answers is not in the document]
pdnooteboom: Read the POP manual section 5.2 and 5.2.1: http://www.cesm.ucar.edu/models/cesm1.2/pop2/doc/sci/POPRefManual.pdf . I understand that W at the top layer changes the sea surface height anomaly \eta, and no tracers can be transported through the surface z=\eta. Is the best thing to interpolate between bottom W and W==0 at z==\eta (instead of z==0) in the upper cell?

parcels/fieldset.py
erikvansebille: Apparently, but it would be good to confirm that with other users

parcels/include/parcels.h
erikvansebille: Well, the tracer interpolation is the same for B-grids and C-grids (tracer is uniform over the grid cell). I'll change the function name to `spatial_interpolation_tracer_bc_grid*`, for clarity

parcels/field.py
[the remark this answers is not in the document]
erikvansebille: Sorry, this was a type-o in the comment. It has to be "Since POP is indexed at cell top, allow linear interpolation of W to zero in _lowest_ cell". That's also what the condition `zi >= self.grid.zdim-2` is for. I've now fixed the comment text

parcels/include/parcels.h
daanreijnders: Ah, of course!

parcels/examples/example_dask_chunk_OCMs.py
CKehl: if you had to change that, then something was wrong - what was the error message here ? what was the actual value of `field_set.U.grid.load_chunk` ?

parcels/examples/example_dask_chunk_OCMs.py
```diff
@@ -226,9 +226,8 @@ def test_pop(mode, chunk_mode):
...
     elif chunk_mode == 'specific':
-        print(field_set.U.grid.chunk_info)
```
CKehl, line 231: yes, that can go - but for analysis, it would be nice to know the output for the above case.

parcels/fieldfilebuffer.py
```diff
@@ -136,51 +136,21 @@ def _check_extend_depth(self, data, di):
...
-                t0 = ti[0]
-                t1 = ti[-1] + 1
-                data = self.lib.concatenate((data[t0:t1, d0:d1 - 1, lat0:lat1, lon0:lon1],
-                                             self.lib.zeros((t1 - t0, 1, lat1 - lat0, lon1 - lon0))), axis=1)
-            else:
-                data = self.lib.concatenate((data[ti, d0:d1 - 1, lat0:lat1, lon0:lon1],
-                                             self.lib.zeros((1, lat1 - lat0, lon1 - lon0))), axis=0)
-
-        return data
-
```
CKehl, line 167: is this code really ONLY there because of POP ? If yes - that info would have been good in the docstring. If that could still occur for other fields, possibly good to leave it in, and extend the doctstring. Or do I lack understanding here: does your MOM5-indexing fix mean that this workaround is just never necessary anymore ? Does your MOM5-indexing fix the root cause of this workaround, instead of just a 'symptome' ? Then, sure: this can go - cool!

erikvansebille: Yes, this code was only there for POP; the whole point of this PR was to indeed remove this code (and make Parcels faster for POP because it doesn't require adding an extra layer anymore now)

parcels/examples/example_dask_chunk_OCMs.py
erikvansebille: The result of this `print(field_set.U.grid.chunk_info)` statement (placed in the `elif chunk_mode == 'auto'` clause) is now `[3, 1, 1, 1, 20, 60, 60]`, while it is `[3, 2, 1, 1, 20, 1, 60, 60]` in the current master. Note that the second entry has changed from `2` to `1` due to this PR. Not sure why..

parcels/examples/example_dask_chunk_OCMs.py
erikvansebille: It is now `[2]`, while it is `[2 0]` in the current `master` branch. Not sure why, but see also https://github.com/OceanParcels/parcels/pull/936#discussion_r509144766 below..

CKehl: well, did you change the test data in their layout ?

parcels/examples/example_dask_chunk_OCMs.py
CKehl: yes, I see the change - I don't see exactly why that's the case unless the data layout has changed ...

parcels/examples/example_dask_chunk_OCMs.py
erikvansebille: No, but the PR did change the size of the FieldSet because we don't add an extra layer to the bottom of the Fields anymore. That could cause this change?

parcels/examples/example_dask_chunk_OCMs.py
CKehl: overall, this here should not occur (i.e. the `load_chunk` field should not have a length of 1) unless (a) the data is in a flattened array or (b) the test data is too small for further chunking. In the latter case, we can check if lowering the chunk buffer size at the start of  that method changes the outcome of this if-check statement.

erikvansebille: Well, the only unit test I had to change is the `chunk_mode='auto'` one. The `specific` still works. I think that indeed, because we have now one depth level less, `auto` now defaults to only one chunk

parcels/fieldset.py
pdnooteboom: For me, as a user of POP, there is two depths provided: 
1. The w_dep, which is the depth of the W-points and has a size which is one larger than velocity and tracer fields in the depth dimension.
2. The depth_t, which is the depth at the T-points and has the same size as the velocity and tracer fields in the depth dimension.

PR #938, opened 2020-10-15
Fixing bug when FieldSet.add_vector_field() called without adding individual Fields

### LINKED ISSUE #937: AttributeError: 'VectorField' object has no attribute 'ccode_name'
Hello,
I'm trying to read the Stokes drift velocity from CMEMS files, and add them to another (existing) fieldset
I tried like this:
`def get_wav_fields3(fieldset,WAV_path,mesh_mask):`
`    basepath = WAV_path + '2018/12/' + '/*.nc'`
`    fnames=[];`
`    fnames += sorted(glob(str(basepath)))`
`    filenames =  {'Us': {'lon': mesh_mask, 'lat': mesh_mask, 'data': fnames}, 'Vs': {'lon': mesh_mask, 'lat': mesh_mask, 'data': fnames}}`
`    dimensions = {'Us': {'lon': 'longitude', 'lat': 'latitude', 'time': 'time'},'Vs': {'lon': 'longitude', 'lat': 'latitude', 'time': 'time'}}`
`    variables =  {'Us': 'VSDX', 'Vs': 'VSDY'}`
`    fs=FieldSet.from_netcdf(filenames,variables,dimensions)`
`    UVs = VectorField('UVs', fs.Us, fs.Vs)`
`    fieldset.add_vector_field(UVs)`
`    return fieldset`

I also tried reading separately `Us` and `Vs` using Field.from_netcdf , then adding them directly to the `fieldset`, as well as creating a `UVs` VectorField and adding that also to `fieldset`

I tried also reading the longitude and latitude coordinates from the datafiles, rather than from a separate coordinates file.

If I'm not mistaken, the kernel can have just 1 fieldset argument, that's why I add the Stokes drift Vectorfield to an existing fieldset.

I then call a custom kernel with 2 velocity components:
`def Dbl_AdvectionRK4(particle, fieldset, time):`
`    # Double advection of particles using fourth-order Runge-Kutta integration`
`    # using velocities from both current and Stokes drift`
`    (u1c, v1c) = fieldset.UV[time, particle.depth, particle.lat, particle.lon]`
`    (u1s, v1s) = fieldset.UVs[time, particle.depth, particle.lat, particle.lon]`
`    u1 = u1c + u1s`
`    v1 = v1c + v1s`
`    lon1, lat1 = (particle.lon + u1*.5*particle.dt, particle.lat + v1*.5*particle.dt)`
and so on.

Whatever the choices above, during execution of the kernel, the following error pops up: _AttributeError: 'VectorField' object has no attribute 'ccode_name'_
Using the regular RK4 kernel works without problems, so the VectorField in the error must be the Stokes drift one.

What I am doing wrong, how can I have the C code find the _ccode_name_ attribute for the UVs VectorField  ?

Thank you !
Luc

PS. The reason for not simply using SummedField, is that the existing velocity in the fieldset is a NestedField ( #927  )

erikvansebille:
This should fix #937


parcels/fieldset.py
```diff
@@ -210,6 +213,9 @@ def add_vector_field(self, vfield):
         :param vfield: :class:`parcels.field.VectorField` object to be added
         """
         setattr(self, vfield.name, vfield)
+        for v in vfield.__dict__.values():
+            if isinstance(v, Field) and (v not in self.get_fields()):
```
CKehl, line 217: what is this ?!
a) isn't that exactly what we have the get_fields()-method for ?
b) how could that if-statement even be true ?

Also, in this case the object-object comparison can be quite expensive. As far as I understand, once a Field is created, it's name is fixed, so why not ?
```
ref_names = [f.name for f in self.get_fields()]
for v in vfield.get_fields():  # if that function doesn't exist, it seams plausible to add it
    if v.name not in ref_names:
        self.add_field(v)
```

Then you (a) don't forcefully treat `vfield` as a dict (which may run its own errors) and (b) avoid complex object-in-object comparisons.

CKehl (changes requested):
The `vfields.__dict__.values()` call comes with some pitfalls as not all objects are dict-generalisable. Hence I would prefer this to be changed. Apart from that, it looks good.

parcels/fieldset.py
erikvansebille: I've now changed the code as you suggested. Do note that the `VectorField.get_fields()` still requires a call to `__dict__`; I think we can assume that users only call this function with Fields; and otherwise get an error

PR #939, opened 2020-10-15
Including an option to pass cartopy features and coastline resolution.

axnsantana:
In some cases, the land with lower reasolution (default) covers the particle set making it impossible to visualize.

Including an option to pass cartopy features to plotparticles and create_parcelsfig_axis.
The parameter cartopy_features is a list of features to be added.
The coastline resolution can be defined setting the land variable as a string (e.g. '10m' or '50m').

Example:

```
import cartopy
import cartopy.feature as cfeature

land_50m = cfeature.NaturalEarthFeature('physical', 'land', '50m',
                                        edgecolor='face',
                                        facecolor=cfeature.COLORS['land'])

plotTrajectoriesFile('track_file.nc', mode='movie2d_notebook', land='50m', cartopy_features=[land_50m])
# or from a ParticleSet object pset
pset.show(field=fieldset.U, land='50m', cartopy_features=[land_50m])
```

### DIFF
--- a/parcels/plotting.py
+++ b/parcels/plotting.py
@@ -42,7 +42,7 @@ def plotparticles(particles, with_particles=True, show_time=None, field=None, do
 
     if field is None:
         spherical = True if particles.fieldset.U.grid.mesh == 'spherical' else False
-        plt, fig, ax, cartopy = create_parcelsfig_axis(spherical, land, projection)
+        plt, fig, ax, cartopy = create_parcelsfig_axis(spherical, land, projection, cartopy_features=kwargs.pop('cartopy_features', []))
         if plt is None:
             return  # creating axes was not possible
         ax.set_title('Particles' + parsetimestr(particles.fieldset.U.grid.time_origin, show_time))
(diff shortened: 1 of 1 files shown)


parcels/plotting.py
```diff
@@ -42,7 +42,7 @@ def plotparticles(particles, with_particles=True, show_time=None, field=None, do
 
     if field is None:
         spherical = True if particles.fieldset.U.grid.mesh == 'spherical' else False
-        plt, fig, ax, cartopy = create_parcelsfig_axis(spherical, land, projection)
+        plt, fig, ax, cartopy = create_parcelsfig_axis(spherical, land, projection, cartopy_features=kwargs.pop('cartopy_features', []))
```
CKehl, line 45: quick and simple - looks good!

CKehl (approved):

reint-fischer (approved):
Simple and clear addition. Note that drawing landfeatures may cover drawn particles even at high resolution, depending on the model resolution: see this figure. Also note that the plotting functionality is not meant to accomodate all user options: see [this tutorial](https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_plotting.ipynb) However, I do not see any reason to exclude this elegant small addition.
SMOC_resolution

Issue #556, opened 2019-03-04
particle.xi in scipy mode


delandmeterp:
While in JIT mode, we store `particle.xi[]`, `particle.yi[]` and `particle.zi[]` that keeps track of the cell in which the particle is located, we don't do it in Scipy mode.
Even if scipy mode is not meant to be as efficient as possible but rather an easy development configuration, unstored particle.xi are a serious drawback largely affecting the efficiency.
We should store them as it is done in JIT mode

philippemiron:
Sorry, could you help me with this, where are the indices (particle.xi[]) use only in the JIT case? I can't seem to find it...

erikvansebille:
In the new SOA implementation of the ParticleSet, ScipyParticles also have `xi` etc (see [here](https://github.com/OceanParcels/parcels/blob/master/parcels/particle.py#L177-L180)). However, these are not used and/or updated in the Field interpolation. I am working on also implementing them in ScipyParticle, but it's not simple to do

erikvansebille:
This is now fixed in #940

PR #940, opened 2020-10-16
Implementing Field[Particle] sampling

erikvansebille:
In this Pull Request, we implement a second way for calling sampling a Field. Instead of `Field[time, depth, lat, lon]`, users can now also Sample a Field with `Field[particle]`. 

Not only is this `Field[particle]` sampling much easier (and perhaps more intuitive), it also means that we can find the `particle.xi` (location in the grid on the last search), which could massively speed up Parcels in Scipy mode, especially on curvilinear grids)! This PR would implement #556

Development still to do
- [x] Also support JIT mode
- [ ] Clean up the field.eval() method (see TODOs in the diff)
- [x] Implement new sampling in `VectorField`, `NestedField`, `SummedField` etc classes too
- [x] Support `FieldSets` with multiple grids
- [x] Change the built-in Advection Kernels to use this new sampling mode
- [x] Assess/time how much faster this new sampling is for Scipy mode

Note that the old method of calling a Field will also be maintained, not only for backwards compatibility, but also because in some use cases (e.g. calculating a derivative of a Field), it is necessary to sample a Field on a different location than the particle's

Note also that this PR does not affect the speed of Field Sampling in JIT mode, since that has always used the particle's location in the grid on the last search.


erikvansebille:
OK, so I think I've now implemented this `Field[particle]` sampling for Scipy. Note that, since e.g. AdvectionRK4 also samples at locations that are different from the particle location/time, I also had to implement `Field[time, depth, lat, lon, particle]` sampling.

I've also been able now to time the improvements. See the updated [jit-vs-scipy tutorial](https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/field_sampling_with_particle_as_argument/parcels/examples/tutorial_jit_vs_scipy.ipynb). The summary is
1. There is **no change** in the speed of Field sampling in **JIT mode** (was not expected, as we already used the `particle.xi` info in JIT anyways)
2. There is **no change** in the speed of Field sampling in **Scipy mode for Rectilinear Grids**. This is because grid index location searching does not use `particle.xi` in Scipy-Rectilinear mode (and I don't really see how to change this)
3. There is a **>>10 times speed improvement** of Field sampling in **Scipy mode for Curvilinear Grids**. So that is what we were hoping for!

The question now is whether this speed improvement for Curvilinear Grids in Scipy mode is a sufficiently important use case to warrant this major change to how we sample Fields. To be honest, I'm not entirely sure about that. Also because even with this change, Scipy mode is still much slower than JIT mode (see point 1 above). @CKehl and others, keen to hear your feedback!

CKehl (approved):

CKehl:
looks good so far - ready to merge, I guess

PR #943, opened 2020-11-03
Updating documentation_MPI notebook to use new chunksize API

erikvansebille:
Note that, because the documentation and tutorials follow the releases of Parcels, this PR should only be merged after #932 is merged into master _and_  this master is released. That is also why this change is a different PR than #932


reint-fischer (approved):
These are the types of updates to documentation/tutorials that are very nice to do immediately and release later

CKehl:
Is this PR now concluded ?

CKehl (approved):

erikvansebille:
We're waiting until the release of the next version of Parcels (v2.2.2) so that the notebook on the website aligns with the conda-version of Parcels. That's why it's not merged yet

reint-fischer:
I have added the documentation of the `populate_indices` function introduced in #954 . @erikvansebille, can you check if this is correct?

angus-g:
The example in the documentation does indeed match the intended use of the function: you only need to call it to perform the initial index search.

erikvansebille:
I've done some minor rewording of the documentation, but it looks good! This is ready for merging after we release version 2.2.2

PR #948, opened 2020-11-10
Treat NaN points as land in invdist interpolation

angus-g:
In addition to zeroes, it's pretty common for land points to be masked out, appearing as NaN within the dataset. This way we can avoid extra post-processing to make these files work with land-aware interpolation.

Hopefully I'm not mistaken in there being any use for this: I'm not sure why 0 was chosen in the first place, so maybe it doesn't make sense to do it this way.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -892,6 +892,7 @@ class Field(object):
             return val
         elif self.interp_method == 'linear_invdist_land_tracer':
             land = np.isclose(self.data[ti, yi:yi+2, xi:xi+2], 0.)
+            land |= np.isnan(self.data[ti, yi:yi+2, xi:xi+2])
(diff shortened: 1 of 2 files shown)


erikvansebille:
Thanks for this PR, @angus-g. The reason that we set land values to zero in Parcels is because the built-in Scipy interpolator can't handle NaNs well. See this snippet of the code in `Field.__init__()`: https://github.com/OceanParcels/parcels/blob/master/parcels/field.py#L156-L158

So unless a user somehow works around this code, land should always be zero. When do you experience otherwise?

angus-g:
Ah thanks, I missed that! I'll have to check whether it's still happening for us (with JIT), but it looks like it should be handled so I'll close this one.

PR #953, opened 2020-11-18
allow multiple c_includes

lucvandenbulcke:
If one uses the following sequence, the C code liste into "c_include" will now be added to the final code generated by Parcels.
kernels = pset.Kernel(AdvectionRK4)
kernels += pset.Kernel(ckernel_something, c_include=c_include)

One can also add different c code with different kernels; they will be concatenated


erikvansebille (approved):
Nice fix

PR #954, opened 2020-11-18
Add a method to pre-seed curvilinear indices using kdtree

### LINKED ISSUE #947: Initial index search improvement
From what I can tell, the index search routines (both rectilinear and curvilinear, but in particular the latter) use an iterative algorithm to find the cell indices containing a given point. In the rectilinear case, this is just a step search in each direction orthogonally, which is probably bounded by `max(xdim, ydim)` iterations. For a curvilinear grid, however, the algorithm equates to finding the target position in the local coordinates of each cell, stepping to an adjacent cell if it ends up outside `[0,1]x[0,1]`. I assume this is then bounded by `xdim * ydim` iterations, however each iteration involves a pretty expensive calculation for the local coordinate transformation.

I've been playing around with data on a curvilinear grid of about 6500x3000 points, and the time to calculate the initial indices for all particles is a fairly significant cost. To get around this, I used a [k-d tree](https://en.wikipedia.org/wiki/K-d_tree), e.g. Scipy's [cKDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html). I actually used a C implementation, and called it as part of `search_indices_curvilinear`, but upon further reflection, I don't think this is necessary. Given the previous cell indices, the search algorithm should only take a couple of steps to find the new cell, assuming the particle didn't move too far.

To help out the index search, doing a 4-neighbour query for each point (hopefully) gives us the indices of the cell bounds. I found that this worked for most of the domain, but some areas required an extra iteration or two of the regular index search to converge on the right place. I dropped the maximum iteration limit `maxIterSearch` to 3 and never exceeded it. Anyway, perhaps this could be a useful addition to ParticleSet initialisation: the k-d tree is fairly cheap to construct and query, and would probably give a sufficiently "close enough" answer even for nested domains to drastically reduce the number of search iterations.

angus-g:
This adds a dependency on the pykdtree library, which seems to be about ~10x faster than scipy's cKDTree (a drop-in replacement). The initial index search is only implemented for curvilinear grids, as the benefit may not be as pronounced vs. construction time for rectilinear grids.

Closes #947.

### DIFF
--- a/parcels/particlesets/particlesetsoa.py
+++ b/parcels/particlesets/particlesetsoa.py
@@ -7,6 +7,7 @@ import numpy as np
 import xarray as xr
 
 from parcels.grid import GridCode
+from parcels.grid import CurvilinearGrid
 from parcels.kernel import Kernel
 from parcels.particle import JITParticle
 from parcels.particlefile import ParticleFile
@@ -20,6 +21,11 @@ try:
     from mpi4py import MPI
 except:
     MPI = None
+# == comment CK: prevents us from adding KDTree as 'mandatory' dependency == #
+try:
+    from pykdtree.kdtree import KDTree
+except:
+    KDTree = None
 
 __all__ = ['ParticleSet', 'ParticleSetSOA']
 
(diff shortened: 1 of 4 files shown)


erikvansebille:
Thanks for adding this code, @angus-g! Am I right that the  `populate_indices()` method isn't actually called by any function at the moment? So users would have to do it themselves? Should it not be called in `ParticleSet.__init__`?

Also, @RoelBrouwer, @qubixes and @CKehl are working on #913, the major refactoring of ParticleSets. Do we want to merge this new code in master and then port also to the new ParticleSets? Or implement directly there?

angus-g:
> Am I right that the populate_indices() method isn't actually called by any function at the moment?

That's right: with pending refactors, I didn't know if it was best to default to a call to this method.

CKehl:
hm - well, so: it will create a merge-conflict with the refactoring, indeed.

Still, the proposed change sounds okay - would be something then either for the `ParticleSetSoA` or `CollectioSoA`. For non-index based structures it wouldn't connect to it - doesn't fit cause they don't work with indices anyway.

I think it would be best to wait with *this* PR until the refactoring is done, and then apply the changes from here.

CKehl:
so, I now merged the current `master` in parcels with Angus' `curvilinear-index-search` branch. The merge now looks fine to me. Made the use of KDTree conditional for now to prevent adding it at this point as mandatory dependency for an 'optional' improvement - may change in the future with particle-particle interaction, but for now let's keep it optional.

The branch has my approval. @angus-g and @erikvansebille , please have a look again and if it's approved, then we can merge the PR (conditional that all ticks are green)

parcels/particlesets/particlesetsoa.py
```diff
@@ -187,6 +193,36 @@ def indexed_subset(self, indices):
         return ParticleCollectionIteratorSOA(self._collection,
                                              subset=indices)
 
+    def populate_indices(self):
```
erikvansebille, line 196: Do we need to also highlight this method in a tutorial or example? @reint-fischer, can you think of a good place to document this?

angus-g:
Thanks for the rebase @CKehl, looks good to me (pending a decision on documentation per the review above)!

erikvansebille:
Just had a chat with @reint-fischer, and we decided to now merge this PR and then put the documentation in another PR (#943) which we can merge _after_ we do a new release, so that this new functionality is only advertised when it is available on a conda release

PR #964, opened 2020-11-26
Create tutorial_boundary_particles.ipynb

reint-fischer:
Mainly typing out the thoughts behind the notebook, these may be trimmed down later.


reint-fischer:
I have implemented the solution I have written for my master thesis and renamed the document to 'documentation' because I use SMOC data that is not in the Parcels examples. I have looked at the NEMO documentation for their implementation of lateral boundaries: https://www.nemo-ocean.eu/doc/node58.html The normal velocity is zero at the boundary and there are different options for the tangential velocity boundary condition. In my profiles both velocities still go to zero at the boundary. In the free-slip and partial-slip conditions in NEMO the tangential velocity is non-zero at the boundary and the profiles are linear. The main question I have (also in bold in C grid section) is how Parcels does the C grid interpolation normal to the velocity component. I suspect this is piecewise constant. In any case I think to be consistent with the computational grid, the profiles should both be linear, but the tangential velocity should not go to zero. To be consistent with reality is another matter...

reint-fischer:
Hi @cpierard, @quintenbohte,

I am writing the documentation notebooks on the Parcels boundary problems. We have decided to make it a multiple part series:
1. Introducing and visualizing the problem and showing the underlying assumptions. [Stuck particle notebook](https://github.com/OceanParcels/parcels/blob/boundary_conditions/parcels/examples/documentation_stuck_particles.ipynb)
2. Documenting the typical workarounds - ## still todo
3. Implementing solutions inside parcels - ## still todo

At this point, I thought I would like to introduce you to the first notebook and ask you if you would want to take a quick look at it for some feedback since you have been thinking about this problem also a bit already. Then maybe for the next notebook (2.) we can work together a bit. Let me know if you have any questions.

Cheers,

cpierard:
Hi @reint-fischer ! 

I like the notebook. I like how it's structured in general and the plots are first class! 👍 👍 

These are my suggestions:

1. I think that is not very relevant to have the code to generate de plots because it takes too much space and it breaks the flow of the reading. Do you think is possible to hide the unnecessary code so is easier to read? I hided it using a nb-extension but not sure if this save to be displayed in an online jupyter-viewer.
2. I didn't notice that the subsections were subsections until I double clicked on the markdown cell. Maybe you could add numbers to the section, just like this: 1. A grid, 1.1 Parcels bilinear interpolation, 1.2 ...  | 2. C grids, 2.1 ...
3. In figure 1a, there are two white nodes in the middle of the ocean, and in figure 1b the island has two more cells. Is this a bug or is it part of the grid? 
4. I expanded a bit some of the explanations in my commit. Hopefully you can see the changes.
5. Highlight each time you mention a figure. e.g. In **figure X** blablabla... It makes it easier to follow. 

That's all so far :)

erikvansebille (approved):

reint-fischer:
Hi @cpierard and @quintenbohte ,

Thanks again for the feedback. As you can see we have made some small fixes and merged the PR. 
To comment on your larger points Claudio:
1. I've looked up the options for nbextensions in nbviewer, but it seems like these tags do not work there unfortunately. Also, one of the goals of the notebook is to show users how to visualise these grids, so maybe it's not so bad if people see the code.
3. that is indeed correct. I have tried to explain the logic under the figure.

Feel free to comment further on the current version btw.

PR #965, opened 2020-11-27
Added functionality such that particles can also have arrays as variables

MariusWiggert:
I needed this functionality for a project where the particles have some pre-specified active moving around to do.
Thought I'll share it with the community.

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -248,10 +248,8 @@ class ParticleSet(object):
             if v.name in ['xi', 'yi', 'zi', 'ti']:
                 ngrid = fieldset.gridset.size if fieldset is not None else 1
                 self.particle_data[v.name] = np.empty((len(lon), ngrid), dtype=v.dtype)
-            elif v.name == 'control_traj':
-                self.particle_data[v.name] = np.empty(kwargs['control_traj'].shape, dtype=v.dtype)
-            elif v.name == 'control_time':
-                self.particle_data[v.name] = np.empty(kwargs['control_time'].shape, dtype=v.dtype)
+            elif v.dtype == object:
+                self.particle_data[v.name] = np.empty(kwargs[v.name].shape, dtype=v.dtype)
             else:
                 self.particle_data[v.name] = np.empty(len(lon), dtype=v.dtype)
 
@@ -701,7 +699,7 @@ class ParticleSet(object):
         assert outputdt is None or outputdt >= 0, 'outputdt must be positive'
         assert moviedt is None or moviedt >= 0, 'moviedt must be positive'
 
-        mintime, maxtime = self.fieldset.gridset.dimrange('time_full')
+        mintime, maxtime = self.fieldset.gridset.dimrange('time_full') if self.fieldset is not None else (0, 1)
         if np.any(np.isnan(self.particle_data['time'])):
             self.particle_data['time'][np.isnan(self.particle_data['time'])] = mintime if dt >= 0 else maxtime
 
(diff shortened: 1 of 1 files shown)


erikvansebille:
Thank you @MariusWiggert, for providing this addition! I hadn't realised it was possible to use arrays as particle variables. And how do you use this in a Kernel then? Can you provide a very, very simple testcase that shows how this works?

Note that @RoelBrouwer and @qubixes are currently refactoring the ParticleSet code, so we'll probably wait with merging your PR until their new code is in master. But still, nice functionality and thanks for contributing!

MariusWiggert:
Hey Erik, 
thank you guys for the great basis to build upon =)

For my application, it is useful as I'm designing a controller for an ocean floating platform for now just one, but soon many platforms.
So for open-loop control (pre-specified when & how much to actuate/turn on the motors), I need a way for each particle to just store the control & time trajectory I'm setting when I initiate the simulation and during execution, the kernel accesses those arrays. 

My particle looks like this:

`class OpenLoopParticle(p.ScipyParticle):
	# target and control trajectory
	control_traj = p.Variable('control_traj', initial=0., to_write=False, dtype=object)     # 2xN long
	control_time = p.Variable('control_time', initial=0., to_write=False, dtype=object)     # 1xN+1 long

	v_max = p.Variable('v_max', dtype=np.float32, initial=1., to_write=False)`

and the kernel then executes the control action in addition to the advection. To better understand the control signal is a 2D array with the first row being the heading as in radiants and the second row the magnitude of control from 0-1.

`def open_loop_control(particle, fieldset, time):
    # find index of the relevant control signal
    index = 1
    while particle.control_time[-index] - time > 0:
        index += 1

    particle.lon += math.cos(particle.control_traj[1, -index + 1]) * particle.control_traj[0, -index + 1] * particle.v_max * particle.dt
    particle.lat += math.sin(particle.control_traj[1, -index + 1]) * particle.control_traj[0, -index + 1] * particle.v_max * particle.dt
`

erikvansebille:
Thanks for this code, @MariusWiggert. But I still don't entirely understand how you then set e.g. the values of `pset.control_time`. You initialise with `0.`, but then later(?) set the values? How do you do that?

MariusWiggert:
Hey Erik, good point.
What I do is that I solve an optimal control problem to get both the control vector u and a vector of control times time_vec. Then I feed them in when I initialize the particle set:

`pset = p.ParticleSet.from_list(fieldset=fieldset,   
                             pclass=particles.OpenLoopParticle,  
                             lon=[x_0[0]],   
                             lat=[x_0[1]],  
                            control_traj=[u],
                            control_time=[time_vec],
                            v_max=[u_max/conv_m_to_deg])`

CKehl:
I have a small question: are those changes tested in JIT-mode or only tested in SciPy-mode ?

MariusWiggert:
Good point @CKehl, I only tested them in SciPy mode yet.

erikvansebille:
I tried to create a unit test for this a few days ago, and also found that the implementation only seems to work if there is only _one_ particle in the `ParticleSet`. Or have you tried it with multiple `Particles` too?

CKehl:
The issue here is that, if you model variables as fields, the whole interface needs to look similar to the `xi,yi,zi` setup, which are fixed-length arrays interfacing in C. The issue here is that your array-variables are not fixed-length (before calling `pset.execute`) and they are length adaptive during the simulation, as far as I see. So, to get that up and running in JIT will require a fair bit more work. It's quite involved, because if we do the step-over in C, those arrays must be somehow fixed-length cause the data need to be C-continuous. Also, you'd need to store each array's length `array.shape[0]` as particle variable, as a starter. 

Your initiative is great and this PR a nice starting point - just saying that I am already thinking regularly how to do that given other developments in parcels (i.e. particle-particle interaction), and I haven't found a 'graceful' (meaning: generic; versatile) way of doing that until now..

erikvansebille:
Closing this stale PR for now

PR #967, opened 2020-12-01
Slight improvements to 2D diffusion

daanreijnders:
I've made some changes to the 2D diffusion kernels and tutorials:
* `Kh_zonal` and `Kh_meridional` are now assumed to be constants when using the `DiffusionUniformKh` kernel. Having them as a field is computationally wasteful, since these uniform fields are interpolated at each timestep. Having these values as constants also lowers the chance that users mistakenly use this kernel with varying diffusivity fields, in which M1 or EM should always be used.
* Removed the combined RK4-EM/RK4-M1 kernels. Their limited use-case (using RK4 since diffusivities can be zero in certain regions) does not justify including them. Including them may cause users to think that those kernels somehow have smaller errors, while the error is dominated by the diffusive step.
* Use normally distributed random Wiener increments. The previous argument for using a uniform distribution (with the correct first and second moment) was that particle increments remain bounded. However, a normal distribution should be used for the convergence of individual particle paths (strong convergence). Note that M1 and EM have the same order of weak convergence (convergence of statistics), but that M1 is of strong convergence of order 1, while EM is of order 0.5. Thus it makes no sense to use M1 while not using normally distributed Wiener increments.
* The tutorials are rewritten so that the use case and connection to Eulerian advection-diffusion is (hopefully) more clear. They now also allude to ongoing developments in 3D advection-diffusion kernels and a new way of computing gradients so that users do not have to worry about `dres` anymore in the future. `dres` is also more clearly defined, while it is also stressed more that its 'correct' value is uncertain. The section with `DiffusionUniformKh` has been removed, since this kernel does not support the varying diffusivity fields that make up the test case anymore.

Left to do:
- [x]  Unit conversion of `Kh_zonal` and `Kh_meridional` when those are constants (currently makes the tests fail)


daanreijnders:
Instead of having `Kh_zonal` and `Kh_meridional` as constants, @erikvansebille implemented a new `FieldSet.add_constant_field()` method, so these 'constants' play nice with the unit converters.

PR #976, opened 2021-01-04
Fixing autochunking

erikvansebille:
The dask autochunking in the latest master was extremely slow since #932. After some investigation, it turned out to be because I asked the `static_name_maps` to be removed; this was not correct. Putting them back now

In the meantime, I'll investigate the dask chunking further

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1165,7 +1165,7 @@ class Field(object):
             else:
                 self.data_chunks[0, :] = None
             self.c_data_chunks[0] = None
-            self.grid.load_chunk[0] = g.chunk_loaded_touched_in_JIT
+            self.grid.load_chunk[0] = g.chunk_loaded_touched
             self.data_chunks[0] = np.array(self.data)
 
     @property
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1146,17 +1146,14 @@ class Field(object):
     def chunk_data(self):
         if not self.chunk_set:
             self.chunk_setup()
-        # self.grid.load_chunk code:
-        # 0: not loaded
-        # 1: was asked to load by kernel in JIT
-        # 2: is loaded and was touched last C call
-        # 3: is loaded
+        g = self.grid
         if isinstance(self.data, da.core.Array):
             for block_id in range(len(self.grid.load_chunk)):
-                if self.grid.load_chunk[block_id] == 1 or self.grid.load_chunk[block_id] > 1 and self.data_chunks[block_id] is None:
+                if g.load_chunk[block_id] == g.chunk_loading_requested \
+                        or g.load_chunk[block_id] in g.chunk_loaded and self.data_chunks[block_id] is None:
                     block = self.get_block(block_id)
                     self.data_chunks[block_id] = np.array(self.data.blocks[(slice(self.grid.tdim),) + block])
-                elif self.grid.load_chunk[block_id] == 0:
+                elif g.load_chunk[block_id] == g.chunk_not_loaded:
                     if isinstance(self.data_chunks, list):
                         self.data_chunks[block_id] = None
                     else:
@@ -1168,7 +1165,7 @@ class Field(object):
             else:
                 self.data_chunks[0, :] = None
             self.c_data_chunks[0] = None
-            self.grid.load_chunk[0] = 2
+            self.grid.load_chunk[0] = g.chunk_loaded_touched_in_JIT
             self.data_chunks[0] = np.array(self.data)
 
     @property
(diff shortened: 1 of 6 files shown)


reint-fischer:
I've tested the impact of these changes in [this simulation on GEMINI](https://github.com/OceanParcels/UtrechtTeam/blob/master/scripts/time_chunking.py). The figure is from [this notebook](https://github.com/OceanParcels/UtrechtTeam/blob/master/scripts/Dask_vs_indices.ipynb). The first two commits are in red, they still include the rechunk_callback_fields. The branch as it is up to commit 0403d91 is shown in yellow. Both have a significant speed up in the chunksizes 128, 256 and auto and removing the rechunk_callback_fields does not seem to have a real impact on the speed.
dask_updates_PR

erikvansebille:
Thanks for this analysis, @reint-fischer! Good to see that with this fix, the odd slowdown for `chunksize='auto'` goes away, and performance is as expected

@ckehl, what do you think? Should we merge this entire branch? Or only the first two commits (and revert 8f982f7b08b085be0c893385e57a43e248c70ef3 and 0403d9189e54ec36b3577d44aa86533e0e5dc165 that removes the `fieldfilebuffer.rechunk_callback_fields`)

parcels/fieldfilebuffer.py
```diff
@@ -568,9 +694,7 @@ def data_access(self):
                         self.chunk_mapping[chunkIndex] = chunkDim
                         chunkIndex += 1
                     self._chunkmap_to_chunksize()
-                    if self.rechunk_callback_fields is not None:
-                        self.rechunk_callback_fields()
-                        self.chunking_finalized = True
+                    self.chunking_finalized = True
```
CKehl, line 697: Here, I have my first objection. This will fail if the rechunking (based on the memory budget in the dask config file) mandates more, smaller chunks than the failsafe was able to establish - which is very likely given that the default chunk budget is 128MB and the failsafe in a 32GB system under theoretical conditions is roughly 10 GB. It doesn't fail the tests because, in order to replicate and test for a specific outcome of the chunking, we set this chunk budget manually in the tests (we have to). 

Overall this line says: If `chunksize='auto'" was chosen, then the final chunksize is determined with the very first file access. In order to resize the `load_chunks` array of the related field to the new chunksize, the field needs to update its chunk information. This is done via the callback that has been removed here.

Please restore the previous version for those 3 lines.

parcels/fieldfilebuffer.py
```diff
@@ -581,9 +705,7 @@ def data_access(self):
                 data = np.array(data)
             else:
                 data = da_data
-            if not self.chunking_finalized and self.rechunk_callback_fields is not None:
-                self.rechunk_callback_fields()
-                self.chunking_finalized = True
+            self.chunking_finalized = True
```
CKehl, line 708: commen see line 571 (697)

parcels/grid.py
```diff
@@ -233,6 +233,26 @@ def computeTimeChunk(self, f, time, signdt):
...
+    @property
+    def chunk_loading_requested(self):
+        return 1
+
+    @property
+    def chunk_loaded_touched_in_JIT(self):
+        return 2
+
+    @property
+    def chunk_loaded_not_touched(self):
```
CKehl, line 249: isn't that the 'chunk_deprecated' state ?

parcels/fieldset.py
```diff
@@ -1119,11 +1121,14 @@ def computeTimeChunk(self, time, dt):
                                 f.data[2, :] = None
                         f.data[1:, :] = f.data[:2, :]
                         f.data[0, :] = data
-                g.load_chunk = np.where(g.load_chunk == 3, 0, g.load_chunk)
+                g.load_chunk = np.where(g.load_chunk == g.chunk_loaded_touched_in_JIT,
+                                        g.chunk_loading_requested, g.load_chunk)
```
CKehl, line 1125: well, this line is redundant - the data are still used by JIT, they are already loaded, so they do not need to be loaded again. This would just request unnecessary I/O for data already in memory.

Please comment / remove the line and test if it still works (which it should).

CKehl (changes requested):
Apart from one naming comment on the chunk load state, and 3 functional comments, the rest is fine by me. I also hope the PR helped a bit understanding the Dask usage.

The 2 (or: 3) functional comments really need to be changed. As Reint was showing, the callback has minimal runtime impact, but will make the auto-chunking fail under real conditions.

parcels/fieldfilebuffer.py
erikvansebille: Well, the problem is that this `self.rechunk_callback_fields()` is called for every new file that is opened, and because the filed.chuck_setup also sets the chunk_info to zero everywhere, the info on which chunks are occupied is lost. This can be a severe slowdown, because it means that Parcles will need to figure out which chunks are occupied and which aren't every time a new file is opened

Perhaps we should fix the if-condition then so that it is only called once per Field, not on every FileBuffer instance?

parcels/fieldset.py
erikvansebille: Not true. This line was necessary to fix a bug in #978; took me a long time to find, but when a new FileBuffer instance is created, in some cases the chunking otherwise goes wrong. So I'm pretty sure now that we need this line

parcels/field.py
```diff
@@ -1357,12 +1354,13 @@ def computeTimeChunk(self, data, tindex):
                 ti = g.ti + tindex
             timestamp = self.timestamps[np.where(ti < summedlen)[0][0]]
 
+        rechunk_callback_fields = self.chunk_setup if isinstance(tindex, list) else None
```
CKehl, line 1357: elegant work-around for the issue with the multi-initialisation call.

CKehl (approved):

reint-fischer:
I ran the performance notebook again with the branch as of the latest commit, and here it seems not much has changed in terms of performance.
dask_updates_19de6ec

Issue #951, opened 2020-11-12
Explore keeping only two time snapshots in memory in DeferredLoad mode


erikvansebille:
In DeferredLoad mode, Parcels now keeps three snapshots in memory, and cycles through these. We use three snapshots because an RK4 step could jump over a snapshot time, instead of step on it. 

However, the [recent paper by @nordam](https://doi.org/10.5194/gmd-2020-154) showed that for accurate numerical integration, it is important to make sure that integration is stopped/started at the time of each snapshot. And in fact, that is already enforced when Parcels is run in DeferredLoad mode, because new snapshots are loaded at each snapshot time.

Does this mean we could keep only two snapshots in memory? If so, this would reduce memory significantly in many cases

erikvansebille:
This has been fixed in #978

PR #978, opened 2021-01-05
Deferred load with two snapshots

erikvansebille:
Presently, Parcels stores three snapshots in memory in deferred_load mode. However, @nordam's [paper](https://gmd.copernicus.org/articles/13/5935/2020/) on time interpolation for particle codes made me realise that perhaps we don't need three snapshots in memory, but can do with only two

While this does not increase speed (all snapshots still need to be loaded), it does reduce the memory load of the `FieldSet` by almost 30%, which in some cases (large 3D global datasets with many `Fields` from e.g. biogeochemistry), could make a big difference on the total memory load

This PR implements this change to only two snapshots in memory in `deferred_load=True` mode


erikvansebille:
Note that this fixes #951

parcels/fieldfilebuffer.py
```diff
@@ -574,8 +574,6 @@ def data_access(self):
                 else:
                     if not self.autochunkingfailed:
                         data = data.rechunk(self.chunk_mapping)
-                    if self.rechunk_callback_fields is not None:
-                        self.rechunk_callback_fields()
```
CKehl: actually, if we now do have the failsafe active, then this callback needs to stay (in my opinion). Reason:

The failsafe comes into play if the user requests `chunksize='auto'`, does not provide a dimension mapping netcdf <-> parcels, and the dimension mapping is not in the static maps (i.e. the dimension name mappings that we know). In that case, we either fully abort, or we do the failsafe (i.e. parse the netcdf file and match the dimensions heuristically). In that case, the `self.chunksize` will be overwritten, from `'auto'` to a safe, specific, strictly 2D-spatial chunksize. Thus, the check in l. 558 will return `False` even though `'auto'` was requested by the user. Now, the field attached to the file doesn't know at this point that we overruled the auto-decision - its chunk information are still uninitialised. Hence, they need to be updated by the callback in either case. And that was the purpose of this line.

I hope you can follow the trail of thoughts - decide for yourself. In my opinion, that line needs to stay, but there's a counter-argument to it, I'm happy to hear it.

erikvansebille: This is an outdated change, I reverted it already in a later commit (593cf10). The discussion about how to call this call_back_fields is better contained in #976

tests/test_fieldset.py
```diff
@@ -521,10 +521,13 @@ def SampleUV2(particle, fieldset, time):
 
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
 @pytest.mark.parametrize('time_periodic', [4*86400.0, False])
+@pytest.mark.parametrize('dt', [-3600, 3600])
 @pytest.mark.parametrize('chunksize', [False, 'auto', {'time': ('time_counter', 1), 'lat': ('y', 32), 'lon': ('x', 32)}])
 @pytest.mark.parametrize('with_GC', [False, True])
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="skipping windows test as windows memory leaks (#787)")
-def test_from_netcdf_memory_containment(mode, time_periodic, chunksize, with_GC):
+def test_from_netcdf_memory_containment(mode, time_periodic, dt, chunksize, with_GC):
```
CKehl, line 528: k, I see - testing forward- and backward simulation in 1 go, nice.

tests/test_fieldset.py
```diff
@@ -905,3 +908,30 @@ def test_fieldset_from_data_gridtypes(xdim=20, ydim=10, zdim=4):
     pset.execute(AdvectionRK4, runtime=1, dt=.5)
     assert np.allclose(plon, pset.lon)
     assert np.allclose(plat, pset.lat)
+
+
+@pytest.mark.parametrize('mode', ['scipy', 'jit'])
+@pytest.mark.parametrize('direction', [1, -1])
+@pytest.mark.parametrize('time_extrapolation', [True, False])
+def test_deferredload_simplefield(mode, direction, time_extrapolation, tmpdir, tdim=10):
```
CKehl, line 916: I assume that's also the test to see that the non-dask `DeferredNetcdfFileBuffer` also works as-expected. Nice.

CKehl (approved):

Issue #970, opened 2020-12-11
Internally compute Field derivatives


daanreijnders:
Parcels currently lacks automatic computation of the derivative of a `Field`. A use case is computing the gradient in a diffusivity field. In the `AdvectionDiffusion` kernels, this is now implemented by sampling the diffusivities around a particle (`±dres` in longitude/latitude) and computing a central difference from these sampled quantities. This implementations is suboptimal because:
* It is unclear what the correct `dres` value is, especially in grids with varying cell sizes (in the vertical direction, cell height often increases with depth).
* Interpolations are a costly step. Computing derivatives under-the-hood can eliminate a couple of interpolations.

Another way to currently use derivatives with Parcels is preprocess the data and construct derivative fields using packages like [xgcm](https://github.com/xgcm/xgcm), but such preprocessing can be costly and requires the derivatives to be stored on disk. The extra I/O this requires when using Parcels is another bottleneck.

## Proposed implementation and syntax
Derivatives can be computed under-the-hood as central differences on the gridded data. The locations at which these derivatives are defined are then the same as in an OGCM: for C-grids, the derivatives of tracers will be defined at the respective cell interface. The derivatives of velocities are located at the cell centers or at the f-points. Since the user currently needs to supply the f-points already, cell centers and the locations of the interface centers can be inferred. In A-grids, the derivative is simply located at the center between two points, which is easily computed. 

Users should be able to choose how the derivative at the particle location is defined: nearest neighbor or linear interpolation. In the case of linear interpolation, the following figure shows a simple 2D example of which tracer cells should be accessed to define the zonal derivative at the particle's location (red) using a flat mesh:
<img src="https://user-images.githubusercontent.com/9340879/101906618-86b61300-3bb9-11eb-95c7-a08622d10853.png" width="400" />
To interpolate the derivative, the derivative values should in this case be computed at the locations of the 4 closest u-points (yellow), meaning that a total of 6 tracer cells are accessed. In 3D grids, this extends to 12 cells. 

In curvilinear fields, a zonal or meridional derivative can be determined through computing the gridded horizontal gradient and projecting it in the meridional or zonal direction (to compute a gradient in 2D, 9 cells need to be accessed, since a derivative is computed in both horizontal directions).

I propose an implementation where a derivative can be readily accessed by kernels as an attribute of a `Field`, i.e.:
```Python
fieldset.field.derivative[dim, time, particle.depth, particle.lat, particle.lon]
```
[...]

PR #979, opened 2021-01-05
Implementing in-kernel computation field derivatives

erikvansebille:
This PR implements #970, to internally compute the derivatives of Fields. 

Proposed syntax is 
```python
field.eval(time, particle.depth, particle.lat, particle.lon, derivative='dir')
```
where `'dir'` can be any of `'lon'`, `'lat'`, `'depth'`, `'x'` (which is different from `'lon'` because it goes in the index-direction, which is not necessarily zonal for curvilinear grids), `'y'`, `'z'`, and perhaps even `'t'` and `'time'`

To-do list
- [x] implementation of the 'plumbing' (i.e. passing of derivative information to the field interpolation calls) in both scipy and JIT
- [ ] correct implementation in scipy of derivative for grids with non-unit mesh sizes, including also `mesh='spherical'`
- [ ] correct implementation in scipy of derivative for `x` and `y`, also for MITgcm and NEMO indexing convention
- [ ] correct implementation in scipy of derivative for `lon` and `lat`, also for curvilinear grids
- [ ] implementation in scipy of derivative for `depth` (and `z`) direction
- [ ] implementation in scipy of derivative for `time` (and `t`) direction
- [ ] implementation of all the above in JIT too

All these tasks on the list will require more unit tests, on more complicated `Fields`. @daanreijnders, can you help with these tasks, especially the implementations in scipy? I will then convert them to JIT where necessary

### DIFF
--- a/parcels/compilation/codegenerator.py
+++ b/parcels/compilation/codegenerator.py
@@ -68,11 +68,12 @@ class FieldEvalCallNode(IntrinsicNode):
 
 
 class FieldEvalNode(IntrinsicNode):
-    def __init__(self, field, args, var, convert=True):
+    def __init__(self, field, args, var, convert=True, derivative=None):
         self.field = field
         self.args = args
         self.var = var  # the variable in which the interpolated field is written
         self.convert = convert  # whether to convert the result (like field.applyConversion)
+        self.derivative = derivative  # for calculating Field derivatives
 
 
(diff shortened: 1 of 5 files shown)


erikvansebille:
Closing this PR for now, as it is not needed anymore for @daanreijnders's project. If others would like to use this feature, please let us know

PR #988, opened 2021-01-25
fixing ever-decreasing p.dt

reint-fischer:
P.dt is partly defined in the particleloop and possibly in kernels like 'AdvectionRK45` in the form of adaptive timestepping. This creates loopholes for p.dt to be ever-decreasing, resulting in an extreme slowdown of the simulation (#746). 

In the particleloop, `p.dt` is set by `dt_pos = min(abs(p.dt), abs(endtime-p.time))`. If `endtime` is not a multiple of `dt`, this can cause p.dt to decrease here. In the next timestep however, this `p.dt` remains this smaller value, because we want to allow for kernels to change `p.dt`. This PR proposes to reset `p.dt` if it was decreased to meet the `endtime` and the kernel has not changed it:

```
if abs(endtime - p.time) < abs(p.dt):
    dt_pos = abs(endtime - p.time)
    reset_dt = p.dt
else:
    dt_pos = abs(p.dt)
    reset_dt = 0

....

if reset_dt and p.dt == pdt_prekernels:
    p.dt = reset_dt
```


parcels/compilation/codegenerator.py
```diff
@@ -1014,8 +1014,12 @@ def generate(self, funcname, field_args, const_args, kernel_ast, c_include):
         update_state = c.Assign("particles->state[pnum]", "res")
         update_pdt = c.If("_next_dt_set == 1",
                           c.Block([c.Assign("_next_dt_set", "0"), c.Assign("particles->dt[pnum]", "_next_dt")]))
+        dt_pos = c.If("fabs(endtime - particles->time[pnum])<fabs(particles->dt[pnum])",
+                            c.Block([c.Assign("__dt", "fabs(endtime - particles->time[pnum])"), c.Assign("reset_dt", "1")]),
+                            c.Block([c.Assign("__dt", "fabs(particles->dt[pnum])"), c.Assign("reset_dt", "0")]))
```
CKehl, line 1019: here also: try the speed-trick from the python version. `abs(...)` becomes `fans(...)` and `math.ceil(...)` becomes `ceil(...)`. `alpha` and `value` may need to be declared here, so this overall becomes a `c.Block([...])`, but that's overall not an issue, I presume.

parcels/kernel.py
```diff
@@ -340,10 +345,17 @@ def execute_python(self, pset, endtime, dt):
...
                     p.update_next_dt()
                     if analytical:
                         p.dt = np.inf
-                    dt_pos = min(abs(p.dt), abs(endtime - p.time))
+                    if abs(endtime - p.time) < abs(p.dt):
+                        dt_pos = abs(endtime - p.time)
+                        reset_dt = True
+                    else:
+                        dt_pos = abs(p.dt)
+                        reset_dt = False
```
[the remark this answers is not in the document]
CKehl: little printing rationale from my console:

```
In [2]: import math                                                                                                                                                                                            
In [3]: endtime = 5                                                                                                                                                                                            
In [4]: ptime = 5.15                                                                                                                                                                                           
In [5]: pdt = 0.25                                                                                                                                                                                             
In [6]: value = ((endtime-ptime)-pdt)                                                                                                                                                                          
In [7]: print(value)                                                                                                                                                                                           
-0.40000000000000036
In [8]: sgnv = math.copysign(1.0, value)                                                                                                                                                                       
In [9]: print(sgnv)                                                                                                                                                                                            
-1.0
In [10]: print(bool(sgnv))                                                                                                                                                                                     
True
In [11]: print(math.ceil(value))                                                                                                                                                                               
0
In [12]: endtime = 5.25                                                                                                                                                                                        
In [13]: value = ((endtime-ptime)-pdt)                                                                                                                                                                         
In [14]: print(math.ceil(value))                                                                                                                                                                               
0
In [15]: endtime = 5.5                                                                                                                                                                                         
In [16]: value = ((endtime-ptime)-pdt)                                                                                                                                                                         
In [17]: print(math.ceil(value))                                                                                                                                                                               
1
In [18]: sgnv = math.copysign(1.0, value)                                                                                                                                                                      
In [19]: print(bool(sgnv))                                                                                                                                                                                     
True
In [20]: ptime = 5.25                                                                                                                                                                                          
In [21]: value = ((endtime-ptime)-pdt)                                                                                                                                                                         
In [22]: print(math.ceil(value))                                                                                                                                                                               
0
In [23]: print(value)                                                                                                                                                                                          
0.0
```

CKehl (changes requested):

CKehl (approved):

parcels/kernel.py
```diff
@@ -431,6 +449,7 @@ def execute(self, pset, endtime, dt, recovery=None, output_file=None, execute_on
 
         # while np.any(error_particles):
         while n_error > 0:
+            print(n_error)
```
erikvansebille, line 452: print statement can go?

parcels/kernel.py
```diff
@@ -340,10 +346,17 @@ def execute_python(self, pset, endtime, dt):
...
                     p.update_next_dt()
                     if analytical:
                         p.dt = np.inf
-                    dt_pos = min(abs(p.dt), abs(endtime - p.time))
+                    if abs(endtime - p.time) < abs(p.dt):
+                        dt_pos = abs(endtime - p.time)
+                        reset_dt = True
+                    else:
+                        dt_pos = abs(p.dt)
+                        reset_dt = False
```
erikvansebille, line 359: Did you try the vectorisation here  that you suggested, @ckehl? Did it not work?

erikvansebille (changes requested):
Code looks good for me. However, do we also need to add a **unit test** that breaks on the current master but works on this new branch, to make sure we've actually fixed the issue (and later don't accidentally introduce it again)?

reint-fischer:
> Code looks good for me. However, do we also need to add a **unit test** that breaks on the current master but works on this new branch, to make sure we've actually fixed the issue (and later don't accidentally introduce it again)?

I am writing that test as we speak :)

parcels/kernel.py
CKehl: I tried the vectorisation suggestion myself - running into numerical accuracy errors and numerical overflows in the `math.ceil(....)` function. So, at the points and tests where it matters, it fails. Probably it could be made working, but the gain doesn't justify then the time investment in this case, and Reint has a working solution - so we better just apply that.

erikvansebille: OK, let's then keep this code as is

@reint-fischer, can you confirm here in the comments that the unit test breaks on the current master, but passes here (i.e. the PR fixes the bug)?
 
Except for fixing another flake8 issue, I think this PR is then ready for merging?

reint-fischer: Yes, the test fails on master, and since it is included here and the checks are ticked, it is fixed by the PR.

erikvansebille (approved):

parcels/kernel.py
CKehl: great - seems ready to be merged! (waiting for that eagerly) :D
Did Cleo also green-light the fix ?

parcels/kernel.py
erikvansebille: I see there's still this print statement here. Can this be removed before we merge?

erikvansebille (changes requested):

erikvansebille (approved):

parcels/kernel.py
cjongedijk: I still observe some fluctuations in my particle.dt in to me unexpected parts of the simulations, which I cannot necessarily explain. My output dt is already a multiple of my computational dt, and furthermore I have to output particle.dt in very high precision to observe it so smells like a round off somewhere.  However, maybe that is for another discussion and since it doesn't any more grind the simulations to an extreme slow down for me the light is green!

PR #989, opened 2021-01-26
Remove explicit loops over particle state

angus-g:
The loops to test and set particle state were implemented using Python primitives, which are extremely slow for large particle counts. We can easily modify `remove_deleted_soa`, because we know the underlying collection supports array indexing. We should also be able to rely on `Collection.__getitem__` returning something with an array-like interface that we can use to set all particle states simultaneously: for SoA the array assignment is significantly faster, and it would be no slower for other collections.

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -388,7 +388,7 @@ class Kernel(object):
         This deletion function is targetted to index-addressable, random-access array-collections.
         """
         # Indices marked for deletion.
-        bool_indices = np.array([p.state == OperationCode.Delete for p in pset])
+        bool_indices = pset.collection.state == OperationCode.Delete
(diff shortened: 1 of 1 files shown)


CKehl (commented):
Well, I see what you did here - nice reading the paper :-)
Well, on the serious point: not a big fan of that adaptation at this point, because the change you did really works exclusively on this SoA array structure. We will soon integrate alternative collections in the main branch, and because the attributes are not in sequential arrays then, you suggestion won't be generalizable.

Ideally, it would be nice to apply those specific changes in a SoA-specific kernel, but that's probably asking too much.

I leave the merge decision to Erik

erikvansebille (approved):

PR #990, opened 2021-01-26
Fixing error message when FieldSet calendar is a cftime.datetime object

erikvansebille:
This issue came up because GlobCurrentv3.0 has a 'julian' calendar, which means that all datetimes for the `ParticleSet` also need to be julian calendars. This is now better handled (requires `cftime>=1.3.1`)

### DIFF
--- a/parcels/tools/converters.py
+++ b/parcels/tools/converters.py
@@ -53,9 +53,19 @@ class TimeConverter(object):
             return (time - self.time_origin) / np.timedelta64(1, 's')
         elif self.calendar in _get_cftime_calendars():
             if isinstance(time, (list, np.ndarray)):
-                return np.array([(t - self.time_origin).total_seconds() for t in time])
+                try:
+                    return np.array([(t - self.time_origin).total_seconds() for t in time])
+                except np.core._exceptions.UFuncTypeError:
+                    raise ValueError(f"Cannot subtract 'time' (a {type(time)} object) from "
+                                     f"a {self.calendar} calendar.\n"
+                                     f"Provide 'time' as a {type(self.time_origin)} object?")
             else:
-                return (time - self.time_origin).total_seconds()
+                try:
+                    return (time - self.time_origin).total_seconds()
+                except ValueError:
+                    raise ValueError(f"Cannot subtract 'time' (a {type(time)} object) from "
+                                     f"a {self.calendar} calendar.\n"
+                                     f"Provide 'time' as a {type(self.time_origin)} object?")
         elif self.calendar is None:
             return time - self.time_origin
         else:
--- a/environment_py3_osx.yml
+++ b/environment_py3_osx.yml
@@ -25,7 +25,7 @@ dependencies:
   - six>=1.10.0
   - xarray>=0.10.8
   - dask>=2.0
-  - cftime
+  - cftime>=1.3.1
   - pytest
   - nbval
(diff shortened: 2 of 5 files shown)


parcels/tools/converters.py
```diff
@@ -53,9 +53,19 @@ def reltime(self, time):
             return (time - self.time_origin) / np.timedelta64(1, 's')
         elif self.calendar in _get_cftime_calendars():
             if isinstance(time, (list, np.ndarray)):
-                return np.array([(t - self.time_origin).total_seconds() for t in time])
+                try:
+                    return np.array([(t - self.time_origin).total_seconds() for t in time])
+                except np.core._exceptions.UFuncTypeError:
+                    raise ValueError(f"Cannot subtract 'time' (a {type(time)} object) from "
+                                     f"a {self.calendar} calendar.\n"
+                                     f"Provide 'time' as a {type(self.time_origin)} object?")
```
CKehl, line 61: this string-concatenate generates a syntax-error - isn't there a '\' missing at the end of the line ?

parcels/tools/converters.py
```diff
@@ -53,9 +53,19 @@ def reltime(self, time):
...
             else:
-                return (time - self.time_origin).total_seconds()
+                try:
+                    return (time - self.time_origin).total_seconds()
+                except ValueError:
+                    raise ValueError(f"Cannot subtract 'time' (a {type(time)} object) from "
+                                     f"a {self.calendar} calendar.\n"
+                                     f"Provide 'time' as a {type(self.time_origin)} object?")
```
CKehl, line 68: this string-concatenate will generate another syntax-error - isn't there a '\' missing at the end of the line ?

CKehl (changes requested):

parcels/tools/converters.py
erikvansebille: Not sure why this breaks PEP8. The code works in my local system, and `flake8` also doesn't throw an error then. I could try going back to old-skool `%` formatting?

parcels/tools/converters.py
erikvansebille: All lines have a beginning and ending `"`, or am I missing something?

CKehl: well, so I just tried 30 min to get an up-to-date version of flake8, but I am stuck with the 1.6.0 version. This seems to be an error interpretation coming with a recent flake8 version. Because parcels relies on flake8 to finish, the linter error needs to be resolved.

Now, I never use this `f"..."` notation, because I am used to the `"... %s ..." % (string)` notation or the `"... {} ... {} ...".format(...)` notation because my world is C/C++. So, if it was me, I would just rewrite the whole thing.

If you like this new notation, then that's also totally fine, but then I can't help you resolve the linter error here - again, my resolution would be a rewrite not using that notation in the first place, and not do line-breaks on those error message. Sorry for that.

parcels/tools/converters.py
erikvansebille: I've now used %-strings again

CKehl:
so, it's a linter-internal mis-interpretation of the standard and a mumbler between Python2 and Python3. See [official github post](https://github.com/PyCQA/pycodestyle/issues/584). possibly your f-string syntax would work now too again.

Either way the PR is approved now.

CKehl (approved):

PR #995, opened 2021-02-09
update docstrings

reint-fischer:
Because of the new release, I have updated the documentation page on the oceanparcels website. Here I have updated some docstrings, so that the documentation page looks correct. The new particlesets folder still contains some functions with lenghty docstrings

### DIFF
--- a/parcels/kernels/TEOSseawaterdensity.py
+++ b/parcels/kernels/TEOSseawaterdensity.py
@@ -7,19 +7,19 @@ __all__ = ['PolyTEOS10_bsq']
 
 def PolyTEOS10_bsq(particle, fieldset, time):
     '''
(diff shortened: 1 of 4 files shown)


parcels/kernels/TEOSseawaterdensity.py
```diff
@@ -6,6 +6,7 @@
 
 
 def PolyTEOS10_bsq(particle, fieldset, time):
+    '''
     # calculates density based on the polyTEOS10-bsq algorithm from Appendix A.2 of
     # https://www.sciencedirect.com/science/article/pii/S1463500315000566
```
erikvansebille, line 11: Good to change to docstring, but the `#`s are then not necessary anymore?

reint-fischer: Yes, good point!

erikvansebille (approved):

Issue #866, opened 2020-06-24
Proper OO-Programming inheritance


erikvansebille:
While some Parcels classes have ok Object-Oriented behaviour (e.g. the `Particle` class and its associated `JITParticles` and `ScipyParticles`, many others don't. 

It would be good to improve the OO inheritance for at least

- [x] the `ParticleSet` class (e.g. 'build', 'add particle', 'remove particle' as methods)
- [x] the `Kernel` class (with e.g. 'build', 'compile', 'load', 'unload', 'execute' as methods)
- [x] the `ParticleFile` class
- [x] ...

Some of this is already implemented in #862, for example in the `Kernels`, where we have a 'BaseKernel' (in kernelbase.py), from which each kind of 'Kernel' (in 'kernel_node.py' and 'kernel_vectorized.py') is derived from.

This will also help fix #618. Thanks to @CKehl for bringing this up

CKehl:
for `ParticleSet` class, see #913 and follow-ups - issue resolved, progressively.

PR #997, opened 2021-02-09
Implement Kernel hierarchy - generalize kernels

CKehl:
Based on the performance branches (#862 , #884 , #957 , #958 and #959 ) and in preparation  for merging #958 , we estanlish a kernel hierarchy, with a base kernel (basekernel.py) from with `ParticleSet`-specific derivatives and None-fieldset derivatives can derive from. The latter one is just meant for larger extension purposes, as a testbed-draft.

This issue is linked with #862 (initial basekernel.py), #958 (target ParticleSet to be incorporated after changes), #866 (proper object-oriented programming) and #618 (earlier starting point).


CKehl:
ready for review

parcels/basekernel.py
```diff
@@ -0,0 +1,327 @@
...
+                all_files_array.append(dyn_src)
+            compiler.compile(self.dyn_srcs, self.lib_file, self.log_file)
+        else:
+            with open(self.src_file, 'w') as f:
+                f.write(self.ccode)
+            all_files_array.append(self.src_file)
+            compiler.compile(self.src_file, self.lib_file, self.log_file)
+        logger.info("Compiled %s ==> %s" % (self.name, self.lib_file))
+        all_files_array.append(self.log_file)
+        # self._cleanup_files = finalize(self, cleanup_remove_files, self.delete_cfiles, self.src_file, self.lib_file, self.log_file)
```
erikvansebille, line 256: stray comment. Line can be removed?

parcels/basekernel.py
```diff
@@ -0,0 +1,327 @@
...
+            # basename = path.join(get_cache_dir(), self._cache_key) if mpi_rank == 0 else None  # very early version
+            # basename = path.join(get_cache_dir(), cache_name) if mpi_rank == 0 else None  # early version
+            basename = cache_name if mpi_rank == 0 else None
+            basename = mpi_comm.bcast(basename, root=0)
+            basename = basename + "_%d" % mpi_rank
+        else:
+            # cache_name = "lib"+self._cache_key    # only required here because loading is done by Kernel class instead of Compiler class
+            cache_name = self._cache_key    # only required here because loading is done by Kernel class instead of Compiler class
+            dyn_dir = get_cache_dir()
+            # basename = path.join(get_cache_dir(), "%s_0" % self._cache_key)
```
erikvansebille, line 226: Stray comment. Line can go?

parcels/basekernel.py
```diff
@@ -0,0 +1,327 @@
...
+            cache_name = self._cache_key  # only required here because loading is done by Kernel class instead of Compiler class
+            dyn_dir = get_cache_dir() if mpi_rank == 0 else None
+            dyn_dir = mpi_comm.bcast(dyn_dir, root=0)
+            # basename = path.join(get_cache_dir(), self._cache_key) if mpi_rank == 0 else None  # very early version
+            # basename = path.join(get_cache_dir(), cache_name) if mpi_rank == 0 else None  # early version
+            basename = cache_name if mpi_rank == 0 else None
+            basename = mpi_comm.bcast(basename, root=0)
+            basename = basename + "_%d" % mpi_rank
+        else:
+            # cache_name = "lib"+self._cache_key    # only required here because loading is done by Kernel class instead of Compiler class
```
erikvansebille, line 223: Is this comment not true anymore?

parcels/basekernel.py
```diff
@@ -0,0 +1,327 @@
...
+        """
+        if MPI:
+            mpi_comm = MPI.COMM_WORLD
+            mpi_rank = mpi_comm.Get_rank()
+            # cache_name = "lib"+self._cache_key    # only required here because loading is done by Kernel class instead of Compiler class
+            cache_name = self._cache_key  # only required here because loading is done by Kernel class instead of Compiler class
+            dyn_dir = get_cache_dir() if mpi_rank == 0 else None
+            dyn_dir = mpi_comm.bcast(dyn_dir, root=0)
+            # basename = path.join(get_cache_dir(), self._cache_key) if mpi_rank == 0 else None  # very early version
+            # basename = path.join(get_cache_dir(), cache_name) if mpi_rank == 0 else None  # early version
```
erikvansebille, line 218: Redundant commented code?

parcels/basekernel.py
```diff
@@ -0,0 +1,327 @@
...
+    def __del__(self):
+        # Clean-up the in-memory dynamic linked libraries.
+        # This is not really necessary, as these programs are not that large, but with the new random
+        # naming scheme which is required on Windows OS'es to deal with updates to a Parcels' kernel.
+        # if self._lib is not None:
+        #     _ctypes.FreeLibrary(self._lib._handle) if platform == 'win32' else _ctypes.dlclose(self._lib._handle)
+        #     del self._lib
+        #     self._lib = None
+        #     if path.isfile(self.lib_file) and self.delete_cfiles:
+        #         [remove(s) for s in [self.src_file, self.lib_file, self.log_file]]
```
erikvansebille, line 97: Why is this code commented out? Can be removed?

erikvansebille (changes requested):
Good cleanup of the code. A few minor comments
- [x] there seem to be a few stray commented lines (from debugging?) left in the code. Remove these for a cleaner code?
- [ ] We previously moved all the particleset*.py files to a new subdirectory. Should we do the same for these kernel*.py files, for consistency?
- [x] Not sure why the CI on Windows is failing?

CKehl:
> Good cleanup of the code. A few minor comments
> 
>     * [ ]  there seem to be a few stray commented lines (from debugging?) left in the code. Remove these for a cleaner code?
> 
>     * [ ]  We previously moved all the particleset*.py files to a new subdirectory. Should we do the same for these kernel*.py files, for consistency?
> 
>     * [ ]  Not sure why the CI on Windows is failing?

(a) first trying to trace and eradicate Win32 compile errors
(b) well, moving the files is difficult. We cannot put e.g. the `kernelsoa.py` in `kernels/`, because this will result in circular imports (i.e. import errors) - that's because because RK4 would import BaseKernel and KernelSOA imports RK4. I'd rather have suggested to create folders such as `soa/`, `aos/` etc., and put particlesets, collections, kernels and partilcefiles (to come) of one structure in the same folder - thus getting rid of the particlesets folder. for now, I leave it as is, see if all runs smoothly.
(c) comment removal - thy will be done.

parcels/basekernel.py
```diff
@@ -0,0 +1,314 @@
...
+                f.write(self.ccode)
+            all_files_array.append(self.src_file)
+            compiler.compile(self.src_file, self.lib_file, self.log_file)
+        logger.info("Compiled %s ==> %s" % (self.name, self.lib_file))
+        all_files_array.append(self.log_file)
+        self._cleanup_files = finalize(self, BaseKernel.cleanup_remove_files, self.lib_file, all_files_array, self.delete_cfiles)
+
+    def load_lib(self):
+        self._lib = npct.load_library(self.lib_file, '.')
+        # self._lib = npct.load_library(self.lib_file, get_cache_dir())
```
erikvansebille, line 248: Still a debug line here?

parcels/basekernel.py
```diff
@@ -0,0 +1,314 @@
...
+                logger.warning_once("compiled library already freed.")
+
+    def remove_deleted(self, pset, output_file, endtime):
+        """
+        Utility to remove all particles that signalled deletion.
+
+        This version is generally applicable to all structures and collections
+        """
+        # Indices marked for deletion.
+        indices = [i for i, p in enumerate(pset) if p.state == OperationCode.Delete]
```
erikvansebille, line 300: So this statement could be very slow and inefficient, right? Is the idea not that we force/encourage developers to provide a more efficient way of locating the particles signalled for deletion in the respective datastructure-implementation?

If so, should we just `pass` here, like the other functions below?

parcels/kernelsoa.py
```diff
@@ -0,0 +1,370 @@
+import inspect
+import math  # noqa
+import random  # noqa
+# import time  # noga
```
erikvansebille, line 4: Stray debug line?

parcels/kernelsoa.py
```diff
@@ -0,0 +1,370 @@
...
+            py_mod.body = [self.py_ast]
+            exec(compile(py_mod, "<ast>", "exec"), user_ctx)
+            self._pyfunc = user_ctx[self.funcname]
+        else:
+            self._pyfunc = pyfunc
+
+        numkernelargs = self.check_kernel_signature_on_version()
+
+        assert numkernelargs == 3, \
+            'Since Parcels v2.0, kernels do only take 3 arguments: particle, fieldset, time !! AND !! Argument order in field interpolation is time, depth, lat, lon.'
```
erikvansebille, line 92: I imagine that this code (but also a lot of the rest in this `__init__` is very general and would apply to _all_ kernels? Could we move it to the BaseKernel? Or is that technically difficult?

parcels/kernelsoa.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            self.execute_python(pset, endtime, dt)
+
+        # Remove all particles that signalled deletion
+        self.remove_deleted(pset, output_file=output_file, endtime=endtime)   # Generalizable version!
+
+        # Identify particles that threw errors
+        n_error = pset.num_error_particles
+
+        # while np.any(error_particles):
```
erikvansebille, line 340: Again a stray debug comment?

erikvansebille (changes requested):
I had some more time to carefully look through the changes, and have two slightly larger comments (sorry for not raising these earlier); as well as still a commented line of old code here and there that could go?

parcels/basekernel.py
CKehl: so, initially, we had 2 incarnations of the `remove_deleted(...)` function: (a) this one, which is a generally-applicable version to remove deleted particles that works for all index-based data structures (e.g. arrays of all sorts), and (b) an SoA-targeted version (called previously `remove_deleted_soa`, see commit history). The SoA-targeted version obviously now remains in `KernelSOA` and _overrides_ this base-class implementation. Essentially: here is a base-implementation that should work for most derivatives - it's not greatly efficient, but it's a working fallback if no more-optimal `remove_deleted(...)` can be developed. Example:

for Array-of-Structure ParticleSets, we do not need to override this function in the derived class, as there is no 'better' option available - slimming down the code and sharing it.

In contrast to the other functions (e.g. `execute`, `exe_python`, `execute_jit`), this function is general enough so it can be shared. Those other functions actually require extensive modifications so a base implementation for them is non-sensical.

parcels/basekernel.py
CKehl: oops - yes, possible. I'll remove that.

parcels/kernelsoa.py
CKehl: yep - that can go now too

parcels/kernelsoa.py
CKehl: nope, we can't because of the instruction flow. Meaning: this kernel-arg number require's the pyfunc to be initialised. That would though require some kernel-specific interpretations.

We can only put things in the `BaseKernel`'s `__ini__` function that is completely (i.e. provably - not just _potentially_) function-independent. That is also why the `pyfunc` is interpreted in the specific- and not the base class.

Specifically: we have those `use_ctx` commands. If we have a structure (e.g. nodes) that require a multi-object, multi-stage compilation, then there are some context-commands that change (e.g. around the `random` context). That makes this `pyfunc` initialisation kernel-/structure-dependent, and hence everything that relies on it needs to follow in the derived class.

CKehl: As you see, in order to get around the problem, the actual check to obtain the `numkernelargs` has been moved to a member function inside the base class - which is the way to go for those cases.

parcels/kernelsoa.py
CKehl: kind-of - and kind-of not, because the `np.any(...)` function is faster than the `n_error` count. The count for the is (or: was) useful in making the `execute(...)` function work when we implemented the new ParticleSet class structure, which is why we left it in.

So: this comment is fast, and it is a feasible alternative to the `n_error` count, but the `n_error` count is easier when tracing errors for the kernel.

I can remove it - but I need to make a note about it for myself when for implementing all the new kernels later.

erikvansebille: OK, no worries. Then leave it in

erikvansebille (approved):
Happy with all the changes if the last two debug lines are removed. Feel free to merge after that

PR #998, opened 2021-02-12
Implement ParticleFile inheritance - generalize ParticleFile

CKehl:
Based on the performance branches (#862 , #884 , #957 , #958 and #959 ) and in preparation  for merging #958 , we establish a a generalized `ParticleFile`, with a base particle file (baseparticlefile.py) from which `ParticleSet`-specific derivatives can inherit. This inheritance and abstract-class implementation is necessary because for array-structures (e.g. AoS, SoA), the `particle.id` and the index of a particle is the same, this the particlefile can write-address and house-keep written entries by either ID or index, at not additional performance cost. For ordered sets, the particle's ID and its index are different - the ID is static, the index is chaning (i.e. dynamic). Thus, for those structures, the `ParticleFile` needs to house-keep and track the `maxindex`, not the `maxid`, which comes at steep performance cost. In order to facilitate these differences, a class hierarchy of `ParticleFile` is necessary to share code and minimize code duplication.

This issue is linked with #958 (target ParticleSet to be incorporated after changes), #866 (proper object-oriented programming) and #618 (earlier starting point).


parcels/baseparticlefile.py
```diff
@@ -0,0 +1,326 @@
...
+                elif v.to_write is True:
+                    self.var_names += [v.name]
+            if len(self.var_names_once) > 0:
+                self.written_once = []
+                self.file_list_once = []
+
+            self.file_list = []
+            self.time_written = []
+            # ==== distinction between ID-based and index-based write done in derived class ==== #
+            # self.maxid_written = -1
```
erikvansebille, line 90: This commented line can go?

erikvansebille (approved):
Except for one small issue with a commented line that could be removed I think, this looks fine

parcels/baseparticlefile.py
CKehl: yes, it can - gonna take care of it today

PR #1022, opened 2021-04-30
Removing some of the logger info messages

erikvansebille:
These logger messages are not very relevant and/or interesting. But users can get confused by them, especially in jupyter notebooks where they show up in red text under the cells. So I'd suggest to remove them, what do you think? And are there any other messages that we could/should remove?

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -490,7 +490,6 @@ class Field(object):
         if not isinstance(data, (np.ndarray, da.core.Array)):
             data = np.array(data)
         if not data.dtype == np.float32:
-            logger.warning_once("Casting field data to np.float32")
             data = data.astype(np.float32)
         lib = np if isinstance(data, np.ndarray) else da
         if transpose:
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -43,14 +43,11 @@ class Grid(object):
         self.lat = lat
         self.time = np.zeros(1, dtype=np.float64) if time is None else time
         if not self.lon.dtype == np.float32:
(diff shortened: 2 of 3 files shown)


reint-fischer (approved):
I agree, I don't think anyone really pays attention to these. Is there any case in which users would want to have these in anything other than `np.float32`? I can't think of any.

I also went through my log-files to see if I could find any other messages that could be removed, but couldn't find any.

Two other options I thought of are:
1. Change these into `logger.info` messages, as in [this line](https://github.com/OceanParcels/parcels/blob/c8957ba256dc7573ac80ea4ad7eb5fb399fbc9b6/parcels/kernel/basekernel.py#L252)
2. Make a switch to turn these kind of messages on and off. That seems like it would be too much work however and only make things more confusing.

CKehl:
> I agree, I don't think anyone really pays attention to these. Is there any case in which users would want to have these in anything other than `np.float32`? I can't think of any.
> 
> I also went through my log-files to see if I could find any other messages that could be removed, but couldn't find any.

One reason: flat meshes and alternative coordinate systems. If you e.g. go with a UTM coordinate system resolved in meters or centimeters, one will need 64-bit lat-lons.

CKehl (approved):

erikvansebille:
Thanks for the feedback, @reint-fischer and @CKehl. I suggest we then simply remove these lines, also because as @reint-fischer mentioned, implementing some kind of switch seems overkill and further obfuscates the code. `logger.info` is not really a solution, as this is also rendered in red text in the jupyter notebooks (see e.g. the compilation info), so still suggests something going wrong..

PR #1026, opened 2021-05-12
Expand new test parameters for particle collection to all tests

CKehl:
After including first test for advection, particle sets and particle files when integrating AoS, we now spread the new test parameters to the other tests files where this change applies to improve test coverage.

files to-do:
- [x] test_diffusion.py
- [x] test_fieldset.py
- [x] test_fieldset_sampling.py
- [x] test_grids.py
- [x] test_kernel_execution.py
- [x] test_kernel_language.py
- [x] test_mpirun.py
- [x] test_particle_collections.py
- [x] test_particles.py
- [x] test_scripts.py


tests/test_mpirun.py
```diff
@@ -10,17 +10,19 @@
...
         stommel_file = path.join(path.dirname(__file__), '..', 'parcels',
                                  'examples', 'example_stommel.py')
         outputMPI = tmpdir.join('StommelMPI.nc')
         outputNoMPI = tmpdir.join('StommelNoMPI.nc')
 
-        system('mpirun -np 2 python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputMPI, repeatdt, maxage))
-        system('python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputNoMPI, repeatdt, maxage))
+        # ehm, this test doesn't work. I remember that we only employ the actual ParticleSet split _when_ # particles >= 1024. Testing it with 4 particles doesn't test MPI at all.
+        system('mpirun -np 2 python %s -p 4 -o %s -r %d -a %d -psm %s' % (stommel_file, outputMPI, repeatdt, maxage, pset_mode))
+        system('python %s -p 4 -o %s -r %d -a %d -psm %s' % (stommel_file, outputNoMPI, repeatdt, maxage, pset_mode))
```
CKehl, line 25: @erikvansebille this part of the test really needs to change cause it doesn't really test MPI ...

CKehl: of course, drawback is that the test will take longer. I just don't know why we would need to test such long runtimes. Perhaps you can explain the rationale tomorrow.

erikvansebille: I checked for your thought that MPI is only used when p>1024, but can't find that in the code at https://github.com/OceanParcels/parcels/blob/master/parcels/collection/collectionsoa.py#L92-L119. So I think it does actually check MPI already

Note that the intention here was to test the _output_ of MPI runs; to make sure that the merging of temporary files from different nodes worked well. It does not test the efficiency etc of MPI; we will need to develop other tests for that

CKehl: so, indeed, we probably got rid of that 1024-min-particleset-size limit when redesigning the collections - which is good.

On your note: yes, sure, we test if it works, not if it's speedy. Still, if the limit was in place, the code would not have done the actual MPI stuff that the test is supposed to 'test', so with the limit the test would be meaningless.

Luckily, the artificial size limit isn't active anymore, so the test works as originally intended. Thanks for looking it up!

CKehl:
ready for general review and merge.

parcels/examples/example_stommel.py
```diff
@@ -122,14 +125,15 @@ class MyParticle(ParticleClass):
     return pset
 
 
+@pytest.mark.parametrize('pset_mode', ['soa', 'aos'])
```
erikvansebille, line 128: Not sure if this is possible in pytest, but for easy further extension with other datastructures, would it be an idea to change these lines to e.g. `@pytest.mark.parametrize('pset_mode', pset_types)` and then define `pset_types = ['soa', 'aos']` somewhere at the top of this (and other) test files? If we then add another ParticleSet type, we only need to change one line?

tests/test_grids.py
```diff
@@ -542,19 +560,20 @@ class MyParticle(ptype[mode]):
...
+    pset = pset_type[pset_mode]['pset'].from_list(fieldset, MyParticle, lon=lonp, lat=latp, depth=.2)
     pset.execute(pset.Kernel(sampleVel), runtime=0, dt=0)
     pset.zonal[0] = fieldset.U.units.to_source(pset.zonal[0], lonp, latp, 0)
     pset.meridional[0] = fieldset.V.units.to_source(pset.meridional[0], lonp, latp, 0)
-    assert abs(pset.zonal[0] - 1) < 1e-3
-    assert abs(pset.meridional[0]) < 1e-3
-    assert abs(pset.vertical[0] - 1) < 1e-3
+    assert (pset[0].zonal-1) < 1e-3
+    assert (pset[0].meridional-1) < 1e-3
+    assert (pset[0].vertical-1) < 1e-3
```
erikvansebille, line 569: Why no `abs()` here anymore?

tests/test_mpirun.py
```diff
@@ -10,17 +10,19 @@
...
+def test_mpi_run(pset_mode, tmpdir, repeatdt, maxage):
     if MPI:
         stommel_file = path.join(path.dirname(__file__), '..', 'parcels',
                                  'examples', 'example_stommel.py')
         outputMPI = tmpdir.join('StommelMPI.nc')
         outputNoMPI = tmpdir.join('StommelNoMPI.nc')
 
-        system('mpirun -np 2 python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputMPI, repeatdt, maxage))
-        system('python %s -p 4 -o %s -r %d -a %d' % (stommel_file, outputNoMPI, repeatdt, maxage))
+        # ehm, this test doesn't work. I remember that we only employ the actual ParticleSet split _when_ # particles >= 1024. Testing it with 4 particles doesn't test MPI at all.
```
erikvansebille, line 23: This comment can go, as we discussed
```suggestion
```

erikvansebille (commented):
Thanks for updating all these test files, @CKehl! Must have been a lot of work... Just a few minor comments and suggestions above

parcels/examples/example_stommel.py
CKehl: hm, yes, that could work, nice suggestion.

tests/test_grids.py
CKehl: nicely spotted! indeed that was a bit of a rework error in the test. fixing it.

parcels/examples/example_stommel.py
CKehl: @erikvansebille have a look at the new edits, see if that is what you had in mind, please.

erikvansebille (approved):
Looks good to me; happy to be merged

PR #1027, opened 2021-05-19
Tutorial on how to get unstuck-unbeached

reint-fischer:
Based on some work by @cpierard, I have started to create a notebook detailing the options to prevent particles from getting stuck. For now, I have only looked at how to displace particles. 

My main question thus far is whether to combine this with the other options, since there is quite some overlap, but on the other hand, this notebook is already quite long


reint-fischer:
I have added a section now explaining how to implement a partial slip boundary condition and the parabolic profile that I used in my thesis. The last figure shows how large the differences can be.

There are several things to look at still. 

For the displacement kernel, the distance to shore and the magnitude of the displacement field will affect the trajectory. How much should we look at this?

For the reduction of the cross-boundary velocity, I need to write some more text to explain the code. I also imagine that we only keep the partial-slip condition. Another challenge is to accurately determine the relative position in the cell. Now I enclose a reference point [x0,y0] and assume that all dx and dy are exactly equal.

cpierard (approved):
Hi Reint,
Sorry for taking so long to review the notebook. I think the notebook is very good as it right now. I couldn't spot any error. Great job!

reint-fischer:
Hi @erikvansebille @cpierard 

I have now commited a version that includes everything I want to include. This would be a nice moment to read through the text especially. I hope this can go online with minor changes. I think I have finally found the cleanest way to implement the boundary conditions similar to C-grids. A new project would be to look at what makes physical sense, even in C-grids, but that is for later and probably someone else.

Also in this version is a correction on where the displacement field is defined. In the previous version the inside corners were not taken into account.

Issue #1023, opened 2021-05-03
Compiling ParcelsRandom Multiple times


axnsantana:
Hi,

I am building a python package that internally uses Parcels, but I am facing a problem. Even though I do not use the part related to Parcel, I still get the compilation of the Parcels Random generator.
`INFO: Compiled ParcelsRandom ==> /tmp/parcels-1000/libparcels_random_0ad87ac4-bed2-4038-ad8c-8bd52cdace77.so`

https://github.com/OceanParcels/parcels/blob/400abb4a1335a1bdc8b0bfeb5e764ecfe452e70e/parcels/rng.py#L128

In some cases, when I make use of Joblib in any part of the package, I get the compilation N times the number of jobs I started with joblib. Jobs not correlated with Parcels.

I am using python 3.8.5 and Parcels 2.2.2.
Any suggestion to avoid this?

A minimal example:
```
# -*- coding: UTF-8 -*-

"""Package __init__ file."""

from . import module_a
from . import module_b

__all__ = ['module_a', 'module_b']
```

module_a.py
```
def mysum(a, b):
    print(F"a + b = {a+b}")
```
module_b.py
```
import parcels
from parcels import FieldSet, ParticleSet, ErrorCode
```
Executing:
```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import use_case_parcels_issue as parcels_issue
INFO: Compiled ParcelsRandom ==> /tmp/parcels-1000/libparcels_random_65fdc468-557a-49ae-83aa-0e6ba384c19f.so
>>> parcels_issue.module_a.mysum(3,4)
a + b = 7
```

axnsantana:
Here is the parallel example using joblib, N = 5 times compiled without uses parcels:

```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from joblib import Parallel, delayed
>>> import use_case_parcels_issue as parcels_issue
INFO: Compiled ParcelsRandom ==> /tmp/parcels-1000/libparcels_random_97ee78cf-56ef-4d8c-b1db-f6473b321377.so
[...]

PR #1044, opened 2021-07-08
Removing the compiling and loading of random libs on import of Parcels

erikvansebille:
Parcels compiles the random libraries on first import statement, even when those libraries are not necessary. This is not only confusing (and redundant), it can also lead to problems like in #1023. 

Here, I've simply removed the compiling and loading step from `rng.RandomC.__init__`, and the library then still is compiled when needed (and all unit tests still seem to pass). 

I'm wondering therefore whether this is a way to remove the on-import compilation, although I also remember that @CKehl mentioned this may cause a problem with MPI?

### DIFF
--- a/parcels/rng.py
+++ b/parcels/rng.py
@@ -70,6 +70,8 @@ extern float pcls_vonmisesvariate(float mu, float kappa){
         self.ccode += self.fnct_expovariate
         self.ccode += self.fnct_vonmisesvariate
         self._loaded = False
+        self.compile()
+        self.load_lib()
 
     def __del__(self):
         self.unload_lib()
@@ -123,16 +125,24 @@ extern float pcls_vonmisesvariate(float mu, float kappa){
         return self._lib
 
 
-_parcels_random_ccodeconverter = RandomC()
+_parcels_random_ccodeconverter = None
+
+
+def _assign_parcels_random_ccodeconverter():
+    global _parcels_random_ccodeconverter
+    if _parcels_random_ccodeconverter is None:
+        _parcels_random_ccodeconverter = RandomC()
 
 
 def seed(seed):
     """Sets the seed for parcels internal RNG"""
+    _assign_parcels_random_ccodeconverter()
     _parcels_random_ccodeconverter.lib.pcls_seed(c_int(seed))
 
 
 def random():
     """Returns a random float between 0. and 1."""
+    _assign_parcels_random_ccodeconverter()
     rnd = _parcels_random_ccodeconverter.lib.pcls_random
     rnd.argtype = []
     rnd.restype = c_float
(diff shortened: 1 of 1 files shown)


CKehl:
Well, so: we can remove it now and I will need to re-integrate it within the next 14 days again. Reason: I need access to that library when doing the multi-stage compilation with nodes. So, instead of just removing the code and celebrating no errors, it would actually be good to fix this by compiling the library not in `from parcels import *`, but when actually creating a particle set or a random-number generator instance. 

To my understanding, the issue rather is that we - on importing `rng.py` - create a global instance of the `RandomC()` class, in `rng.py` line 128. 

```
_parcels_random_ccodeconverter = RandomC()
```

Bad code practice anyway, those global variables. Now, I understand this is done so the user is allowed to call those random-number generation functions at any time, and it always uses the parcels RNG. Now, if that IS actually the desired working principle, it is logical and fine that this library compiles on `from parcels import *`, because we agree that when using parcels generally, the user should also use our RNG. 

Reminder why we even want that: the larger issue here on github (of which I don't know the issue code) that discusses that pre-initialised RNG seeds should always give the same outcome (in terms of randomized number sequences) in SciPy and JIT, and that is only possible if both use the same C-based RNG in the background. Which is why we prohibt the mix of our C-based RNG with NumPy RNG's or Python RNG's. Personally, I think that's overly presumptuous, but that's probably just my personal opinion and who am I to judge in this case. 

Still, if this really IS what one expects for parcels, this whole PR makes no sense cause parcels operates the way it should: when the user decides to use parcels, the consistent C-based RNG for parcels is generated so that the user can do the following:

```
from parcels import ParcelsRandom as random
random.seed(0).
print(random.random())
lat = [random.random() for i in range(100)]
```

```
from parcels import ParticleSet

[...]

def kernel(particle, time, fieldset):
    particle.lat = random()

[...]
pset.execute(kernel, runtime=1.0, dt=1.0)
```

and expect that both code snippets (i.e. both notebook cells) draw numbers from the same sequence. By the way: this is not at all covered in current tests because this approach not working would not appear in the test scripts but in a notebook style code arrangement, for example.

Now, if you want to relax that and say: the user can use any RNG he/she wants, but once he/she actively uses `ParcelsRandom` the library should compile, then we rather need to kill `rng.py` line 128, initialise that `global` instance of the code generate with `None`, and test in each random function call if that global variable was already initialised, and if it's still None, create the instance of that `RandomC()` object before proceeding. That seems from my perspective the more reliable 'fix' rather than removing the code compilation without tracking the actual problem here.

CKehl (commented):
see comment above - forgot to frame that as a review response.

erikvansebille:
Thanks @ckehl, for this extensive feedback! Exactly why I pinged you in this very simplistic PR ;-)

I think I agree with your proposed solution in the last paragraph, will implement that and see if it works!

CKehl:
comment above from the commit: oops, wrong issue number in the reference - fixing that in future commits.

CKehl (approved):
Yes, that looks much better with regard to the actual issue in question.

PR #1045, opened 2021-07-09
Particle-tracer interaction notebook

fedeguerr:
A simple tutorial about how to have particles interacting with a Field object.


erikvansebille:
Very nice tutorial, @fedeguerr! I made some minor changes to the code and text, and have only one small comment left: perhaps it would be nice to colour the trajectory of the particle in cell 8by its tracer value? Using e.g. `plt.scatter`? Of course you wouldn't be able to use the `plotTrajectoriesFile()` function anymore, but it would be quite interesting and insightful to show this trajectory and its values, perhaps even on top of the map of what is now cell 9?

erikvansebille (approved):

PR #1048, opened 2021-07-14
Fixing bug in bwd run when starttime between two snapshots

erikvansebille:
The PR fixes a small bug when particles are run in backward mode and have a starttime between two snapshots of the netcdf file for deferred_load fieldsets. This gave an OutofTime error before, but is now fixed.


CKehl (approved):
does not seem to affect other procedures - looks good, ready to merge.

Issue #1046, opened 2021-07-12
Implement writing of Fields in-step with ParticleSets


erikvansebille:
As we are implementing support for Particle-Field interaction (see #1045) with @fedeguerr, we also require a functionality to write a `Field`, at least in Scipy mode, at the same `outputdt` as the `ParticleSet`. This will not be fast, but will be necessary for post-simulation analysis of the interaction between the particles and the tracer.

erikvansebille:
This has been implemented in #1049

PR #1049, opened 2021-07-14
Implementing method for writing Fields on each outputdt

erikvansebille:
This PR implements an option to write specific Fields at each `outputdt`, using the new `Field.to_write=True` flag. This relevant for Particle-Field interaction, see also #1046.


erikvansebille:
I think this PR is ready now; the only somewhat ugly/hacky part if the creation of the Field outputfilename: 
```python
fldfilename = str(output_file.name).replace('.nc', '_%.4d' % fld.to_write)
```
Note that the `fld.write()` method will then append an additional `"f{Field.name}.nc"` to the filename

If the Particle output filename is `out.nc`, the filenames for `U` will therefore be `out_0001U.nc`, `out_0002U.nc`, ..., `out_9999U.nc`. 
Is this 'good enough'?

CKehl (approved):

PR #1051, opened 2021-07-16
Some edits to show the new .to_write method for fields

fedeguerr:
The notebook has been slightly changed to demonstrate the new Field method (developed by @erikvansebille) that enables Field writing at each outputdt.


fedeguerr:
#1051 is now part of #1049

Issue #1064, opened 2021-07-20
Worse then linear scaling with time


JamiePringle:
I have been using ocean parcels to track particles released from small locations within the Mercator 1/12 flow fields, and I am finding that when the duration of the run increases, the run time increases much more then I would expect.  For example, when I increase the integration time from 12 days to 60 days, a factor of 5, the run time increases by a factor of 10. This issue becomes much worse when I try to, for example, model larval release over a season.  This effect is making it dramatically harder to do my analyses, and is making me start to think about stoping and restarting each run after about 10 days to improve the performance; this would be cumbersome and inconvenient.

Do you understand why the scaling is worse than the expected linear, and how I might fix it? Is this expected behavior?

I am running parcels 2.2.2 on python 3.6.13

The top panel of the attached figure shows the run-time as a function of the duration of the simulation of a single release of a large number of particles in a small area on a single day. The dashed line shows the expected linear scaling.  The bottom panel shows the time it takes to integrate a day as a function of the model-day. You can see quite dramatically as the run goes on the time to calculate each day increases.  All disk caches have been cleared before each run with “sync; echo 2 > /proc/sys/vm/drop_caches”
![scalingFigure](https://user-images.githubusercontent.com/12818667/126354788-3ad288fa-6716-434c-a9f4-5aeb7784bc25.png)

I have spent some time trying to eliminate some possible issues.  The input fields are global 1/12 degree Mercator/NEMO runs. The data is in compressed netCDF fields chunked by 1 in the vertical and 256 by 256 in the horizontal. There is one file/day. I am using the standard netCDF4 engine in the field set, and chunking by 1,512,512, which I have found to be optimal.  Most runs are made with 144 particles at each depth released at 8 depths in the top few meters in a horizontally dense patch. To make sure the issue is not related to the initial particle patch expanding so that more chunks of data are read in,  the particles are released so closely in depth and horizontally that they do not disperse by more than 1 degree in the longest integration (well within 1 netcdf chunk which is about 20 degrees of latitude squared). If the initial particle distribution is larger, the issue remains. 

The effect is seen when running in both MPI and in scaler mode, and is not sensitive to the number of MPI processes. The output frequency is daily, so the output load will scale linearly and steadily with the number of days of integration. 

I would be happy to take any suggestions, or to provide model code or details to illustrate the problem. The code that launches the particles is included. 

Do you have any ideas on why this is occurring?

Thanks,
Jamie Pringle

[...]

PR #1066, opened 2021-07-21
Fixing performance issue when writing many files

erikvansebille:
This commit fixes the performance issue reported by @JamiePringle in #1064. 

Instead of writing `time_written` to the `pset_info.npy` file every `pset.write()`, this commit calculates all the times written from inspecting the individual npy pset dictionary dumps. This will probably be somewhat slower when calling `pfile.close()` (because every dictionary dump has to be opened), but is _much_ faster during `pset.execute()` so in the end will lead to much faster Parcels simulations for long runs with many output times written


CKehl (approved):

CKehl:
The changes look ok, seems fine. Question is now: does this perform substantially better than the previous state ?

erikvansebille:
Yes, check the graph at https://github.com/OceanParcels/parcels/issues/1064#issuecomment-884150611

CKehl (approved):
k, sounds good! I need to make a mental note again that I also need to adapt that then when I rebase the node to the new master.

PR #1072, opened 2021-08-10
Reducing interpolation errors in JIT

### LINKED ISSUE #1069: What actually is an interpolation error?
I'm trying to debug a `Field interpolation error`/`InterpolationError` I am getting at apparently random locations and times with a SummedField. However, I am not sure what actually triggers this error as I can't find any documentation for this error and looking through the code, I can't work out what triggers it either. Common causes of errors when interpolating include querying out of bounds (but this would trigger a `OutOfBoundError`, NaNs in the dataset (but this cannot be the issue in this case since there are no NaNs in the data and in any case, particles are triggering the error in the middle of the ocean), or incorrectly setting up arrays (but this also cannot be the issue here since Parcels starts fine, `InterpolationError`s just get triggered at random points during execution). I believe that the code carrying out interpolations for JIT mode is `parcels.h` but I am not sure how parcels works out which error to trigger when interpolating.

Some insights into which conditions trigger this error would be very much appreciated!

**Edit:** I think I've found the relevant code in `index_search.h`, looks like it's something to do with parcels trying to interpolate a particle outside of the identified cell? Not sure why this would have been triggered but will do some more digging...

erikvansebille:
This PR reduces the number of InterpolationErrors in JIT, by setting `xsi`, `eta` and `zeta` to 0 or 1 if they fall only slightly outside (within double precision) these values. Note that for values that are not within double precision of 0 or 1, the Interpolation Error is still raised. Nevertheless, it is expected that this greatly reduces the number of Interpolation Errors in the wild. This should also fix #1069 

The PR also includes a more explicit Warning message when latitudes are flipped in Parcels


parcels/include/index_search.h
```diff
@@ -260,6 +300,13 @@ static inline StatusCode search_indices_rectilinear(type_coord x, type_coord y,
   else
     *zeta = 0;
 
+  if ( (*xsi < 0)  && (is_zero_dbl(*xsi)) )       {*xsi = 0.;}
+  if ( (*xsi > 1)  && (is_close_dbl(*xsi, 1.)) )  {*xsi = 1.;}
+  if ( (*eta < 0)  && (is_zero_dbl(*eta)) )       {*eta = 0.;}
+  if ( (*eta > 1)  && (is_close_dbl(*eta, 1.)) )  {*eta = 1.;}
+  if ( (*zeta < 0) && (is_zero_dbl(*zeta)) )      {*zeta = 0.;}
+  if ( (*zeta > 1) && (is_close_dbl(*zeta, 1.)) ) {*zeta = 1.;}
```
CKehl, line 308: That's indeed quite good - indeed should prevent some barycentric interpolation errors close the the cells' edges.

CKehl (approved):

PR #1074, opened 2021-08-17
Speedup C-contiguous arrays

erikvansebille:
This PR implements the 'engineering speedup' proposed by @CKehl in #957 by making sure data arrays are C-contiguous from the start (so that they don't need to be transposed on the fly)


parcels/kernel/basekernel.py
```diff
@@ -334,13 +334,13 @@ def load_fieldset_jit(self, pset):
                                         g.chunk_loaded_touched, g.load_chunk)
                 if len(g.load_chunk) > g.chunk_not_loaded:  # not the case if a field in not called in the kernel
                     if not g.load_chunk.flags.c_contiguous:
```
erikvansebille, line 336: Above, we change this to
```suggestion
                    if not g.load_chunk.flags['C_CONTIGUOUS']:
```
but here we leave it as was. Should we also change here (and in the lines below)?

CKehl (commented):
Well, just to add the final note to this PR:
For now, I highly discourage those changes of array-copy changes in the main branch (especially before wrapping up the node-list particle set), as it will invalidate any benchmarking already done, and hence make me redo any publishable measurements. 

The changes made until now look quite good. I suggest pausing this PR and continuing with it as soon as the performance benchmarks can be easily separated from main again. At that point, the engineering improvements can safely be pursued without invalidating publishable measurements.

erikvansebille:
Sure, I understand and was not planning to merge this PR before you had completed your benchmarks. I was planning to use this branch to do some benchmarking of its speedup on of my own simulations. 

Let's then merge this PR asap (surely before the next Parcels release) so that users can take advantage of it too; would be a shame to deny them this potential speed-up for their simulations!

PR #1076, opened 2021-08-17
Support Field.data in np.float64 format

### LINKED ISSUE #1075: Support for casting Field data to float64
For some applications, (especially in Particle-Field interactions, such as performed by @fedeguerr), it is important to work with higher precision for `Field` data. Therefore, it would be useful to have an option to stored `Field.data` as `np.float64`; although using this option would of course come at the cost of increased memory load

erikvansebille:
This PR adds a keyword option in `Field` (and `FieldSet`) creation to _not_ convert `Field.data` to np.float32. This can be used when `Field.data` needs to be available in float64 format, for example for reasons of precision

This PR fixes #1075

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -70,6 +70,7 @@ class Field(object):
     :param transpose: Transpose data to required (lon, lat) layout
     :param vmin: Minimum allowed value on the field. Data below this value are set to zero
     :param vmax: Maximum allowed value on the field. Data above this value are set to zero
+    :param usefloat32: Cast Field data to float32 for more efficient memory use (at expense of precision)
     :param time_origin: Time origin (TimeConverter object) of the time axis (only if grid is None)
     :param interp_method: Method for interpolation. Options are 'linear' (default), 'nearest',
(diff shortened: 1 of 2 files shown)


parcels/kernel/basekernel.py
```diff
@@ -322,6 +322,8 @@ def load_fieldset_jit(self, pset):
             for f in pset.fieldset.get_fields():
                 if type(f) in [VectorField, NestedField, SummedField]:
                     continue
+                if f.data.dtype != np.float32:
+                    raise RuntimeError('Field %s data needs to be float32 in JIT mode' % f.name)
```
CKehl, line 326: important statement, as the field itself is cast regardlessly.

CKehl (approved):
What is written right now on this PR is functional and acceptable. 2 things that worry me the way it's done is (a) the field is hard-cast to either float64 or float32 (no flexibility here) on `Field` initialisation and (b) there is no testing on how this will affect the writing of fields (are you writing it in the original format or the recast ?) or the fusion of fields as `VectorField` or such (if U and V are 32-bit and W is 64-bit, or vice versa, which ramifications does that have ?).

This, I think for the purpose it is meant to be used, it's fine.

PR #1081, opened 2021-09-17
Separate GitHub Actions workflows for unit and integration tests

erikvansebille:
Because the GitHub Actions workflow is taking an hour on each commit now, @CKehl and I discussed whether we could make testing more efficient by separating the GitHub Actions into a unittesting workflow (triggered on every commit to a PR) and a integration testing workflow that is only triggered for Pull Requests that are ready for review

Before merging, it would then also be good to

- [ ] Move some of the longer-running/larger test scripts in `tests/` to `examples/` so they are run on the integration test workflow

### DIFF
--- a/.github/workflows/ci-workflow-integrationtests.yml
+++ b/.github/workflows/ci-workflow-integrationtests.yml
@@ -1,0 +1,49 @@
+name: integration-tests
+on:
+  pull_request:
+    types: [review_requested, ready_for_review]
+jobs:
+  test_macOS:
+    name: Integration testing on macOS python latest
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-workflow-integrationtests.yml
```diff
@@ -0,0 +1,49 @@
+name: integration-tests
+on:
+  pull_request:
+    types: [review_requested, ready_for_review]
```
CKehl, line 4: very cool. looks much simpler than I expected.

CKehl (approved):
very nice change

PR #1092, opened 2021-10-14
Chunkwise netcdf export

nvogtvincent:
Suggestion (WIP) for automatically 'chunking' netcdf export step if intermediate variables would exceed available system memory. Importantly, the code below currently assumes that the number of unique time steps in the first chunk == number of unique time steps for all chunks, which will often but not always be true. I am not currently sure how to calculate this cheaply.


erikvansebille:
Closing this PR, as the conversion from the `npy` dumps to NetCDF will not be necessary anymore once we save directly to `zarr` format (#1199).

PR #1096, opened 2021-10-19
Interaction periodic bc

pdnooteboom:
After merging, it is possible to apply particle-particle interaction with periodic boundary conditions in the zonal direction. This means that the calculation of particle neighbours (i.e. to determine which particles are within the maximum interaction distance of a particle) in the particle-particle interaction, takes into account zonally periodic boundaries.
Meridional periodic boundaries are not implemented yet.

### DIFF
--- a/parcels/interaction/neighborsearch/base.py
+++ b/parcels/interaction/neighborsearch/base.py
@@ -14,7 +14,7 @@ class BaseNeighborSearch(ABC):
     """
 
     def __init__(self, inter_dist_vert, inter_dist_horiz,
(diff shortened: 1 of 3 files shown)


erikvansebille:
Nice addition to the interaction-particles feature, @pdnooteboom! I've just written a unit test to check if the code indeed works. It does for `mesh='flat'`, but is not implemented yet for `mesh='spherical'` so the CI breaks on that unit test. 

Could you try implementing for spherical meshes too, while you're at it? Then we really have a feature-complete PR!

PR #1105, opened 2021-11-12
Modify PolyTEOS10_bsq to accept negative and positve depths

cpierard:
Hi, 

I added the math.fabs() function in order to accept positive or negative depths for the TEOS10 function.

### DIFF
--- a/parcels/application_kernels/TEOSseawaterdensity.py
+++ b/parcels/application_kernels/TEOSseawaterdensity.py
@@ -22,7 +22,7 @@ def PolyTEOS10_bsq(particle, fieldset, time):
     Oceanic Technology, 20, 730-741.
     '''
 
-    Z = math.fabs(particle.depth)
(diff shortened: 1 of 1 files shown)


parcels/application_kernels/TEOSseawaterdensity.py
```diff
@@ -22,7 +22,7 @@ def PolyTEOS10_bsq(particle, fieldset, time):
     Oceanic Technology, 20, 730-741.
     '''
 
-    Z = - particle.depth  # note: use negative depths!
+    Z = math.fabs(particle.depth)
```
erikvansebille, line 25:
```suggestion
    Z = - math.fabs(particle.depth)
```
I think the point was that `Z` should always be a negative number?

cpierard: I was confused about that because of the comment on the same line.

PR #1034, opened 2021-06-01
Integration of the double-linked node list ParticleSet structure to 'master'

CKehl:
major effort to include now the double-linked bode list as collection data structure for ParticleSets. Necessary prelude to the `particle.add(..)` function extension (as well as `particle.split(...)`) for internal features like fragmentation.

PR #1113, opened 2021-12-16
Use actual particle datatype for netcdf export

nvogtvincent:
At present, during conversion to netcdf format, all variables are converted to float32, resulting in a loss of information in certain cases. This PR ensures that the correct dtype is used during netcdf export (and sets the fill value to the minimum possible value for ints).

### DIFF
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -221,7 +221,6 @@ class BaseParticleFile(ABC):
             self.z.units = "m"
             self.z.positive = "down"
 
-
         for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
                 fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -224,7 +224,7 @@ class BaseParticleFile(ABC):
 
         for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
-                fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
+                fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
                 nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
                 setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, coords, fill_value=fill_value))
                 getattr(self, vname).long_name = ""
@@ -232,7 +232,7 @@ class BaseParticleFile(ABC):
                 getattr(self, vname).units = "unknown"
 
         for vname, dtype in zip(self.var_names_once, self.var_dtypes_once):
-            fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
+            fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
             nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
             setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, "traj", fill_value=fill_value))
             getattr(self, vname).long_name = ""
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -100,10 +100,12 @@ class BaseParticleFile(ABC):
             for v in self.particleset.collection.ptype.variables:
                 if v.to_write == 'once':
                     self.var_names_once += [v.name]
-                    self.var_dtypes_once += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
+                    self.var_dtypes_once += [v.dtype.__name__]
+                    # self.var_dtypes_once += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
                 elif v.to_write is True:
                     self.var_names += [v.name]
-                    self.var_dtypes += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
+                    self.var_dtypes += [v.dtype.__name__]
+                    # self.var_dtypes += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
             if len(self.var_names_once) > 0:
                 self.written_once = []
                 self.file_list_once = []
@@ -222,15 +224,17 @@ class BaseParticleFile(ABC):
 
         for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
-                fill_value = np.nan if dtype[0] == 'f' else -2**(8*int(dtype[-1])-1)
-                setattr(self, vname, self.dataset.createVariable(vname, dtype, coords, fill_value=fill_value))
+                fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
+                nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
+                setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, coords, fill_value=fill_value))
                 getattr(self, vname).long_name = ""
                 getattr(self, vname).standard_name = vname
                 getattr(self, vname).units = "unknown"
 
         for vname, dtype in zip(self.var_names_once, self.var_dtypes_once):
-            fill_value = np.nan if dtype[0] == 'f' else -2**(8*int(dtype[-1])-1)
-            setattr(self, vname, self.dataset.createVariable(vname, dtype, "traj", fill_value=fill_value))
+            fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
+            nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
+            setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, "traj", fill_value=fill_value))
             getattr(self, vname).long_name = ""
             getattr(self, vname).standard_name = vname
             getattr(self, vname).units = "unknown"
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -98,11 +98,11 @@ class BaseParticleFile(ABC):
             self.var_names_once = []
             self.var_dtypes_once = []
             for v in self.particleset.collection.ptype.variables:
-                if v.to_write == 'once':
+                if v.to_write == 'once' and self.write_ondelete == False:
                     self.var_names_once += [v.name]
                     self.var_dtypes_once += [v.dtype.__name__]
                     # self.var_dtypes_once += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
-                elif v.to_write is True:
+                elif v.to_write is True or v.to_write == 'once':
                     self.var_names += [v.name]
                     self.var_dtypes += [v.dtype.__name__]
                     # self.var_dtypes += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -101,11 +101,9 @@ class BaseParticleFile(ABC):
                 if v.to_write == 'once' and self.write_ondelete == False:
                     self.var_names_once += [v.name]
                     self.var_dtypes_once += [v.dtype.__name__]
-                    # self.var_dtypes_once += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
                 elif v.to_write is True or v.to_write == 'once':
                     self.var_names += [v.name]
                     self.var_dtypes += [v.dtype.__name__]
-                    # self.var_dtypes += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
             if len(self.var_names_once) > 0:
                 self.written_once = []
                 self.file_list_once = []
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -63,8 +63,10 @@ class BaseParticleFile(ABC):
     time_origin = None
     lonlatdepth_dtype = None
     var_names = None
+    var_dtypes = None
     file_list = None
     var_names_once = None
+    var_dtypes_once = None
     file_list_once = None
     maxid_written = -1
     tempwritedir_base = None
@@ -92,12 +94,16 @@ class BaseParticleFile(ABC):
             self.time_origin = self.particleset.time_origin
             self.lonlatdepth_dtype = self.particleset.collection.lonlatdepth_dtype
             self.var_names = []
+            self.var_dtypes = []
             self.var_names_once = []
+            self.var_dtypes_once = []
             for v in self.particleset.collection.ptype.variables:
                 if v.to_write == 'once':
                     self.var_names_once += [v.name]
+                    self.var_dtypes_once += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
                 elif v.to_write is True:
                     self.var_names += [v.name]
+                    self.var_dtypes += [v.dtype.__name__[0] + str(int(int(v.dtype.__name__[-2:])/8))]
             if len(self.var_names_once) > 0:
                 self.written_once = []
                 self.file_list_once = []
@@ -213,15 +219,18 @@ class BaseParticleFile(ABC):
             self.z.units = "m"
             self.z.positive = "down"
 
-        for vname in self.var_names:
+
+        for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
-                setattr(self, vname, self.dataset.createVariable(vname, "f4", coords, fill_value=np.nan))
+                fill_value = np.nan if dtype[0] == 'f' else -2**(8*int(dtype[-1])-1)
+                setattr(self, vname, self.dataset.createVariable(vname, dtype, coords, fill_value=fill_value))
                 getattr(self, vname).long_name = ""
                 getattr(self, vname).standard_name = vname
                 getattr(self, vname).units = "unknown"
 
-        for vname in self.var_names_once:
-            setattr(self, vname, self.dataset.createVariable(vname, "f4", "traj", fill_value=np.nan))
+        for vname, dtype in zip(self.var_names_once, self.var_dtypes_once):
+            fill_value = np.nan if dtype[0] == 'f' else -2**(8*int(dtype[-1])-1)
+            setattr(self, vname, self.dataset.createVariable(vname, dtype, "traj", fill_value=fill_value))
             getattr(self, vname).long_name = ""
             getattr(self, vname).standard_name = vname
             getattr(self, vname).units = "unknown"
--- a/parcels/particlefile/baseparticlefile.py
+++ b/parcels/particlefile/baseparticlefile.py
@@ -100,10 +100,10 @@ class BaseParticleFile(ABC):
             for v in self.particleset.collection.ptype.variables:
                 if v.to_write == 'once' and self.write_ondelete is False:
                     self.var_names_once += [v.name]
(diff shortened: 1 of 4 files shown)


erikvansebille:
Thanks for this PR, @nvogtvincent. I just came across a similar comment by @CKehl while reviewing #1034:
https://github.com/OceanParcels/parcels/blob/190d317658381f7c9b0ee8987c4020e51e756ea2/parcels/particlefile/particlefilenodes.py#L120-L123
I'm not entirely sure what part of the CF conventions @CKehl refers to, but would be good to check with him

nvogtvincent:
Thanks for pointing this out @erikvansebille. Could I follow up on this @CKehl? I'm not very familiar with the CF conventions but I couldn't find any reference to data type limitations in [Appendix H4](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html)!

CKehl:
Hi everyone - so, the comment I made in the code there is approximately half a year old. I just reviewed the [CF-convention](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#discrete-sampling-geometries) (which is also quoted in-text in the docstring of baseparticlefile.py, BaseParticleFile::open_netcdf_file(self, data, shape) ). 

The above-mentioned docstring passage also mentions the 'convention data layout (cdl)" [file (i.e. incomplete trajectory declaration)](https://www.nodc.noaa.gov/data/formats/netcdf/v2.0/trajectoryIncomplete.cdl) the current implementation adheres to. This is a textfile-like listing of all accepted variables and their (common) data type - thus: we need to adhere to that specification. That file defines the following things:
```
variables:
        int trajectory(trajectory); //....................................... RECOMMENDED - If using the attribute below: cf_role. Data type can be whatever is appropriate for the unique feature type.
[...]
        double time(trajectory,obs) ;//........................................ Depending on the precision used for the variable, the data type could be int or double instead of float.
[...]
        float lat(trajectory,obs) ;//....................................... Depending on the precision used for the variable, the data type could be int or double instead of float. 
[...]
       float lon(trajectory,obs) ; //........................................ Depending on the precision used for the variable, the data type could be int or double instead of float. 
[...]
        float z(trajectory,obs) ;//........................................ Depending on the precision used for the variable, the data type could be int or double instead of float. Also the variable "z" could be substituted with a more descriptive name like "depth", "altitude", "pressure", etc.
[...]
```
Thus, this declaration file actually defines the accepted datatype that is in accordance with the CF-convention and specification of Parcels - at least the way it is documented currently.

TL;DR: 

- trajectories (i.e. their IDs, interpreted as index by NetCDF) are **32-bit signed integers (_i4_)**
- time is stored as a special character, but receives as input a **64-bit floating-point number (_f4_)**
- lat, lon and depth are stored as **32-bit floating-point numbers (_f4_)**

If that is incorrect, that I somehow missed the point of that convention data layout (cdl) file. If I am wrong, then I think we may need some clarification of the authors of that convention.

nvogtvincent:
Hi @CKehl - thanks for your response. Storing lon/lat/depth as 32-bit floats when `lonlatdepth_dtype=np.float64` seems strange to me (particularly when the cdl specifies that the datatype of lon/lat/depth "depends on the precision used for the variable"... which would be a 64-bit float in that case). But this wasn't actually my main concern - my main concern is with arbitrary user-defined variables, which should be consistent between the declaration in Parcels, and the netcdf output. And indeed, the cdl you linked states that geophysical variables (which presumably extends to any arbitrary user-defined variable) should have a "data type which is appropriate for the variable".

CKehl:
Hi @nvogtvincent - indeed you're right about it that a bit more flexibility at that point wouldn't hurt
Personally, my major concern was initially rather with the IDs or indices (signed 32-bit integer is quite a limitation).
I agree it would be a good wish-list item for the future to be more adaptive for (lon, lat, depth) to respect the `lonlatdepth_type` definition of the particle set (or: it's underlying collection). This issue didn't receive much attention in the past.
For the user-defined variables you are interested in, the store-loop referenced above would complicate quite a bit cause we'd need to `if ... elif ... else` the variable's `dtype` for each an every variable. Also there: did not receive any attention in the past, but I agree with the idea that data should be stored in the datatype they are represented in memory.

Another issue with that 'plan' is the `read_npy_file` function, which doesn't name any datatype of the stored variables. No problem if all data are later stored the same as **_f4_**, but it shows that making that arbitrary-variable storage respecting the datatype also has consequences for later code segments.

nvogtvincent:
> Hi @nvogtvincent - indeed you're right about it that a bit more flexibility at that point wouldn't hurt Personally, my major concern was initially rather with the IDs or indices (signed 32-bit integer is quite a limitation). I agree it would be a good wish-list item for the future to be more adaptive for (lon, lat, depth) to respect the `lonlatdepth_type` definition of the particle set (or: it's underlying collection). This issue didn't receive much attention in the past. For the user-defined variables you are interested in, the store-loop referenced above would complicate quite a bit cause we'd need to `if ... elif ... else` the variable's `dtype` for each an every variable. Also there: did not receive any attention in the past, but I agree with the idea that data should be stored in the datatype they are represented in memory.
> 
> Another issue with that 'plan' is the `read_npy_file` function, which doesn't name any datatype of the stored variables. No problem if all data are later stored the same as **_f4_**, but it shows that making that arbitrary-variable storage respecting the datatype also has consequences for later code segments.

I think I'm missing something because I don't understand your first point on having to go through an `if ... elif` chain for dtype? 

For the `read_from_npy` function, this is a good point, I didn't notice that. I think the issue you mean is that we would have to set the dtype of `data` on [line 75](https://github.com/OceanParcels/parcels/blob/3329c2d60e821db8ca9b40c9957ebd0b07d7fea8/parcels/particlefile/particlefilesoa.py#L75) because at present, the datatype is automatically converted to whatever the datatype of `data` is? I think the updated PR I've just pushed solves this issue?

parcels/particlefile/baseparticlefile.py
```diff
@@ -213,15 +221,20 @@ def _create_trajectory_records(self, coords):
             self.z.units = "m"
             self.z.positive = "down"
 
-        for vname in self.var_names:
+
+        for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
-                setattr(self, vname, self.dataset.createVariable(vname, "f4", coords, fill_value=np.nan))
+                fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
```
CKehl, line 227: I am not entirely confident about the `np.iinfo(np.dtype(dtype)).min` filling value. That would result for unsigned integer types in `0`, which can easily conflict with an actually useful value. I think `np.iinfo(np.dtype(dtype)).max` would be more resilient. Of course, it's a design decision - the convention doesn't give a firm limit on the fill-value.

parcels/particlefile/baseparticlefile.py
```diff
@@ -213,15 +221,20 @@ def _create_trajectory_records(self, coords):
...
+                nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
+                setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, coords, fill_value=fill_value))
                 getattr(self, vname).long_name = ""
                 getattr(self, vname).standard_name = vname
                 getattr(self, vname).units = "unknown"
 
-        for vname in self.var_names_once:
-            setattr(self, vname, self.dataset.createVariable(vname, "f4", "traj", fill_value=np.nan))
+        for vname, dtype in zip(self.var_names_once, self.var_dtypes_once):
+            fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).min
```
CKehl, line 235: see comment l. 227

parcels/particlefile/baseparticlefile.py
nvogtvincent: I think that's sensible - I just went with the minimum value since I have rarely (if ever) come across unsigned ints in a netcdf, but for versatility this seems like a good idea.

CKehl (approved):
what the PR attempts works, I think.

nvogtvincent:
The tests have failed with handling the boolean data type... at the moment, I assume parcels just converts it to a float like everything else upon export. I don't think netcdf can handle booleans, what about converting it to a 16-bit integer (the smallest allowed integer)?

parcels/particlefile/baseparticlefile.py
```diff
@@ -92,12 +94,16 @@ def __init__(self, name, particleset, outputdt=np.infty, write_ondelete=False, c
             self.time_origin = self.particleset.time_origin
             self.lonlatdepth_dtype = self.particleset.collection.lonlatdepth_dtype
             self.var_names = []
+            self.var_dtypes = []
             self.var_names_once = []
+            self.var_dtypes_once = []
             for v in self.particleset.collection.ptype.variables:
-                if v.to_write == 'once':
+                if v.to_write == 'once' and self.write_ondelete is False:
```
erikvansebille, line 101: I don't understand why this change is needed. Does it not work without this change?

parcels/particlefile/baseparticlefile.py
```diff
@@ -92,12 +94,16 @@ def __init__(self, name, particleset, outputdt=np.infty, write_ondelete=False, c
...
                     self.var_names_once += [v.name]
-                elif v.to_write is True:
+                    self.var_dtypes_once += [v.dtype.__name__]
+                elif v.to_write is True or v.to_write == 'once':
```
erikvansebille, line 104: Same as above, why does this `elif' statement need to change?

parcels/particlefile/baseparticlefile.py
```diff
@@ -92,12 +94,16 @@ def __init__(self, name, particleset, outputdt=np.infty, write_ondelete=False, c
...
             self.var_names = []
+            self.var_dtypes = []
             self.var_names_once = []
+            self.var_dtypes_once = []
             for v in self.particleset.collection.ptype.variables:
-                if v.to_write == 'once':
+                if v.to_write == 'once' and self.write_ondelete is False:
                     self.var_names_once += [v.name]
-                elif v.to_write is True:
+                    self.var_dtypes_once += [v.dtype.__name__]
```
erikvansebille, line 103: Shy not store the `dtype` object, instead of its name?
```suggestion
                    self.var_dtypes_once += [v.dtype]
```

parcels/particlefile/baseparticlefile.py
```diff
@@ -92,12 +94,16 @@ def __init__(self, name, particleset, outputdt=np.infty, write_ondelete=False, c
...
+                elif v.to_write is True or v.to_write == 'once':
                     self.var_names += [v.name]
+                    self.var_dtypes += [v.dtype.__name__]
```
erikvansebille, line 106: Same as above, why not store the object?
```suggestion
                    self.var_dtypes += [v.dtype]
```

parcels/particlefile/baseparticlefile.py
```diff
@@ -213,15 +219,19 @@ def _create_trajectory_records(self, coords):
             self.z.units = "m"
             self.z.positive = "down"
 
-        for vname in self.var_names:
+        for vname, dtype in zip(self.var_names, self.var_dtypes):
             if vname not in self._reserved_var_names():
-                setattr(self, vname, self.dataset.createVariable(vname, "f4", coords, fill_value=np.nan))
+                fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
+                nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
```
erikvansebille, line 225: This could be nicer with a dictionary? Something like declaring at the top of the function
```python
fmt_map = {'float32': 'f4', 'float64': 'f8', 'int32': 'i4', 'int64', 'i8'}
fill_value_map = {'float32': np.nan, 'float64': np.nan, 'int32': np.iinfo(np.int32).max, 'int64': np.iinfo(np.int64).max}
```
And then 
```suggestion
                fill_value = fill_value_map[str(dtype)]
                nc_dtype_fmt = fmt_map[str(dtype)]
```

parcels/particlefile/baseparticlefile.py
```diff
@@ -213,15 +219,19 @@ def _create_trajectory_records(self, coords):
...
+                setattr(self, vname, self.dataset.createVariable(vname, nc_dtype_fmt, coords, fill_value=fill_value))
                 getattr(self, vname).long_name = ""
                 getattr(self, vname).standard_name = vname
                 getattr(self, vname).units = "unknown"
 
-        for vname in self.var_names_once:
-            setattr(self, vname, self.dataset.createVariable(vname, "f4", "traj", fill_value=np.nan))
+        for vname, dtype in zip(self.var_names_once, self.var_dtypes_once):
+            fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
+            nc_dtype_fmt = dtype[0] + str(int(int(dtype[-2:])/8))
```
erikvansebille, line 233: See suggested change with dictionaries above?

parcels/particlefile/particlefilesoa.py
```diff
@@ -71,7 +72,11 @@ def read_from_npy(self, file_list, n_timesteps, var):
         :param var: name of the variable to read
         """
         max_timesteps = max(n_timesteps.values()) if n_timesteps.keys() else 0
-        data = np.nan * np.zeros((len(n_timesteps), max_timesteps))
+        fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
+        if dtype[0] == 'f':
+            data = fill_value * np.zeros((len(n_timesteps), max_timesteps), dtype=dtype)
+        else:
+            data = fill_value * np.ones((len(n_timesteps), max_timesteps), dtype=dtype)
```
erikvansebille, line 79: See comment on this change in `particlefileaos.py`

erikvansebille (commented):
Nice addition! See some comments below

erikvansebille:
> The tests have failed with handling the boolean data type... at the moment, I assume parcels just converts it to a float like everything else upon export. I don't think netcdf can handle booleans, what about converting it to a 16-bit integer (the smallest allowed integer) or an unsigned byte?

This could be cleanly implemented with the dictionary approach that I suggested in my comments? Then you could map an `np.bool` to e.g. an `i4` (or even an `i1`?) in the netcdf file?

parcels/particlefile/particlefileaos.py
```diff
@@ -71,7 +72,11 @@ def read_from_npy(self, file_list, n_timesteps, var):
         :param var: name of the variable to read
         """
         max_timesteps = max(n_timesteps.values()) if n_timesteps.keys() else 0
-        data = np.nan * np.zeros((len(n_timesteps), max_timesteps))
+        fill_value = np.nan if dtype[0] == 'f' else np.iinfo(np.dtype(dtype)).max
+        if dtype[0] == 'f':
+            data = fill_value * np.zeros((len(n_timesteps), max_timesteps), dtype=dtype)
+        else:
+            data = fill_value * np.ones((len(n_timesteps), max_timesteps), dtype=dtype)
```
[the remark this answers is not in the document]
nvogtvincent: No difference at all - this is a remnant from an earlier solution that I forgot to remove. Thank you for spotting it!

parcels/particlefile/baseparticlefile.py
nvogtvincent: No, it seems to crash when a particle is set to write `'once'` and `write_ondelete=True`. I got the same crash from the master branch. I'm not entirely sure what the cause of the crash is (and I do not know whether this is a niche crash with my complex configuration or something more basic) but this seemed to be an easy fix that worked.

parcels/particlefile/baseparticlefile.py
nvogtvincent: As above

parcels/particlefile/baseparticlefile.py
nvogtvincent: Great idea - have implemented this (along with using the dtype object rather than dtype name, you're right that there's no good justification for that)

parcels/particlefile/particlefileaos.py
```diff
@@ -111,6 +113,13 @@ def export(self):
             if MPI.COMM_WORLD.Get_rank() > 0:
                 return  # export only on threat 0
 
+        # Create dictionary to translate datatypes and fill_values
+        self.fmt_map = {np.float32: 'f4', np.float64: 'f8',
+                        np.bool_: 'i1', np.int16: 'i2', np.int32: 'i4', np.int64: 'i8'}
+        self.fill_value_map = {np.float32: np.nan, np.float64: np.nan,
+                               np.bool_: np.iinfo(np.bool_).max, np.int16: np.iinfo(np.int16).max,
```
erikvansebille, line 120: There seems to be a bug in `np.bool_: np.iinfo(np.bool_).max`. Max value for a bool is perhaps a rather strange concept?
Anyways, when I changed locally to np.nan, it worked
```suggestion
                               np.bool_: np.nan, np.int16: np.iinfo(np.int16).max,
```

parcels/particlefile/baseparticlefile.py
erikvansebille: Hmm, but the change doesn't make sense from a conceptual angle: your changes to the `if/else` change the meaning of this statement

I suggest we return to the original code in this PR, and if needed you open another Issue on the bug you're experiencing

parcels/particlefile/particlefileaos.py
nvogtvincent: Looks good - for my future reference, is there a way to run these tests locally so I can avoid pushing modifications with silly mistakes?

erikvansebille: Yes, you can run tests locally with the `pytest` from command line (e.g. `pytest tests/*.py`)

CKehl: @nvogtvincent yes, `np.nan` and `np.bool` makes conceptually no sense. `np.nan` and any non-float datatype is anyway non-sense, because integer datatypes can express any integer numbers - there are no integer-NaN's. There is something like '_overflow_', but that's basically just a binary clip after <x> bits. Hence, `nan` and `inf` are purely floating-point concepts. Numpy internally though is smart enough to map `np.nan` to `0` for integers.

I think it would be smarter to cleanly define the fill-value for bool: after all, it can only be `0` or `1`, and in this context setting it to `0` makes more sense than to `1`.

nvogtvincent: @CKehl But this fill value is for the netcdf file, which can't handle boolean datatypes anyway? The closest netcdf has to a boolean (as far as I'm aware) is an 8-bit integer, in which case I think it would be most consistent to have the fill value set to `np.iinfo(np.int8).max` not `0` (which would surely be incorrect anyway since a `False` boolean is not the same thing as the absence of data)?

CKehl: yes, indeed @nvogtvincent - I am also not arguing about the `i1` choice ;-) So, 8-bit (signed) integer is fine.
The point with the fill value is that it is a (somehow) _valid_ value in the datatype you represent information. In the file, the boolean would be 8-bit integer, so `np.iinfo(np.int8).max` would be fine. But if you then read the file into an array again, then `np.iinfo(np.int8).max` will be an overflow and will result in auto-`1` (or auto-`True`) the value. I don't know how good that maps the intention. Perhaps @erikvansebille has some feelings about it.

But overall, the 'fill value for bool' question is really boiling down to viewpoint, because in the end you really only have the choice of the 2 values `True` and `False` - and deciding for one also automatically means deciding against the other. Personally, I'd rather choice for `0` than `1`, cause after all what's otherwise the purpose of expressing `1` (i.e. `True`) anyway ...

nvogtvincent: Sorry @CKehl, I'm a bit confused what you mean by "read the file into an array again". If you load this 8-bit integer variable from the netcdf into python (or whatever), it should load it as a (masked) 8-bit integer array. You'd only get an overflow if you attempt to convert the unmasked array (raw data) to a boolean datatype. I completely agree that this consequence of converting fill values to `True` is not good and happy to see what @erikvansebille thinks about this but from my perspective, being able to distinguish between missing values and `False` values, albeit with the warning that some care must be taken when reading the file, is preferable to not being able to distinguish between between missing values and `False` values at all!

CKehl: Practical point response ;-):
- masked array conversion or clipping: depends on how the data are written. As we write it as contiguous array per attribute, you're right with the masking. This, though, it makes no difference. The only valid values for bools are `0 (False)` or `1(True)`.
- because of boolean being boolean, there is by definition no space to have a missing value skewing the data, so there will not be a way to distinguish the missing value either from a defined `True` (if decided to use `1` as missing value) or separating it from `False` (if deciding to use `0` as missing value). One of the two options is will be.
- on the warning: good point. Would you actually see it as 'important; to be informed about it for every simulation ?

erikvansebille (approved):
Very nice feature, @nvogtvincent! I am happy to now accept your PR and will merge as soon as it has green ticks on the CI

Issue #1122, opened 2022-01-13
SampleParticle time offset


apatlpo:
When using a sample particle, it seems the output sampled field is offset by one timestep compared to the input field.
Is this expected behavior?

See the following example:

```
from parcels import FieldSet, ParticleSet, JITParticle, Variable

class SampleParticle(JITParticle):
    p = Variable('p', dtype=np.float32)

def SampleP(particle, fieldset, time):
    particle.p = fieldset.P[time, particle.depth, particle.lat, particle.lon]

dims = [5, 4, 2]
dx, dy = 1./dims[0], 1./dims[1]
dimensions = {'lon': np.linspace(0., 1., dims[0], dtype=np.float32),
              'lat': np.linspace(0., 1., dims[1], dtype=np.float32),
              'time': np.arange(dims[2], dtype=np.float32),
             }
data = {'U': np.zeros(dims, dtype=np.float32),
        'V': np.zeros(dims, dtype=np.float32),
        'P': np.ones((dims[0], dims[1]))[:,:,None]*np.arange(dims[2])[None, None, :]        
       }
fieldset = FieldSet.from_data(data, dimensions, mesh='flat', transpose=True)

xv, yv = np.meshgrid(np.linspace(0, 1, 8), np.linspace(0, 1, 8))

pset = ParticleSet(fieldset, pclass=SampleParticle, lon=xv.flatten(), lat=yv.flatten())
pset.execute(SampleP, endtime=1, dt=1)
pset
```

Outputs:

```
P[320](lon=0.000000, lat=0.000000, depth=0.000000, p=0.000000, time=1.000000)
P[321](lon=0.142857, lat=0.000000, depth=0.000000, p=0.000000, time=1.000000)
...
```
i.e. `p` is 0 at `t=1` while this corresponds to its value at `t=0` but not `t=1`.

erikvansebille:
Hmm, this is a very interesting point; thanks for raising it, @apatlpo. Indeed, the order of operations in Parcels is 

Kernel evaluations -> updating time -> writing output. That has so far gone well, but in this case leads to unexpected behaviour. If you give `endtime=2` in the `pset.execute()`, you will see that `p=1` (but `t=2`)

What use-case do you have for this. Is there a way to circumvent this issue?

erikvansebille:
One fairly simple fix (hack?) could be to change your sampling kernel to
```python
def SampleP(particle, fieldset, time):
[...]

PR #1127, opened 2022-01-17
Adding section to interpolation tutorial on sampling time-varying fields

erikvansebille:
Following the Issue #1122 reported by @apaltpo


CKehl (approved):
Physically I can't provide a good feedback. If there was the demand for this tutorial the tests work out, then I think the PR is ok.

Issue #1128, opened 2022-01-19
Fieldset.from_xarray_dataset with moving vertical grid


apatlpo:
We are running simulations with a moving vertical grid.
It works fine with netcdf files but fails when we attempt to create the fieldset from an xarray dataset (see error message at the bottom).
Looking at the code, it is pretty clear this has not been implemented yet.
Being able to feed xarray dataset would be very convenient for us and I was thus wondering whether there was a fundamental reason for it not being implemented or if it was simply that nobody had needed to do that thus far?

```
variables = {'U': 'u',
             'V': 'v',
             'W': 'w',
             'depth_psi': 'z_psi',
            }

dimensions = {'U':{'lon': 'lon_psi', 'lat': 'lat_psi', 'depth':'not_yet_set', 'time': 'time_counter'},
              'V':{'lon': 'lon_psi', 'lat': 'lat_psi', 'depth':'not_yet_set', 'time': 'time_counter'},
              'W':{'lon': 'lon_psi', 'lat': 'lat_psi', 'depth':'not_yet_set', 'time': 'time_counter'},
              'depth_psi':{'lon': 'lon_psi', 'lat': 'lat_psi', 'depth':'not_yet_set', 'time': 'time_counter'},
             }

fieldset = FieldSet.from_xarray_dataset(ds,variables,dimensions,
                                allow_time_extrapolation=True,
                                interp_method='cgrid_velocity',
                                       )
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/.miniconda3/envs/equinox/lib/python3.8/site-packages/xarray/core/dataarray.py in _getitem_coord(self, key)
    730         try:
--> 731             var = self._coords[key]
    732         except KeyError:

KeyError: 'not_yet_set'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
/dev/shm/pbs.8777261.datarmor0/ipykernel_26609/1988345921.py in <module>
----> 1 fieldset = FieldSet.from_xarray_dataset(ds,variables,dimensions,
      2                                 allow_time_extrapolation=True,
      3                                 interp_method='cgrid_velocity',
      4                                        )

~/.miniconda3/envs/equinox/lib/python3.8/site-packages/parcels/fieldset.py in from_xarray_dataset(cls, ds, variables, dimensions, mesh, allow_time_extrapolation, time_periodic, **kwargs)
    926             cls.checkvaliddimensionsdict(dims)
    927 
[...]

PR #1131, opened 2022-01-20
Support time varying depth dimensions for xarray datasets

apatlpo:
Follows from issue #1128

Test notebook is on the [following gist](https://gist.github.com/apatlpo/b85198b0ce36f7edf07e931d07ed4e89)

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -495,7 +495,7 @@ class Field(object):
                 depth = np.empty((0, len(da[_dim_depth].values), len(lat), len(lon)))
                 kwargs['depth_field'] = 'not_yet_set'
             else:
-                depth = da[dimensions['time']].values
+                depth = da[dimensions['depth']].values
         else:
             depth = np.zeros(1)
 
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -485,13 +485,11 @@ class Field(object):
         interp_method = kwargs.pop('interp_method', 'linear')
 
         time = da[dimensions['time']].values if 'time' in dimensions else np.array([0])
-        #depth = da[dimensions['depth']].values if 'depth' in dimensions else np.array([0])
         lon = da[dimensions['lon']].values
         lat = da[dimensions['lat']].values
 
         if 'depth' in dimensions:
             if dimensions['depth'] == 'not_yet_set':
-                #depth = filebuffer.depth_dimensions
                 _lon_lat_time = [dimensions[d] for d in ['lon', 'lat', 'time']]
                 _dim_depth = [d for d in da.dims if d not in _lon_lat_time][0]
                 depth = np.empty((0, len(da[_dim_depth].values), len(lat), len(lon)))
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -485,9 +485,22 @@ class Field(object):
         interp_method = kwargs.pop('interp_method', 'linear')
 
         time = da[dimensions['time']].values if 'time' in dimensions else np.array([0])
-        depth = da[dimensions['depth']].values if 'depth' in dimensions else np.array([0])
+        #depth = da[dimensions['depth']].values if 'depth' in dimensions else np.array([0])
         lon = da[dimensions['lon']].values
         lat = da[dimensions['lat']].values
+
+        if 'depth' in dimensions:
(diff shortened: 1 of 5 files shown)


erikvansebille:
I just pushed a small fix to the PR; it turned out that the depth fields were never connected to the grid because the `Field.dataFiles is None` for Fields created from xarray and therefore the following lines evaluate True https://github.com/OceanParcels/parcels/blob/f6e2ed7f7a2e45eed9acbb2603caf3c124a3c831/parcels/fieldset.py#L271-L272  

This seems to be fixed now by changing the last bit to
```python
or (f.dataFiles is None and f.creation_log != 'from_xarray_dataset'):
```

Can you check if this now works as expected?

apatlpo:
I think it does indeed, see the [gist](https://gist.github.com/apatlpo/b85198b0ce36f7edf07e931d07ed4e89) !
Thanks for your help.

What next, should we try to adresse the large number of tests that are failing?
or work on a tutorial?

I also have a somewhat indirectly related question: could you please confirm using xarray does leverage its lazy capability (i.e. does not load full datasets in memory) ?

erikvansebille:
Great to hear that this fix seems to work, @apatlpo! I've just pushed a set of fixes (one related to PEP8, the other to a bug in reading the `depth` dimension from the dictionary when depth _is_ set) so hopefully the CI should now give green ticks

I don't think we need a tutorial for this, but I've just added a sentence that this works in `FieldSet.from_xarray_dataset()` in the timevarying depth dimension tutorial. We do need a unit test though, I'll create that based on your gist and put it in `tests/test_grid.py`

erikvansebille:
> I also have a somewhat indirectly related question: could you please confirm using xarray does leverage its lazy capability (i.e. does not load full datasets in memory) ?

Hmm, good question. To be honest I don't know. We haven't used `FieldSet.from_xarray_dataset()` in real-life applications yet, so no need to test/check this. I'd be very keen to hear the answer though; hopefully the lazy capabilities work out of the box!

apatlpo:
Ok, I'll give a shot to test the lazyness of the code and report back

apatlpo:
Ok, I'll give a shot to test the lazyness of the code and report back

parcels/field.py
```diff
@@ -485,10 +485,20 @@ def from_xarray(cls, da, name, dimensions, mesh='spherical', allow_time_extrapol
...
         time = da[dimensions['time']].values if 'time' in dimensions else np.array([0])
-        depth = da[dimensions['depth']].values if 'depth' in dimensions else np.array([0])
         lon = da[dimensions['lon']].values
         lat = da[dimensions['lat']].values
 
+        if 'depth' in dimensions:
+            if dimensions['depth'] == 'not_yet_set':
+                _lon_lat_time = [dimensions[d] for d in ['lon', 'lat', 'time']]
+                _dim_depth = [d for d in da.dims if d not in _lon_lat_time][0]
+                depth = np.empty((0, len(da[_dim_depth].values), len(lat), len(lon)))
```
CKehl, line 495: well, this is nice but overly simplistic and not generalized sufficiently, I think.

Say you have a NetCDF file (or an xarray structure with a similar structure) with the following 'dimensions' (i.e. 'variables' in NetCDF):

- lon
- lat
- time
- level (of detail)
- depth_index
- depth (== `not_yet_set`)

(in that order)
Now, this code would not identify `depth`, but `_dim_depth` would be set to `level`, and the process would break. Sure, we are using the `from_xarray(...)` function here, but the xarray can be arranged in the same format as the file. Auto-parsing this dimensions (from a NetCDF) and actually identifying them is quite intricate - have a look into `parcels/fieldfilebuffer.py`, class `DaskFileBuffer`, function `_is_dimension_in_dataset(self, ...)`, line 401.

For the chosen (i.e. picked) test cases, it works - but it can be made to break fairly easy this way. Just saying as a comment.

CKehl (approved):

parcels/field.py
apatlpo: apologizes, I've had to drop this momentarily, please give me a couple of days to get back to you (on that and on the xarray lazyness issue which I've started to investigate).

erikvansebille:
Closing this PR for now, as there was no action for a long time

PR #1152, opened 2022-03-17
Extended netcdf export format dictionary to unsigned ints

nvogtvincent:
The format dictionary we set up a while ago was missing unsigned integers + 8-bit integers, so I've just added those.


erikvansebille:
Nice addition, but should these types then also be tested in test_particlefile.py? 

I.e. add them to the line
https://github.com/OceanParcels/parcels/blob/65e5d595ba4913d08c73b01eae3accc3a9bd57c1/tests/test_particle_file.py#L180

erikvansebille (approved):

PR #1160, opened 2022-04-11
Add to diffusion tutorial

RomanIMMSP:
Dear Parcels community

In many lagrangian simulations, the precalculated velocity fields are used. They allow simulating the transfer of lagrangian particles due to currents (advection). However, sometimes diffusion processes are also important. Using uniform diffusion coefficients Kh is the simplest way to solve this problem. But, in the case of complex currents, the coefficient of horizontal diffusion is not uniform - Kh(i,j). There are several methods allowing calculate Kh using values of velocities and lon-lat resolutions. One of the simplest ways is using the Smagorinsky method.
I suppose to add a new example in the Diffusion tutorial describing the movement of the particles in the GlobalCurrents velocity fields for the region around South Africa taking into account the horizontal diffusion calculated by the Smagorinsky method. The corresponding example is described in the file 'example_smagdiff.py'. I think it could be useful for new Parcels users. The source code was created with the help of Daan Reijnders.


daanreijnders (approved):
Thanks for adding this tutorial, Roman! It's very instructive for future users.

I've added a small explanation and the formulation of the Smagorinsky parameterization, along with some small comments. I also removed `yres` and `xres`. Those were remnants in the code from @pdnooteboom that I sent you earlier, from when the cell areas conversion between meters and degrees was done differently. With these changes, it's all good to merge imo.

Thanks!

RomanIMMSP:
Thank you, Daan, for your contribution.

PR #1166, opened 2022-04-21
Support curvilinear MITgcm-like indexing

daanreijnders:
Currently, MITgcm/ROMS-like indexing is only supported for rectilinear grids, although the code may already be able to work with curvilinear grids too. In this PR we:
- [x] Turn off the error that checks whether `gridindexingtype = 'mitgcm'`
- [x] Add a unit test for curvilinear MITgcm-like indexing.

### DIFF
--- a/tests/test_grids.py
+++ b/tests/test_grids.py
@@ -634,59 +634,8 @@ def test_popgrid(pset_mode, mode, vert_discretisation, deferred_load):
 @pytest.mark.parametrize('pset_mode', pset_modes)
 @pytest.mark.parametrize('mode', ['scipy', 'jit'])
 @pytest.mark.parametrize('gridindexingtype', ['mitgcm', 'nemo'])
-def test_cgrid_indexing(pset_mode, mode, gridindexingtype):
-    xdim, ydim = 151, 201
-    a = b = 20000  # domain size
-    lon = np.linspace(-a / 2, a / 2, xdim, dtype=np.float32)
-    lat = np.linspace(-b / 2, b / 2, ydim, dtype=np.float32)
-    dx, dy = lon[2] - lon[1], lat[2] - lat[1]
-    omega = 2 * np.pi / delta(days=1).total_seconds()
-
-    index_signs = {'nemo': -1, 'mitgcm': 1}
-    isign = index_signs[gridindexingtype]
-
-    def calc_r_phi(ln, lt):
-        return np.sqrt(ln ** 2 + lt ** 2), np.arctan2(ln, lt)
-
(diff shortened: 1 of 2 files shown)


erikvansebille (commented):
Thanks for adding this unit test for curvilinear grids, @daanreijnders. But most of the code is similar to the `test_cgrid_indexing()` function at https://github.com/OceanParcels/parcels/blob/39aabb163398da5cfc6459a65f73e301f39b7668/tests/test_grids.py#L637

So would it not be easier to add an extra pytest argument (`'curvilinear', [True, False]`) that switches between the rectilinear and curvilinear version of this test (where I think the only extra bits for the curvilinear test are
https://github.com/OceanParcels/parcels/blob/39aabb163398da5cfc6459a65f73e301f39b7668/tests/test_grids.py#L700-L708

and the `alpha` at 
https://github.com/OceanParcels/parcels/blob/39aabb163398da5cfc6459a65f73e301f39b7668/tests/test_grids.py#L721-L724

Or, another option is to make `'alpha', [0, 15*np.pi/180]` a pytest parameter?

erikvansebille (approved):

tests/test_grids.py
```diff
@@ -645,24 +646,53 @@ def test_cgrid_indexing(pset_mode, mode, gridindexingtype):
...
+            R = np.zeros(lat.shape, dtype=np.float32)
+            for i in range(lat.shape[1]):
+                for j in range(lat.shape[0]):
+                    r, phi = calc_r_phi(lon[j, i], lat[j, i])
+                    R[j, i] = r
+                    r, phi = calc_r_phi(lon[j, i] + isign * (dx / 2) * np.cos(alpha), lat[j, i] - isign * (dx / 2) * np.sin(alpha))
+                    V[j, i] = np.sin(alpha) * (omega * r * np.cos(phi)) + np.cos(alpha) * (-omega * r * np.sin(phi))
+                    r, phi = calc_r_phi(lon[j, i] + isign * (dy / 2) * np.sin(alpha), lat[j, i] + isign * (dy / 2) * np.cos(alpha))
+                    U[j, i] = np.cos(alpha) * (omega * r * np.cos(phi)) - np.sin(alpha) * (-omega * r * np.sin(phi))
+            return U, V, R
```
daanreijnders, line 692: I've now merged the two tests. I didn't opt for simply having an `alpha` argument, as I think it's also good to differentiate the the case when `lon` and `lat` are 1D or 2D arrays, as I guess this is the biggest way in which rectilinear versus curvilinear grid handling differ.

Issue #1172, opened 2022-05-20
keyError when trying to sampling velocity 


slawchune:
Hi Eric and co-workers, 

I encounter an issue that I can't solve. I Try to do a basic velocity sampling, but fall on a error. Here is the behaviour (I have successfully loaded my fieldset and initial (lon,lat,time) of my pset): 

```
def Sample_Velocity(particle, fieldset, time):
    # sample velocity in m/s instead of m/s 
    particle.U = fieldset.U.eval(time, particle.depth, particle.lat, particle.lon, applyConversion=False)
    particle.V = fieldset.V.eval(time, particle.depth, particle.lat, particle.lon, applyConversion=False)

class SampleParticle(JITParticle):         # Define a new particle class
    U = Variable('U', initial=0, to_write=True)
    V = Variable('V', initial=0, to_write=True)

# initialise the set of particle in Parcels kernel
pset = ParticleSet(fieldset=fieldset, pclass=SampleParticle,
                      lon=launch_lon, lat=launch_lat, time=launch_date)

ksample_UV = pset.Kernel(Sample_Velocity)    # Casting the Sample_UV function to a kernel.
pset.execute(ksample_UV, dt=0) # by only executing the sample kernel we record the initial UV of the particle

```

at the execution: 

```

In [63]:     pset.execute(ksample_UV, dt=0) # by only executing the sample kernel we
    ...: 
---------------------------------------------------------------------------
KernelError                               Traceback (most recent call last)
Input In [63], in <cell line: 1>()
----> 1 pset.execute(ksample_UV, dt=0)

File ~/.conda/envs/py3_parcels/lib/python3.10/site-packages/parcels/particleset/baseparticleset.py:458, in BaseParticleSet.execute(self, pyfunc, pyfunc_inter, endtime, runtime, dt, moviedt, recovery, output_file, movie_background_field, verbose_progress, postIterationCallbacks, callbackdt)
    456 # If we don't perform interaction, only execute the normal kernel efficiently.
    457 if self.interaction_kernel is None:
--> 458     self.kernel.execute(self, endtime=next_time, dt=dt, recovery=recovery, output_file=output_file,
    459                         execute_once=execute_once)
    460 # Interaction: interleave the interaction and non-interaction kernel for each time step.
    461 # E.g. Inter -> Normal -> Inter -> Normal if endtime-time == 2*dt
    462 else:
    463     cur_time = time

[...]

PR #1175, opened 2022-05-27
Fixing sampling of fieldset.UV

erikvansebille:
As reported in #1172, there was a bug in the sampling of velocities. When `fieldset.U` and `fieldset.V` are sampled independently, they return the velocities in `i` and `j` index, respectively; which on a curvilinear grid is not per se the same as the velocities in longitude and latitude direction. Furthermore, boundary conditions such as free slip and partial slip are not taken into account either when sampling `fieldset.U` and `fieldset.V` separately. 

The _correct_ way to sample a velocity is
```python
(particle.u, particle.v) = fieldset.UV[time, particle.depth, particle.lat, particle.lon]
```
Or, when wanting to avoid unit conversion
```python
(particle.u, particle.v) = fieldset.UV.eval(time, particle.depth, particle.lat, particle.lon, applyConversion=False)
```

This PR:
- [x] implements support for the second `UV.eval()` option (the first was already implemented, and is used by the default advection kernels).
- [ ] updates the sampling and unitconversion tutorials to explain that `fieldset.UV` should be sampled when requesting velocities at a location
- [ ] implements a warning when users still sample `fieldset.U` or `fieldset.V` separately (as there are hardly any use cases)

### DIFF
--- a/parcels/compilation/codegenerator.py
+++ b/parcels/compilation/codegenerator.py
@@ -78,17 +78,31 @@ class FieldEvalNode(IntrinsicNode):
 
 
 class VectorFieldNode(IntrinsicNode):
+    def __getattr__(self, attr):
+        if attr == "eval":
+            return VectorFieldEvalCallNode(self)
+        else:
+            raise NotImplementedError('Access to VectorField attributes are not (yet) implemented in JIT mode')
+
     def __getitem__(self, attr):
         return VectorFieldEvalNode(self.obj, attr)
 
 
+class VectorFieldEvalCallNode(IntrinsicNode):
+    def __init__(self, field):
+        self.field = field
+        self.obj = field.obj
+        self.ccode = ""
+
+
 class VectorFieldEvalNode(IntrinsicNode):
-    def __init__(self, field, args, var, var2, var3):
+    def __init__(self, field, args, var, var2, var3, convert=True):
         self.field = field
         self.args = args
         self.var = var  # the variable in which the interpolated field is written
         self.var2 = var2  # second variable for UV interpolation
         self.var3 = var3  # third variable for UVW interpolation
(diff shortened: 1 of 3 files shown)


erikvansebille:
@daanreijnders, could you check the changes in the Smagorinsky kernel in the tutorial_diffusion notebook, and the other kernels in `advectiondiffusion.py`?

erikvansebille:
@daanreijnders, could you check the changes in the kernels in `advectiondiffusion.py` and in the Smagorisky kernel in the diffusion tutorial?

parcels/application_kernels/advectiondiffusion.py
```diff
@@ -66,16 +65,18 @@ def AdvectionDiffusionEM(particle, fieldset, time):
     dWx = ParcelsRandom.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
     dWy = ParcelsRandom.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
 
+    u, v = fieldset.UV[time, particle.depth, particle.lat, particle.lon]
+
     Kxp1 = fieldset.Kh_zonal[time, particle.depth, particle.lat, particle.lon + fieldset.dres]
     Kxm1 = fieldset.Kh_zonal[time, particle.depth, particle.lat, particle.lon - fieldset.dres]
     dKdx = (Kxp1 - Kxm1) / (2 * fieldset.dres)
-    ax = fieldset.U[time, particle.depth, particle.lat, particle.lon] + dKdx
+    ax = u + dKdx
```
daanreijnders, line 73: Looks good. Even clearer this way, as it more easily shows what the deterministic term is composed of.

daanreijnders (approved):
The diffusion kernels all look fine, also in the notebook.

CKehl (approved):

PR #1182, opened 2022-06-13
Projection is never None at this point of the code and the if-clause is ill defined

ocefpaf:
If a user sets projection to None the `create_parcelsfig_axis` will create one with PlateCaree, making this if-clause ill defined and it will fail in some cases with the cryptic `ValueError: A LinearRing must have at least 3 coordinate tuples` that will confuse newcomers.

I'm not sure if this is the best solution but it does fix an issue some users I'm helping are experiencing. For now they are defining a projection explicitly to avoid setting the x-axis with the wrong values.

### DIFF
--- a/parcels/plotting.py
+++ b/parcels/plotting.py
@@ -48,7 +48,7 @@ def plotparticles(particles, with_particles=True, show_time=None, field=None, do
             return  # creating axes was not possible
         ax.set_title('Particles' + parsetimestr(particles.fieldset.U.grid.time_origin, show_time))
         latN, latS, lonE, lonW = parsedomain(domain, particles.fieldset.U)
-        if cartopy is None or projection is None:
+        if cartopy is None:
             if domain is not None:
                 if isinstance(particles.fieldset.U.grid, CurvilinearGrid):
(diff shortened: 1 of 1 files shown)


erikvansebille:
Thanks for reporting this Issue, @ocefpaf. And also for proposing this quick fix!
I dug in a bit this morning, and have now proposed a slightly more thorough implementation; can you check if this version also fixes your Issue? if not, please let me know when and where it goes wrong, so I can try fix it again

ocefpaf:
> Thanks for reporting this Issue, @ocefpaf. And also for proposing this quick fix! I dug in a bit this morning, and have now proposed a slightly more thorough implementation; can you check if this version also fixes your Issue? if not, please let me know when and where it goes wrong, so I can try fix it again

It does fix the issue. I did not check all the ramifications of that change. Looking at your PR I believe that parcels could probably use a re-factor. That is not something I can do and I believe most parcels devs are also quite busy.

Have you consider participating in outreachy or Google Summer of Code? I can help you with that. If you are interested feel free to contact me.

PR #1222, opened 2022-08-18
Using f-string in logger info for plotting

erikvansebille:
To support more types for savefile path (e.g. PathLike-objects); as suggested in #1221


VeckoTheGecko:
Yes that would fix it, LGTM! (will make sure I get a dev install up and running next time so I can directly checkout these fixes 😄 )

PR #1231, opened 2022-08-31
Equal distribution of particles to MPI processors

JakeNewmanUEA:
When running simulations in MPI mode involving the repeated release of small quantities of particles (e.g. 100 particles across 10 MPI processors), we would occasionally receive the error 'Cannot initialise with fewer particles than MPI processors'. The cause of this appeared to be the way that K-means was distributing the particles among the MPI processors. Specifically, K-means does not return clusters of a fixed or minimum size, thus it is highly possible that some MPI processors will be allocated fewer than the minimum number of required particles, especially if the number of particles is small and the number of MPI processors approaches the maximum for a given number of particles (e.g. 10 MPI processors is the maximum for 100 particles, as 100 particles / 10 processors = 10 particles per processor). To overcome this, we equally distribute the available particles among the MPI processors using Numpy's linspace method. We create a linearly spaced array that is the same length as the number particles. The array contains decimals ranging from 0 to the number of MPI processors, and that array is then rounded down so that the values represent integer indices to MPI processors. As before, this change requires that the user has requested at least the minimum required number of MPI processors, but it ensures that the minimum number is always sufficient.

This change has only been tested for collectionsoa.py but not collectionaos.py.

### DIFF
--- a/parcels/collection/collectionaos.py
+++ b/parcels/collection/collectionaos.py
@@ -101,9 +101,10 @@ class ParticleCollectionAOS(ParticleCollection):
                 if partitions is not False:
                     if self._pu_indicators is None:
                         if mpi_rank == 0:
-                            coords = np.vstack((lon, lat)).transpose()
-                            kmeans = KMeans(n_clusters=mpi_size, random_state=0).fit(coords)
-                            self._pu_indicators = kmeans.labels_
+                            # distribute particles equally among MPI processors
+                            labels = np.linspace(0, mpi_size, lon.size, endpoint=False)
+                            labels = np.floor(labels)
+                            self._pu_indicators = labels
                         else:
                             self._pu_indicators = None
                         self._pu_indicators = mpi_comm.bcast(self._pu_indicators, root=0)
--- a/parcels/collection/collectionaos.py
+++ b/parcels/collection/collectionaos.py
@@ -3,6 +3,7 @@ from operator import attrgetter  # NOQA
 
 from ctypes import c_void_p
 
+import copy
 import numpy as np
 
 from parcels.collection.collections import ParticleCollection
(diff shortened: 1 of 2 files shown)


erikvansebille:
Thanks for creating this PR, @JakeNewmanUEA. Interesting issue that you found; although I wonder what use cases would warrant distribution of 100 particles on 10 cores?

Your solution is in principle elegant; but the problem is that it doesn't take into account the efficiency of the spatial distribution of particles per processor. Since MPI is often run together with FieldSet chunking, it is _highly_ beneficial to put particles that are spatially close together on the same processor, so to minimise the number of chunks that each processor has to load. 

See also the [Load balancing section of the MPI documentation](https://nbviewer.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_MPI.ipynb#Future-developments:-load-balancing). Note that actual 'on-the-fly' load-balancing is still a future development; but the `KMeans`-distribution is a way to at least do some load balancing at the initial distribution of particles over the processors. 

So can you come up with a solution that is spatially-aware?

JakeNewmanUEA:
> Thanks for creating this PR, @JakeNewmanUEA. Interesting issue that you found; although I wonder what use cases would warrant distribution of 100 particles on 10 cores?
> 
> Your solution is in principle elegant; but the problem is that it doesn't take into account the efficiency of the spatial distribution of particles per processor. Since MPI is often run together with FieldSet chunking, it is _highly_ beneficial to put particles that are spatially close together on the same processor, so to minimise the number of chunks that each processor has to load.
> 
> See also the [Load balancing section of the MPI documentation](https://nbviewer.org/github/OceanParcels/parcels/blob/master/parcels/examples/documentation_MPI.ipynb#Future-developments:-load-balancing). Note that actual 'on-the-fly' load-balancing is still a future development; but the `KMeans`-distribution is a way to at least do some load balancing at the initial distribution of particles over the processors.
> 
> So can you come up with a solution that is spatially-aware?

I've reinstated the k-means allocation but now it assigns an equal number of particles to each centroid. Probably still not ideal, but hopefully this is a better solution than no spatial awareness at all or imbalanced particle allocation. At least it still overcomes the original issue I described (the case for which my colleague has offered to elaborate on soon).

erikvansebille:
Thanks for this update, @JakeNewmanUEA. But I wonder now whether it would not be more efficient and easier to implement another spatial clustering algorithm altogether, that seeks roughly equal clusters.

This is not my expertise, but from [this introduction website](https://geodacenter.github.io/workbook/9c_spatial3/lab9c.html), it seems that the SKATER algorithm could be a good candidate? It's implemented in [pygeoda](https://geodacenter.github.io/pygeoda/spatial_clustering.html), which would then become a dependency for Parcels. 

Do you have experience with this? Do you want to give it a try whether `pygeoda.skater()` works for your case?

erikvansebille:
This has now been fixed/implemented in #1414 and #1424, so closing this PR

PR #1247, opened 2022-09-27
Fix field plotting

### LINKED ISSUE #741: "compute_on_defer" not working with dask array

CKehl:
Added mini-test for loading and plotting NEMO data with dask for error replication and fix #741

Comment: little clean-up of open branches. This one is a small addition adding more stability to the plotting by checking that plotting works correctly. Valuable addition with little extra hassle - hence: merge-request.


parcels/scripts/plottrajectoriesfile.py
```diff
@@ -71,7 +71,8 @@ def plotTrajectoriesFile(filename, mode='2d', tracerfile=None, tracerfield='P',
     if mode == '3d':
         from mpl_toolkits.mplot3d import Axes3D  # noqa
         plt.clf()  # clear the figure
-        ax = fig.gca(projection='3d')
+        # ax = fig.gca(projection='3d')
+        ax = fig.gca()
```
erikvansebille, line 75: I don't understand this change. I think this is what breaks the unit-tests. The `ax` is now not set as 3D anymore, so there is no `set_zlevel` method. Revert back to original?

CKehl: The change was necessary cause that line broke [bccb214](https://github.com/OceanParcels/parcels/pull/1247/commits/bccb2140274886f0ecbbd56dbfb019c036b2a3d1). In short: in Windows, the workflow obtains a newer version of matplotlib, and in matplotlib, the function `figure.gca()` has no `projection` parameters. Unexpected change, but therefore I needed to make this `projection` option part of the figure creation procedure. I guess help would be welcome from [bccb214](https://github.com/OceanParcels/parcels/pull/1247/commits/bccb2140274886f0ecbbd56dbfb019c036b2a3d1) onwards.

parcels/scripts/plottrajectoriesfile.py
```diff
@@ -71,8 +71,7 @@ def plotTrajectoriesFile(filename, mode='2d', tracerfile=None, tracerfield='P',
     if mode == '3d':
         from mpl_toolkits.mplot3d import Axes3D  # noqa
         plt.clf()  # clear the figure
-        # ax = fig.gca(projection='3d')
-        ax = fig.gca()
+        ax = plt.axes(projection='3d')
```
CKehl, line 74: nice

environment_py3_linux.yml
```diff
@@ -28,4 +28,5 @@ dependencies:
   - nbval
   - scikit-learn
   - pykdtree
+  - cartopy
```
erikvansebille, line 31: Why is cartopy now need? That was not the case before, and I suggest not to keep it
```suggestion
```

environment_py3_win.yml
```diff
@@ -19,10 +19,11 @@ dependencies:
...
   - tqdm
   - xarray>=0.5.1
-  - dask>=2.0
+  - dask<=2022.9.0
   - cftime>=1.3.1
   - ipykernel
   - pytest
   - nbval
   - pykdtree
+  - cartopy
```
erikvansebille, line 28: See above, I suggest not to make cartopy an explicit dependency (as we had not done before)
```suggestion
```

parcels/plotting.py
```diff
@@ -101,7 +102,7 @@ def plotparticles(particles, with_particles=True, show_time=None, field=None, do
 
 
 def plotfield(field, show_time=None, domain=None, depth_level=0, projection='PlateCarree', land=True,
-              vmin=None, vmax=None, savefile=None, **kwargs):
+              vmin=None, vmax=None, savefile=None, use3D=False, **kwargs):
```
erikvansebille, line 105: Is this variable still needed now? I am a bit confused what this does

parcels/plotting.py
```diff
@@ -264,21 +268,26 @@ def plotfield(field, show_time=None, domain=None, depth_level=0, projection='Pla
...
+        ax.grid()
+    elif spherical and projection:
         try:
             import cartopy
         except:
             logger.info("Visualisation of field with geographic coordinates is not possible. Cartopy not found.")
             return None, None, None, None  # creating axes was not possible
 
         projection = cartopy.crs.PlateCarree(central_longitude) if projection == 'PlateCarree' else projection
+        # projection = '3d' if use3D else projection
```
erikvansebille, line 290: Stray debugging comment
```suggestion
```

tests/test_plotting.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+if __name__ == '__main__':
+    fset = test_field_from_netcdf()
+    print(fset)
+    pset = test_pset_create_field(fset)
+    kernels = pset.Kernel(AdvectionRK4) + periodicBC
+    pset.execute(kernels, dt=delta(hours=1), output_file=None,
+                 recovery={ErrorCode.ErrorOutOfBounds: DeleteParticle})
+    pset.show(field=fset.U)
```
erikvansebille, line 54: What does this new set of functions test that was not tested before? Why is it needed?

environment_py3_linux.yml
CKehl: that will make the _new_ test fail that actually tests platecarree plotting

parcels/plotting.py
CKehl: possibly not, indeed

erikvansebille:
I think that this was now fixed in 0326f773e0faee0cfb1400ad2bd0385833123936

PR #1273, opened 2023-01-06
Remove broken badge from README

VeckoTheGecko:
Removed broken badge from README. No such workflow `test-suite`.


README.md
```diff
@@ -24,7 +24,6 @@ See [oceanparcels.org](http://oceanparcels.org/) for further information about [
 ### Launch Parcels Tutorials on mybinder.org
 
 [![Binder](https://mybinder.org/badge.svg)](https://mybinder.org/v2/gh/OceanParcels/parcels_examples_binder/master?urlpath=lab/tree/parcels_examples/parcels_tutorial.ipynb)
-[![Build Status](https://github.com/oceanparcels/parcels/workflows/test-suite/badge.svg?branch=master)](https://github.com/OceanParcels/parcels/actions?query=workflow%3Atest-suite+branch%3Amaster)
```
erikvansebille, line 27: Why remove the badge all together? I think that it would work well to remove the workflow and testsuite arguments, e.g.
```
[![Build Status](https://github.com/oceanparcels/parcels/workflows/test-suite/badge.svg)](https://github.com/OceanParcels/parcels/actions)

VeckoTheGecko: From my understanding of badges, they can be used only to monitor the success of individual workflows ([docs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge)). With the `test-suite` keyword, I thought it was monitoring an outdated workflow. I wasn't sure which workflow to replace it with.

Does `test-suite` have a different meaning? The link you gave (https://github.com/oceanparcels/parcels/workflows/test-suite/badge.svg) does lead to a badge, but I've never seen that URL format before for badges.

erikvansebille:
Now that the [weekly schedule for unit tests and integration tests works](https://github.com/OceanParcels/parcels/actions) (thanks @VeckoTheGecko!), perhaps we can set the badge on the README to track the weekly unit-tests action? 

I don't think we can use one badge to track both unit-tests and integration-tests, and two badges may be overdoing it a bit. What do you think, @VeckoTheGecko?

VeckoTheGecko:
One badge monitoring `unit-tests` sounds good. I have updated the readme

Issue #1274, opened 2023-01-06
Failing test "test_randomvonmises" on master


VeckoTheGecko:
The `test_randomvonmises` test is failing in the build process on master (as seen in the workflow output of #1273 ).

https://github.com/OceanParcels/parcels/blob/56e988ae0f627ec8994c656f5f4bab71154a2e47/tests/test_diffusion.py#L120-L148

VeckoTheGecko:
I think it might be to do with this line, and potentially a change in Scipy (since no change in `parcels` seems to have triggered the test failing).

https://github.com/OceanParcels/parcels/blob/56e988ae0f627ec8994c656f5f4bab71154a2e47/tests/test_diffusion.py#L146

erikvansebille:
Thanks for reporting this, @VeckoTheGecko! Very strange that the von mises now fails. Seems indeed to be related to the line you mention above. Where before the average of this `scipy_mises` list was `mu` (`=np.pi`), it is now closer to `0`. I have no idea why this doesn't work on the scipy version used in the Github Actions??

I'm not in favour of simply disabling the unit test; better to check what is going on and fix it where/how we can. I'll dig a bit deeper

PR #1275, opened 2023-01-06
Disable failing test

VeckoTheGecko:
Disable failing test (`test_randomvonmises`) reported in #1274 until it can be properly addressed.

### DIFF
--- a/tests/test_diffusion.py
+++ b/tests/test_diffusion.py
@@ -117,6 +117,7 @@ def test_randomexponential(mode, pset_mode, lambd, npart=1000):
     assert np.allclose(np.mean(depth), expected_mean, rtol=.1)
 
 
+@pytest.mark.xfail(reason="Test failing as reported in issue #1274")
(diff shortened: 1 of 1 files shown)


erikvansebille:
As I mentioned in https://github.com/OceanParcels/parcels/issues/1274#issuecomment-1373809656, I prefer to fix the actual unit test, rather than disabling it. Something is going wrong and we should try to understand/solve it?

VeckoTheGecko:
@erikvansebille Yes, I was just having this exact conversation in another forum where they said the same thing. Thinking about it, disabling tests in master just seems like bad practice 😬😅

VeckoTheGecko:
@erikvansebille another note, for me this workflow failed as well. This time on the the unittesting workflow for Windows. Weirdly there are [no logs that I can see](https://user-images.githubusercontent.com/36369090/211113082-14a6bdb6-1a6b-4248-8cef-0de518f54831.png). Have you encountered this before?

PR #1278, opened 2023-01-07
Clean up CI/CD workflows

VeckoTheGecko:
Rewrote workflows to be more concise, leveraging the [setup-miniconda](https://github.com/conda-incubator/setup-miniconda) GitHub Action.

Changes:
- Updated miniconda install to use `setup-miniconda`
  - Removed now obsolete scripts
- Separated linting into its own file
- Updated workflow triggers for linting and testing (adding `test-me/*` branch namespace for future testing of CI/CD)
- Generate HTML reports for tests, and save as artifacts


VeckoTheGecko:
In addition to the failing test in `master`, the mac tests aren't passing for `JIT` compilation due to the removal of this line:
https://github.com/OceanParcels/parcels/blob/56e988ae0f627ec8994c656f5f4bab71154a2e47/.github/scripts/osx_unittest.sh#L5

This can be hardcoded into the action by checking the platform, which I'm happy to do in the coming days.

VeckoTheGecko:
Testing is now working on mac. Ready for feedback (sorry for the noise)

erikvansebille:
Thanks @VeckoTheGecko, for this great improvement in the testing functionality!

A few quick comments:
- [ ] The [integration-tests](https://github.com/OceanParcels/parcels/actions/runs/3867092567/jobs/6591653447) doesn't seem to work yet?
- [ ] Are the three warnings at e.g. [Unittesting on ubuntu with python latest](https://github.com/OceanParcels/parcels/actions/runs/3867092563)
`Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/upload-artifact@v2` something to worry about? Something that can be fixed?
- [ ] There still seems to be a lot of code duplication between the unit-tests and the integration-tests; can that not be streamlined in some way?
- [ ] I like the report in html a lot! Didn't know that was possible. But any way to not have it be zipped before downloading? (so that it's even simpler to open it)

VeckoTheGecko:
> * [ ]  The [integration-tests](https://github.com/OceanParcels/parcels/actions/runs/3867092567/jobs/6591653447) doesn't seem to work yet?

Not sure what happened there. Seems like they were skipped for some reason. We can check again on the next run.

> * [ ]  Are the three warnings ...

Easy fix. Thanks for raising that

> * [ ]  There still seems to be a lot of code duplication between the unit-tests and the integration-tests; can that not be streamlined in some way?

Actually, yes! Just went down a bit of a rabbit hole into composite actions, and updated accordingly. It should be working, let me know what you think.

> * [ ]  I like the report in html a lot! Didn't know that was possible. But any way to not have it be zipped before downloading? (so that it's even simpler to open it)

All artifacts are zipped, so you can't just download plain HTML (or "open in new tab"). I know for coverage there are platforms to upload coverage reports. There could be similar platforms for testing, but I think it might not be worth it. Runs will hopefully only fail occasionally (and when they do, downloading the report isn't much overhead).

Speaking of coverage, I'll implement it sometime after this PR has merged.

---

On another note, I think the Windows builds are all failing without logs (even on PR #1273, prior to the workflow changes). Would it be worth disabling Windows builds? I'm not sure what the problem is ([relevant issue](https://github.com/actions/runner-images/issues/1776))

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  push:
+    branches:
+      - "master"
+      - "test-me/*"
+  pull_request:
+    branches:
+      - "*"
+  schedule:
+    - cron: "0 7 * * 1" # Run every Monday at 7:00 UTC
```
erikvansebille, line 11: Don't you need to specify here which branches need to run the cron-job? Only master, I assume?

VeckoTheGecko: Crons only run on `master`

.github/actions/install-parcels/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        auto-activate-base: false
+        auto-update-conda: true
+    - run: conda info
+      shell: bash -el {0}
+    - name: Install parcels
+      run: python setup.py install
+      shell: bash -el {0}
+    - name: Set env variables for macos
+      run: |
+        if [[ "${{ runner.os }}" == mac* ]]; then
```
VeckoTheGecko, line 31:
```suggestion
        if [[ "${{ runner.os }}" == macOS ]]; then
```
No need for wildcard match as values of `runner.os` are `macOS`, `Linux`, or `Windows` as per the [docs](https://docs.github.com/en/actions/learn-github-actions/variables#:~:text=example%2C%20Hosted%20Agent-,RUNNER_OS,-The%20operating%20system)

erikvansebille:
I've now fixed the vonmises issue in PR #1279, but don't get green ticks with Windows because of a [strange error](https://github.com/OceanParcels/parcels/actions/runs/3879215870/jobs/6624344113)
```
D:\a\parcels\parcels>call activate parcels 
[446](https://github.com/OceanParcels/parcels/actions/runs/3879215870/jobs/6624344113#step:3:447)

[447](https://github.com/OceanParcels/parcels/actions/runs/3879215870/jobs/6624344113#step:3:448)
EnvironmentNameNotFound: Could not find conda environment: parcels
[448](https://github.com/OceanParcels/parcels/actions/runs/3879215870/jobs/6624344113#step:3:449)
You can list all discoverable environments with `conda info --envs`.
```

@VeckoTheGecko, can you try to merge #1279 into this one, to see if it works then?

VeckoTheGecko:
Just merged, hopefully those tests pass now 🤞

I'm going to dial back my activity on this PR in favour of other areas of the codebase (debugging these workflows is doing my head in).

VeckoTheGecko:
Looks like the tests are still failing on Windows, and I don't know if this will be a simple fix. If the macos/linux tests pass, I think we should:

1. revert memory swap fix
1. remove Windows from testing matrix
1. create new issue for Windows CI/CD
1. merge this PR if everything else is in order

erikvansebille:
I've now fixed it in #1274 (with the solution at https://github.com/OceanParcels/parcels/pull/1279/commits/36cdbd3af36270505caf3a797ad02db0cef4ab90); can you implement that too here? That should give us green ticks for CI?

VeckoTheGecko:
Tests should pass now. If all is in order, feel free to squash and merge this (as most commits here are just noise).

PR #1283, opened 2023-01-22
Separate dev dependencies

VeckoTheGecko:
I am looking to include a variety of dev dependencies to improve the quality of this codebase. Eg:

- sphinx (+ related packages)
- pre-commit
- coverage
...

It would be good to isolate these dependencies so that they are not shipped to individual users' installation.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -38,6 +38,7 @@ jobs:
         with:
           environment-file: environment_py3_${{ matrix.os-short }}.yml
           environment-name: py3_parcels
+      - run: pip install -r requirements-dev.txt
       - name: Integration test
         run: |
           parcels_get_examples examples
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -38,7 +38,7 @@ jobs:
         with:
           environment-file: environment_py3_${{ matrix.os-short }}.yml
           environment-name: py3_parcels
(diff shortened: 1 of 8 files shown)


erikvansebille:
Good idea to create a separate requirements file for developers. I see you've now already moved pytest and flake there; what about sphinx? Or that will come later? Is this ready to merge now?

.github/workflows/integration-tests.yml
```diff
@@ -38,6 +38,7 @@ jobs:
         with:
           environment-file: environment_py3_${{ matrix.os-short }}.yml
           environment-name: py3_parcels
+      - run: pip install -r requirements-dev.txt
```
erikvansebille, line 41: The rest of install is doen through conda; why use pip here? Would it not be more consistent to use condo for these packages too? I have some poor experiences with mixing condo and pip...

VeckoTheGecko: It would be more consistent, but I'm afraid that not all development dependencies will be available through conda channels.

I doubt this would cause any issues as these are pure Python dependencies (and don't require compilers etc). Even if issues arise, it would only be for developers and not users

VeckoTheGecko:
@erikvansebille I've refactored to use Conda forge. I found surprisingly little online about specifying dev dependencies in Conda, so this is a bit of my own doing.

They can be installed by running `conda env update -f requirements-dev.yml` with the parcels environment activated.

PR #1290, opened 2023-01-25
Fix Readthedocs

### LINKED ISSUE #1288: Fix documentation website deployment
Currently it seems that there are two deployments of the documentation website:
- [OceanParcels website](https://oceanparcels.org/gh-pages/html/): Manually updated, complete
- [ReadTheDocs deployment](https://parcels.readthedocs.io/en/latest/): Automatically updated, broken (missing most documentation)

I think it would be better to just fix the ReadTheDocs implementation, and phase out the deployment to `.../gh-pages/html`

This would allow for not only auto-updating docs, but also looking at the documentation from any version (as opposed to just the latest version) so that people can see how the package has changed.

VeckoTheGecko:
Fixes #1288 

Deployment on my fork is visible here: https://veckoparcels.readthedocs.io/en/patch-rtd/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,7 @@
+version: 2
+build:
+  os: ubuntu-22.04
+  tools:
+    python: mambaforge-4.10
+conda:
+  environment: environment_py3_linux.yml
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,7 +19,7 @@ import shlex
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('../parcels'))
+sys.path.insert(0, os.path.abspath('..'))
 
 # -- General configuration ------------------------------------------------
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -198,17 +198,17 @@ parcels.plotting module
     :members:
     :undoc-members:
 
-scripts.plottrajectoriesfile module
+parcels.scripts.plottrajectoriesfile module
 -----------------------------------
 
-.. automodule:: scripts.plottrajectoriesfile
+.. automodule:: parcels.scripts.plottrajectoriesfile
     :members:
     :undoc-members:
 
-scripts.get_examples module
+parcels.scripts.get_examples module
 ---------------------------
 
-.. automodule:: scripts.get_examples
+.. automodule:: parcels.scripts.get_examples
     :members:
     :undoc-members:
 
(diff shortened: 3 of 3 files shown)


erikvansebille:
Thanks for this PR, @VeckoTheGecko. Your idea to automatically link through readthedocs.io is a good one! I forgot that we had a place for parcels there in the past...

One question; does the `.readthedocs.yaml` file need to be in the project root? Is there no way to put it in the `docs` folder and point the readthedocs server there? That would keep the project file tree a bit cleaner

VeckoTheGecko:
I had the same thought, but unfortunately [it has to be in the project root](https://docs.readthedocs.io/en/stable/config-file/v2.html#:~:text=The%20configuration%20file%20must%20be%20in%20the%20root%20directory%20of%20your%20project%20and%20be%20named%20.readthedocs.yaml.).

VeckoTheGecko:
Good to go!

VeckoTheGecko:
Looks like the [official deployment isn't updating](https://parcels.readthedocs.io/en/latest/). I've updated my fork to master and [that deployment is good](https://veckoparcels.readthedocs.io/en/latest/).

It would be good to check the config/builds on RTD.

erikvansebille:
I manually triggered the build, so the pages seem to now be updated. I'll have to check how to auto-build (via a hook?)

erikvansebille:
Hmm, and I now also see that the left side panel (with the table of content etc) doesn't work at https://parcels.readthedocs.io/en/latest/#. Did you do anything special to get that working on your fork, @VeckoTheGecko?

VeckoTheGecko:
Strange. Are there any errors visible in the build logs?

I didn't do any additional configuration on my side outside of this PR.

I don't have hooks/automation rules, though I do [have an integration](https://user-images.githubusercontent.com/36369090/215091630-5a1f08a2-022a-4675-9567-6ec474ab9405.png) (which I think was added by default on setup). Perhaps that could be it?

erikvansebille:
I haven't changed anything, but the disbar seems to work now. So all good after all. I just also updated the links on oceanparcels.org, to link directly to https://parcels.readthedocs.io/

VeckoTheGecko:
Super minor detail, but might be worth disabling prior versions listed [here](https://user-images.githubusercontent.com/36369090/215371824-8285df3c-aec1-4cdf-a617-fd6e265c378e.png) as the deployment is broken on those tags.

I don't think its worth merging this branch into those tags, but just to disable those builds in RTD and use the tags for versions moving forward.

erikvansebille:
> I don't think its worth merging this branch into those tags, but just to disable those builds in RTD and use the tags for versions moving forward.

Good point, I now removed all these older versions for which the RTD integration was not working yet

Issue #1285, opened 2023-01-22
Pre-commit hooks


VeckoTheGecko:
Pre-commit allows for the automated running of linting workflows before a commit is finalised, even applying automatic fixes. This is particularly useful for linting, so that linting errors can be corrected before a commit is even made (hence reduced reliance on the linting action to pickup linting errors).

I propose the following hooks as a baseline (a list of popular hooks are available [here](https://github.com/pre-commit/pre-commit-hooks)):
- check-yaml (verify yaml syntax)
- end-of-file-fixer (adds newline to end of files)
- trailing-whitespace (trims trailing whitespace)
- black (autoformats python files, related to #1284)

VeckoTheGecko:
Pre-commit is "opt-in" by default, meaning that code quality errors can slip through if someone isn't using pre-commit. [Pre-commit CI](https://pre-commit.ci/) prevents this by using the pre-commit config to run CI on the codebase. Precommit runs on pull requests, and will automatically make commits to fix issues.

erikvansebille:
Good idea, @VeckoTheGecko. Some of these are already implemented in my own IDE (I use PyCharm). 

But I don't entirely understand why you open an Issue about this. What is broken in Parcels itself and/or what do we need to change?

PR #1292, opened 2023-01-25
Revert separating dev reqs

VeckoTheGecko:
Revert #1283 as per #1289

### DIFF
--- a/environment_py3_linux.yml
+++ b/environment_py3_linux.yml
@@ -5,9 +5,7 @@ channels:
 dependencies:
   - python>=3.8
   - cgen
-  - coverage
   - ffmpeg>=3.2.3
-  - flake8>=2.1.0
   - git
   - gcc_linux-64
   - mpi4py>=3.0.1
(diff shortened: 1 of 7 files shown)


environment_py3_linux.yml
```diff
@@ -28,3 +26,15 @@ dependencies:
...
+  - pytest
+  - pytest-html
+  - coverage
+  
+  # Linting
+  - flake8>=2.1.0
+
+  # Docs
+  - sphinx<6
+
```
erikvansebille, line 40: Strange End of Line (End of File) character here? (Also in the other three files). Do you know what's going on?

VeckoTheGecko: Its used by GitHub to denote files that don't end with a "\n" character. There was some whitespace at the end of the file left over from my IDE. I've fixed this now.

This is also partly my motivation for #1285 (which I'm using for my local development anyway now).

[[StackOverflow] Why text files should end with newline](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline), [[StackOverflow] What's the significance of the "No newline at end of file" log?](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log)

PR #1303, opened 2023-02-05
Allow for Zarr stores being passed to ParticleFile

willirath:
Let's allow for directly passing a `zarr.storage.Store` object to `ParticleFile`. This will allow for a couple of advanced output strategies:

- diskless capture of the output into a [`MemoryStore`](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.MemoryStore)
- streaming data to a Redis store using [`RedisStore`](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.RedisStore)
- ~~capturing output in a zip file using a [`ZipZtore`](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.ZipStore)~~ (Possible but inefficient and throwing lots of warnings.)
- etc.

Todos:
- [x] make sure `ParticleFile` works for Zarr stores directly.
- [x] document changes
- [x] add tutorial / example notebook with some of the above stores


parcels/particlefile/baseparticlefile.py
```diff
@@ -93,16 +93,22 @@ def __init__(self, name, particleset, outputdt=np.infty, chunks=None, write_onde
...
-        if extension in ['.nc', '.nc4']:
-            raise RuntimeError('Output in NetCDF is not supported anymore. Use .zarr extension for ParticleFile name.')
-        if MPI and MPI.COMM_WORLD.Get_size() > 1:
-            self.fname = os.path.join(name, f"proc{self.mpi_rank:02d}.zarr")
-            if extension in ['.zarr']:
-                logger.warning(f'The ParticleFile name contains .zarr extension, but zarr files will be written per processor in MPI mode at {self.fname}')
+        if issubclass(type(name), zarr.storage.Store):
+            # If we already got a Zarr store, we won't need any of the naming logic below.
+            # But we need to handle incompatibility with MPI mode for now:
+            if MPI:
```
erikvansebille, line 99: Also need to check if we're actually in MPI mode (i.e. more than one processor)
```suggestion
            if MPI and MPI.COMM_WORLD.Get_size() > 1:
```

erikvansebille:
Thanks for this PR, @willirath, it looks very good! 

Apart from the small change proposed above on the MPI-handling, my only comment would be to rename the `tutorial_zarr_advanced` notebook to `documentation_advanced_zarr`. In Parcels, documentations are typically more advanced and/or for background, and they are not integration-tested. Since you already have a unit-test too (thanks!), I'm not sure integration testing is needed too. Or do you think that adds value?

Also, I'm pinging @JamiePringle, so he can have a quick look at this new feature too

willirath:
@erikvansebille I've renamed to notebook as suggested and also fixed the MPI check.

Re this also being a possible integration test: I think it wouldn't hurt as it also does not rely on any input data (but produces some output) and is relatively quick (30 seconds in total?). But seeing the already substantial time spent on automated tests, I can also see that it makes sense not to run everything.

erikvansebille (approved):

PR #1321, opened 2023-02-28
Documentation: flake8-docstrings, theme migration, minor theme features

VeckoTheGecko:
## Implemented improvements
- [x] Migrate to PyData Sphinx theme (this is the theme used by PyData projects (pandas, numpy, etc). It is modern, familiar, and open source/free to use).
- [x] flake8 for docstrings (ignoring certain codes, and fixing to comply with others. See codes [here](http://www.pydocstyle.org/en/stable/error_codes.html))
- [x] Have `[source]` links on the documentation website link directly to lines in the GitHub repo
- [x] Have the version number in docs be autopopulated from `parcels.__version__`

Part of #1324


VeckoTheGecko:
@erikvansebille I installed `flake8-docstrings` locally and ran `flake8`. This is a list of the unique `D*` error codes that were raised:
| code   | description                                                      |
|:-------|:-----------------------------------------------------------------|
| D100   | Missing docstring in public module                               |
| D101   | Missing docstring in public class                                |
| D102   | Missing docstring in public method                               |
| D103   | Missing docstring in public function                             |
| D104   | Missing docstring in public package                              |
| D105   | Missing docstring in magic method                                |
| D107   | Missing docstring in __init__                                    |
| D200   | One-line docstring should fit on one line with quotes            |
| D202   | No blank lines allowed after function docstring                  |
| D204   | 1 blank line required after class docstring                      |
| D205   | 1 blank line required between summary line and description       |
| D207   | Docstring is under-indented                                      |
| D208   | Docstring is over-indented                                       |
| D209   | Multi-line docstring closing quotes should be on a separate line |
| D210   | No whitespaces allowed surrounding docstring text                |
| D300   | Use “””triple double quotes”””                                   |
| D400   | First line should end with a period                              |
| D401   | First line should be in imperative mood                          |
| D401   | First line should be in imperative mood; try rephrasing          |
| D403   | First word of the first line should be properly capitalized      |

It would be good to discuss which codes:
- You want to ignore outright for the codebase
- You want to ignore for now, (implementing fixes in the future to remove the ignore statement)
- You want me to patch in this PR :)

You can use `flake8 | grep "D300"` to filter down the output to see where (file+line) in the code the different error codes are raised.

erikvansebille:
Wow, thanks so much for opening this PR @VeckoTheGecko! I really like the new these at https://veckoparcels.readthedocs.io/en/docs/; much cleaner indeed.

And the flak8 results are rather overwhelming. `flake8 | wc -l` returns 1850(!) flake8 issues in the docstrings...

I'm not are there are any codes that we want to ignore outright. Especially if we want clean and readable documentation; then following these suggestions might help a lot. I can imagine that the `D2??`, `D300`, `D400` and `D403` codes are the easiest to fix now; and that the `D1??` and `D401` codes require more work in (re)writing the actual text. 

So perhaps if we (temporarily) ignore the `D1??` and `D401` codes for now, and work on all others? Does that sound like a plan?

VeckoTheGecko:
I think with some of the "easy to fix" codes, there will be edge cases which would require intervention. (eg. I've been working through D400. A bit non-trivial since a lot of methods don't have the first line as a summary. I wrote a summary for a couple, before giving up and kicking the can down the road.)

Also as a general code quality thing, does `parcels` consistently use the convention of a leading `_` to denote private functions? I saw docstrings like [this](https://github.com/VeckoTheGecko/parcels/blob/67367b9e2c8056302227ed95211322579af2dc27/parcels/fieldfilebuffer.py#L398-L403) saying `[private function - not to be called from outside the class]`, which seemed to be doubling up in what they're saying.

I guess as a research package, maybe the lines between private and public functions blur. Just wanting to know what the go with that is. If a lot of private functions are missing the leading `_`, it will also affect the amount of docstrings needing to be written to satisfy `D100-D104`.

I'll continue working on this PR on the weekend :)

VeckoTheGecko:
Also let me know if you want me to break this PR into multiple. Just realising that it might be a nightmare to review otherwise (mixing minor changes, with more major changes).

erikvansebille:
Thanks you so much for starting to fix some of the flake8 issues with the docstrings, @VeckoTheGecko; this is invaluable work! To answer your questions above one by one

> I think with some of the "easy to fix" codes, there will be edge cases which would require intervention. (eg. I've been working through D400. A bit non-trivial since a lot of methods don't have the first line as a summary. I wrote a summary for a couple, before giving up and kicking the can down the road.)

Then I suggest you leave these open, and I can fix these myself later


> Also as a general code quality thing, does parcels consistently use the convention of a leading _ to denote private functions? I saw docstrings like [this](https://github.com/VeckoTheGecko/parcels/blob/67367b9e2c8056302227ed95211322579af2dc27/parcels/fieldfilebuffer.py#L398-L403) saying [private function - not to be called from outside the class], which seemed to be doubling up in what they're saying.

Hmm, we ought to follow private and public functions with `_`. Whenever you encounter the `[private function - not to be called from outside the class]`, I suggest you can prepend a `_` to the function name. Or do you want me to do that?

> Also let me know if you want me to break this PR into multiple. Just realising that it might be a headache for you to review otherwise (mixing minor changes alongside more major changes).

I wouldn't worry too much about that. I'll keep track of your individual commits and check them as they come in. In the end, we'll have to review all changes anyways. So no need from my side to separate into multiple PRs.

VeckoTheGecko:
All good.

> Or do you want me to do that?

No need 😄 . The instances I saw already had a leading `_`, so I'll just remove the comment.

erikvansebille:
> The instances I saw already had a leading _, so I'll just remove the comment.

Ah, now I understand your question. Yes, if the function already starts with a `_`, then it's redundant to also put that message in the docstring. So you can remove the messages

VeckoTheGecko:
Ok. Done with all the easy fixes. These are the errors I'm ignoring for now (until someone can come past and write the docstrings/summaries).

https://github.com/OceanParcels/parcels/blob/d223534bc2e820a8ce34ee73515057093c3c80b8/setup.cfg#L33-L53

parcels/collection/collections.py
```diff
@@ -789,8 +786,8 @@ class ParticleCollection(Collection):
     _data = None  # formerly: particle_data
 
     def __init__(self):
-        """
-        ParticleCollection - Constructor
+        """Constructor for ParticleCollection.
+
```
VeckoTheGecko, line 790: Not sure about this. The capitalisation of the class name was making D403 act up, but this rewording is now inconsistent with other parts. Same for the edit just below

erikvansebille: Hmm, I don't understand the problem here. This passes `flake8` now, right? But you would want to revert it? Seems fine to me

VeckoTheGecko: Passes with flake8, just now its inconsistent with:
https://github.com/VeckoTheGecko/parcels/blob/0ee5bd2f82cb25650560af9b59eceb575d147620/parcels/collection/collections.py#L22

https://github.com/VeckoTheGecko/parcels/blob/0ee5bd2f82cb25650560af9b59eceb575d147620/parcels/collection/collections.py#L31

https://github.com/VeckoTheGecko/parcels/blob/0ee5bd2f82cb25650560af9b59eceb575d147620/parcels/collection/collectionaos.py#L151

https://github.com/VeckoTheGecko/parcels/blob/0ee5bd2f82cb25650560af9b59eceb575d147620/parcels/collection/collectionsoa.py#L165

erikvansebille: OK I see. I think that at some point we'll have to carefully go through the docstrings again anyway, and then we can homogenise this inconsistency. It doesn't bother me that much.

erikvansebille:
Thanks for making all these changes @VeckoTheGecko, a Herculean effort! Is this PR ready for review now? If so, can you convert from draft status and merge the latest changes on master?

VeckoTheGecko:
The docs are still a WIP as I have more improvements/bugfixes I would like to make. It would be good to release the revamped docs all at once to the community when it's more stable (rather than incrementally, potentially messing with people's workflows if they rely on the docs website deployed from `master` for their work).

Would it be more appropriate to target this merge into a `new-docs` branch on `parcels`? That way I can incrementally open/review/merge more PRs into `new-docs` (implementing type annotations/other changes), and then we can merge that into `master` once the docs are more stable. Also you can enable the build in readthedocs for the `new-docs` branch if people want to use while it's still WIP.

That would also be nicer since I'd like to take care when reviewing (especially) the type annotations (including them in this PR is a bit of a nightmare for me to review with the flake stuff soiling the diffs 😅).

parcels/collection/collections.py
VeckoTheGecko: Easy, that works. It was a minor point anyway :)

erikvansebille:
Yep, very good idea to create a new branch `new_docs`. I just created it; see https://github.com/OceanParcels/parcels/tree/new_docs

PR #1328, opened 2023-03-04
Upgrade syntax to Python 3.8

VeckoTheGecko:
Upgrade syntax to Python 3.8 project-wide using [`pyupgrade`](https://github.com/asottile/pyupgrade) (including f-strings!).


VeckoTheGecko:
I've separated the changes into 3 commits:
- [`a75e15c`](https://github.com/OceanParcels/parcels/pull/1328/commits/a75e15c05c49519792fb25110dbd278dfc8c817f) Is just running `pyupgrade` with no manual edits.
- [`e67292b`](https://github.com/OceanParcels/parcels/pull/1328/commits/e67292bda66d06b37fcd58ba94c7775ddac90b8e) I manually edit the remaining old f-strings to work with `pyupgrade`, or otherwise edit them before rerunning.
- [`d962a43`](https://github.com/OceanParcels/parcels/pull/1328/commits/d962a434d8ec514137871b791e99ef8ca9a824aa) Handling some notebook f-strings, and `%d` fstrings.

`a75e15c` is OK. `e67292b` and `d962a43` would require more focus.

VeckoTheGecko:
Some Python 2 format strings were left in because it was unclear whether converting them would break logic (mainly in `codegenerator.py`).

VeckoTheGecko:
In this PR I've heavily relied on `pyupgrade` to do the conversion to reduce human error potentially introducing bugs. Hence its not a 100% migration (in my opinion a 100% migration isn't worth the time to convert + test these edge cases, especially since the code is already working).

## The commits
- https://github.com/OceanParcels/parcels/pull/1328/commits/b8a84b268de087db0467e5751e8dfc6bbe261e2b Run pyupgrade with no manual intervention (100% OK)
- https://github.com/OceanParcels/parcels/pull/1328/commits/8bef7fe76186bcc45f2602ed8d5762c0cc2f891e Run pyupgrade with some minor tweaks to help it along (only done for print statements, log messages, or unambiguous areas to avoid introducing bugs)
- https://github.com/OceanParcels/parcels/pull/1328/commits/0f94fec99ac0b32c1d5065ba20641fd9a8a18972 Convert example notebooks to only use f-strings

parcels/fieldfilebuffer.py
```diff
@@ -72,7 +72,7 @@ def parse_name(self, name):
                     name = nm
                     break
         if isinstance(name, list):
-            raise IOError('None of variables in list found in file')
+            raise OSError('None of variables in list found in file')
```
erikvansebille, line 75: Why does IOError not exist anymore? This error really is about the input files, not about the OS?

parcels/fieldfilebuffer.py
```diff
@@ -286,8 +286,7 @@ def __enter__(self):
                 self.dataset = xr.open_dataset(str(self.filename), decode_cf=True, engine=self.netcdf_engine, chunks=init_chunk_dict, lock=False)
             self.dataset['decoded'] = True
         except:
-            logger.warning_once("File %s could not be decoded properly by xarray (version %s).\n         It will be opened with no decoding. Filling values might be wrongly parsed."
-                                % (self.filename, xr.__version__))
+            logger.warning_once(f"File {self.filename} could not be decoded properly by xarray (version {xr.__version__}).\n         It will be opened with no decoding. Filling values might be wrongly parsed.")
```
erikvansebille, line 289: Also remove the line break here?
```suggestion
            logger.warning_once(f"File {self.filename} could not be decoded properly by xarray (version {xr.__version__}). It will be opened with no decoding. Filling values might be wrongly parsed.")
```

parcels/particlefile/baseparticlefile.py
```diff
@@ -17,7 +17,7 @@
 try:
     from parcels._version import version as parcels_version
 except:
-    raise EnvironmentError('Parcels version can not be retrieved. Have you run ''python setup.py install''?')
+    raise OSError('Parcels version can not be retrieved. Have you run ''python setup.py install''?')
```
erikvansebille, line 20: Does EnvironmentError not exist anymore? This really is an issue with the Environment!

parcels/rng.py
```diff
@@ -111,7 +111,7 @@ def compile(self, compiler=None):
             with open(self.src_file, 'w+') as f:
                 f.write(self.ccode)
             ccompiler.compile(self.src_file, self.lib_file, self.log_file)
-            logger.info("Compiled %s ==> %s" % ("ParcelsRandom", self.lib_file))
+            logger.info("Compiled {} ==> {}".format("ParcelsRandom", self.lib_file))
```
erikvansebille, line 114: Why not an f-string here?

parcels/scripts/get_examples.py
```diff
@@ -76,7 +76,7 @@ def _still_to_download(file_names, target_path):
 def download_files(source_url, file_names, target_path):
     """Mirror file_names from source_url to target_path."""
     _maybe_create_dir(target_path)
-    print("Downloading %s ..." % (source_url.split("/")[-1]))
+    print("Downloading {} ...".format(source_url.split("/")[-1]))
```
erikvansebille, line 79: Why not an f-string here?

parcels/tools/statuscodes.py
```diff
@@ -44,10 +44,10 @@ def __init__(self, x, y, z, field=None):
         self.x = x
         self.y = y
         self.z = z
-        message = "%s sampled at (%f, %f, %f)" % (
+        message = "{} sampled at ({:f}, {:f}, {:f})".format(
```
erikvansebille, line 47: Why not an f-string? (same for other messages in this file)

tests/test_kernel_language.py
```diff
@@ -337,8 +337,8 @@ class TestParticle(ptype[mode]):
                                         lat=np.zeros(npart) + 0.5)
     series = random_series(npart, rngfunc, rngargs, mode)
     rnglib = 'ParcelsRandom' if mode == 'jit' else 'random'
-    kernel = expr_kernel('TestRandom_%s' % rngfunc, pset,
-                         '%s.%s(%s)' % (rnglib, rngfunc, ', '.join([str(a) for a in rngargs])), pset_mode)
+    kernel = expr_kernel(f'TestRandom_{rngfunc}', pset,
+                         '{}.{}({})'.format(rnglib, rngfunc, ', '.join([str(a) for a in rngargs])), pset_mode)
```
erikvansebille, line 341: Why not an f-string?

erikvansebille:
> Good to go (if no other feedback).

I just reviewed the files, and added a few comments and questions. Can you address these?

> PS. Don't know why the Integration Tests on this branch show "skipped". My [`test-me/test`](https://github.com/VeckoTheGecko/parcels/tree/test-me/test) branch is a duplicate and has all green ticks (except Windows integration test, which seems to be flakey for some reason)

Hmm, it could be because you triggered it from your fork, perhaps before you were part of the organisation? All tests seem to run now, after I updated to latest master

parcels/fieldfilebuffer.py
VeckoTheGecko: As of Python 3.3, [PEP3151](https://docs.python.org/3/whatsnew/3.3.html#pep-3151) changed such that `IOError` and `EnvironmentError` are only aliases for `OSError`. They are only supported for backwards compatability, keeping it in the code doesn't provide more informative error messages.

```
Python 3.10.6 | packaged by conda-forge | (main, Aug 22 2022, 20:36:39) [GCC 10.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> raise EnvironmentError("Invalid environment!")
  File "<stdin>", line 1, in <module>
OSError: Invalid environment!
>>>
```

parcels/particlefile/baseparticlefile.py
VeckoTheGecko: See my other https://github.com/OceanParcels/parcels/pull/1328#discussion_r1127736973

VeckoTheGecko:
> it could be because you triggered it from your fork

Hmm maybe. Really not sure

---
Re. the f-string comments, any instances where it hasn't converted to an f-string has been due to limitations of `pyupgrade`. `pyupgrade` tries to be safe in the way it operates, so doesn't really dig deep to take care of edge cases.

eg. It won't format `"{x}".format("HELLO".lower())`, as both strings have double quotes so it can't directly sub in.

I can manually upgrade these to f-strings, but I just didn't want to risk introducing bugs (eg. by forgetting an `f`) as I won't really be running the code where I'm changing to f-strings.

erikvansebille:
> EDIT: Would you like me to manually upgrade to f-strings too as part of this PR?

Ah, I only now see your update. I've just spent 15 minutes updating the f-strings myself, in https://github.com/OceanParcels/parcels/pull/1328/commits/92fd8e5d217b8121a8c43a2c17486df4582cb86e. I think I caught them all...

Issue #1324, opened 2023-03-03
Documentation revamp


VeckoTheGecko:
I'll use this issue to outline the ideas I have for the documentation, so that discussion can be centralised here with PRs linking back to this. Any of these points are up for discussion. I'm happy to work on all of these (although may need some additional help with config). If anyone else would like to contribute to a bulletpoint, drop a comment 🙂 (although there's no expectation on my end).

As these changes get implemented, I will edit this description adding the PR numbers to the bullet points.

# Changes
## Major changes
- [x] Documentation restructure
  - Add sectioning to the documentation, breaking apart the documentation into its conceptual components so that it isn't one large page.
- [x] Migrate to PyData theme (#1321)
  - The PyData theme is a much cleaner theme, with more features.
- [x] Create contribution guidelines for `parcels`
  - Having guidelines that contributors can follow would be good to communicate the standards adopted by the codebase.
  - The contributing [guidelines from geopandas](https://geopandas.org/en/stable/community/contributing.html) might be good to copy and slightly adapt to our purposes (ie. we only support Python 3.8+ (@erikvansebille it would be good to confirm this, I'm going off the `environment_*.yml` files), we don't use `black`, and [commit message conventions](https://geopandas.org/en/stable/community/contributing.html#commit-message-conventions) might be a bit extra for the size of this project).
  - The contribution guidelines can be its own page on the documentation website.
  - Perhaps [oceanparcels.org](https://oceanparcels.org) can be updated to link to the contributing page for dev installs.
- [x] Use [`flake8-docstrings`](https://pypi.org/project/flake8-docstrings/) to monitor docstring quality (#1321)
- [x] Migrate docstring format to the [Numpy Docstring Standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)
  - The `rst` format is quite difficult to read when displayed in IDEs.
  - Due to this ^, the Numpy Docstring Standard has been adopted by many other scientific packages (numpy, pandas, geopandas, xarray, to name a few).
  - Existing docstrings can be easily converted from `rst` format into Numpy format using [`pyment`](https://pypi.org/project/pyment/).
  - Compiling and [validation](https://numpydoc.readthedocs.io/en/latest/validation.html) of Numpy docstrings is possible in the build process through Sphinx with the `numpydoc` package. This validation is in extension to `flake8-docstrings`, making sure that the docstring is in the Numpy standard format.
[...]

Issue #1335, opened 2023-03-12
Integration tests being skipped


VeckoTheGecko:
In some of the workflow runs the integration tests are outright skipped, providing no error logs.

I've never seen this issue before in other codebases. Its cropped up in this repo for a couple months, most recently in #1328 (relevant comments [1](https://github.com/OceanParcels/parcels/pull/1328#issuecomment-1457842321) and [2](https://github.com/OceanParcels/parcels/pull/1328#issuecomment-1458008488)) and in #1331 ([screenshot](https://user-images.githubusercontent.com/36369090/224548747-7082a83b-a4f0-4409-9afb-1d3c0b1392a6.png) of associated [job run](https://github.com/OceanParcels/parcels/actions/runs/4387936468/jobs/7683829540)).

Its occuring on forks, as well as on internal repo branches.

VeckoTheGecko:
Whelp. Turns out I just completely missed the `if: github.event.pull_request.draft == false` line in the integration test workflow

PR #1341, opened 2023-03-26
DOC: Numpy docstrings migration

VeckoTheGecko:
This PR focusses on the following features (in order of implementation):
- [x] Initial conversion of `rst` docstrings to `numpydoc` format (with the help of pyment)
- [x] `numpydoc` docstring validators to ensure docstring quality
  - [x] Install `numpydoc` docstring validators as part of the Sphinx build process
  - [x] Ensure docstrings are compliant with chosen validators
  - [x] Tying in docstring validation with CI/CD
- [x] Add types to docstring parameters (particularly focussing on popular modules).

PR contributes to #1324, where you can read more on this featureset.

Draft PR for visibility (and to trigger RTD builds).

(given the size of these diffs, let me know if you want me to break into multiple PRs)

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -74,7 +74,7 @@ class FieldSet:
                (e.g. dimensions['U'], dimensions['V'], etc).
         :param transpose: Boolean whether to transpose data on read-in
         :param mesh: String indicating the type of mesh coordinates and
-               units used during velocity interpolation, see also `this tutorial <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_unitconverters.ipynb>`_:
+               units used during velocity interpolation, see also `this tutorial <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_unitconverters.ipynb>`__:
 
                1. spherical (default): Lat and lon in degree, with a
                   correction for zonal velocity U near the poles.
@@ -88,13 +88,13 @@ class FieldSet:
         Usage examples
         ==============
 
-        * `Analytical advection <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_analyticaladvection.ipynb>`_
+        * `Analytical advection <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_analyticaladvection.ipynb>`__
 
-        * `Diffusion <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_diffusion.ipynb>`_
+        * `Diffusion <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_diffusion.ipynb>`__
 
-        * `Interpolation <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_interpolation.ipynb>`_
+        * `Interpolation <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_interpolation.ipynb>`__
 
-        * `Unit converters <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_unitconverters.ipynb>`_
+        * `Unit converters <https://nbviewer.jupyter.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_unitconverters.ipynb>`__
 
         """
         fields = {}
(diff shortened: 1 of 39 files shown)


VeckoTheGecko:
Done with initial conversion of docstrings to numpy format. Now moving onto docstring validation.

VeckoTheGecko:
Numpydoc validators have been included in the build process (basically linting through docstrings, looking for things like "is the argument in the docstring the same as the one in the function call", ensuring the docstring is rendered correctly).

The full list of possible validation checks are outlined in the [error codes](https://numpydoc.readthedocs.io/en/latest/validation.html). Some of the codes are incompatible with the codebase (either outright, or due to the state of many docstrings). Implementing all these codes is a very very low priority, but the small subset of codes at the moment are a good starting point (and additional codes can be included easily down the line if desired).

VeckoTheGecko:
Just confirming @erikvansebille , is this option checked in the `Advanced Settings` section of the RTD parcels project panel?
image

erikvansebille:
> Just confirming @erikvansebille , is this option checked in the `Advanced Settings` section of the RTD parcels project panel?

Yep, that checkbox is on (see below), but the default branch is not set. Should I do that?

Screenshot 2023-03-27 at 19 12 38

VeckoTheGecko:
Following on from [this (10 year old) conversation](https://github.com/readthedocs/readthedocs.org/issues/239), it doesn't look like builds are triggered from PRs originating from branches on forks. No matter though. My RTD build for this branch is available [here](https://github.com/readthedocs/readthedocs.org/issues/239). We'll review properly when merging `new_docs` -> `master`.


> the default branch is not set. Should I do that?
No need 😄

erikvansebille:
Thanks for digging into this

> My RTD build for this branch is available [here](https://github.com/readthedocs/readthedocs.org/issues/239). We'll review properly when merging `new_docs` -> `master`.

That link seems to point to the RTD Issue? Can you update to the correct link?

VeckoTheGecko:
Opps, yes, just updated

VeckoTheGecko:
Used the regex `\s:$` to search across the codebase for parameters with missing types. Added many types, will leave the rest to be gradually done over a longer timeframe.

PR ready for feedback 😄

parcels/fieldset.py
```diff
@@ -426,62 +482,76 @@ def from_nemo(cls, filenames, variables, dimensions, indices=None, mesh='spheric
...
+            Watch out: NEMO is discretised on a C-grid:
+            U and V velocities are not located on the same nodes (see https://www.nemo-ocean.eu/doc/node19.html ).
+
+            +-----------------------------+-----------------------------+-----------------------------+
+            |                             |         V[k,j+1,i+1]        |                             |
+            +-----------------------------+-----------------------------+-----------------------------+
+            |U[k,j+1,i]                   |W[k:k+2,j+1,i+1],T[k,j+1,i+1]|U[k,j+1,i+1]                 |
+            +-----------------------------+-----------------------------+-----------------------------+
+            |                             |         V[k,j,i+1]          +                             |
+            +-----------------------------+-----------------------------+-----------------------------+
```
erikvansebille, line 519: These tables don't show as nice as they do in the old version? Compare https://veckoparcels.readthedocs.io/en/docstrings/#parcels.fieldset.FieldSet.from_c_grid_dataset to https://parcels.readthedocs.io/en/latest/#parcels.fieldset.FieldSet.from_c_grid_dataset

VeckoTheGecko:
> The title in your https://veckoparcels.readthedocs.io/en/docstrings/ version mentions `v2.3.0`; will that be the latest after we merge?

As of 398ba3b the version number is determined automatically using the `parcels.__version__` attribute. `new_docs` (the target of this PR) is at `v2.3.0`. When we re-sync and merge `new_docs` into `master` it will be at the intended version.

> The copyright statement on the page bottom could/should be updated to 2023?

Can do. Should we dynamically determine the year for the copyright statement? i.e. update the copyright statement from:

```py
copyright = '2022, The OceanParcels Team'
```
to
```py
copyright = f"{datetime.datetime.now().year}, The OceanParcels Team"
```

Happy to fix this in a separate PR.

> Should we add these commits (especially the really large ones) to the `.git-blame-ignore-revs` file too?

I don't think so. The purpose of `.git-blame-ignore-revs` is to ignore (large) *superfluous* changes that in the end have no effect on the function of the code (eg. introducing black, introducing isort, whitespace. These are mainly created by automated formatting tools) and would otherwise soil the blame. These commits mark a shift in how the codebase works with docstrings, and introduce additional edits to the docstrings, which are worth keeping in blame.

erikvansebille:
OK is fine, let's then just go for it. The integration tests have skipped (not sure why), but I guess we can merge without these in this case, because only docstrings are affected in this PR?

VeckoTheGecko:
> The integration tests have skipped (not sure why)

#1335 rearing its head again 😕

And yes, only docstrings/doc related changes.

PR #1353, opened 2023-04-28
Moving tutorial and documentation notebooks to the new documentation

erikvansebille:
Now that we have a [new shiny documentation](https://parcels.readthedocs.io/en/latest/), it might be an idea to move some (all?) of the tutorial and documentation notebooks to the new readthedocs page. This is mostly a matter of exporting as markdown, but links don't often work so need to be double-checked, and heading levels are sometimes messed up.

Furthermore, we need to decide if we still also want to CI (some) of the notebooks. I think there are tools for that, but needs to be explored how well they work. Note that the documentation-notebooks are not CI-ed at the moment anyways.

Currently, this is just a proposal. Look forward to hearing what people think!


VeckoTheGecko:
Missed this (my notifications weren't set for PRs before 😁). Just wondering about the motivation for this, since the tutorial notebooks/documentation is linked on the main OceanParcels website.

1. Is it to consolidate (having all the documentation in one place)?
1. Is it to make them easier to browse in a website-like format as opposed to nbviewer?
1. Is it to reduce the reliance on the OceanParcels website in general?

Also, what impact does this have on Google analytics?

If we compile the notebooks to md for the docs site, it should be done in CI during building the docs (I'm pretty sure there Sphinx tooling for this). I'm opposed to commiting md files for notebooks as that introduces duplication making updating notebooks a headache.

Also regarding link checking (i.e. checking 404s and other errors), that can be automated using a Sphinx extension I recently learned about. Happy to open a PR for that.

Let me know if you want me to help out with this

erikvansebille:
Thanks for the feedback, @VeckoTheGecko; you were right that I didn't clearly explain my motivation for this move. Of your three options, the first and second are indeed correct; I'm not sure about reducing reliance on the oceanparcels.org domain (which I do like to keep as the landing place for info on Parcels)

My main motivation was to bring the Parcels documentation more in line with documentation style of other packages such as array; where it seems that tutorials are often also on rtd? The styling is then nicer and more flexible. 

An added advantage is that removing notebooks would also change the 'Languages' summary in the right-hand column of GitHub; which now suggests that Parcels is 93.5% Jupyter Notebook. Although that may not be a really big issue

I can also see some drawbacks. Exposure for oceanparcels.org on Google might take a hit if more content is on RTD?

My original idea was to remove the `ipynb` files and replace them all with `md` files; so we wouldn't have duplicates. I didn't know of the Sphinx-notebook extension; that might be a good solution?! And link checking would also be very nice to implement yes. 

SO yes, a little help is appreciated. I started this PR purely to see what a RTD-tutorial would look like; haven't completely made up my mind if we want to go this way. If you're happy to help with trying the ipynb-in-rtd workflow, that would be interesting to test?

VeckoTheGecko:
> I can also see some drawbacks. Exposure for oceanparcels.org on Google might take a hit if more content is on RTD?

I'm not sure how SEO works, or what keywords people use to get to parcels (I imagine anyone Googling "oceanparcels" will still be led to the right place 😄 ). My comment was more about the Google Analytics linktracking that exists on the oceanparcels website currently (allowing you to see usage stats on a per-notebook basis from the clicking of the cards). Migrating to RTD would mean that traffic will be split between the tracked ones on the oceanparcels website (if they're still going to exist there) and the ones in RTD. Although RTD does [have its own traffic analysis solution](https://docs.readthedocs.io/en/stable/analytics.html) if that is something the OP team is interested in collecting.

> My original idea was to remove the `ipynb` files and replace them all with `md` files; so we wouldn't have duplicates

I see. I think the best would be to leave the `ipynb` notebooks and have better integration for them into the RTD site. Having the `ipynb` notebooks is really handy because it has that level of interactivity that you can't get from `md` files. Including the ipynb files in the RTD site (using automated tooling to convert the ipynb file to a webpage) will have the output cells shown properly (even with interactive elements in the case of xarray/pandas widgets). The [pandas link](https://pandas.pydata.org/docs/user_guide/style.html) I sent above is a good illustration of what it could look like.

Regarding the language summary, GitHub uses Linguist to determine that, and it can be [manually overriden](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) to ignore certain files and directories. I think its a simple matter of creating a `.gitattributes` file with:
```
parcels/examples/* linguist-documentation
```
which will mark these files as docs, and hence ignore them in the summary.

Happy to work on both these in a PR, and then you can see if its preferable.

> link checking would also be very nice to implement

Can do.

erikvansebille:
Oh wow, you can do so much more than I expected! Yes, RTD on a custom domain sounds like a great option. And so does the linguist-documentation solution. The traffic analysis is not a deal-breaker I guess; we\re not using it that much anymore. But maybe we could even move it to RTD too? 

This is all very exciting!

PR #1373, opened 2023-06-06
Update umamba gha

ocefpaf:
Using mamba-org/setup-micromamba@v1 instead of the deprecated mamba-org/provision-with-micromamba.


VeckoTheGecko (approved):
Thanks for the PR! CI failure is unrelated.

PR #1374, opened 2023-06-07
Update linkcheck_ignore

### LINKED ISSUE #1372: Broken readthedocs for geospatial tutorial
It seems that the readthedocs for the geospatial documentation causes an error, see https://readthedocs.org/projects/parcels/builds/20933429/

```
(examples/documentation_geospatial: line   39) broken    https://lxml.de/ - HTTPSConnectionPool(host='lxml.de', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f444f1a8110>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

This seems to prevent the v2.4.2 documentation to be built? Can you check what's going on, @VeckoTheGecko?

VeckoTheGecko:
Fixes #1372

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,6 +70,7 @@ linkcheck_ignore = [
     r"https://pubs\.acs\.org/doi/10\.1021/acs\.est\.0c01984",  # Site doesn't allow crawling
     r"https://aip\.scitation\.org/doi/10\.1063/1\.4982720",  # Site doesn't allow crawling
     r"https://www\.sciencedirect\.com/.*",  # Site doesn't allow crawling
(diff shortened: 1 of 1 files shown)


VeckoTheGecko:
Also added http://marine.copernicus.eu/ to the ignore as its unresponsive at the moment. Just some clarification on workflow, I am much more inclined to add stuff to `linkcheck_ignore` than to outright remove links from docs/notebooks.

Reasoning:
- Site could be temporarily down
- Is easier to revert if the site works at a later stage
- Its easier to maintain
- A broken link is better than no link (as it can be used with the Wayback Machine)

Of course, if the link is broken because the resource has moved, the link can be updated to the new one.

PR #1368, opened 2023-05-29
Updated Binder

VeckoTheGecko:
With the merging of #1355, consolidating binder support into this repo is possible (in fact, https://github.com/OceanParcels/parcels_examples_binder/ won't work anymore for future versions of parcels).

- [x] Update binder badge in README
- [x] add `.binder/environment.yml`

### Optional
- [ ] Archive https://github.com/OceanParcels/parcels_examples_binder/
- [x] Surface link to Binder somewhere prominent in doc website

Blocked by #1355

PR #1378, opened 2023-06-08
Fix Binder filepath typo

VeckoTheGecko:
Fixes https://github.com/OceanParcels/parcels/pull/1368#issuecomment-1581932278

Path should be `docs/examples` not `parcels/docs/examples`

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See [oceanparcels.org](http://oceanparcels.org/) for further information about [
 
 ### Launch Parcels Tutorials on mybinder.org
 
-[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/OceanParcels/parcels/master?labpath=parcels%2Fdocs%2Fexamples)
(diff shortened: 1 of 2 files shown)


erikvansebille:
Do you know if there's a way to have the binder default open with the filelist? It now opens with the landing page where users can select a clean notebook; but that is probably not what most users want. If they're new to Jupyter, they may not realise they have top click the file tab to see all the tutorials and examples

VeckoTheGecko:
I don't think theres an option to have the tree open automatically. Maybe if we have the notebook open on `parcels_tutorial.ipynb` then they can know that its working, and see the folder icon in the top-left to explore further?

erikvansebille:
> I don't think theres an option to have the tree open automatically. Maybe if we have the notebook open on `parcels_tutorial.ipynb` then they can know that its working, and see the folder icon in the top-left to explore further?

Yes, that's the way it used to be. I'll implement that option on oceanparcels.org; can you change it here on the parcels repo?

VeckoTheGecko:
Done

erikvansebille (approved):

Issue #1377, opened 2023-06-08
Cleaning python install process and setup.py


erikvansebille:
Issue #1376 made me realise that we now don't have any more scripts to install for Parcels. The `parcels_convert_npydir_to_netcdf` script had gone with version 2.4.0 in #1199, now the `parcels_get_examples` is also gone in v2.4.2 with #1355. 

So I guess this means that the code below in `setup.py` can go
https://github.com/OceanParcels/parcels/blob/bb98d113d89f1ea64020e8c7b03918577d0e8baf/setup.py#L21-L22

Furthermore, the versioning is now covered by [`parcels/_version.py`](https://github.com/OceanParcels/parcels/blob/master/parcels/_version.py), so I also don't think we need the `setuptools_scm` anymore?
https://github.com/OceanParcels/parcels/blob/bb98d113d89f1ea64020e8c7b03918577d0e8baf/setup.py#L17

Do does that mean that we don't need a [`setup.py`](https://github.com/OceanParcels/parcels/blob/master/setup.py) script al all? 

And that e.g. the call to [`python setup.py install --prefix=PREFIX`](https://github.com/OceanParcels/oceanparcels_website/blob/master/index.html#L589-L592) in the  Installation for developers section at https://oceanparcels.org/#installing can be removed?

PR #1379, opened 2023-06-08
Cleaning setup script

erikvansebille:
Following #1377, the setup.py script can be cleaned a bit

- [x] Remove legacy `parcels_convert_npydir_to_netcdf` from setup.py
- [ ] Investigate whether it's possible to remove the setup tools version system
- [ ] Investigate whether it's possible to remove setup.py altogether
- [ ] Update installation instructions on the website

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,4 @@ setup(name='parcels',
       python_requires='>=3.8',
       packages=find_packages(),
       package_data={'parcels': ['include/*']},
-      entry_points={'console_scripts': [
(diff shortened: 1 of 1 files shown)


VeckoTheGecko:
Just a note that implementing a `pyproject.toml` or `setup.cfg` file for the metadata seems to be more in line with current best practice (with `setup.py` just almost bare).

```py
# setup.py
from __future__ import annotations

from setuptools import setup
setup()
```

erikvansebille:
This has been implemented in #1392

PR #1392, opened 2023-07-05
Migrate to `pyproject.toml` (project metadata, tool config cleanup)

### LINKED ISSUE #1377: Cleaning python install process and setup.py

VeckoTheGecko:
Consolidate project metadata and tool configs into `pyproject.toml`.

Currated with reference to:
- [Python packaging guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata)
- [PEP621](https://peps.python.org/pep-0621/#name-the-project-urls-table-project-project-urls)
- [Pytest config](https://docs.pytest.org/en/7.1.x/reference/customize.html), [isort config](https://pycqa.github.io/isort/docs/configuration/config_files.html), [pydocstyle config](https://www.pydocstyle.org/en/stable/usage.html#configuration-files) 
- [Stackoverflow: Package data](https://stackoverflow.com/questions/69647590/specifying-package-data-in-pyproject-toml)
- https://github.com/pypa/setuptools_scm/issues/756#issuecomment-1199858274

Installation instructions on the website remain the same.

Fixes #1377 
Supercedes #1379


VeckoTheGecko:
From #1377 

> Furthermore, the versioning is now covered by parcels/_version.py, so I also don't think we need the setuptools_scm anymore?

@erikvansebille What is the origin behind the `parcels/_version.py` implementation, and its use in favour of the `setuptools_scm` implementation? I'm a bit weirded out about `parcels/_version.py` determining version at runtime (and feels a bit more custom than it needs to be).

erikvansebille:
> @erikvansebille What is the origin behind the `parcels/_version.py` implementation, and its use in favour of the `setuptools_scm` implementation? I'm a bit weirded out about `parcels/_version.py` determining version at runtime (and feels a bit more custom than it needs to be).

The reason was that users running with the `master` branch via git would then still get the right version number; and would need to reinstall every time they did a `git pull`.

VeckoTheGecko:
I see. I'm going to investigate what the behaviour of scm is without the dynamic behaviour in `parcels/_version.py`. I think its either `scm` or both, as relying solely on the dynamic behaviour has side-effects as mentioned above in my editted message

VeckoTheGecko:
Done with investigating, I think we'll need to leave the version selection as is.

erikvansebille:
> Done with investigating, I think we'll need to leave the version selection as is.

OK, can you briefly outline why you think we can't remove `scm`? What goes wrong if we do?

.flake8
erikvansebille, line 1: Why add these flake-ignores here? What is the reason for them in this PR?

VeckoTheGecko:
> OK, can you briefly outline why you think we can't remove `scm`? What goes wrong if we do?

Removing scm results in `conda env export` reporting `0.0.0` for the parcels version. I think the scm mechanism is the primary mechanism for determining the version (if the package isn't installed from a git repo)

.flake8
VeckoTheGecko: The flake config was specified in `setup.cfg` before alongside metadata/other tooling. All the other tooling/metadata has been now moved to `pyproject.toml` (flake8 is incompatible with pyproject.toml as mentioned bb05726). I renamed `setup.cfg` to be a `.flake8` file (another valid config file for flake8) to be a bit more explicit, and because the file now only deals with flake8.

erikvansebille (approved):

PR #1403, opened 2023-07-27
Adding installation instructions to documentation

erikvansebille:
Now that the tutorials are all on readthedocs, it makes sense to also move the installation instructions there. In this PR, I created an RST version of the installation instructions at https://oceanparcels.org/#installing, and also considerably cleaned them up / simplified them

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,7 +8,7 @@ This document outlines some guidelines for contributing to the parcels codebase.
 
 ## Getting Started
 
-To get started contributing to parcels, you'll need to install the developer version of Parcels following [the instructions](installation#installation-for-developers). This will install the latest version of Parcels from the `master` branch of the repository.
+To get started contributing to parcels, you'll need to install the developer version of Parcels following [our installation instructions](installation.rst#installation-for-developers). This will install the latest version of Parcels from the `master` branch of the repository.
 
 **Optional:** Setup `pre-commit` hooks by running `pre-commit install`. This ensures that code is formatted correctly before you commit it.
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -5,21 +5,12 @@ Even if you don't want to contribute directly to the codebase, getting involved 
 
 This document outlines some guidelines for contributing to the parcels codebase. If you have any questions or suggestions for how we can improve this document, please let us know.
 
+
 ## Getting Started
 
-To get started contributing to parcels, you'll need to do a few things:
+To get started contributing to parcels, you'll need to install the developer version of Parcels following [the instructions](installation#installation-for-developers). This will install the latest version of Parcels from the `master` branch of the repository.
 
-1. [Fork our repository](https://github.com/OceanParcels/parcels/fork) on GitHub.
-
-2. Clone your fork of the repository to your local machine, and `cd` into your project folder.
-
-3. Install [Anaconda](https://www.anaconda.com/download) (if you haven't already).
-
-4. Install the development Anaconda environment by installing from the environment file corresponding to your OS (`environment_py3_linux.yml`, `environment_py3_osx.yml` or `environment_py3_win.yml`) then activate it. E.g. on Linux: `conda env create -f environment_py3_linux.yml` then `conda activate py3_parcels`.
-
-5. Install your local copy of parcels in editable mode by running `pip install -e .`.
-
-6. **Optional:** Setup `pre-commit` hooks by running `pre-commit install`. This ensures that code is formatted correctly before you commit it.
+**Optional:** Setup `pre-commit` hooks by running `pre-commit install`. This ensures that code is formatted correctly before you commit it.
 
 
 ## Making Changes
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,7 +10,7 @@ Here you'll find tutorials, documentation, and the API reference for Parcels. Yo
 
    Home <self>
    Installation <installation>
-   Tutorials and documentation <documentation>
+   Tutorials & Documentation <documentation>
    API reference <reference>
    Contributing <contributing>
    OceanParcels website <https://oceanparcels.org/>
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -2,7 +2,7 @@ Parcels documentation
 ===================================
 
 Welcome to the documentation of Parcels.
-Here you'll find tutorials, documentation, and the API reference for Parcels. You can browse the documentation for older versions by using the version switcher in the left sidebar.
+Here you'll find `installation instructions <installation.rst>`_, `tutorials and documentation <documentation.rst>`_, and the `API reference <reference.rst>`_ for Parcels. You can browse the documentation for older versions by using the version switcher in the left sidebar.
 
 .. toctree::
    :maxdepth: 2
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,7 +9,7 @@ Here you'll find tutorials, documentation, and the API reference for Parcels. Yo
    :hidden:
 
    Home <self>
-   Installing Parcels <installation>
+   Installation <installation>
    Tutorials and documentation <documentation>
    API reference <reference>
    Contributing <contributing>
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,6 +9,7 @@ Here you'll find tutorials, documentation, and the API reference for Parcels. Yo
    :hidden:
 
    Home <self>
+   Installing Parcels <installation>
    Tutorials and documentation <documentation>
    API reference <reference>
    Contributing <contributing>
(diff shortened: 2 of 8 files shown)


docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+    .. code-block:: bash
+
+      jupyter notebook
+
+#. The next time you start a terminal and want to work with Parcels, activate the environment with:
+
+    .. code-block:: bash
+
+      conda activate parcels  # Linux / macOS
+      activate parcels        # Windows
```
VeckoTheGecko, line 52: `activate ...` works in Windows, but I feel that its non-standard compared to `conda activate ...` which also works for Windows.

```suggestion
#. The next time you start a terminal and want to work with Parcels, activate the environment by typing ``conda activate parcels``.
```

docs/index.rst
```diff
@@ -9,6 +9,7 @@ Here you'll find tutorials, documentation, and the API reference for Parcels. Yo
    :hidden:
 
    Home <self>
+   Installing Parcels <installation>
    Tutorials and documentation <documentation>
```
VeckoTheGecko, line 13:
```suggestion
   Tutorials & Documentation <documentation>
```

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+#. Install Anaconda's Miniconda following the steps at https://conda.io/docs/user-guide/install/, making sure to select the Python-3 version. If you're on Linux /macOS, it also assumes that you installed Miniconda-3 to your home directory.
+
+#. Start a terminal (Linux / macOS) or the Anaconda prompt (Windows). Activate the root (or base) environment of your Miniconda and create an environment containing Parcels, all its essential dependencies, and the nice-to-have cartopy and jupyter packages:
+
+    .. code-block:: bash
+
+      conda activate root  # Linux / macOS
+      activate root        # Windows
+
+      conda create -n parcels -c conda-forge parcels cartopy
```
VeckoTheGecko, line 17: Is root used for Anaconda? First I've heard of it; maybe something from older versions, or is it specific to Miniconda? Maybe worth just mentioning `base`?

```suggestion
#. Start a terminal (Linux / macOS) or the Anaconda prompt (Windows). Activate the ``base`` environment of your Miniconda and create an environment containing Parcels, all its essential dependencies, and the nice-to-have cartopy and jupyter packages:

    .. code-block:: bash

      conda activate base
      conda create -n parcels -c conda-forge parcels cartopy
```

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+    .. note::
+
+        For some of the examples, ``pytest`` needs to be installed with ``conda install -n parcels pytest``
+
+#. Activate the newly created Parcels environment:
+
+    .. code-block:: bash
+
+      conda activate parcels  # Linux / macOS
+      activate parcels        # Windows</code></pre>
```
VeckoTheGecko, line 28: See other comment.

```suggestion
#. Activate the newly created Parcels environment by typing ``conda activate parcels``.
```

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+    .. code-block:: bash
+
+      conda activate root  # Linux / macOS
+      activate root        # Windows
+
+      conda create -n parcels -c conda-forge parcels cartopy
+
+    .. note::
+
+        For some of the examples, ``pytest`` needs to be installed with ``conda install -n parcels pytest``
```
VeckoTheGecko, line 21: Clarification (1. I was unfamliar with `-n` in `conda install`, and 2. previously may have been misread as `conda install parcels pytest` which installs `parcels` and `pytest`).

```suggestion
        For some of the examples, ``pytest`` also needs to be installed. This can be quickly done with ``conda install -n parcels pytest`` which installs ``pytest`` directly into the newly created ``parcels`` environment.
```

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+    .. code-block:: bash
+
+      git clone https://github.com/OceanParcels/parcels.git
+      conda env create -f environment_py3_<OS>.yml  # where <OS> is either linux, osx or win
+
+Then, add the directory where you stored the Parcels code to your ``$PYTHONPATH`` environment variable. For example, if you cloned the Parcels code to ``/home/username/parcels``, add the following line to your ``.bashrc`` or ``.zshrc`` file:
+
+    .. code-block:: bash
+
+      export PYTHONPATH=/home/username/parcels:$PYTHONPATH
```
VeckoTheGecko, line 69: Havent extensively tested, `pip install -e /home/username/parcels` seems to work. How does this approach differ? I'm a bit hesitant to directly modify `PYTHONPATH`

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Installation for developers
+===========================
+
+Parcels depends on a working Python installation, a netCDF installation, a C compiler, and various Python packages. If you prefer to maintain your own Python installation providing all this, download one of the environment files for `linux <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_py3_linux.yml>`_, for `macOS <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_py3_osx.yml>`_ or for `Windows <https://raw.githubusercontent.com/OceanParcels/parcels/master/environment_py3_win.yml>`_ and install it with:
+
+    .. code-block:: bash
+
+      git clone https://github.com/OceanParcels/parcels.git
+      conda env create -f environment_py3_<OS>.yml  # where <OS> is either linux, osx or win
```
VeckoTheGecko, line 63: I'm not sure about the wording of this section. Is it necessary to include info on downloading the environment files? (they would be in the clone anyway). Perhaps something along the line of 

```suggestion
If you would prefer to have a development installation of Parcels (i.e., where the code can be actively editted), you can do so by cloning the Parcels repo, installing dependencies using the environment file (which includes Python, netCDF tooling, a C compiler, and various Python packages), and then installing Parcels in an editable mode such that changes to the cloned code can be tested during development.

    .. code-block:: bash

      git clone https://github.com/OceanParcels/parcels.git
      conda env create -f environment_py3_<OS>.yml  # where <OS> is either linux, osx or win
```

VeckoTheGecko (changes requested):
I've made some comments.

Sidenote, @erikvansebille do you use `sphinx-autobuild docs docs/_build` for your docs local development? I'm getting an infinite rendering loop on my end, and just wondering if that's high enough priority for me to patch.

erikvansebille:
Tanks for these comments, @VeckoTheGecko! Really useful! Especially the fact that `conda activate` now works all platforms makes the instructions cleaner. Shows that we based these instructions on a very old condo version...

> Sidenote, @erikvansebille do you use `sphinx-autobuild docs docs/_build` for your docs local development? I'm getting an infinite rendering loop on my end, and just wondering if that's high enough priority for me to patch.

No, I don't use sphinx-auto build myself. Am doing manual building for now.

docs/installation.rst
erikvansebille: OK, I now added `pip install -e .` as default option (this was actually suggested too by Github Copilot); but want to keep the `$PYTHONPATH` option as a fall-back (in the past I've had issues with `pip install`)

docs/installation.rst
erikvansebille: Good point about using `condo activate` in all cases, but I prefer to show it in a `code-block`

docs/installation.rst
erikvansebille: Good point about using `condo activate` in all cases, but I prefer to show it in a `code-block`

docs/contributing.md
```diff
@@ -5,21 +5,12 @@ Even if you don't want to contribute directly to the codebase, getting involved
...
 
-2. Clone your fork of the repository to your local machine, and `cd` into your project folder.
-
-3. Install [Anaconda](https://www.anaconda.com/download) (if you haven't already).
-
-4. Install the development Anaconda environment by installing from the environment file corresponding to your OS (`environment_py3_linux.yml`, `environment_py3_osx.yml` or `environment_py3_win.yml`) then activate it. E.g. on Linux: `conda env create -f environment_py3_linux.yml` then `conda activate py3_parcels`.
+## Getting Started
 
-5. Install your local copy of parcels in editable mode by running `pip install -e .`.
+To get started contributing to parcels, you'll need to install the developer version of Parcels following [the instructions](installation#installation-for-developers). This will install the latest version of Parcels from the `master` branch of the repository.
```
erikvansebille, line 11: I can't get this link to work; could you help me how to format this link so it points to installation.html#installation#installation-for-developers, @VeckoTheGecko?

docs/installation.rst
VeckoTheGecko: Doing a quick search, I come acrross [this SO article](https://stackoverflow.com/a/49478207) going over developper installation for Conda packages using `conda develop .` ([see help documentation for `conda develop`](https://docs.conda.io/projects/conda-build/en/main/resources/commands/conda-develop.html)).

Has this route been investigated? Happy to do some further investigating if not.

Part of my hesitancy for `$PYTHONPATH` is mainly since I'm not sure how that would affect other Conda environments (i.e. globally "installing" it into all conda/python environments?).

erikvansebille: OK thanks for digging into this. I can confirm that `pip install --no-build-isolation --no-deps -e .` also seems to work on my system. Have updated it now in the instructions

VeckoTheGecko (approved):
Pushed a small edit, but LGTM otherwise!

PR #1471, opened 2023-11-27
Changing conda selectors syntax

erikvansebille:
Changing the syntax for the conda [unix] selector (for MPI to install only on linux and osx), because `sel(unix)` was not widely supported and sometimes led to install problems


VeckoTheGecko:
Just wondering about the motivation for 9bd78f0. From my understanding the mamba project is a much more performant version of the conda project, and operates as a drop in replacement.

erikvansebille:
> Just wondering about the motivation for [9bd78f0](https://github.com/OceanParcels/parcels/commit/9bd78f0053104341e7d6a129be82a5706c40d719). From my understanding the mamba project is a much more performant version of the conda project, and operates as a drop in replacement.

I was trying to get the original `sel(unix)` commands to work on conda too (because I hoped that would work with the new libmamba solver). It's needed to make the [the instructions for our local system at UU](https://github.com/OceanParcels/UtrechtTeam/wiki/How-to-run-parcels-on-lorenz-or-gemini#the-developer-version-of-parcels) work, as well as make the [installation instructions for developers](https://docs.oceanparcels.org/en/latest/installation.html#installation-for-developers) simpler.

As I couldn't get that `sel(unix)` to work, I tried to revert back to the `# [unix]` preprocessing selectors (see [here](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#preprocessing-selectors) and therefore (temporarily) reverted back to conda instead of mamba. Nevertheless, that didn't seem to work in environment files either?

Anyways, I agree this PR is a bit of a mess, but trying to get installation instructions as simple as possible for conda. Any ideas how?

VeckoTheGecko:
Looks like selectors aren't supported in `environment.yml` files (only `meta.yml`) https://github.com/conda/conda/issues/8089. Not sure what the best solution to this is unfortunately. I guess we migrated away from maintaining 3 different `environment.yml` files?

PR #1494, opened 2024-01-11
Expanding explanation of output chunks parameter

erikvansebille:
Following the discussion at https://github.com/OceanParcels/parcels/discussions/1473#discussioncomment-8071430, this PR expands the explanation of the `ParticleFile`  `chunks` parameter in the parcels overview tutorial


JamiePringle (commented):
I would add something like "In either case, it will generally be much more efficient if the product of the two chunk sizes is greater than several thousand. Often much greater. The details will depend on the nature of the filesystem the data is being written to, and often some experimentation can help greatly increase the efficiency of the runs."

PR #1552, opened 2024-04-30
Compilation error message fix for local variable name that is also FieldSet constant

### LINKED ISSUE #1512: Confusing compilation error message when variable in Kernel has same name as FieldSet constant
When a user writes a Kernel with a variable that has the same name as a FieldSet contestant, that throws an unclear error. E.g.

```python
fieldset.add_constant('vertical_speed', 0.1)

 def ArgoVerticalMovement(particle, fieldset, time):
    vertical_speed = fieldset.vertical_speed
```

Throws an output like
```
/var/folders/1n/500ln6w97859_nqq86vwpl000000gr/T/parcels-504/982d2e61fa4f6033c50aae39bf44d63b_0.c:73:18: warning: explicitly assigning value of variable of type 'float' to itself [-Wself-assign]
  vertical_speed = vertical_speed;
  ~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~
```

This could be caught in the `codeconverter.py` with a more elegant and understandable error

erikvansebille:
Fixing the compilation error message when a local Kernel variable has same name as a FieldSet constant

This fixes #1512

### DIFF
--- a/tests/test_kernel_language.py
+++ b/tests/test_kernel_language.py
@@ -5,6 +5,7 @@ import numpy as np
 import pytest
 
 from parcels import (
+    Field,
     FieldSet,
     JITParticle,
     Kernel,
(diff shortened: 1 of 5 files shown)


tests/test_kernel_language.py
```diff
@@ -186,6 +187,31 @@ def kernel_pnum(particle, fieldset, time):
         assert error_thrown
 
 
+def test_varname_as_fieldname():
```
VascoSch92, line 190: You can use `pytest.raises(Exception)`. It is a much easier way to test that the code raise exception.

See also: https://stackoverflow.com/questions/23337471/how-do-i-properly-assert-that-an-exception-gets-raised-in-pytest

tests/test_kernel_language.py
```diff
@@ -186,6 +187,27 @@ def kernel_pnum(particle, fieldset, time):
...
+    def kernel_particlename(particle, fieldset, time):
+        particle.speed = fieldset.speed[particle]  # noqa
+
+    pset.execute(kernel_particlename, endtime=1, dt=1.)
+    assert pset[0].speed == 10
+
+    def kernel_varname(particle, fieldset, time):
+        vertical_speed = fieldset.vertical_speed  # noqa
+
+    with pytest.raises(Exception):
```
VascoSch92, line 207:
```suggestion
    with pytest.raises(NotImplementedError):
```

PR #1558, opened 2024-05-08
Throwing an error when dt has too high precision

erikvansebille:
As found by @sruehs and @ezekiel-lemur, parcels output can 'skip' timesteps when `dt` has a very high precision (i.e. many digits behind the comma). This is very likely because of the test done when considering which particles to write

https://github.com/OceanParcels/parcels/blob/a0ea4044c3ced351b65ec3a19ea3bd8d29f8bd3d/parcels/particledata.py#L322-L328

Solution is to throw an error when users provide a `dt` that has higher precision than 1e-6. Note that parcels before already threw an error when `dt < 1e-6`, now a value like `dt=1+1e-6` would also be caught

The new testcase in example_globcurrent (with `dt = 81.2584344538292`) indeed shows how the output file misses a few time-snapshots if this error is not raised

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -880,7 +880,7 @@ class ParticleSet(ABC):
             dt = dt.total_seconds()
         if abs(dt) <= 1e-6:
             raise ValueError('Time step dt is too small')
-        if dt*1e6 % 1 != 0:
+        if (dt * 1e6) % 1 != 0:
             raise ValueError('Output interval should not have finer precision than 1e-6 s')
         outputdt = output_file.outputdt if output_file else np.infty
         if isinstance(outputdt, delta):
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -878,8 +878,10 @@ class ParticleSet(ABC):
             runtime = runtime.total_seconds()
         if isinstance(dt, delta):
             dt = dt.total_seconds()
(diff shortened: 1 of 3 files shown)


parcels/particleset.py
```diff
@@ -878,8 +878,10 @@ def execute(self, pyfunc=AdvectionRK4, pyfunc_inter=None, endtime=None, runtime=
             runtime = runtime.total_seconds()
         if isinstance(dt, delta):
             dt = dt.total_seconds()
-        if dt > 0 and dt <= 1e-6:
+        if abs(dt) <= 1e-6:
             raise ValueError('Time step dt is too small')
+        if dt*1e6 % 1 != 0:
```
VascoSch92, line 883: What do you want to check here?

erikvansebille: As I say in the comment of the PR, we've seen that Parcels unexpectedly fails (outputfiles miss some timeslices) when `dt` has a precision finer than 1e-6. 

So `dt=3.001` works, but `dt=3.0000001` doesn't. 

I want to catch these latter cases

erikvansebille: Do you have a proposal for a cleaner code to do the same?

VascoSch92: ah ok sorry. Now is clear. Perhaps just add parenthisis: `(dt * 1e6) % 1`

Issue #1564, opened 2024-05-14
Zarr 2.18 Breaks ParticleSet.execute(): ValueError: buffer source array is read-only


jamespolly:
The latest release of Zarr is [v2.18.0](https://github.com/zarr-developers/zarr-python/releases/tag/v2.18.0) and was released last week.

Using the [parcels development environment](https://docs.oceanparcels.org/en/latest/installation.html#installation-for-developers), the following error is generated when performing `particleset.execute()`.

Rather than include some other links pointing to possibly unrelated instances of this error, I'll leave the traceback below, and say that pinning the zarr dependency to `zarr<2.18.0` in `environment.yml` will fix this.

Zarr v3.0.0 is soon in the works for this summer and maybe this behavior will change.

```
INFO: Output files are stored in [...]out.zarr.
  0%|                                                                                                                                                          | 0/18000.0 [00:00<?, ?it/s]---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/[...]:121
--> 121 pset.execute(
    122     kernels,
    123     runtime=timedelta(hours=5),
    124     dt=timedelta(minutes=5),
    125     output_file=output_file,
    126 )

File ~/repos/parcels/parcels/particleset.py:988, in ParticleSet.execute(self, pyfunc, pyfunc_inter, endtime, runtime, dt, output_file, verbose_progress, postIterationCallbacks, callbackdt, delete_cfiles)
    986         output_file.write_latest_locations(self, time)
    987     else:
--> 988         output_file.write(self, time_at_startofloop)
    989 if np.isfinite(outputdt):
    990     next_output += outputdt * np.sign(dt)

File ~/repos/parcels/parcels/particlefile.py:273, in ParticleFile.write(self, pset, time, indices)
    271         else:
    272             if max(obs) >= Z[varout].shape[1]:
--> 273                 self._extend_zarr_dims(Z[varout], store, dtype=self.vars_to_write[var], axis=1)
    274             Z[varout].vindex[ids, obs] = pset.particledata.getvardata(var, indices_to_write)
    276 pset.particledata.setvardata('obs_written', indices_to_write, obs+1)

File ~/repos/parcels/parcels/particlefile.py:184, in ParticleFile._extend_zarr_dims(self, Z, store, dtype, axis)
    182     obs = zarr.group(store=store, overwrite=False)["obs"]
    183     if len(obs) == Z.shape[1]:
[...]

PR #1567, opened 2024-05-18
unpin zarr version upper bound

jamespolly:
Per #1564 a simple change of the environment.yml back to `zarr>=2.11.0`.

Still seeing a couple of failed parameterized tests for `test_mpi_run` but this may be a local issue.

Thought I'd just put this PR up here for discussion of those tests or whatever else we may want to do here. Feel free to disregard if not helpful.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -25,7 +25,7 @@ dependencies:
   - nbval
   - scikit-learn
   - pykdtree
-  - zarr>=2.11.0,<2.18.0
+  - zarr>=2.11.0
 
   # Formatting
   - black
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -25,7 +25,7 @@ dependencies:
   - nbval
   - scikit-learn
   - pykdtree
-  - zarr>=2.11.0,<2.18.0
+  - zarr>=2.11.0
```
jakirkham, line 28: If you like, it may be worth excluding the problematic version as a safety measure

```suggestion
  - zarr>=2.11.0,!=2.18.0
```

jakirkham (commented):
Thanks James! 🙏

Had a minor suggestion below. Though leave it to you as to whether or not to include it

jamespolly:
@jakirkham Thanks very much for the suggestion! It's been incorporated :+1:

PR #1574, opened 2024-05-24
PyPi release updates

### LINKED ISSUE #1550: Missing dependencies in pyproject.toml
`pyproject.toml` is missing dependencies. These are present in `environment.yml`. See the following list:

```
cgen
dask
cftime
psutil
netCDF4
zarr
tqdm
pymbolic (not sure but I think this is actually a missing cgen dependency?)
```

As an example, adding them can be done like so:
```
[project]
(...)
dependencies = [
"cgen",
"dask >= 2.0",
]

andrew-s28:
This adds dependencies to the pyproject.toml files (except for the c compiler dependency, which cannot be installed from pip) and includes a basic test in the pypi-release workflow that uses conda to install a c compiler then pip installs the Parcels release from PyPi and runs the example_peninsula.py file. This is meant to mimic the workflow in the [conda recipe](https://github.com/conda-forge/parcels-feedstock/blob/main/recipe/meta.yaml) and should hopefully simplify efforts to keep pyproject.toml up to date. 

The updated workflow was tested on my [personal test.pypi.org release](https://github.com/andrew-s28/parcels/actions/runs/9228339614).

Closes #1550


erikvansebille:
Thanks for adding this workflow, @andrew-s28! However, it doesn't seem to be triggered by Actions. Not sure why, I already renamed it from pypi-release.yaml to pypi.release.yml. Do you know what I should do to automatically have it checked by Actions?

VascoSch92:
> Thanks for adding this workflow, @andrew-s28! However, it doesn't seem to be triggered by Actions. Not sure why, I already renamed it from pypi-release.yaml to pypi.release.yml. Do you know what I should do to automatically have it checked by Actions?

How you tried? 
because the workflow is triggered just if the job `upload-to-pypi` is executed. This last job is executed just when there is a release. i.e., you need to release a version to check this job.

erikvansebille:
Thanks @VascoSch92, I didn't realise that the Action will only be triggered upon release. Before merging, I would want to check if it indeed passes all tests, as intended, though. So I now (temporarily?) added a check for every pull-request. But the `test-pypi_release` was skipped so not tested; do you know why?

VascoSch92:
> Thanks @VascoSch92, I didn't realise that the Action will only be triggered upon release. Before merging, I would want to check if it indeed passes all tests, as intended, though. So I now (temporarily?) added a check for every pull-request. But the `test-pypi_release` was skipped so not tested; do you know why?



> Thanks @VascoSch92, I didn't realise that the Action will only be triggered upon release. Before merging, I would want to check if it indeed passes all tests, as intended, though. So I now (temporarily?) added a check for every pull-request. But the `test-pypi_release` was skipped so not tested; do you know why?

Because the job `upload-to-pypi` is trigger just when `if: github.event_name == 'release'`. and the test-pypi_release need this job.

I think the beast way to test it is to delete `needs: upload-to-pypi`
In this way you can see if the job work and before merging you can re-add it to the workflow

VascoSch92:
Nice :-)

PR #1605, opened 2024-07-16
Removing redundant functions related to old AoS architecture

erikvansebille:
This PR removed some redundant functions in `parcels.h` related to old AoS architecture, which was retired in v3.0.0 (https://github.com/OceanParcels/parcels/pull/1423). 

These functions are not used anymore, so better to be removed to reduce clutter


VeckoTheGecko (approved):
LGTM!

Issue #1608, opened 2024-07-22
Kernel can't be converted to C depending on method of import


VeckoTheGecko:
In pursuit of #1581, I encountered the following error message:

```
NotImplementedError: Cannot convert 'parcels' used in kernel to C-code
```

See the following minimal example. Doing the suggested change causes the error message:

```diff
import parcels
- from parcels import ParcelsRandom

example_dataset_folder = parcels.download_example_dataset("GlobCurrent_example_data")
filenames = {
    "U": f"{example_dataset_folder}/20*.nc",
    "V": f"{example_dataset_folder}/20*.nc",
}
variables = {
    "U": "eastward_eulerian_current_velocity",
    "V": "northward_eulerian_current_velocity",
}
dimensions = {"lat": "lat", "lon": "lon", "time": "time"}

fieldset = parcels.FieldSet.from_netcdf(filenames, variables, dimensions)

def nudge_kernel(particle, fieldset, time):
-   dlat = ParcelsRandom.normalvariate(0.0, 1.0)
+   dlat = parcels.ParcelsRandom.normalvariate(0.0, 1.0)
    particle_dlat += dlat

time = np.arange(0, 43200, 3600)
lon = [29] * len(time)
lat = [-33] * len(time)
pset = parcels.ParticleSet(fieldset=fieldset, pclass=parcels.JITParticle, lon=lon, lat=lat, time=time)


pset.execute(
    nudge_kernel,
    runtime=timedelta(days=7),
    dt=timedelta(minutes=5),
    output_file=output_file,
)
```

Likely culplit is code here:
https://github.com/OceanParcels/parcels/blob/035159a1b4002bc1550cebd275f93c0adaa8513d/parcels/compilation/codegenerator.py#L251-L265

PR #1610, opened 2024-07-23
Documentation improvements (update import statements, include scripts)

### LINKED ISSUE #1581: Changing Parcels -> parcels and importing all of parcels throughout
Most python packages are all-lowercase ([as suggested by pep8](https://peps.python.org/pep-0008/#package-and-module-names)), so we should also change the writing of `parcels` to lowercase throughout in the examples, documentation, docstrings and website. 

While we're at it, we could also change the notebooks in the examples to refer to the classes as `parcels.ClassName`, to better highlight that these are specific to `parcels`. So e.g. change
```python
from parcels import FieldSet

fset = FieldSet(...)
```
to
```python
import parcels

fset = parcels.FieldSet(...)
```

### LINKED ISSUE #1636: Removing `delta` alias for `timedelta`
The codebase uses `delta` as an alias for `datetime.timedelta` through the codebase and through the notebooks. This is a bit unique to `parcels`, and can lead to confusion to those famliar with Python wanting to use parcels for simulation.

It would be good to remove this in favour of using `timedelta`.

VeckoTheGecko:
This PR implements the following changes:
- Update the parcels import across notebooks and example scripts (fixes #1581)
- Fixes #1636 
- Include script examples in documentation website
- Replace deprecated prettier hook
- Update sphinx builder to dirhtml. This makes documentation links go from `/pagename.html` or `/pagename/index.html` to `/pagename`. Redirects are automatically handled by RTD so no broken links are created
- Bugfix geospatial notebook `LineString` error
- Add Black autoformatting for examples/*.py files
- Remove dedentation code (contributes to #1620)

Blocked by #1608


docs/conf.py
```diff
@@ -316,7 +316,11 @@ def linkcode_resolve(domain, info):
 # html_use_smartypants = True
 
 # Custom sidebar templates, maps document names to template names.
-# html_sidebars = {}
+
+html_sidebars = {
+    "**": ["sidebar-nav-bs", "sidebar-ethical-ads"],
```
erikvansebille, line 321: This now seems to give us the ads twice? That's perhaps overdoing thing?

docs/conf.py
```diff
@@ -316,7 +316,11 @@ def linkcode_resolve(domain, info):
...
+    "documentation/additional_examples": []
```
erikvansebille, line 322: So no additional examples yet?

docs/documentation/additional_examples.rst
erikvansebille, line 1: I don't see this page in the top-bar where the other navigation is?

docs/documentation/index.rst
```diff
@@ -0,0 +1,76 @@
...
+   ../examples/documentation_homepage_animation.ipynb
+
+
+Python Example Scripts
+----------------------
+
+
+.. toctree::
+
+   additional_examples
```
erikvansebille, line 76: This part does not show up on https://parcels--1610.org.readthedocs.build/en/1610/documentation/?

parcels/kernel.py
```diff
@@ -178,15 +170,23 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None, funccode=None, p
...
+        self.funccode = (  # Remove parcels. prefix (see #1608)
+            self.funccode
+            .replace('parcels.rng', 'rng')
+            .replace('parcels.ParcelsRandom', 'ParcelsRandom')
+            .replace('parcels.StatusCode', 'StatusCode')
+        )
+
         # Parse AST if it is not provided explicitly
-        self.py_ast = py_ast or parse(BaseKernel.fix_indentation(self.funccode)).body[0]
+        self.py_ast = py_ast or parse(textwrap.dedent(self.funccode)).body[0]  # Dedent allows for in-lined kernel definitions
```
erikvansebille, line 181: Why this new method? What was wrong with the old one?

VeckoTheGecko: `BaseKernel.fix_indentation()` does the exact same as `textwrap.dedent()`. For maintainability and documentation, I would prefer to rely on standard library code rather than writing custom code.

As an aside I think simple string utilities like this shouldn't be included as public static methods, but should instead be private utility functions. `BaseKernel.fix_indentation()` has no purpose in the public API

docs/documentation/index.rst
VeckoTheGecko: its a section at the bottom of the page (https://parcels--1610.org.readthedocs.build/en/1610/documentation/#python-example-scripts). I have updated the paragraph at the top introducing the documentation with a mention to it

docs/documentation/additional_examples.rst
VeckoTheGecko: I thought it would be better to include it in the documentation page instead of adding it as another item at the top

docs/conf.py
VeckoTheGecko: This change is only to do with the sidebars themselves. `"**": ["sidebar-nav-bs", "sidebar-ethical-ads"]` is the default setting, showing the page navigation (left) and the section navigation (right). On the additional_examples page there was horizontal scrolling which made it difficult to read the code, so I disabled the sidebars.

docs/conf.py
VeckoTheGecko: I can't replicate this in my browser :( . It shouldn't be doing so, could you attach a screenshot?

parcels/kernel.py
```diff
@@ -121,15 +122,6 @@ def _cache_key(self):
...
 
-    @staticmethod
-    def fix_indentation(string):
-        """Fix indentation to allow in-lined kernel definitions."""
-        lines = string.split('\n')
-        indent = re.compile(r"^(\s+)").match(lines[0])
-        if indent:
-            lines = [line.replace(indent.groups()[0], '', 1) for line in lines]
-        return "\n".join(lines)
-
```
VeckoTheGecko, line 132: Note that this would be a change to the public API. As mentioned https://github.com/OceanParcels/parcels/pull/1610#discussion_r1696604280 I don't see a usecase where a parcel user directly calls this code, so I don't think this removal would be disruptive.

docs/conf.py
erikvansebille: Sure, see below
Screenshot 2024-07-30 at 14 34 36

VeckoTheGecko: Ok, this is really strange. I have gone back and forth between getting 1 ad and 2 ads, even changing based on the page reload or the tab I have open. The placement of the ad is determined by a div with `id="ethical-ad-placement"`, and sometimes when reloading there are two of these divs.

I'm not sure if this is a problem on ReadTheDocs side (where it isn't properly identifying the div, so it assumes it doesn't exist and inserts its own), or if its a problem with the PyData theme and its documented defaults.

I'm going to try removing the explicit placement and see if that helps.

VeckoTheGecko: 10c6824 looks to have fixed it. Feel free to test it on your end @erikvansebille

erikvansebille: Indeed, seems fixed now

VeckoTheGecko:
I guess force pushing while referencing a commit glitches out the UI a bit. This is just a single commit in the git log image

VeckoTheGecko:
Thought I'd wrap #1636 into here

erikvansebille (approved):
In principle OK, but I don't entire understand why we still use `from datetime import timedelta`. 

Given also the other changes, would it not be more consistent to do `import datetime` and then use `datetime.timedelta` throughout?

VeckoTheGecko:
The changes for the import statements are related to how these imports are typically done by the Python community, and reducing confusion for software developers. The imports for `os`, and `ast` resulted in functions/modules like `parse()` and `path` being imported into the global namespace which can lead to confusion as `path` as a variable is mainly used as to represent a filepath object/string, and `parse` could be thought of as a user made function. Writing `ast.` or `os.` is not much hardly more of a burden to type out, and signal clearly where the function is coming from.

Similarly, changing `delta` to `timedelta` also reduces confusion. `from datetime import timedelta` is a common import, so seeing `timedelta` by itself already signals its from `datetime`. We could do `import datetime` and `datetime.timedelta`, but its a bit of a pain to type out especially since this method is used a lot in parcels. We could shorten that by doing `import datetime as dt` (which is another common alias), but `dt` is used quite a bit in our function calls so not optimal. I thought `timedelta` was best.

Issue #1549, opened 2024-04-30
Add a tutorial to help users choose an appropriate advection kernel


erikvansebille:
Following on from the RK45 update (#1539), it might be good to create a tutorial that helps users identify which Advection kernels are most appropriate for their work. This could build on the paper by [Nordam & Duran 2020](https://gmd.copernicus.org/articles/13/5935/2020/) and some recent work by @daanreijnders. 

The tutorial could include comparisons between the Euler-Forward, RK4, RK45, and Analytical schemes that already exist, as well as perhaps add RK2 and RK23 (aka Bogacki–Shampine). 

The tutorial could also link to the diffusion tutorial, and perhaps even have some discussion on how to choose appropriate values of `dt` (for explicit methods)

erikvansebille:
Some of the bits from @michaeldenes fantastic notebook for the Parcels event at https://github.com/OceanParcels/10year-anniversary-session2 could be used for this

PR #1612, opened 2024-07-26
Analytical Advection in jit

erikvansebille:
This PR implements a C version of the `AnalyticalAdvection` kernel for use with `JITParticles`

Recent work by @daanreijnders found that the [Analytical algorithm for advection](https://docs.oceanparcels.org/en/latest/examples/tutorial_analyticaladvection.html) is far more accurate than RK4 advection, mainly because of discontinuities at grid ceel boundaries. Hence, we want to be able to provide a fully functional AnalyticalAdvection kernel to Parcels users

- [x] make unit tests work
- [ ] confirm AnalyticalAdvection also works on spherical grids
- [ ] performance-test and validate AnalyticalAdvection in real-world scenarios
- [ ] (potentially) make AnalyticalAdvection the default kernel in Parcels; or at least encourage its use in a new 'Choosing-Advection-Kernels-tutorial' (see also #1549)


michaeldenes:
> @erikvansebille is [47f50a0](https://github.com/OceanParcels/Parcels/commit/47f50a0676b1bb2194447c33d41855aff26eabfe) the commit that will fix #1689 ?

Looks like you already placed the fix in master! https://github.com/OceanParcels/Parcels/pull/1614

erikvansebille:
This is a very old PR that will not be implemented in v4 (both because it depends on JIT-mode, and because AnalyticalAdviection is not scheduled for the first v4 release candidates. Closing for now (but not deleting branch so we can learn from it when working on AnalyticalAdvection for v4)

PR #1615, opened 2024-07-26
Refactoring GridCode to be called GridType

erikvansebille:
The terms `GridType` (`gtype`) and `GridCode` (`gcode`) used for differentiating between Rectilinear- and Curvilinear-, and Z- and S-grids, were used interchangeably in the Parcels codebase, leading to confusion. 

Therefore, this refactoring changes all `GridCode` to `GridType` and `gcode` to `gtype`.

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -7,7 +7,7 @@ import numpy as np
 from parcels.tools.converters import TimeConverter
 from parcels.tools.loggers import logger
 
-__all__ = ['GridType', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'CGrid', 'Grid']
+__all__ = ['GridType', 'GridCode', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'CGrid', 'Grid']
 
 
 class GridType(IntEnum):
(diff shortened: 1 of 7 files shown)


parcels/grid.py
```diff
@@ -7,10 +7,10 @@
 from parcels.tools.converters import TimeConverter
 from parcels.tools.loggers import logger
 
-__all__ = ['GridCode', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'CGrid', 'Grid']
+__all__ = ['GridType', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'CGrid', 'Grid']
```
VeckoTheGecko, line 10:
```suggestion
__all__ = ['GridType', 'GridCode', 'RectilinearZGrid', 'RectilinearSGrid', 'CurvilinearZGrid', 'CurvilinearSGrid', 'CGrid', 'Grid']
```

Since this is a change to the public API, I wondering if we should still accept `.GridCode`? Otherwise people's simulation code may break in the new version.

Should just be a
```python
# GridCode has been renamed to GridType for consistency.
# TODO: Remove alias in Parcels v4
GridCode = GridType
```
after the enum declaration.

I think it would be worth discussing our approach to deprecating items. I don't think we need to be as stable as numpy as this is a research package (in fact, I feel being that stable may be detrimental to development) but I think if we do/plan to deprecate things moving forward it would be good to raise warnings/deprecation notices.

Otherwise LGTM and happy for this to be merged pending green checks

VeckoTheGecko (changes requested):

VeckoTheGecko (approved):

PR #1618, opened 2024-07-26
Support use of parcels.rng in Kernels

### LINKED ISSUE #1608: Kernel can't be converted to C depending on method of import

erikvansebille:
This PR fixes #1608. it supports using `parcels.rng` directly in Kernels (note, `parcels.ParcelsRandom` is also supported, but since the module itself it called `rng` better to use `parcels.rng` for traceability)

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -23,6 +23,7 @@ except ModuleNotFoundError:
     MPI = None
 
 import parcels.rng as ParcelsRandom  # noqa
+from parcels import rng  # noqa
 from parcels.application_kernels.advection import (
     AdvectionAnalytical,
     AdvectionRK4_3D,
@@ -178,6 +179,7 @@ class Kernel(BaseKernel):
         else:
             self.funcvars = None
         self.funccode = funccode or inspect.getsource(pyfunc.__code__)
+        self.funccode = self.funccode.replace('parcels.', '')  # Remove parcels. prefix (see #1608)
         # Parse AST if it is not provided explicitly
         self.py_ast = py_ast or parse(BaseKernel.fix_indentation(self.funccode)).body[0]
         if pyfunc is None:
@@ -187,6 +189,7 @@ class Kernel(BaseKernel):
                 user_ctx = stack[-1][0].f_globals
                 user_ctx['math'] = globals()['math']
                 user_ctx['ParcelsRandom'] = globals()['ParcelsRandom']
+                user_ctx['rng'] = globals()['rng']
                 user_ctx['random'] = globals()['random']
                 user_ctx['StatusCode'] = globals()['StatusCode']
             except:
--- a/parcels/application_kernels/advectiondiffusion.py
+++ b/parcels/application_kernels/advectiondiffusion.py
@@ -4,7 +4,7 @@ See `this tutorial <../examples/tutorial_diffusion.ipynb>`__ for a detailed expl
 """
 import math
 
-import parcels.rng as ParcelsRandom
+import parcels
 
 __all__ = ['DiffusionUniformKh', 'AdvectionDiffusionM1', 'AdvectionDiffusionEM', ]
 
@@ -25,8 +25,8 @@ def AdvectionDiffusionM1(particle, fieldset, time):
     mean and a standard deviation of sqrt(dt).
     """
     # Wiener increment with zero mean and std of sqrt(dt)
-    dWx = ParcelsRandom.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
-    dWy = ParcelsRandom.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
+    dWx = parcels.rng.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
+    dWy = parcels.rng.normalvariate(0, math.sqrt(math.fabs(particle.dt)))
 
     Kxp1 = fieldset.Kh_zonal[time, particle.depth, particle.lat, particle.lon + fieldset.dres]
     Kxm1 = fieldset.Kh_zonal[time, particle.depth, particle.lat, particle.lon - fieldset.dres]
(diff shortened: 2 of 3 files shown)


VeckoTheGecko:
Have managed to very briefly look at this before EOD. If possible, I would like to make kernel behaviour completely robust to how kernel functions are imported. As such, I'm apprehensive to do string manipulations on contents of the kernel functions. If such robustness isn't possible, we at least need to have complete compatibility with prior documentation to avoid regressing and breaking existing code.

I'll investigate possibilities more next week.

erikvansebille:
> If such robustness isn't possible, we at least need to have complete compatibility with prior documentation to avoid regressing and breaking existing simulation code.

Yes I agree that we have to be careful here. Perhaps we should then also change `ParcelsRandom` -> `parcels.rng` throughout the tutorials and examples (while keeping support for `ParcelsRandom` at least until v4). That would also align with #1581

parcels/kernel.py
```diff
@@ -178,6 +179,8 @@ def __init__(self, fieldset, ptype, pyfunc=None, funcname=None, funccode=None, p
         else:
             self.funcvars = None
         self.funccode = funccode or inspect.getsource(pyfunc.__code__)
+        self.funccode = self.funccode.replace('parcels.rng', 'rng')  # Remove parcels. prefix (see #1608)
```
VeckoTheGecko, line 182: If this is the entry point, would it be easier to just replace this with 'ParcelsRandom'? It would result in less changes to the codebase and everything would just flow through as if its `ParcelsRandom` and also wouldn't interfere in the odd chance they already have a variable called `rng`. So maybe its just this one-liner?

VeckoTheGecko (changes requested):
Ok, I've managed to look through this fully now. I have also written some test cases (see 
[changes.patch](https://github.com/user-attachments/files/16410620/changes.patch); do `git apply changes.patch` to bring them into your working directory), so I'll refer to those hereon.

1. `test_explicit_ParcelsRandom` (i.e., `from parcels import ParcelsRandom`)
2. `test_parcels_dot_ParcelsRandom` (i.e., using `parcels.ParcelsRandom`)
3. `test_parcels_dot_rng` (i.e., using `parcels.rng`)
4.  `test_custom_ParcelsRandom_alias` (i.e., `from parcels import ParcelsRandom as my_custom_name`)
5. Doing direct function import (i.e., `from parcels.rng import normalvariate`). No tests written.

(1) was the only supported implementation before, and doing any other sort of import broke the kernel conversion code. #1608 is requiring us to extend functionality to (2) as well. Ideally, it would be Pythonic to support all 1-5 so that users don't get surprised, but looking into the code conversion code it doesn't look feasible/worth it. 

The invention of `ParcelsRandom` as an aliased module in itself (instead of using `parcels.rng`) I'm happy with as it highlights these RNGs are parcels specific, and also because this is familiar to users. Maybe in a V4, if there is already significant API changes, we can rename `rng.py` to `parcels_random.py` and then remove the alias, but that's a Python convention nitpick. I'm not sure what you mean by "traceability", so maybe you could elaborate there.

In the meanwhile, I don't mind only supporting `ParcelsRandom` (hence (3) shouldn't work; which is reflected in the test case I wrote). (4) and (5) shouldn't work due to limitations.

parcels/kernel.py
erikvansebille: True, but I actually like to start using `parcels.rng` a bit more throughout the codebase, as `ParcelsRandom` was just an alias around it. Hence the other changes

erikvansebille:
Thanks for the patch, @VeckoTheGecko. I adjusted it a bit (adding an assert so that we can check that the random number is indeed added as expected) and pushed it in https://github.com/OceanParcels/parcels/pull/1618/commits/b12b3abd8bc2f150ac9ee60b5b543cd2c4df70ed

Note that your third test _does_ work, because we do import `rng`. This is handled in https://github.com/OceanParcels/parcels/blob/b12b3abd8bc2f150ac9ee60b5b543cd2c4df70ed/parcels/kernel.py#L193

I prefer to move to `parcels.rng` throughout the codebase (which means a bit more changes, but clarifies that this is associated with the `parcels/rng.py` file)

parcels/kernel.py
VeckoTheGecko: I see. Sounds good!

VeckoTheGecko (approved):

PR #1627, opened 2024-07-30
Improve pset.execute time management

### LINKED ISSUE #1625: Smarter time management in pset.execute() loop
The time management in the `pset.execute()` loop can be made a bit smarter. For example

1. If the `pset` is empty, execution can in principle be stopped. Since we don't have a possibility to add particles at runtime (except if `repeatdt` is used), a run with an empty `pset` can be stooped. This would also avoid the warning (in particlefile that "ParticleSet is empty on writing as array at time")
2. If there are no particle to be executed in the next loop, that loop can be fast-forwarded. For example, if the time counter is at t=3 (with a dt=1) and the next particle to start is at t=20, then the time can directly be moved to t=20. This saves I/O time for loading in unnecessary Field data

erikvansebille:
This PR fixes #1625


VeckoTheGecko:
converting to draft until I've managed to take a look at refactoring the pset.execute() loop

VeckoTheGecko (approved):

Issue #1620, opened 2024-07-26
Remove v2 backward compatability


VeckoTheGecko:

erikvansebille:
Thanks for flagging this, @VeckoTheGecko. I quickly browsed through the code, but don't know of any other pre-v2 or even pre-v3 code snippets.

VeckoTheGecko:
This was primarily for v3. Relevant for v4 - but will be addressed in other issues

PR #1638, opened 2024-08-05
Migrate to Ruff, and extend black formatting

### LINKED ISSUE #1609: Migrate to ruff tooling for Python linting and formatting
Currently we're running `flake8`, `isort`, `pydocstyle`, and (notebooks only) `black` via pre-commit. With the deprecation of the [`pydocstyle` project,](https://github.com/pycqa/pydocstyle) I think its time to migrate to the [ruff linter/code formatter](https://docs.astral.sh/ruff/).

This would also act as a replacement to these other tools, greatly cleaning up our pre-commit configuration, future proofing development, and aligning ourselves with modern Python practices.

VeckoTheGecko:
- Closes #1609 
- Contributes to #1620

See [here](https://docs.astral.sh/ruff/rules/) for a full list of rules supported by Ruff. The ruleset chosen is nearly identical to the one that was already in the codebase.


erikvansebille (approved):
Changes look good to me. 

Note we will also need to update the list of GHA status checks that are required, in https://github.com/OceanParcels/parcels/settings/branch_protection_rules/315306.

VeckoTheGecko:
> update the list of GHA status checks

Sounds good. I've now updated it to require pre-commit to pass.

PR #1641, opened 2024-08-07
Support for CROCO 3D velocities

erikvansebille:
This PR implements support for 3D velocities in CROCO (and potentially also ROMS, but not tested). Users can simply call `FieldSet.from_croco()` and provide the `w` field from CROCO output in their `variables` list and call the `AdvectionRK4_3D` kernel. Under the hood, Parcels will then replace the kernel with `AdvectionRK4_3D_CROCO`, which works slightly different from the normal 3D advection kernel because it converts the vertical velocity in sigma-units. 

In particular, the following algorithm is used (note that the RK4 version is slightly more complex than this Euler-Forward version, but the idea is identical)

```python
    # calculate local sigma level of particle, by scaling depth by local ocean depth H
    sigma = particle.depth / fieldset.H[time, particle.depth, particle.lat, particle.lon]

    (u, v, w) = fieldset.UVW[time, particle.depth, particle.lat, particle.lon, particle]  

    # scaling the w with the sigma level of the particle
    w1 *= sigma 

    lon_new = particle.lon + u*particle.dt
    lat_new = particle.lat + v*particle.dt

    # calculating new sigma level
    sigma_new = sigma + w/fieldset.H[time, particle.depth, particle.lat, particle.lon]*particle.dt 
    
    # Converting back from sigma to depth, at _new_ location
    depth_new = sigma_new * fieldset.H[time, particle.depth, lat_new, lon_new]  
```

We have tested this algorithm on an idealised setup (created by @vesnaber) of a purely zonal flow over a sloping bottom. Particles indeed stay on their depth levels (so cross sigma-levels) as expected, with this algorithm. So we think this is an appropriate scheme.

To make this work under the hood, we had to adapt Parcels to convert from particle depth (in m) to sigma-coordinates in Field interpolations. In a call such as `fieldset.U[time, particle.depth, particle.lat, particle.lon]`, the `particle.depth` will thus under-the-hood be converted to `sigma` (`= particle.depth / fieldset.H[time, particle.depth, particle.lat, particle.lon]`).

Todo before this PR can be merged
- [x] Further implement `FieldSet.from_croco()` for seamless use
- [x] Test 2D CROCO surface flow too (so without providing `w` field)
- [x] Create tutorial how to use Parcels with CROCO
- [x] Support for sampling of individual Fields (not velocities)
- [x] Test the new algorithm in more realistic simulations


parcels/tools/exampledata_utils.py
```diff
@@ -72,6 +72,7 @@
         "field_0065557.nc",
     ],
     "WOA_data": [f"woa18_decav_t{m:02d}_04.nc" for m in range(1, 13)],
+    "CROCOidealized_data": ["CROCO_idealized.nc"],
```
VeckoTheGecko, line 75: Should we remove this dataset from the 'test_data' folder and just rely on the data fetching for both the notebook and the test cases?

tests/test_data/fieldset_CROCO2D.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+import parcels
+
+
+def create_fieldset(indices=None):
```
VeckoTheGecko, line 6:
```suggestion
def create_fieldset():
```

tests/test_advection.py
```diff
@@ -193,6 +195,44 @@ def test_advection_RK45(lon, lat, mode, rk45_tol):
     print(fieldset.RK45_tol)
 
 
+@pytest.mark.parametrize("mode", ["scipy", "jit"])
+def test_advection_3DCROCO(mode):
+    data_path = os.path.join(os.path.dirname(__file__), "test_data/")
+    fieldset = FieldSet.from_modulefile(data_path + "fieldset_CROCO3D.py")
```
VeckoTheGecko, line 201: Use `from tests.utils import TEST_DATA` instead of creating `data_path`


```suggestion
    fieldset = FieldSet.from_modulefile(TEST_DATA / "fieldset_CROCO3D.py")
```

parcels/fieldfilebuffer.py
```diff
@@ -105,6 +106,9 @@ def lonlat(self):
         else:
             xdim = lon.size if len(lon.shape) == 1 else lon.shape[-1]
             ydim = lat.size if len(lat.shape) == 1 else lat.shape[-2]
+            if self.gridindexingtype in ["croco"]:
+                xdim -= 1
+                ydim -= 1
```
VeckoTheGecko, line 111: I guess this is due to how croco is represented under the hood?

parcels/compilation/codegenerator.py
```diff
@@ -819,6 +826,16 @@ def visit_FieldEvalNode(self, node):
         self.visit(node.field)
         self.visit(node.args)
         args = self._check_FieldSamplingArguments(node.args.ccode)
+        statements_croco = []
+        if "croco" in node.field.obj.gridindexingtype and node.field.obj.name != "H":  # TODO needs to be sigma
```
VeckoTheGecko, line 830: is this TODO still current?

parcels/compilation/codegenerator.py
```diff
@@ -421,6 +426,8 @@ def __init__(self, fieldset=None, ptype=JITParticle):
         self.fieldset = fieldset
         self.ptype = ptype
         self.field_args = collections.OrderedDict()
+        if isinstance(fieldset.U, Field) and fieldset.U.gridindexingtype == "croco" and hasattr(fieldset, "H"):
+            self.field_args["H"] = fieldset.H  # CROCO requires H field
```
VeckoTheGecko, line 430: Could you elaborate on this if statement? The isinstance check for U being a field feels out of place -- do we expect cases where U isn't a field? Do we expect `hasattr(fieldset, 'H') == False` at this point of the code?

tests/test_fieldset_sampling.py
```diff
@@ -616,6 +617,21 @@ def test_sampling_out_of_bounds_time(mode, allow_time_extrapolation):
             pset.execute(SampleP, runtime=0.1, dt=0.1)
 
 
+@pytest.mark.parametrize("mode", ["scipy", "jit"])
+def test_sampling_3DCROCO(mode, npart=10):
+    data_path = os.path.join(os.path.dirname(__file__), "test_data/")
+    fieldset = FieldSet.from_modulefile(data_path + "fieldset_CROCO3D.py")
```
VeckoTheGecko, line 623:
```suggestion
def test_sampling_3DCROCO(mode):
    data_path = os.path.join(os.path.dirname(__file__), "test_data/")
    fieldset = FieldSet.from_modulefile(data_path + "fieldset_CROCO3D.py")
```

parcels/compilation/codegenerator.py
erikvansebille: No, was a stray comment. Removed the TODO

parcels/fieldfilebuffer.py
erikvansebille: Yes exactly; the last row and column can't be used in croco

tests/test_data/fieldset_CROCO2D.py
erikvansebille: Oops, forgot to propagate indices as keyword argument to `parcels.FieldSet.from_croco()`

parcels/compilation/codegenerator.py
erikvansebille: There are cases when `fieldSet.U` is a `NestedField`, this check avoided querying the `gridindexingtype` attribute for these cases.

tests/test_data/fieldset_CROCO2D.py
```diff
@@ -0,0 +1,22 @@
...
+    }
+    fieldset = parcels.FieldSet.from_croco(
+        file,
+        variables,
+        dimensions,
+        allow_time_extrapolation=True,
+        mesh="flat",
+    )
+
+    return fieldset
```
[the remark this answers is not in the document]
erikvansebille: Well, this is also an opportunity to test the 'modulefile system'. We have a few more of these module files in `test_data`, so I'd prefer to keep it here?

parcels/tools/exampledata_utils.py
erikvansebille: Good idea; done now

parcels/fieldfilebuffer.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         self.cast_data_dtype = kwargs.pop("cast_data_dtype", np.float32)
         self.ti = None
         self.interp_method = interp_method
+        self.gridindexingtype = kwargs.pop("gridindexingtype", "nemo")
```
VeckoTheGecko, line 38: Should we update `_FileBuffer` signature and associated init?

```diff

class _FileBuffer:
    def __init__(
        self,
        filename,
        dimensions,
        indices,
        timestamp=None,
        interp_method: InterpMethodOption = "linear",
        data_full_zdim=None,
-        **kwargs,
+       cast_data_dtype = np.float32,
+       gridindexingtype = "nemo",
    ):
    ...
```
I would like to shift away from using `kwargs` where named arguments are better suited (i.e, unless we're using them to propagate through to other functions, or need arbitrarily named keyword arguments. [SO post talking about it](https://stackoverflow.com/a/1415830/15545258)

VeckoTheGecko (changes requested):

tests/test_data/fieldset_CROCO2D.py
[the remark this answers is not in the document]
VeckoTheGecko: Sounds good

parcels/compilation/codegenerator.py
erikvansebille: Well, that test is already done here
https://github.com/OceanParcels/Parcels/blob/c376f0d2056f1bafbd65b08acef6266732a69dd3/parcels/fieldset.py#L738-L740

parcels/fieldfilebuffer.py
erikvansebille: OK, but should we do that here, or in another PR/refactor? Note that the `DaskFileBuffer(NetcdfFileBuffer)` also has some `kwargs.pop()` calls; and purest breaks if we don't put these also in the `_FileBuffer()` class.
https://github.com/OceanParcels/Parcels/blob/c376f0d2056f1bafbd65b08acef6266732a69dd3/parcels/fieldfilebuffer.py#L304-L307

So it's a bit more work than just this change?

VeckoTheGecko: yeah, I think migrating away from `kwargs` entirely for all classes would be difficult given some of the inheritence structures that we have. This comment was mainly for the `_FileBuffer()` which I think is a simple fix.

Other changes here can be left for future refactorings (which will hopefully also include removing this large inheritence structure, as it results in a lot of coupling between the classes -- I think a composition approach would be worth exploring)

parcels/compilation/codegenerator.py
erikvansebille: No we can't, because CROCO does not have an H field if it is run in _2D mode_. The check above is only if there is a `depth` dimension. 

The alternative is that we use that same test here, but I'm not sure how easy it is to check if there is a depth dimension at this point in the codeconverter...

VeckoTheGecko: Ah, I see. All good

VeckoTheGecko (approved):

PR #1644, opened 2024-08-08
Revert from dirHtml builder, update version default, clarify support for fieldset=None in ParticleSet API, doc copy edits

VeckoTheGecko:
**Code**
- Officially remove support for fieldset=None in ParticleSet API
- Update version retrieval to avoid git call by default
- pset.execute() notes and clean up some variable instantiation
  - Will delay a refactoring until I have a better overview of the codebase

**Docs**
- Revert from dirHtml builder to original builder
  - I was encountering difficulty with the version switcher was not linking the pages properly between the new and old documentation versions, leading to people encountering 404s. I thought this could be accounted for fully using redirects, but I don't think so.
- Add version warning on tutorial homepage
- Copy edits to `parcels_tutorial.ipynb`


VeckoTheGecko:
Looks like making the output less verbose is more complicated than I thought 😂. The problem I think is the progress bar in `.execute()`. Leaving for now

parcels/particleset.py
```diff
@@ -823,21 +816,25 @@ def execute(self, pyfunc=AdvectionRK4, pyfunc_inter=None, endtime=None, runtime=
...
         callbackdt :
             Optional, in conjecture with 'postIterationCallbacks', timestep interval to (latest) interrupt the running kernel and invoke post-iteration callbacks from 'postIterationCallbacks' (Default value = None)
         pyfunc_inter :
             (Default value = None)
         delete_cfiles : bool
             Whether to delete the C-files after compilation in JIT mode (default is True)
+
+        Notes
+        -----
+        Parcels interoperates between Python and (in the case of JIT mode) compiled C code. ParticleSet.execute() acts as the main entrypoint for simulations, and provides the simulation time-loop. This method encapsulates logic deciding when control flow is handled to C (for kernel execution), and Python (for writing output files, reading in fields for new timesteps, for adding new particles to the simulation domain, for stopping the simulation, and for executing custom functions `postIterationCallbacks` provided by the user).
```
erikvansebille, line 837: It is not entirely true that the logic for deciding between Scipy and JIT is in ParticleSet.execute(). That's actually in the Kernel class. But users don't per se interact with it, so for them it makes sense this way

parcels/particleset.py
```diff
@@ -927,28 +924,34 @@ def execute(self, pyfunc=AdvectionRK4, pyfunc_inter=None, endtime=None, runtime=
...
+        next_callback = starttime * np.sign(dt)
+
         tol = 1e-12
-        while (time < endtime and dt > 0) or (time > endtime and dt < 0) or dt == 0:
+        time = starttime
+
+        while (time < endtime and dt > 0) or (time > endtime and dt < 0):
             time_at_startofloop = time
 
+            next_input = self.fieldset.computeTimeChunk(time, dt) if self.fieldset is not None else np.inf
```
erikvansebille, line 952: So here fieldset can still be None?

parcels/particleset.py
VeckoTheGecko: Happy to edit this. Any suggestion on a rewording?

parcels/particleset.py
```diff
@@ -165,18 +162,14 @@ def ArrayClass_init(self, *args, **kwargs):
 
         if time.size > 0 and type(time[0]) in [datetime, date]:
             time = np.array([np.datetime64(t) for t in time])
-        self.time_origin = fieldset.time_origin if self.fieldset is not None else 0
```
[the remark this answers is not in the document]
VeckoTheGecko: Yes, I moved it up to line 107

parcels/particleset.py
VeckoTheGecko: Fixed in b684c21

parcels/particleset.py
```diff
@@ -823,21 +809,25 @@ def execute(self, pyfunc=AdvectionRK4, pyfunc_inter=None, endtime=None, runtime=
...
         callbackdt :
             Optional, in conjecture with 'postIterationCallbacks', timestep interval to (latest) interrupt the running kernel and invoke post-iteration callbacks from 'postIterationCallbacks' (Default value = None)
         pyfunc_inter :
             (Default value = None)
         delete_cfiles : bool
             Whether to delete the C-files after compilation in JIT mode (default is True)
+
+        Notes
+        -----
+        Parcels interoperates between Python and (in the case of JIT mode) compiled C code. ParticleSet.execute() acts as the main entrypoint for simulations, and provides the simulation time-loop. This method encapsulates logic deciding when control flow is handled to C (for kernel execution), and Python (for writing output files, reading in fields for new timesteps, for adding new particles to the simulation domain, for stopping the simulation, and for executing custom functions `postIterationCallbacks` provided by the user).
```
erikvansebille, line 830:
```suggestion
        ParticleSet.execute() acts as the main entrypoint for simulations, and provides the simulation time-loop. This method encapsulates the logic controlling the switching between kernel execution (where control in handed to C in JIT mode), output file writing, reading in fields for new timesteps, adding new particles to the simulation domain, stopping the simulation, and executing custom functions (`postIterationCallbacks` provided by the user).
```

docs/examples/tutorial_parcels_structure.ipynb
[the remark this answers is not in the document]
VeckoTheGecko: I'm all for _supporting_ `parcels.rng`, but I'm hesitant to _promote_ the use of it in notebooks as it might look quite unfamiliar to those who are familiar working with parcels. I think in v4 it would be good to officially deprecate `ParcelsRandom`.

Happy to discuss more about it.

parcels/particleset.py
VeckoTheGecko: Looks good! Updated the docstring to also use RST formatting.

erikvansebille (approved):

PR #1647, opened 2024-08-09
Ruff formatting

VeckoTheGecko:
Currently `pyproject.toml` specifies Python files for Ruff to ignore during formatting. During refactoring, some of these will become ready for formatting. In this sense, "ready for formatting" means that there aren't really long lines that will be spread out and become unreadable.

This draft PR is to be used to collect the names of these files so that Ruff can format it in one commit down the line.

- parcels/tools/global_statics.py
- 

Contributes to #1620


VeckoTheGecko:
Superceded by #1655

PR #1648, opened 2024-08-09
Enable Bugbear and other rules in Ruff linting

VeckoTheGecko:
Enables [Bugbear error codes in Ruff linting](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b).

> A plugin for finding likely bugs and design problems in your program. Contains warnings that don't belong in pyflakes and pycodestyle.

Leaving the codes B011, B016, B904 as I want to incorporate them into a larger PR at a later date looking at reworking the parcels error system (i.e., so we aren't raising Runtime and Assertion errors). Discussed in #1620

Also added in some other rules I thought would be good. Notably RUF100 which removes unused `# noqa` flags.

Contributes to #1620

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1269,8 +1269,8 @@ class Field:
         self.grid.load_chunk = np.zeros(npartitions, dtype=c_int, order='C')
         # self.grid.chunk_info format: number of dimensions (without tdim); number of chunks per dimensions;
         #      chunksizes (the 0th dim sizes for all chunk of dim[0], then so on for next dims
-        self.grid.chunk_info = [[len(self.nchunks)-1], list(self.nchunks[1:]), sum(list(list(ci) for ci in chunks[1:]), [])]
-        self.grid.chunk_info = sum(self.grid.chunk_info, [])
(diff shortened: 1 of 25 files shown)


parcels/field.py
```diff
@@ -1269,8 +1269,8 @@ def chunk_setup(self):
         self.grid.load_chunk = np.zeros(npartitions, dtype=c_int, order='C')
         # self.grid.chunk_info format: number of dimensions (without tdim); number of chunks per dimensions;
         #      chunksizes (the 0th dim sizes for all chunk of dim[0], then so on for next dims
-        self.grid.chunk_info = [[len(self.nchunks)-1], list(self.nchunks[1:]), sum(list(list(ci) for ci in chunks[1:]), [])]
-        self.grid.chunk_info = sum(self.grid.chunk_info, [])
+        self.grid.chunk_info = [[len(self.nchunks)-1], list(self.nchunks[1:]), sum(list(list(ci) for ci in chunks[1:]), [])]  # noqa: RUF017 # TODO: Perhaps avoid quadratic list summation here
+        self.grid.chunk_info = sum(self.grid.chunk_info, [])  # noqa: RUF017
```
erikvansebille, line 1273: Yep, good idea to do this (at a later stage?)

erikvansebille (approved):

parcels/field.py
VeckoTheGecko: Yes, we can do this at a later stage

PR #1656, opened 2024-08-13
Update smagdiff function

coquereau:
The Smagdiff function didn't update the particle's position because it didn't modify the lat and lon attributes of the "particle" object. This problem has now been solved.

### DIFF
--- a/docs/examples/tutorial_diffusion.ipynb
+++ b/docs/examples/tutorial_diffusion.ipynb
@@ -479,8 +479,8 @@
     "        2 * math.fabs(particle.dt) * Kh\n",
     "    )\n",
     "\n",
-    "    particle_dlat += dlat\n",
-    "    particle_dlon += dlon"
+    "    particle.lat += dlat\n",
+    "    particle.lon += dlon"
    ]
   },
   {


erikvansebille:
Thanks for your involvement in parcels development and for creating this PR, @coquereau! 

However, I'm not sure it does what you expect. Since Parcels v3.0, changes to particle position need to be given through he `particle_dlon` and `particle_dlat` variables; see [the kernel loop tutorial](https://docs.oceanparcels.org/en/latest/examples/tutorial_kernelloop.html#The-Parcels-Kernel-loop). This is the only way to facilitate vectorial summing of displacements.

If the Smagorinski kernel does not work for you, that's likely because you're running a version pre-v3.0? In that case, check the documentation for your version (using the version switcher on the bottom right) and you'll see the right version of the kernel at e.g. https://docs.oceanparcels.org/en/v2.4.2/examples/tutorial_diffusion.html (for version 2.4.2)

coquereau:
Thank you for your comments. 
You're absolutely right, I was using an older version of Parcels, which is why it wasn't working for me. I've updated it, everything works now.
Sorry for the loss of time. And thanks for Parcels!

VeckoTheGecko:
All good, we're glad you got it working 😄

PR #1667, opened 2024-08-22
Combine workflow files, add codecov token

VeckoTheGecko:
As in title. I just realised we don't have any coverage on unit tests in CodeCov since we were being rate limited before. Specifying a token should help with this.


VeckoTheGecko:
Yay! Bot integration looks to be working

erikvansebille (approved):

PR #1673, opened 2024-08-29
Refactoring of CField struct to include gridindexingtype and interp_method

erikvansebille:
Following #1612, where we need to provide the `gridindexingtype` on the `CField` struct for technical reasons, it makes sense to also use it throughout the rest of the C-code. Hence, this PR refactors the code to not require the `gridindexingtype` and `interp_method` when calling a Field evaluation in C, but get them from the `CField` itself

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -1371,7 +1371,7 @@ class Field:
 
     def ccode_eval(self, var, t, z, y, x):
         self._check_velocitysampling()
-        ccode_str = f"temporal_interpolation({x}, {y}, {z}, {t}, {self.ccode_name}, &particles->xi[pnum*ngrid], &particles->yi[pnum*ngrid], &particles->zi[pnum*ngrid], &particles->ti[pnum*ngrid], &{var}, {self.interp_method.upper()}, {self.gridindexingtype.upper()})"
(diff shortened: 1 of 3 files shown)


parcels/field.py
```diff
@@ -1476,6 +1477,18 @@ class CField(Structure):
...
+            "NEAREST": 1,
+            "CGRID_VELOCITY": 2,
+            "CGRID_TRACER": 3,
+            "BGRID_VELOCITY": 4,
+            "BGRID_W_VELOCITY": 5,
+            "BGRID_TRACER": 6,
+            "LINEAR_INVDIST_LAND_TRACER": 7,
+            "PARTIALSLIP": 8,
+            "FREESLIP": 9,
+        }
```
erikvansebille, line 1491: This is still a bit ugly, because it mirrors the information from 
https://github.com/OceanParcels/Parcels/blob/997e3f86399cf319341a3d3f4fb1c5dc339e2580/parcels/include/index_search.h#L22-L30

It would be cleaner to either
1. Use the string here in the CField, instead of an integer. I tried using `c_char_p`, but that didn't work (yet)
2. Somehow import the typedefs from `index_search.h` (above) here
In any case, I think we want to avoid happing two places where these mappings between strings and integers occur...

erikvansebille:
Will be solved in v4

PR #1680, opened 2024-08-29
Enable pyupgrade, add Grid repr, and other changes

VeckoTheGecko:
- Upgrade to f-strings (wasn't done automatically by pyupgrade)
- Enable pyupgrade
- Remove code targeting Python 2
- Add repr for Grid class (and some tooling for testing reprs)
- Type annotations and typos

related to #1620 and #1693

## Breaking changes
- `Field.from_netcdf()` no longer supports `full_load` as a parameter. This was a legacy parameter from before version 2 of Parcels and hasn't been documented in tutorials and docstrings since. Use `deferred_load` instead.

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -1317,7 +1317,7 @@ class FieldSet:
         """
         # check if filename exists
         if not os.path.exists(filename):
-            raise IOError(f"FieldSet module file {filename} does not exist")
+            raise OSError(f"FieldSet module file {filename} does not exist")
 
         # Importing the source file directly (following https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly)
         spec = importlib.util.spec_from_file_location(modulename, filename)
@@ -1326,10 +1326,10 @@ class FieldSet:
         spec.loader.exec_module(fieldset_module)
 
         if not hasattr(fieldset_module, modulename):
-            raise IOError(f"{filename} does not contain a {modulename} function")
+            raise OSError(f"{filename} does not contain a {modulename} function")
         fieldset = getattr(fieldset_module, modulename)(**kwargs)
         if not isinstance(fieldset, FieldSet):
-            raise IOError(f"Module {filename}.{modulename} does not return a FieldSet object")
+            raise OSError(f"Module {filename}.{modulename} does not return a FieldSet object")
         return fieldset
 
     def get_fields(self):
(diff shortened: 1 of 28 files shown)


parcels/fieldset.py
```diff
@@ -1326,10 +1326,10 @@ def from_modulefile(cls, filename, modulename="create_fieldset", **kwargs):
         spec.loader.exec_module(fieldset_module)
 
         if not hasattr(fieldset_module, modulename):
-            raise IOError(f"{filename} does not contain a {modulename} function")
+            raise OSError(f"{filename} does not contain a {modulename} function")
```
erikvansebille, line 1329: Is this really an OSError? Perhaps better a ValueError?

parcels/fieldset.py
```diff
@@ -1326,10 +1326,10 @@ def from_modulefile(cls, filename, modulename="create_fieldset", **kwargs):
...
         fieldset = getattr(fieldset_module, modulename)(**kwargs)
         if not isinstance(fieldset, FieldSet):
-            raise IOError(f"Module {filename}.{modulename} does not return a FieldSet object")
+            raise OSError(f"Module {filename}.{modulename} does not return a FieldSet object")
```
erikvansebille, line 1332: Is this really an OSError? Perhaps better a ValueError?

parcels/kernel.py
```diff
@@ -239,9 +239,10 @@ def __init__(
 
         numkernelargs = self.check_kernel_signature_on_version()
 
-        assert (
-            numkernelargs == 3
-        ), "Since Parcels v2.0, kernels do only take 3 arguments: particle, fieldset, time !! AND !! Argument order in field interpolation is time, depth, lat, lon."
+        if numkernelargs != 3:
+            raise ValueError(
+                "Since Parcels v2.0, kernels do only take 3 arguments: particle, fieldset, time !! AND !! Argument order in field interpolation is time, depth, lat, lon."
```
erikvansebille, line 244: I thought we wanted to keep this `ValueError`, because otherwise running old Kernels can really get users into trouble...

parcels/fieldset.py
VeckoTheGecko: This was just an update by pyupgrade. I'd prefer to give attention to all the errors in one go (as there are a lot of assertion errors that need to turn into ValueErrors)

parcels/kernel.py
VeckoTheGecko: Just changed it from assertion to value error. TODO: Add unit test so we avoid regression

parcels/fieldset.py
VeckoTheGecko: This was just an update by pyupgrade. I'd prefer to give attention to all the errors in one go (as there are a lot of assertion errors that need to turn into ValueErrors)

erikvansebille (approved):

PR #1683, opened 2024-09-03
Add deprecation decorators

VeckoTheGecko:
Adds deprecation decorators `@deprecated()` and `@deprecated_made_private`. Note the brackets on `deprecated()` when using it. This is because it accepts an optional custom message that is appended to the warning message (to be used to denote an alternative to users, or any other information that may be of use).

Note: When using in conjunction with other decorators (e.g., `@staticmethod` `@property`) the deprecation method should be the innermost one (i.e., closest to the `def ...`) so that they're applied first.

Examples _(updated)_:
```python
from parcels.tools._helpers import deprecated, deprecated_made_private

class SomeClass:
    @deprecated("Please use some_method_other_method instead.") # deprecated in v3.1 # TODO: Remove in future version
    def some_method(self, x, y):
        return x + y
    
    @deprecated_made_private # deprecated in v3.1 # TODO: Remove in future version
    def private_method(self):
        pass


x = SomeClass().some_method(1, 2)
SomeClass().private_method()

print(x)
```
output:
```
<filename>:<line_num>: DeprecationWarning: `SomeClass.some_method` is deprecated and will be removed in a future release of Parcels. Please use some_method_other_method instead.
  x = SomeClass().some_method(1, 2)
<filename>:<line_num>: DeprecationWarning: `SomeClass.private_method` is deprecated and will be removed in a future release of Parcels. It was identified as part of the internal API as it is not expected to be directly used by the end-user. If you feel that you use this code directly in your scripts, please comment on our tracking issue at <>.
  SomeClass().private_method()

3
```


VeckoTheGecko:
The creation of the `util` directory is a bit unjustified in this PR. Perhaps it would be better to put somewhere else. I wasn't too fussed about organisation since I anticipate needing to create more private only helper functions/modules and didn't mind kicking that can down the road. Open to ideas :')

parcels/util/_helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    def some_old_function(x, y):
+        return x + y
+
+    @deprecated()
+    def some_other_old_function(x, y):
+        return x + y
+    ```
+    """
+    if msg:
+        msg = " " + msg
```
erikvansebille, line 31: Should this not be an f-string?

tests/util/test_helpers.py
```diff
@@ -0,0 +1,56 @@
+import pytest
+
+from parcels.util._helpers import deprecated, deprecated_made_private
+
+# Write me tests to test deprecated and depreacted_made_private decorators
+# They should raise a DeprecationWarning, and also have the function name (or in the case of a method, {class_name}.{method_name}) in the warning message
```
erikvansebille, line 6: Is there a reason you leave this (GenAI prompt?) here?

parcels/util/_helpers.py
VeckoTheGecko: I think this is one of the first cases I've had where I found `+` to be more readable (just reading like english; "adding a space to the front"). Happy to update if you disagree.

tests/util/test_helpers.py
VeckoTheGecko: Nope! Will update

parcels/util/_helpers.py
erikvansebille: OK, is fine then

erikvansebille:
> The creation of the `util` directory is a bit unjustified in this PR. Perhaps it would be better to put somewhere else. I wasn't too fussed about organisation since I anticipate needing to create more private only helper functions/modules and didn't mind kicking that can down the road. Open to ideas :')

I'm fine with creating this subdirectory; we can move other tests in there too at a later stage

VeckoTheGecko:
My motivation for creating the `parcels/util` directory was more "ah, we'll probably need some more internal tooling down the line". On second thought - rather than anticipating a need that might not arise - I think it would be better to just put it in `tools`. Since its internal tooling we can move it at a later stage without expecting disruptions.

> we can move other tests in there too at a later stage

I didn't quite understand this comment. Do we need to reconsider the folder structure of the tests?

erikvansebille:
> I didn't quite understand this comment. Do we need to reconsider the folder structure of the tests?

Maybe it was a bit vague. I meant to say we can reconsider where to put which test files at a later stage. Like what you suggest above, so all fine

erikvansebille (approved):

PR #1702, opened 2024-09-12
Adding explicit explanation how to add tracers in FieldSet.from_nemo()

### LINKED ISSUE #1700: Adding information on how to define dimensions for tracers on C-grid
In Parcels, when using c-grid data, dimensions for tracers (T, S) must be specified identical to those for U,V,W [Erik, personal communication]. It would be great to add respective info to the documentation and (NEMO/MITgcm) tutorials.

erikvansebille:
Fixes #1700


docs/examples/tutorial_nemo_3D.ipynb
```diff
@@ -161,7 +171,9 @@
    "source": [
     "## Adding other fields like cell edges\n",
     "\n",
-    "It is quite straightforward to add other gridded data, on the same curvilinear or any other type of grid, to the fieldset. Because it is good practice to make no changes to a `FieldSet` once a `ParticleSet` has been defined in it, we redefine the fieldset and add the fields with the cell edges from the coordinates file using `FieldSet.add_field()`.\n"
+    "It is quite straightforward to add other gridded data, on the same curvilinear or any other type of grid, to the fieldset. Because it is good practice to make no changes to a `FieldSet` once a `ParticleSet` has been defined in it, we redefine the fieldset and add the fields with the cell edges from the coordinates file using `FieldSet.add_field()`.\n",
+    "\n",
+    "Noe that including tracers like `temperature` and `salinity` needs to be done at the f-points, so on the same grid as the velocity fields. See also the section above.\n"
```
VeckoTheGecko:
```suggestion
    "Note that including tracers like `temperature` and `salinity` needs to be done at the f-points, so on the same grid as the velocity fields. See also the section above.\n"
```

VeckoTheGecko (commented):
Are there any usecases then for providing T-points to the Nemo constructor?

Do you think we should also add additional validation to the `from_nemo` function? Maybe if there's no usecase for T-points, we can warn if they provide >2 unique dimensions for grids.

erikvansebille:
> Do you think we should also add additional validation to the `from_nemo` function? Maybe if there's no usecase for T-points, we can warn if they provide >2 unique dimensions for grids.

Well, users may want to add wind or stokes drift from other models, that would then be on a different grid. So I don't think we can enforce only 1 Grid when the velocities are from_nemo

VeckoTheGecko (approved):

PR #1736, opened 2024-10-22
Refactors and bugfixes

VeckoTheGecko:
- `timedelta_to_float` helper
- patch raise statements
- update var names to align with conventions
- type annotations
- #1620 related cleanups

### DIFF
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -561,9 +561,11 @@ class Kernel(BaseKernel):
     def cleanup_remove_files(lib_file, all_files_array, delete_cfiles):
         if lib_file is not None:
             if os.path.isfile(lib_file):  # and delete_cfiles
-                [os.remove(s) for s in [lib_file] if os.path is not None and os.path.exists(s)]
-            if delete_cfiles and len(all_files_array) > 0:
-                [os.remove(s) for s in all_files_array if os.path is not None and os.path.exists(s)]
+                os.remove(lib_file)
+            if delete_cfiles:
+                for s in all_files_array:
+                    if os.path.exists(s):
+                        os.remove(s)
 
     @staticmethod
     def cleanup_unload_lib(lib):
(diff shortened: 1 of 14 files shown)


VeckoTheGecko:
Wanting to wrap some more refactorings into here

VeckoTheGecko:
I didn't fix the #1738 into this. I wanted to do a bit of a deeper refactoring of `.reltime` and the `TimeConverter` class in general, but was running into some roadblocks

parcels/kernel.py
```diff
@@ -562,9 +561,11 @@ def from_list(cls, fieldset, ptype, pyfunc_list, *args, **kwargs):
...
         if lib_file is not None:
             if os.path.isfile(lib_file):  # and delete_cfiles
-                [os.remove(s) for s in [lib_file] if os.path is not None and os.path.exists(s)]
-            if delete_cfiles and len(all_files_array) > 0:
-                [os.remove(s) for s in all_files_array if os.path is not None and os.path.exists(s)]
+                os.remove(lib_file)
+            if delete_cfiles:
+                for s in all_files_array:
+                    if os.path.exists(s):
+                        os.remove(s)
```
erikvansebille, line 568: Why don't we need to check for `os.path is not None` here anymore? Was that old, redundant code?

parcels/kernel.py
```diff
@@ -601,8 +602,7 @@ def load_fieldset_jit(self, pset):
                     g._load_chunk == g._chunk_loading_requested, g._chunk_loaded_touched, g._load_chunk
                 )
                 if len(g._load_chunk) > g._chunk_not_loaded:  # not the case if a field in not called in the kernel
-                    if not g._load_chunk.flags["C_CONTIGUOUS"]:
-                        g._load_chunk = np.array(g._load_chunk, order="C")
+                    g._load_chunk = np.array(g._load_chunk, order="C")
```
erikvansebille, line 605: Why is the if-statement not necessary anymore, here?

parcels/tools/_helpers.py
```diff
@@ -56,3 +59,13 @@ def deprecated_made_private(func: Callable) -> Callable:
 
 def patch_docstring(obj: Callable, extra: str) -> None:
     obj.__doc__ = f"{obj.__doc__ or ''}{extra}".strip()
+
+
+@np.vectorize
```
erikvansebille, line 64: Is it really necessary to vectorise this function? Is it really so compute-intense? 
Or is this a start of more of these decorators?

erikvansebille (approved):

PR #1744, opened 2024-10-25
Bugfix warn_outputdt_release_desync for NaN release times

erikvansebille:
Fixing bug when the `_warn_outputdt_release_desync()` warning is raised if particle release times are NaN

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -1078,7 +1078,7 @@ class ParticleSet:
         outputdt = output_file.outputdt if output_file else np.inf
         if isinstance(outputdt, timedelta):
             outputdt = outputdt.total_seconds()
-        if outputdt is not None:
+        if np.isfinite(outputdt) is not None:
             _warn_outputdt_release_desync(outputdt, self.particledata.data["time_nextloop"])
 
         if isinstance(callbackdt, timedelta):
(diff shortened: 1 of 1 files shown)


parcels/particleset.py
```diff
@@ -1078,7 +1078,7 @@ def execute(
         outputdt = output_file.outputdt if output_file else np.inf
         if isinstance(outputdt, timedelta):
             outputdt = outputdt.total_seconds()
-        if outputdt is not None:
+        if np.isfinite(outputdt) is not None:
```
VeckoTheGecko, line 1081:
```suggestion
        if np.isfinite(outputdt):
```

parcels/particleset.py
```diff
@@ -1235,7 +1235,7 @@ def execute(
 
 def _warn_outputdt_release_desync(outputdt: float, release_times: Iterable[float]):
     """Gives the user a warning if the release time isn't a multiple of outputdt."""
-    if any(t % outputdt != 0 for t in release_times):
+    if any((np.isfinite(t) and t % outputdt != 0) for t in release_times):
```
VeckoTheGecko, line 1238: Is this change required? If `t = np.nan` then `t % outputdt != 0` would always be False anyway.

VeckoTheGecko (changes requested):

parcels/particleset.py
erikvansebille: Well, without this extra statement the Parcels tutorial gives these warnings whenever the time is not set
Screenshot 2024-10-28 at 08 37 01

VeckoTheGecko: 👍  :)

VeckoTheGecko (approved):

VeckoTheGecko (approved):

PR #1751, opened 2024-11-05
Add Unstructured Grid class

fluidnumerics-joe:
Properties and procedures that are common to unstructured grids and structured grids are defined in the `BaseGrid` class. This includes `lon`,`lat`, `time`, `time_origin`, and `mesh` properties.

The `Grid` class is now defined as a type extension of `BaseGrid`. This configuration is compatible with the previous commit of Parcels (all existing tests pass). The `_zonal_periodic`, `_zonal_halo`, `_meridional_halo`, and `_lat_flipped` properties are seen as attributes specific to the structured grid `Grid` class.

The `UGrid` class is defined as a type extension of the `BaseGrid`. This child class adds the `face_node_connectivity` property that relates vertices to the 2-D lateral grid faces. The lon and lat attributes refer to the `node_lon` and `node_lat` that are defined in the `uxarray.Grid` structure; these are the corner node vertices. Note that there are other possible nodes, including the edge vertices (centers of the edges), and `face_vertices` that define the element centroids; these are currently not defined.


fluidnumerics-joe:
@erikvansebille @VeckoTheGecko 
I'm using this PR to provide visibility to some of the proposed changes to Parcels to support unstructured grids. So far, I've only inserted a `BaseGrid` class that the `Grid` and `UGrid` are child classes of. All of the tests under `tests/` pass, which is good.

From the state it's in, I'll need to work on a `creategrid` method for the `UGrid` class and I suppose it'd be ideal to sort out how we'll bring in `UXArray` support for loading the grid. 

This PR is meant to go to the ELPHE-WW Fork's main branch so that we have a safe "playground" to explore.

Your comments and feedback are most welcome before further changes are made.

erikvansebille:
Thanks @fluidnumerics-joe, for starting this work. Your plan sounds great to me. I quickly went through the code, and that also looks fine. Some issues to fix with the CI obviously, but I don't think these are huge problems

erikvansebille:
Will be reworked in v4

PR #1761, opened 2024-11-13
Benchmarking with asv

VeckoTheGecko:
This PR introduces benchmarking infrastructure to the project via [asv](https://asv.readthedocs.io/en/latest/). Benchmarks can be run on a pull request by adding the `run-benchmarks` label to it. Two environments will be created in the CI runner with the prior and proposed changes, and both suites of benchmarks will be run and compared against each other.

Note that this PR only has example benchmarks for the timebeing until we can discuss benchmarks of interest.

The running of the benchmarks in CI is only one aspect of the benchmarking (ie, only for core parcels functionality). Using asv, we can create different suites of benchmarks (e.g., one for CI, and one for more heavy simulations). The benefit of using asv is everything else that comes out of the box with it, some being:

- being able to run benchmarks easily across several commits, visualising them in a dashboard
- easily create and manage these benchmarking environments
- profiling support to dive into locations of slowdowns
- community support

Changes:
- asv configuration (conf file, benchmarks folder, and CI workflow)
- asv documentation (available via the community page in the maintainer section)

---

I have done some testing of the PR label workflow in https://github.com/VeckoTheGecko/parcels/pull/10 . We can only test this for PRs in OceanParcels/parcels when its in master

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,76 @@
+# This workflow was adapted from xarray. See licenses/XARRAY_LICENSE for license details.
+name: Benchmark
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
+  workflow_dispatch:
+
+env:
+  PR_HEAD_LABEL: ${{ github.event.pull_request.head.label }}
+
+jobs:
+  benchmark:
+    if: ${{ contains( github.event.pull_request.labels.*.name, 'run-benchmark') && github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
+    name: Linux
+    runs-on: ubuntu-20.04
+    env:
+      ASV_DIR: "./asv_bench"
+      CONDA_ENV_FILE: environment.yml
+
+    steps:
+      # We need the full repo to avoid this issue
+      # https://github.com/actions/checkout/issues/23
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up conda environment
+        uses: mamba-org/setup-micromamba@v2
+        with:
+          environment-file: ${{env.CONDA_ENV_FILE}}
+          environment-name: xarray-tests
+          cache-environment: true
+          cache-environment-key: "${{runner.os}}-${{runner.arch}}-py${{env.PYTHON_VERSION}}-${{env.TODAY}}-${{hashFiles(env.CONDA_ENV_FILE)}}-benchmark"
+          # add "build" because of https://github.com/airspeed-velocity/asv/issues/1385
+          create-args: >-
+            python-build
+            mamba
+
+      - name: Run benchmarks
+        shell: bash -l {0}
+        id: benchmark
+        env:
+          OPENBLAS_NUM_THREADS: 1
+          MKL_NUM_THREADS: 1
+          OMP_NUM_THREADS: 1
+          ASV_FACTOR: 1.5
+          ASV_SKIP_SLOW: 1
+        run: |
+          set -x
+          # ID this runner
+          asv machine --yes
+          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} ($PR_HEAD_LABEL)"
+          # Run benchmarks for current commit against base
+          ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
+          asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
+              | sed "/Traceback \|failed$\|PERFORMANCE DECREASED/ s/^/::error::/" \
+              | tee benchmarks.log
+          # Report and export results for subsequent steps
+          if grep "Traceback \|failed\|PERFORMANCE DECREASED" benchmarks.log > /dev/null ; then
+              exit 1
+          fi
+        working-directory: ${{ env.ASV_DIR }}
+
+      - name: Add instructions to artifact
+        if: always()
+        run: |
+          cp benchmarks/README_CI.md benchmarks.log .asv/results/
+        working-directory: ${{ env.ASV_DIR }}
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: asv-benchmark-results-${{ runner.os }}
+          path: ${{ env.ASV_DIR }}/.asv/results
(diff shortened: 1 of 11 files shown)


VeckoTheGecko:
@erikvansebille On the topic of performance, are you also experiencing it taking something like 10s occasionally to run `import parcels`?

erikvansebille:
> @erikvansebille On the topic of performance, are you also experiencing it taking something like 10s occasionally to run `import parcels`?

Yes, I also experience this slow import sometimes. Not sure why...

VeckoTheGecko:
Setting to draft until we have some actual benchmarks that we can include in this.

VeckoTheGecko:
From meeting:
We can use `tutorial_nemo_curvilinear.ipynb` as well

erikvansebille:
The Argo tutorial at https://docs.oceanparcels.org/en/latest/examples/tutorial_Argofloats.html is also a quite nice simulation for benchmarking, as it has a 'complex' kernel. It took approximately 20s to run on v3.1 in JIT mode, and no takes 50s on my local computer to run in Scipy-mode.

@danliba or @willirath, could you add the Argo tutorial to the benchmark stack?

asv_bench/benchmarks/Argofloat_benchmark.py
```diff
@@ -0,0 +1,97 @@
+## Argo float benchmark
```
willirath, line 1: Rename file to `benchmark_....py`?

asv_bench/benchmarks/Argofloat_benchmark.py
```diff
@@ -0,0 +1,97 @@
...
+        data = {
+            "U": np.ones((xdim, ydim, zdim), dtype=np.float32),
+            "V": np.zeros((xdim, ydim, zdim), dtype=np.float32),
+        }
+        data["U"][:, :, 0] = 0.0
+        fieldset = FieldSet.from_data(data, dimensions, mesh="flat", transpose=True)
+        fieldset.mindepth = fieldset.U.depth[0]
+
+        # Define a new Particle type including extra Variables
+        self.ArgoParticle = JITParticle.add_variables(
```
willirath, line 75: `ArgoParticle` does not need to be a class attribute. So drop the `self.` here.

asv_bench/benchmarks/Argofloat_benchmark.py
```diff
@@ -0,0 +1,97 @@
...
+                # Variable('temp', dtype=np.float32, initial=np.nan),
+            ]
+        )
+
+        self.pset = ParticleSet(fieldset=fieldset, pclass=ArgoParticle, lon=[0], lat=[0], depth=[0])
+
+        # combine Argo vertical movement kernel with built-in Advection kernel
+        self.kernels = [ArgoVerticalMovement, AdvectionRK4]
+
+    def time_run_single_timestep(self):
```
willirath, line 96: Running for only one time step doesn't make sense in the context of the Argo example, because we have behaviour playing out over ca. 10 days. So add another timing method for a longer integration.

willirath (changes requested):
@danliba I've left a few comments.

asv_bench/benchmarks/Argofloat_benchmark.py
willirath: I think this is what let the benchmark fail? Later the `ArgoParticle` is not defined when the filedset is created.

danliba: ok!

willirath:
So I've spent some time with this over the weekend. Here's a few insights:

- ASV implicitly assumes that benchmarks don't change. This is most obvious from the fact that `asv run` will always discover benchmarks from `$PWD/benchmarks/` with whatever is present in this directory at the time of invoking ASV. Hence adapting semantically identical benchmarks to changing API needs work in the `.setup()` method of benchmark suites.

- ASV's more shiny features (accumulation and publication of benchmarks results over long times and across multiple versions) have never really been adopted widely. All the scipy and pydata projects I've checked ([dask](https://github.com/dask/dask-benchmarks/commits/main/), [numpy](), [asv-runner](https://github.com/asv-runner)) have ASV benchmarks defined, but their auto generated and published reports are outdated for years. Also, all the example benchmark results given in the ASV docs didn't see any update for at least 4 years (see their [astropy example](https://www.astropy.org/astropy-benchmarks/#summarylist?sort=3&dir=desc), their [numpy example](https://pv.github.io/numpy-bench/#summarylist?sort=3&dir=desc), the [other numpy example](https://github.com/HaoZeke/asv-numpy) and their [scipy example](https://pv.github.io/scipy-bench/#summarylist?sort=3&dir=desc)).

- ASV has a way of compiling relative results of benchmarks for different commits taken in one run using `asv continuous`. This is what's done in the github [workflow from an earlier commit](https://github.com/OceanParcels/Parcels/blob/f8a37ef8d59a3e87a8d6dbf91878bda5ce3ceafd/.github/workflows/benchmarks.yml#L59) and is used in xarray for detecting performance degradations introduced PRs. This workflow is, however, rarely used and was only requested in [in 93 PRs out of more than 2000 PRs in the same time range](https://github.com/pydata/xarray/labels/run-benchmark).

- Then env management in ASV appears to be in a phase of restructuring and especially the (faster) `mamba` based envs only work with enforcing pretty old versions of Mamba. It looks as if the ASV devs have given up on Mamba in favour of moving towards `rattler` support. This, however, also is nowhere near fully implemented. As a result, the only env management that worked smoothly with Parcels' compiler dependencies and did not result in thousands of lines of warnings and errors was `"conda"`.

My recommendation is _to forget about anything along the lines of "automation", "continuous", etc._  ASV provides a great way of defining benchmarks, running them against a bunch of similar well-behaved (i.e. no API changes or dependency changes) commits and then comparing the results with a rather narrow scope along the time / version axis.

VeckoTheGecko:
> My recommendation is _to forget about anything along the lines of "automation", "continuous", etc._
> 
> ASV provides a great way of defining benchmarks, running them against a bunch of similar well-behaved (i.e. no API changes or dependency changes) commits and then comparing the results with a rather narrow scope along the time / version axis. We should focus on this core functionality and leave the step of running and interpreting the benchmarks to the individual dev for now.

Yes, I think this sounds good. I was talking with an xarray maintainer, and he mentioned that ASV isn't really used much in CI - the intermitency of github runners adds quite some noise which means that unless a regression degrades performance by a factor, it won't be obvious in output.



> ASV provides a great way of defining benchmarks, running them against a bunch of similar well-behaved (i.e. no API changes or dependency changes) commits and then comparing the results with a rather narrow scope along the time / version axis.

I think so as well. Allowing us to define benchmarks for v3, port them to v4, then work in v4 with them will allow devs to have a targeted lens on performance (even if its just local - which is perfectly fine).

VeckoTheGecko:
Another note: As part of this PR can we remove dependence on the `parcels/tools/timer.py::Timer` class? (which is currently used in a couple examples as simple benchmarking)

Ideally once this is merged we can completely remove the `Timer` class from the codebase

PR #1767, opened 2024-11-20
CI: micromamba and environment caching

VeckoTheGecko:
Migrating to micromamba with environment caching to reduce CI times.

Environment cache key is based on:
- OS
- runner architecture
- python version
- date
- environment file

### DIFF
--- a/.github/actions/install-parcels/action.yml
+++ b/.github/actions/install-parcels/action.yml
@@ -1,7 +1,6 @@
 name: Setup Conda and install parcels
 description: >
-  In-repo composite action to setup Conda and install parcels. Installation of parcels relies on
-  `setup.py` file being available in the root. For general setup of Anaconda environments, just use
+  In-repo composite action to setup micromamba and install parcels. For general setup of Anaconda environments, you can also use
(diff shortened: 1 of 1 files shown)


VeckoTheGecko:
Well, looks like it only saves a minute (and its the windows tests that are the main slowdown anyway, and those aren't affected).

---

<details><summary>Details</summary>
<p>

Before:
image

After (cold start):
image

After (warm start):
image</p>
</details>

VeckoTheGecko:
Thoughts on this @erikvansebille ? No significant speedup, but I do have a very slight preference for micromamba for CI . I'm happy either way, just want to clear this out of the open PRs :)

erikvansebille:
> Thoughts on this @erikvansebille ? No significant speedup, but I do have a very slight preference for micromamba for CI . I'm happy either way, just want to clear this out of the open PRs :)

I'd be prefer to keep the installation method of our CI as close to that advised in the [installation instructions for users/developers](https://docs.oceanparcels.org/en/latest/installation.html), so that our CI also tests installation procedure. So rather stick with conda for now...

VeckoTheGecko:
Roger that :)

PR #1770, opened 2024-11-25
Fixture in `test_examples.py` to clean up generated data files

### LINKED ISSUE #1764: Avoiding pytest suite side-effects (dataset saving)
Running the test suite creates the following zarr files:

```
./Peninsula.zarr
./MIT_particles_scipy.zarr
./nemo_particles.zarr
./DecayingMovingParticle.zarr
./StommelParticle.zarr
./EddyParticle.zarr
./RadialParticle.zarr
./MIT_particles_jit.zarr
```

cluttering the working directory. Changing the tests to clean up these files after the tests have run (e.g., by using temporary files) would improve the quality of the test suite.

VeckoTheGecko:
Fixes #1764

No more side effects from running the test suite 🎉

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1,11 +1,41 @@
+import os
 import runpy
+import shutil
 import sys
+import time
 from pathlib import Path
 
 import pytest
 
 example_folder = (Path(__file__).parent / "../docs/examples").resolve()
 example_fnames = [path.name for path in example_folder.glob("*.py")]
+
+
+@pytest.fixture(autouse=True)
+def cleanup_generated_data_files():
+    """Clean up generated data files from test run.
+
+    Records current folder contents before test, and cleans up any generated `.nc` files
+    and `.zarr` folders afterwards. For safety this is non-recursive. This function is
+    only necessary as the scripts being run aren't native pytest tests, so they don't
+    have access to the `tmpdir` fixture.
+
+    """
+    folder_contents = os.listdir()
+    yield
+    time.sleep(0.1)  # Buffer so that files are closed before we try to delete them.
+    for fname in os.listdir():
+        if fname in folder_contents:
+            continue
+        if not (fname.endswith(".nc") or fname.endswith(".zarr")):
+            continue
+
+        path = Path(fname)
+        if path.is_dir():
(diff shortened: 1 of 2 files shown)


VeckoTheGecko:
Not 100% sure if the buffer is necessary. I get a "PermissionError" on a file one time but that was a once off.

tests/test_examples.py
```diff
@@ -8,6 +11,33 @@
...
+            continue
+        if not (fname.endswith(".nc") or fname.endswith(".zarr")):
+            continue
+
+        path = Path(fname)
+        if path.is_dir():
+            shutil.rmtree(path)
+        else:
+            path.unlink()
+        print(f"Removed {path}")
```
erikvansebille, line 38: Is this print really necessary?

erikvansebille (commented):
Nice solution! Will this now also work on other files in the test_suite (e.g. test_particlfile.py, or many of the notebooks in docs/examples/*.ipynb that generate output zarr files)? Or does the fixture only work in the file (test_examples.py) in which it is created? Would be nice to have this for the entire test_suite?

VeckoTheGecko:
The fixture is scoped to `test_examples.py` as it is unique to the `test_example_script()` function and how it invokes the scripts as these scripts aren't normal pytest tests. Normal Pytest tests should use the `tmpdir` fixture to handle cleanup.

Running `pytest` (no arguments - this is the unit tests - which runs the functions named `test_*` in `tests` and `docs/examples` as well as runs the `__main__` scipts in `docs/examples/example_*.py`) now doesn't generate any loose data files.

Running the tutorial notebooks will still generate loose data files. I don't think its worth cleaning these up automatically as they're saved to `docs/examples` instead of the project root so are out of the way, and also the notebooks aren't normally run via pytest anyway. Devs can do `find docs/examples -name '*.zarr' | xargs rm -r` to remove this generated data.

tests/test_examples.py
VeckoTheGecko: Not particularly, but I think its nice for debugging. It only displays when running Pytest with `--capture=no`.

erikvansebille (approved):

PR #1769, opened 2024-11-22
Adding new tutorial on the impact of grid and velocity interpolation scheme on trajectories

erikvansebille:
This new tutorial was developed with @sruehs and @michaeldenes during discussions on the relation between grids and interpolation schemes.

Issue #1762, opened 2024-11-13
`test_mitgcm` test failing with new dask release


VeckoTheGecko:
`docs/examples/example_dask_chunk_OCMs.py::test_mitgcm` has been failing in CI this week with the release of dask [`v2024.11.0`](https://docs.dask.org/en/stable/changelog.html#v2024-11-1).

Downgrading to v2024.10.0 fixes the error, but it would be good to investigate and patch the test.

erikvansebille:
This has to do with the `chunksize=auto`-chunking. This is a quite tricky part of the code to comprehend; not sure it'll be easy to find why/what changed. Alternative is to relax the unit-test? The most important tests here are that the `chunk_mode=specific*` tests pass, which they still seem to do
https://github.com/OceanParcels/Parcels/blob/eb97b325bb782b3748f567e2ca45b32affb882e5/docs/examples/example_dask_chunk_OCMs.py#L564-L567

VeckoTheGecko:
Yeah, I'm struggling to understand what the test (and underlying fieldfilebuffer.py) is doing. The fact that we have failure here:

```py
if chunk_mode != "specific_different":
    assert len(fieldset.U.grid._load_chunk) == len(fieldset.V.grid._load_chunk)
```
leads me to think that when using auto chunking that for some reason the U and V grids don't have the same number of chunks. Not sure what the implications of that are though.

@erikvansebille any immediate ideas how do you think we could relax the tests? I think these tests could do with a refactor down the line to improve readability.

xref #853

erikvansebille:
> @erikvansebille any immediate ideas how do you think we could relax the tests? I think these tests could do with a refactor down the line.

Yes, a refactor would be good down the line; also when we get uxarray-support in Parcels v4? But until then, perhaps we can just `xfail` these tests?

PR #1771, opened 2024-11-25
CI: Add py310 to integration test matrix

VeckoTheGecko:
In https://github.com/OceanParcels/Parcels/pull/1769#pullrequestreview-2457032302 there was some py310 incompatible code. Adding py310 to integration test matrix so that it fails in CI in future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,9 @@ jobs:
       matrix:
         os: [macos, ubuntu, windows]
         python-version: ["3.12"]
+        include:
+          - os: ubuntu
+            python-version: "3.10"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


VeckoTheGecko:
Sidenote: @erikvansebille Is the purpose of the integration tests to make sure that the tutorials are working correctly, or to actually test the codebase? If the former, how do you feel about removing `windows` and `mac` from the integration tests? (as windows/mac incompatibility should arise in the unit tests)

Has Windows integration tests failed by themselves before?

erikvansebille (approved):

erikvansebille:
> Has Windows integration tests failed by themselves before?

Yes, if I remember well this has happened on the docs/examples/example_dask_chunk_OCMs.py test before; but these are rather flaky anyways (they also failed recently with a new array version in #1762). So I think these integration tests are more for external dependency deprecations than our own code-development
 
> If we're unsure perhaps we should lean towards keeping

We could perhaps run the windows and macOS integration tests only every month or so. That will reduce our CI-burden (and save some energy).

VeckoTheGecko:
> if I remember well this has happened on the docs/examples/example_dask_chunk_OCMs.py

I see. I thought that integration tests were just the notebooks.

erikvansebille:
> I see. I thought that integration tests were just the notebooks.

No, the notebooks _and_ the docs/examples/*.py files

VeckoTheGecko:
Yes. Though technically the `__main__` parts of the `docs/examples/example*.py` scripts are unit tests since `test_example_script()` is in `tests/test_examples.py`.

What has been the rule of thumb for whether something is a unit or integration test?

erikvansebille:
> What has been the rule of thumb for whether something is a unit or integration test?

The integration tests use 'real-world' hydrodynamic datasets (GlobCurrent, MITgcm, OFAM, ...) and/or are full-fledged example simulations (e.g. Peninsula, Stommel, Brownian).

PR #1772, opened 2024-11-25
Implementing correct depth-to-sigma calculation 

### LINKED ISSUE #1763: from_croco() enhancement; correct conversion from z to sigma grid
In Parcels v3.1.0, we implemented support for CROCO sigma grids. We assumed that the conversion from the depth-coordinate (z) to the terrain-following coordinate (sigma) was simply `sigma(t, y, x) = z(t, y, x) / H(y, x)`, where `H` is the local seafloor depth. 

However, following the discussion in https://github.com/OceanParcels/Parcels/discussions/1752, we now realise that this conversion is a bit more complex, and requires not only the free-surface field (`eta`), but also a a critical depth controlling the stretching `depth_c` , and a dimensionless vertical coordinate stretching function `C[zi]` as a function of vertical grid

See http://cfconventions.org/cf-conventions/cf-conventions.html#_ocean_s_coordinate_generic_form_1 for detailed information about this conversion

It would be good to implement this conversion under the hood in Parcels too, so that the mapping between `sigma` and `z` is more in line with CROCO numerics

erikvansebille:
This PR fixes the way that the conversion is done for CROCO fieldsets from particle.depth (in meters) to local sigma-coordinate. Where this was initially assumed to be a simple linear conversion, this turned out to be much more complex. 

After some discussion on the myroms forum https://www.myroms.org/forum/viewtopic.php?p=25752#p25752, the best option is to calculate the local depth of each sigma layer and then linearly interpolate the particle depth to find the fractional sigma depth.

This fixes #1763


parcels/fieldset.py
```diff
@@ -723,11 +724,12 @@ def from_croco(
...
+        the vertical coordinate is scaled by the bathymetry (``h``) and sea-surface height (``zeta``)
+        fields from CROCO, in order to account for the sigma-grid.
+        The horizontal interpolation uses the MITgcm grid indexing as described in FieldSet.from_mitgcm().
 
-        The sigma grid scaling means that FieldSet.from_croco() requires a variable ``H: h`` to work.
+        In 3D, when there is a ``depth`` dimension, the sigma grid scaling means that FieldSet.from_croco()
+        requires variables ``H: h`` and ``Zeta: zeta``, as well as he ``hc`` parameter to work.
 
         See `the CROCO 3D tutorial <../examples/tutorial_croco_3D.ipynb>`__ for more infomation.
         """
```
VeckoTheGecko, line 735: Should we add info here on the `hc` param and what it is?

tests/test_advection.py
```diff
@@ -195,6 +195,45 @@ def test_advection_RK45(lon, lat, mode, rk45_tol):
...
 
 
+def test_conversion_3DCROCO():
+    """Test of the (SciPy) version of the conversion from depth to sigma in CROCO
+
+    Values below are retrieved using xroms and hardcoded in the method (to avoid dependency on xroms):
+    # x, y = 10, 20
+    # s_xroms = ds.s_w.values
+    # z_xroms = ds.z_w.isel(time=0).isel(eta_rho=y).isel(xi_rho=x).values
+    # lat, lon = ds.y_rho.values[y, x], ds.x_rho.values[y, x]
```
VeckoTheGecko, line 205:
```suggestion
    ```py
    x, y = 10, 20
    s_xroms = ds.s_w.values
    z_xroms = ds.z_w.isel(time=0).isel(eta_rho=y).isel(xi_rho=x).values
    lat, lon = ds.y_rho.values[y, x], ds.x_rho.values[y, x]
    ```
```

parcels/field.py
```diff
@@ -617,18 +637,23 @@ def from_netcdf(
...
+            ) as filebuffer:
+                lon, lat = filebuffer.lonlat
+                indices = filebuffer.indices
+                # Check if parcels_mesh has been explicitly set in file
+                if "parcels_mesh" in filebuffer.dataset.attrs:
+                    mesh = filebuffer.dataset.attrs["parcels_mesh"]
+        else:
+            lon = 0
+            lat = 0
+            mesh = "flat"
```
VeckoTheGecko, line 656: Was this for the 1D depth array that you were mentioning @erikvansebille ?

tests/test_advection.py
VeckoTheGecko, line 1: Do you think it would be worth to also add an integration test for this with CROCO data to make sure the advection is working as expected? (or was that added in a prior PR?)

VeckoTheGecko:
Looks good! Just some things that I think would be good to clarify before merging.

I'm not 100% happy with the API changes with `from_croco` (I feel that having the `hc=` parameter is a deviation from the rest of the `from_*` functions. Though I don't have a better suggestion on how we could implement it given our limitations 🤔

parcels/field.py
erikvansebille: Yep indeed, I needed this to be able to parse 1D Fields that have no `lon` or `lat` dimensions. Like the `Cs_w` Field in CROCO/Roms

tests/test_advection.py
erikvansebille: Nice solution, didn't know this was possible in docstrings!

parcels/fieldset.py
erikvansebille: Thanks, I just clarified the doctoring a bit more

tests/test_advection.py
erikvansebille: The test is already in test_advection:
https://github.com/OceanParcels/Parcels/blob/ef9eaa93322b6284908c52010cf75777475d52ce/tests/test_advection.py#L198-L229

erikvansebille:
> I'm not 100% happy with the API changes with `from_croco`. I feel that having the `hc=` parameter is a deviation from the rest of the `from_*` functions. Though I don't have a better suggestion on how we could implement it given our limitations 🤔

Yes, I fully understand; I'm not 100% happy either. 

This would actually be another reason to move to xarray for FieldSet creation in Parcels... The xarray DataSet can be expected to contain all these variables (hc, Cs_w, Zeta, H, etc) if the data comes from CROCO/Roms, so we can take them directly from there...

VeckoTheGecko:
> This would actually be another reason to move to xarray for FieldSet creation in Parcels...

I 100% agree. I'm currently scoping that out

tests/test_advection.py
VeckoTheGecko: Ah I see. And I guess since it was a 2D test there wasn't a need to update it in this PR

erikvansebille: Well, there are unit tests for both 2D and 3D. This PR only covers 3D, the big change to make the 3D unit test work is in the tests/test_data/fieldset_CROCO3D.py change

VeckoTheGecko (approved):

PR #1774, opened 2024-11-25
Make `filefieldbuffer.py` private

### LINKED ISSUE #1773: Make `filefieldbuffer.py` private
Rename `filefieldbuffer.py` to `_filefieldbuffer.py` to make it private. There is no need for users to directly use `DaskFileBuffer` `DeferredDaskFileBuffer` `NetcdfFileBuffer` `_FileBuffer` in their scripts (from what I know)

VeckoTheGecko:
Fixes #1773

(easier to review by looking at commit diffs)


VeckoTheGecko:
Uhh. Looks like we have `DaskFileBuffer.add_to_dimension_name_map_global()` and `DaskFileBuffer.add_to_dimension_name_map()` which are flagged with `[externally callable]`

Marking as draft while I have a think

erikvansebille:
Not needed in v4

PR #1778, opened 2024-11-27
Adding support for grid.negate_depth() method

### LINKED ISSUE #1746: Create a `grid.negate_depth` method to swap the sign of depth fields?
Since the sign/orientation of the depth vector is not standardised in oceanography, users may sometimes need to negate a `grid.depth` field. See e.g. https://github.com/OceanParcels/virtualship/pull/75. Especially since `grid._depth` is a private attribute, this is not trivial to users.

It might thus be useful to add a simple `grid.negate_depth()`method that simply does
```python
def negate_depth(self):
    self._depth = -self._depth
```

erikvansebille:
This PR adds a `grid.negate_depth()` method, which can be useful when the depth needs to be flipped from positive to negative values. This is the case for example in the `VirtualShip` package:
https://github.com/OceanParcels/virtualship/blob/44eae19e55545166c446e8eabe7c76ab13d7344b/src/virtualship/expedition/input_data.py#L98-L100

Note that this method does _not_ change the direction of the vertical velocity; for that users need to add a `fieldset.W.set_scaling_factor(-1.0)`

This fixes #1746

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -114,7 +114,7 @@ class Grid:
     def depth(self):
         return self._depth
 
-    def negate_depth(self):
+    def negate_depth(self) -> None:
         self._depth = -self._depth
 
     @property
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -114,6 +114,9 @@ class Grid:
     def depth(self):
         return self._depth
 
+    def negate_depth(self):
+        self._depth = -self._depth
+
     @property
     def mesh(self):
         return self._mesh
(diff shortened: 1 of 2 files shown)


parcels/grid.py
```diff
@@ -114,6 +114,9 @@ def lat(self):
     def depth(self):
         return self._depth
 
+    def negate_depth(self):
```
VeckoTheGecko, line 117:
```suggestion
    def negate_depth(self) -> None:
```

VeckoTheGecko (approved):
LGTM! Maybe we could add a simple docstring detailing some of the stuff that you have mentioned here about why it's useful? Also maybe specifying that the grid classes don't really support grid manipulation beyond this simple case (as that is beyond the scope of Parcels).

PR #1779, opened 2024-11-29
Cleanup macOS debugging files

### LINKED ISSUE #1637: Files with `.so.dSYM` extension aren't cleaned up after C compilation

VeckoTheGecko:
Files of `.so.dSYM extension` are debugging files creating during C compilation on MacOS systems. This PR checks for them and cleans them up if they exist. This PR also does some refactoring of dead code related to `Kernel.dyn_srcs`.

During my debugging in a Jupyter notebook, I did realise some erroneous `.so` files were left behind sometimes. A bit inconsistent, as these should have been garbage collected. Couldn't recreate in the tests though. Perhaps something to do with the Jupyter notebook environment and how items are garbage collected? Thoughts @erikvansebille ?

Fixes #1637

### DIFF
--- a/parcels/_version.py
+++ b/parcels/_version.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import warnings
 
 try:
     from parcels._version_setup import version
@@ -13,9 +14,11 @@ except ModuleNotFoundError:
             .decode("ascii")
             .strip()
         )
-    except subprocess.SubprocessError as e:
-        e.add_note(
+    except subprocess.SubprocessError:
+        warnings.warn(
             "Looks like you're trying to do a development install of parcels. "
             "This needs to be in a git repo so that version information is available. "
+            "Setting version to 'unknown'.",
+            stacklevel=2,
         )
-        raise e
+        version = "unknown"
--- a/parcels/kernel.py
+++ b/parcels/kernel.py
@@ -439,14 +439,13 @@ class Kernel(BaseKernel):
             dyn_dir = get_cache_dir()
             basename = f"{cache_name}_0"
         lib_path = "lib" + basename
-        src_file_or_files = None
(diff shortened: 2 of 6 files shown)


parcels/kernel.py
```diff
@@ -453,39 +426,26 @@ def get_kernel_compile_files(self):
             dyn_dir = get_cache_dir()
             basename = f"{cache_name}_0"
         lib_path = "lib" + basename
-        src_file_or_files = None
-        if type(basename) in (list, dict, tuple, ndarray):
-            src_file_or_files = [""] * len(basename)
-            for i, src_file in enumerate(basename):
-                src_file_or_files[i] = f"{os.path.join(dyn_dir, src_file)}.c"
-        else:
-            src_file_or_files = f"{os.path.join(dyn_dir, basename)}.c"
```
VeckoTheGecko, line 462: `basename` from `get_kernel_compile_files()` looks to be always a string. I don't know why there are checks for it being an iterable (introduced in c38e927450ff699ba0f4ca08b9c7816bf18cd678)

parcels/_version.py
```diff
@@ -13,9 +14,11 @@
...
-        e.add_note(
+    except subprocess.SubprocessError:
+        warnings.warn(
             "Looks like you're trying to do a development install of parcels. "
             "This needs to be in a git repo so that version information is available. "
+            "Setting version to 'unknown'.",
+            stacklevel=2,
         )
-        raise e
+        version = "unknown"
```
VeckoTheGecko, line 24: I thought that a warning here would be less obtrusive than erroring completely, since version info is only important for particlefile metadata.

Not that easy to test, but also not really something important to support anyway.

erikvansebille (approved):

PR #1784, opened 2024-12-03
Add `tmp_zarr` fixture

VeckoTheGecko:
- Add `tmp_zarr` fixture
- specify warning in test_file_warnings()

Minor change that adds a tmp_zarr fixture. I think relying on simple fixtures like this can help simplify and avoid repeated small tasks during test setup.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,6 @@ import pytest
 
 
 @pytest.fixture()
-def tmp_zarr(tmp_path, request):
+def tmp_zarrfile(tmp_path, request):
     test_name = request.node.name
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture()
+def tmp_zarr(tmp_path, request):
```
erikvansebille, line 5: Perhaps use `tmp_zarrfile`, to clarify this yields a file-path?

VeckoTheGecko: done :)

erikvansebille (approved):

PR #1785, opened 2024-12-03
Add Python 3.13 support

VeckoTheGecko:


VeckoTheGecko:
I'll update the required checks after merging

PR #1788, opened 2024-12-05
Update default value of particlefile chunks to length of pset

erikvansebille:
This fixes a bug where zarr output file writing can be _very_ slow when only one particle is released on the first day of simulation, since the default `chunks` was set to the number of particles to write the first time, instead of the length of the `ParticleSet`.

Note that the new code also aligns with the information box on output chunking at https://docs.oceanparcels.org/en/latest/examples/tutorial_parcels_structure.html#4.-Execution-and-output


VeckoTheGecko (approved):
Looks good. Only comment is that if someone uses `repeatdt` this fix isn't 100% going to work since a new pset is created and then combined with the first.

https://github.com/OceanParcels/Parcels/blob/7542e4c311840d0ac09e271475eb66108668ef39/parcels/particleset.py#L1216-L1232

erikvansebille:
> Looks good. Only comment is that if someone uses `repeatdt` this fix isn't 100% going to work (from what I can tell) since a new pset is created and then combined with the first.
> 

Yes agree, but that's exactly what the note in https://docs.oceanparcels.org/en/latest/examples/tutorial_parcels_structure.html#4.-Execution-and-output is about; it even explicitly mentions this. Users will just have to change chunks themselves (which is a good strategy anyways to improve Parcels performance)

PR #1805, opened 2025-01-05
remove duplicate cftime dependency

observingClouds:
Hi 👋 ,

Thank you for this great package. I just tried installing it with a new package manager ([pixi](https://pixi.sh/latest/)), which failed due to the duplicate `cftime` dependency.

Cheers,
Hauke

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ dependencies = [
   "cftime",
   "numpy",
   "dask",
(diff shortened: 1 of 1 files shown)


erikvansebille (approved):

erikvansebille:
Thanks for this PR, @observingClouds! Good catch that we had a duplicate `cftime` dependency

PR #1807, opened 2025-01-06
Fixing non-second timedelta64 bug

### LINKED ISSUE #1738: `TimeConverter.reltime` with `time_origin = np.timedelta64` sensitive to precision
### Parcels version

master

### Description

The internals of `TimeConverter` with `np.timedelta64` assumes that the precision is in seconds, when that might not actually be the case. Hence `np.timedelta64(1, "D") - 1` gives 0 when in reality it should be treated in seconds all the way.

What would be a suitable precision for the object? I assume nanoseconds and then convert to float seconds for output?

### Code sample

```python
import cftime
import numpy as np

from parcels.tools.converters import TimeConverter, _get_cftime_datetimes

DAY = 24 * 60 * 60

def test_TimeConverter_timedelta64_float():
    tc = TimeConverter(np.timedelta64(0, "s"))
    assert tc.reltime(1 * DAY) == 1 * DAY

    tc = TimeConverter(np.timedelta64(0, "D"))
    assert tc.reltime(1 * DAY) == 1 * DAY

```

```text
________________________________________________ test_TimeConverter_timedelta64_float _________________________________________________

    def test_TimeConverter_timedelta64_float():
        tc = TimeConverter(np.timedelta64(0, "s"))
        assert tc.reltime(1 * DAY) == 1 * DAY
    
        tc = TimeConverter(np.timedelta64(0, "D"))
>       assert tc.reltime(1 * DAY) == 1 * DAY
E       assert np.float64(7464960000.0) == (1 * 86400)
E        +  where np.float64(7464960000.0) = reltime((1 * 86400))
E        +    where reltime = 0 days.reltime

tests/tools/test_converters.py:24: AssertionError
======================================================= short test summary info =======================================================
FAILED tests/tools/test_converters.py::test_TimeConverter_timedelta64_float - assert np.float64(7464960000.0) == (1 * 86400)
```

erikvansebille:
This fixes #1738

### DIFF
--- a/parcels/tools/converters.py
+++ b/parcels/tools/converters.py
@@ -54,14 +54,18 @@ class TimeConverter:
     """
 
     def __init__(self, time_origin: float | np.datetime64 | np.timedelta64 | cftime.datetime = 0):
-        self.time_origin = time_origin
         self.calendar: str | None = None
         if isinstance(time_origin, np.datetime64):
+            self.time_origin = time_origin
             self.calendar = "np_datetime64"
         elif isinstance(time_origin, np.timedelta64):
+            self.time_origin = time_origin.astype("timedelta64[s]")
             self.calendar = "np_timedelta64"
         elif isinstance(time_origin, cftime.datetime):
+            self.time_origin = time_origin
             self.calendar = time_origin.calendar
+        else:
+            self.time_origin = time_origin
 
     def reltime(self, time: TimeConverter | np.datetime64 | np.timedelta64 | cftime.datetime) -> float | npt.NDArray:
         """Method to compute the difference, in seconds, between a time and the time_origin
--- a/tests/tools/test_converters.py
+++ b/tests/tools/test_converters.py
@@ -40,6 +40,18 @@ def test_TimeConverter_reltime_one_day():
         assert tc.reltime(time) == ONE_DAY
 
 
+def test_TimeConverter_timedelta64_float():
+    ONE_DAY = 24 * 60 * 60
+    tc = TimeConverter(np.timedelta64(0, "s"))
+    assert tc.reltime(1 * ONE_DAY) == 1 * ONE_DAY
+
+    tc = TimeConverter(np.timedelta64(0, "D"))
+    assert tc.reltime(1 * ONE_DAY) == 1 * ONE_DAY
+
+    tc = TimeConverter(np.timedelta64(0, "ns"))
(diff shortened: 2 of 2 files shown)


VeckoTheGecko:
Maybe updating

https://github.com/OceanParcels/Parcels/blob/c079963de25f842cbe9b0500319feaa7f8056c98/parcels/particlefile.py#L354
to 

```diff
-                obs = pset.particledata.getvardata("obs_written", indices_to_write)
+                obs: list[int] = pset.particledata.getvardata("obs_written", indices_to_write)

```
Will fix mypy?

parcels/tools/converters.py
```diff
@@ -57,10 +57,13 @@ def __init__(self, time_origin: float | np.datetime64 | np.timedelta64 | cftime.
         self.time_origin = time_origin
         self.calendar: str | None = None
         if isinstance(time_origin, np.datetime64):
+            self.time_origin = time_origin
             self.calendar = "np_datetime64"
         elif isinstance(time_origin, np.timedelta64):
+            self.time_origin = time_origin.astype("timedelta64[s]")
```
VeckoTheGecko, line 63: Is it a concern that this may lose subsecond resolution?

```
>>> np.timedelta64(110, "ns").astype('timedelta64[s]')
np.timedelta64(0,'s')
```

VeckoTheGecko:
Looking under the hood at xarray, I think that we can rely on xarray internals for this.

https://github.com/pydata/xarray/blob/49502fcde4db6ea3da1f60ead589580cfdad5c98/xarray/coding/times.py

parcels/tools/converters.py
erikvansebille: Now using nanoseconds for timedelta64; better like this?

VeckoTheGecko: The issue stems with the actual comparison of datetime objects in `.reltime()`. I'm not entirely sure what datetime objects are coming into (or expected to come into) that function, so something more sophisticated may be needed.

erikvansebille: OK is fine; then let's discuss in person. Or feel free to propose something else, if you want

VeckoTheGecko:
I think looking at how xarray does it and adopting a similar approach would be a good way to go (though whether we want to do that before v4 is another matter - not sure if this bug is encountered in production)

erikvansebille:
Not needed in v4

PR #1808, opened 2025-01-06
Fixing a bug with default ParticleFile outputdt value

### LINKED ISSUE #1806: Default ParticleFile output frequency results in infinite loop
### Parcels version

3.1.2.dev4

### Description

The default output frequency of `ParticleFile` (np.inf) causes the parcel simulation in the execution to end up in an infinite loop without progressing.

I would expect from the default value that the output at all timesteps is written to file.

### Code sample

The following code snippet is identical to [test_fieldset_from_xarray](https://github.com/OceanParcels/Parcels/blob/4f1f4c006e55e9d22521351547c08fe66597596b/tests/test_fieldset.py#L1043) with the exception that the `output_file` attribute is set.

```python
import numpy as np
import xarray as xr
from parcels import FieldSet, ParticleSet, JITParticle, AdvectionRK4, ParticleFile

def generate_dataset(xdim, ydim, zdim=1, tdim=1):
    lon = np.linspace(0.0, 12, xdim, dtype=np.float32)
    lat = np.linspace(0.0, 12, ydim, dtype=np.float32)
    depth = np.linspace(0.0, 20.0, zdim, dtype=np.float32)
    if tdim:
        time = np.linspace(0.0, 10, tdim, dtype=np.float64)
        Uxr = np.ones((tdim, zdim, ydim, xdim), dtype=np.float32)
        Vxr = np.ones((tdim, zdim, ydim, xdim), dtype=np.float32)
        for t in range(Uxr.shape[0]):
            Uxr[t, :, :, :] = t / 10.0
        coords = {"lat": lat, "lon": lon, "depth": depth, "time": time}
        dims = ("time", "depth", "lat", "lon")
    else:
        Uxr = np.ones((zdim, ydim, xdim), dtype=np.float32)
        Vxr = np.ones((zdim, ydim, xdim), dtype=np.float32)
        for z in range(Uxr.shape[0]):
            Uxr[z, :, :] = z / 2.0
        coords = {"lat": lat, "lon": lon, "depth": depth}
        dims = ("depth", "lat", "lon")
    return xr.Dataset(
        {"Uxr": xr.DataArray(Uxr, coords=coords, dims=dims), "Vxr": xr.DataArray(Vxr, coords=coords, dims=dims)}
    )


def test_fieldset_from_xarray(tdim):
    ds = generate_dataset(3, 3, 2, tdim)
    variables = {"U": "Uxr", "V": "Vxr"}
    if tdim:
        dimensions = {"lat": "lat", "lon": "lon", "depth": "depth", "time": "time"}
    else:
        dimensions = {"lat": "lat", "lon": "lon", "depth": "depth"}
    fieldset = FieldSet.from_xarray_dataset(ds, variables, dimensions, mesh="flat")
    assert fieldset.U._creation_log == "from_xarray_dataset"

    pset = ParticleSet(fieldset, JITParticle, 0, 0, depth=20)

    outputfile = ParticleFile("test_fieldset_from_xarray", pset)

    pset.execute(AdvectionRK4, dt=1, runtime=10, output_file=outputfile)
    if tdim == 10:
        assert np.allclose(pset.lon_nextloop[0], 4.5) and np.allclose(pset.lat_nextloop[0], 10)
    else:
        assert np.allclose(pset.lon_nextloop[0], 5.0) and np.allclose(pset.lat_nextloop[0], 10)

test_fieldset_from_xarray(10)

```

```text
INFO: Output files are stored in test_fieldset_from_xarray.zarr.
 10%|█         | 1.0/10.0 [00:00<00:06,  1.49it/s]
```

Setting the output frequency to e.g. 1 (`outputfile = ParticleFile("test_fieldset_from_xarray", pset, outputdt=1)`) results in the expected results.

```text
INFO: Output files are stored in test_fieldset_from_xarray.zarr.
100%|██████████| 10.0/10.0 [00:00<00:00, 18.33it/s]
```

erikvansebille:
Removing the `outputdt=np.inf` default value in `ParticleFile.__init__`. This fixes #1806

### DIFF
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -360,7 +360,7 @@ class ParticleFile:
                         if len(once_ids) > 0:
                             Z[varout].vindex[ids_once] = pset.particledata.getvardata(var, indices_to_write_once)
                     else:
-                        if max(obs) >= Z[varout].shape[1]:
+                        if max(obs) >= Z[varout].shape[1]:  # type: ignore[type-var]
                             self._extend_zarr_dims(Z[varout], store, dtype=self.vars_to_write[var], axis=1)
                         Z[varout].vindex[ids, obs] = pset.particledata.getvardata(var, indices_to_write)
 
--- a/parcels/particlefile.py
+++ b/parcels/particlefile.py
@@ -47,7 +47,7 @@ class ParticleFile:
         ParticleFile object that can be used to write particle data to file
     """
 
-    def __init__(self, name, particleset, outputdt=np.inf, chunks=None, create_new_zarrfile=True):
+    def __init__(self, name, particleset, outputdt, chunks=None, create_new_zarrfile=True):
         self._outputdt = timedelta_to_float(outputdt)
         self._chunks = chunks
         self._particleset = particleset
--- a/tests/test_particlefile.py
+++ b/tests/test_particlefile.py
@@ -35,7 +35,7 @@ def fieldset():
 def test_metadata(fieldset, mode, tmp_zarrfile):
     pset = ParticleSet(fieldset, pclass=ptype[mode], lon=0, lat=0)
 
-    pset.execute(DoNothing, runtime=1, output_file=pset.ParticleFile(tmp_zarrfile))
+    pset.execute(DoNothing, runtime=1, output_file=pset.ParticleFile(tmp_zarrfile, outputdt=1))
 
     ds = xr.open_zarr(tmp_zarrfile)
     assert ds.attrs["parcels_kernels"].lower() == f"{mode}ParticleDoNothing".lower()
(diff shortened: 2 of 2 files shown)


VeckoTheGecko (approved):
Looks good. I think it was introduced in 23f0bd760a7d11d25d852253826622e27a1c2e8d , @erikvansebille do you know what the interpretation of `np.inf` for outputdt was for? (e.g., not writing the outputs?)

erikvansebille:
> @erikvansebille do you know what the interpretation of `np.inf` for outputdt was for? (e.g., not writing the outputs?) Its been a while since that commit, so perhaps just drift


The timings of the different callbacks in pset.execute (output, input, repeated, callbackdt etc) are set to `np.inf` if they are not used, see the line below
https://github.com/OceanParcels/Parcels/blob/07428683c2620030ed626037e2608746f4ffe544/parcels/particleset.py#L1086
and
https://github.com/OceanParcels/Parcels/blob/07428683c2620030ed626037e2608746f4ffe544/parcels/particleset.py#L1152

PR #1810, opened 2025-01-06
Implement warning for particles initialised outside time domain

### LINKED ISSUE #1694: Warning when particles are initialised outside time domain
@sruehs was encountering a bug with her simulation where the number of the particles in the output weren't the same as in the particle set. Turns out not all particles in the particleset were initialised within the time domain of `pset.execute(runtime=...)`. Having an informative warning here would be helpful. Note that the warning would be raised if `pset.execute()` is used in a loop, but that can now be silenced with a filter from `warnings`.

cc @michaeldenes

erikvansebille:
This PR adds a warning when some particles are set to be released outside the `FieldSet`'s time bounds

It also adds a new `ParticleSet` warning to the list of Parcels custom warnings

This fixes #1694

### DIFF
--- a/parcels/particleset.py
+++ b/parcels/particleset.py
@@ -1256,13 +1256,13 @@ def _warn_outputdt_release_desync(outputdt: float, starttime: float, release_tim
 
 def _warn_particle_times_outside_fieldset_time_bounds(time: np.ndarray, fieldset: FieldSet):
     if np.any(time):
-        if np.any(time < fieldset.U.grid.time_full[0]) and not fieldset.U.allow_time_extrapolation:
+        if np.any(time < fieldset.U.grid.time_full[0]) and not fieldset.U.allow_time_extrapolation:  # type: ignore[attr-defined]
             warnings.warn(
                 "Some particles are set to be released before the fieldset's first time and allow_time_extrapolation is set to False.",
                 ParticleSetWarning,
                 stacklevel=2,
             )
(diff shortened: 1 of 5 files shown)


parcels/particleset.py
```diff
@@ -1247,6 +1249,22 @@ def _warn_outputdt_release_desync(outputdt: float, starttime: float, release_tim
...
+                "Some particles are set to be released before the fieldset's first time and allow_time_extrapolation is set to False.",
+                ParticleSetWarning,
+                stacklevel=2,
+            )
+        if np.any(time > fieldset.U.grid.time_full[-1]) and not fieldset.U.allow_time_extrapolation:  # type: ignore[attr-defined]
+            warnings.warn(
+                "Some particles are set to be released after the fieldset's last time and allow_time_extrapolation is set to False.",
+                ParticleSetWarning,
+                stacklevel=2,
+            )
```
VeckoTheGecko, line 1270: I think that this can be refactored to

```py
def _warn_particle_times_outside_fieldset_time_bounds(release_times: np.ndarray, time_full: np.ndarray):
    ...
```
and then the `if not allow_time_extrapolation` can be lifted up a level such that

```py
if not allow_time_extrapolation:
    _warn_particle_times_outside_fieldset_time_bounds(time, fieldset)
```

I think that would reduce the coupling of this function

erikvansebille: Good idea; just implemented!

VeckoTheGecko (approved):

Issue #1685, opened 2024-09-04
"only length-1 arrays can be converted to Python scalars" error when setting periodic boundaries


sruehs:
### Parcels version

3.0.3

### Description

I encountered an error when running Parcels with b-SOSE (MITgcm) fields on Lorenz, which I have not seen before. It seems to be linked to some kind of conversion? Any help how to approach this is appreciated!

### Code sample

```python
import numpy as np
from glob import glob
from datetime import timedelta

from parcels import AdvectionRK4_3D, FieldSet, JITParticle, ParticleSet

filepath_u_5daily = sorted(
    glob("/nethome/ruhs0001/DATA/b-SOSE-iter154/5daily/Uvel_bsoseI154_*_5day.nc")
)
filepath_v_5daily = sorted(
    glob("/nethome/ruhs0001/DATA/b-SOSE-iter154/5daily/Vvel_bsoseI154_*_5day.nc")
)
filepath_w_5daily = sorted(
    glob("/nethome/ruhs0001/DATA/b-SOSE-iter154/5daily/Wvel_bsoseI154_*_5day.nc")
)
gridpath = "/nethome/ruhs0001/DATA/b-SOSE-iter154/grid.nc"


output_dir = "/nethome/ruhs0001/CLIMSATE_SouthAtlanticTransports/dev-lorenz/processed_data/traj_simulations/b-SOSE-iter154/Test/"


def get_fieldset(filepath_u, filepath_v, filepath_w):

    filenames = {
        "U": {
            "lon": gridpath,
            "lat": gridpath,
            "depth": filepath_w[0],
            "data": filepath_u,
        },
        "V": {
            "lon": gridpath,
            "lat": gridpath,
            "depth": filepath_w[0],
            "data": filepath_v,
        },
        "W": {
            "lon": gridpath,
            "lat": gridpath,
            "depth": filepath_w[0],
            "data": filepath_w,
        },
    }
    variables = {
        "U": "UVEL",
        "V": "VVEL",
        "W": "WVEL",
    }

    c_grid_dimensions = {"lon": "XG", "lat": "YG", "depth": "Zl", "time": "time"}
    dimensions = {
        "U": c_grid_dimensions,
        "V": c_grid_dimensions,
        "W": c_grid_dimensions,
    }
    fieldset = FieldSet.from_mitgcm(
        filenames,
        variables,
        dimensions,
        mesh="spherical",
        tracer_interp_method="cgrid_tracer",
[...]

Issue #1240, opened 2022-09-21
Failed advection past east-west domain boundaries in global HYCOM


selipot:
I am releasing particles in a global HYCOM simulation but I am encountering an issue where Parcels is not able to find the next position of particles that would be advected east or west past the domain boundaries. I am attaching a figure that shows the longitude values on the 7055 by 9000 grid of my simulation. In my case, particles are not advected across the 1 or 9000 X-positions in the Indian Ocean and I get a warning at execution. I expected Parcels to be able to deal with the cyclic boundary condition. I am defining my fieldset from NetCDF files that contains the velocity fields and the Latitude(Y,X) and Longitude(Y,X) arrays with coordinates Y from 1 to 7055 and X from 1 to 9000.

 Is there something else I should specify explicitly in defining my fieldset or kernel? Any advice would be greatly appreciated!

![longitude](https://user-images.githubusercontent.com/7576346/191560379-5d0e8630-0218-4d61-9bc5-e6e8d8b1115e.png)

selipot:
It looks like there's a solution already, thank you @philippemiron for the pointer! https://nbviewer.org/github/OceanParcels/parcels/blob/master/parcels/examples/tutorial_periodic_boundaries.ipynb

selipot:
The `tutoral_periodic_boundaries.ipynb` is a start but I don’t think it is readily applicable when (i) the longitude at the end of the domain varies as a function of latitude, and (ii) when the limits of the domain do not correspond to the min and max of the longitude (0,359) or (-180,179) as an example. 

With respect to (i), in that same tutorial, the kernel `periodicBC` conducts a comparison between the particle longitude and a boundary longitude but in my case this fails because the boundary longitude, the *halo* I believe, is actually a vector ... and with respect to (ii), many particles will have longitude greater or less than the longitude at the boundary but will still be in the domain (because of the way it is organized as shown above). Would a solution be to track the X, Y indices instead of longitude?

erikvansebille:
Hi @selipot. I think that @VictorOnink has used global HYCOM fields for particle advection (also across the domain boundary) is his paper at https://iopscience.iop.org/article/10.1088/1748-9326/abecbd. Perhaps you can find some hints how he did it in his code at https://github.com/VictorOnink/Modeling-Global-Plastic-Beaching?

selipot:
[...]

PR #1811, opened 2025-01-06
Deprecating curvilnear halo support

erikvansebille:
As reported in #1685, the `Field.add_periodic_halo()` method was quite buggy for Fields on Curvilinear Grids. In order to avoid users using broken code, perhaps better to deprecate the method?

### DIFF
--- a/parcels/grid.py
+++ b/parcels/grid.py
@@ -694,53 +694,7 @@ class CurvilinearGrid(Grid):
         halosize : int
             size of the halo (in grid points). Default is 5 grid points
         """
-        if zonal:
-            lonshift = self.lon[:, -1] - 2 * self.lon[:, 0] + self.lon[:, 1]
-            if not np.allclose(self.lon[:, 1] - self.lon[:, 0], self.lon[:, -1] - self.lon[:, -2]):
-                warnings.warn(
-                    "The zonal halo is located at the east and west of current grid, "
-                    "with a dx = lon[1]-lon[0] between the last nodes of the original grid and the first ones of the halo. "
-                    "In your grid, lon[1]-lon[0] != lon[-1]-lon[-2]. Is the halo computed as you expect?",
-                    FieldSetWarning,
-                    stacklevel=2,
-                )
-            self._lon = np.concatenate(
-                (
-                    self.lon[:, -halosize:] - lonshift[:, np.newaxis],
-                    self.lon,
-                    self.lon[:, 0:halosize] + lonshift[:, np.newaxis],
-                ),
-                axis=len(self.lon.shape) - 1,
-            )
-            self._lat = np.concatenate(
-                (self.lat[:, -halosize:], self.lat, self.lat[:, 0:halosize]), axis=len(self.lat.shape) - 1
-            )
-            self._zonal_periodic = True
-            self._zonal_halo = halosize
-        if meridional:
-            if not np.allclose(self.lat[1, :] - self.lat[0, :], self.lat[-1, :] - self.lat[-2, :]):
-                warnings.warn(
-                    "The meridional halo is located at the north and south of current grid, "
-                    "with a dy = lat[1]-lat[0] between the last nodes of the original grid and the first ones of the halo. "
-                    "In your grid, lat[1]-lat[0] != lat[-1]-lat[-2]. Is the halo computed as you expect?",
-                    FieldSetWarning,
-                    stacklevel=2,
-                )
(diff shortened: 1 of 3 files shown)


VeckoTheGecko:
I get that for a lot of simulations with curvilinear grids (especially arbitrary curvilinear ones) the approach is buggy/there is no way to sensibly implement periodicity in this case. However, what about people running global simulations?

erikvansebille:
> I get that for a lot of simulations with curvilinear grids (especially arbitrary curvilinear ones) the approach is buggy/there is no way to sensibly implement periodicity in this case. However, what about people running global simulations?

Well, this is described at the start of the [Periodic Boundaries tutorial](https://docs.oceanparcels.org/en/latest/examples/tutorial_periodic_boundaries.html):

_In case you have a global model on a C-grid, then there is typically not much that needs to be done. The only thing you have to be aware of, is that there is no ‘gap’ between the longitude values on the right side of the grid and on the left side of the grid (modulo 360). In other words, that_
```python
fieldset.U.lon[:, 0] >= fieldset.U.lon[:, -1]
```
_This is the case for most [curvilinear NEMO model outputs on a C-grid](https://docs.oceanparcels.org/en/latest/examples/tutorial_nemo_curvilinear.html), so that periodic boundary conditions work out of the box, without any special code required._

_For [Hycom models](https://github.com/OceanParcels/parcels/issues/1240), we have found that it can help to add an extra column of longitudes at the end of the longitude array:_
```python
lon_c = np.expand_dims(fieldset.U.grid.lon[:, 0], axis=1)
fieldset.U.grid.lon = np.hstack((fieldset.U.grid.lon, lon_c))
```

VeckoTheGecko (approved):
I see. Thanks for the clarification

parcels/grid.py
```diff
@@ -694,53 +694,7 @@ def add_periodic_halo(self, zonal, meridional, halosize=5):
...
-                ),
-                axis=len(self.lat.shape) - 2,
-            )
-            self._lon = np.concatenate(
-                (self.lon[-halosize:, :], self.lon, self.lon[0:halosize, :]), axis=len(self.lon.shape) - 2
-            )
-            self._meridional_halo = halosize
-        if isinstance(self, CurvilinearSGrid):
-            self._add_Sdepth_periodic_halo(zonal, meridional, halosize)
+        raise NotImplementedError("CurvilinearGrid does not support add_periodic_halo")
```
VeckoTheGecko, line 697: Perhaps let's also add a link to this PR

```suggestion
        raise NotImplementedError("CurvilinearGrid does not support add_periodic_halo. See https://github.com/OceanParcels/Parcels/pull/1811")
```

noemieplanat:
Good morning, 
I work with a curvilinear grid in the Arctic, that is twisted so that it has no pole-hole (see figure). 
I cannot make my particles cannot cross the -180/180 line : if I initiate them around -175, I get an out-of-bound error when they reach -179.95. 
I tried modifying all longitudes so that they range between 0 and 360 with the following (in my advection problem, I don't care if they are killed at the 0/360 line) : 
    fset.U.grid.lon.data = np.where(fset.U.grid.lon < 0, fset.U.grid.lon+360, fset.U.grid.lon)
    fset.V.grid.lon.data = np.where(fset.V.grid.lon < 0, fset.V.grid.lon+360, fset.V.grid.lon)
    fset.W.grid.lon.data = np.where(fset.W.grid.lon < 0, fset.W.grid.lon+360, fset.W.grid.lon)

which seems to partially work : initiating from -175 they manage to cross the 180 line, and also manage to come back. I did visualise the fset.U.grid.lon and it has no discontinuity now. 

However, there are still some grid issues for some particles that are very close to the line [for exemple : 
parcels.tools.statuscodes.FieldOutOfBoundError: Field sampled out-of-bound, at (depth=0.8745151251824836, lat=69.56177349375857, lon=179.99605165659113)]

I did not find a way to open the zarr files after the crash (to try to visualize the trajectory before it crashes) ? 
Do you ave any idea how I could dig into this whole problem ? Is there a way to re-use the old "period condition" ? would that help ? 
Thank you, 
Noemie
PS : I also had troubles at initialisation because the index_search.h function was failing to understand my grid. So I am working with a modified version of this function that I twisted for my grid.

btw : fieldset.U.lon[:, 0] >= fieldset.U.lon[:, -1] is not valid in my case, as my xcoordinates goes from America to Europe. 


grid_CREG

PR #1812, opened 2025-01-06
Harmonise Field dimensions order

### LINKED ISSUE #1646: Harmonise Field interpolation order to consistently be (time, depth, lat, lon) throughout the code
Since Parcels v2, the order for Field interpolation has been (time, depth, lat, lon). However, under the hood, there are still multiple cases (especially in JIT-mode) where that order is different. E.g. 

https://github.com/OceanParcels/parcels/blob/b6dd5c852134ac78d9e19af575dd8793fbce0fc7/parcels/field.py#L1236-L1240

Or 
https://github.com/OceanParcels/parcels/blob/b6dd5c852134ac78d9e19af575dd8793fbce0fc7/parcels/field.py#L1552

For consistency, it might be good to harmonise the order of variables throughout the code-base

erikvansebille:
As discussed om #1646, the order for Field interpolation has been [time, depth, lat, lon] since Parcels v2, . However, under the hood, there are still multiple cases where that order is different.

This PR fixes #1646 by harmonising the order of the dimensions in any Field operation to be `[time, depth, lat, lon]`

### DIFF
--- a/parcels/application_kernels/advection.py
+++ b/parcels/application_kernels/advection.py
@@ -184,8 +184,8 @@ def AdvectionAnalytical(particle, fieldset, time):
         time_i = np.linspace(0, fieldset.U.grid.time[ti + 1] - fieldset.U.grid.time[ti], I_s)
         ds_t = min(ds_t, time_i[np.where(time - fieldset.U.grid.time[ti] < time_i)[0][0]])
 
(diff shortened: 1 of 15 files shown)


VeckoTheGecko (approved):
Looks good to me, though admittedly I'm not the most familiar with this section of the codebase

PR #1813, opened 2025-01-06
Reordering dimensions to [time, depth, lat, lon] in Grid class

erikvansebille:
Building on PR #1812 that harmonises the Field dimension order, this PR harmonises the `Grid` dimension order too. However, `Grid` is a public class, so this is a breaking change. Not sure, therefore, when/whether we also want to merge this PR... What are your thoughts, @VeckoTheGecko?


erikvansebille:
Not needed in v4

PR #1816, opened 2025-01-08
Refactoring field interpolation and allow custom interpolation methods in Scipy mode

### LINKED ISSUE #1823: Refactor interpolation and indexing methods out of `field.py`
This will help for maintainability, v4 development, testing and also starts paving the way for users to use custom interpolation schemes on fields.

VeckoTheGecko:
This PR refactors many of the indexing methods and interpolation methods out of `field.py`, and moves indexing code to a separate file to make things more manageable and reduce the coupling with the Field class.

This PR also allows users to easily overwrite the behaviour of existing interpolation methods or (untested) define new interpolation methods in Scipy mode. This can be done via the `register_2d_interpolator(...)` and `register_3d_interpolator(...)` decorators. **This behaviour is in beta and is subject to change.**

This also makes the interpolation functions more easily testable by providing only the required data.

Fixes #1823


parcels/_interpolation.py
```diff
@@ -0,0 +1,265 @@
...
+        + xsi * (1 - eta) * data[yi, xi + 1]
+        + xsi * eta * data[yi + 1, xi + 1]
+        + (1 - xsi) * eta * data[yi + 1, xi]
+    )
+    if self.interp_method == "bgrid_w_velocity" and self.gridindexingtype == "mom5" and zi == -1:
+        # Since MOM5 is indexed at cell bottom, allow linear interpolation of W to zero in uppermost cell
+        return zeta * f1
+    else:
+        return (1 - zeta) * f0 + zeta * f1
+"""
```
VeckoTheGecko, line 259: This refactoring isn't as simple as "moving the code" like the other ones, as this needs to be separated into a couple interpolation functions. Do you want me to be rigorous with the refactoring here (i.e., create test cases for this isolated code, do the refactor and make sure they still pass), or should I do the refactoring relying on the existing test suite?

parcels/tools/statuscodes.py
VeckoTheGecko, line 1: Some of the changes here are about removing the custom inits. This was initially done to enable an approach to `add_note` (I opted for a different implementation of `add_note` anyway which didn't rely on this though).

Having custom init's for exception classes isn't used much in Python (and imo needlessly limits the exception class), but I wouldn't say its an important change. The only place I'd see its necessary is if its important that the user can programmatically access the exact x,y,z of the interpolation failure (as this was previously saved as state on the exception).

VeckoTheGecko (commented):
Thoughts on the below @erikvansebille ? Let me know if you want me to break this into separate PRs

parcels/_interpolation.py
```diff
@@ -0,0 +1,276 @@
...
+    data = ctx.data[ti, zi, :, :]
+    f0 = (
+        (1 - xsi) * (1 - eta) * data[yi, xi]
+        + xsi * (1 - eta) * data[yi, xi + 1]
+        + xsi * eta * data[yi + 1, xi + 1]
+        + (1 - xsi) * eta * data[yi + 1, xi]
+    )
+    if ctx.gridindexingtype == "pop" and zi >= ctx.grid.zdim - 2:
+        # Since POP is indexed at cell top, allow linear interpolation of W to zero in lowest cell
+        return (1 - zeta) * f0
```
VeckoTheGecko, line 258: I'm looking at this code, and would like to remove the dependency on the `grid` class. Do you know whether chunking also happens vertically @erikvansebille? wondering whether `data.shape[1]` would be the equivalent of `grid.zdim`.

Otherwise I might need to have a deeper think on this

VeckoTheGecko: I can't see any handling of chunks in the scipy version of the code

erikvansebille: > I can't see any handling of chunks in the scipy version of the code, at least not in field.py or these new files, so perhaps `data` is full?

Yep, indeed there is no explicit handling of the chunks in script; to be honest I'm not sure whether that's because the data is converted to full at some point, or whether that magically happens somewhere by dask/numpy under the hood...

erikvansebille: > I'm looking at this code, and would like to remove the dependency on the `grid` class. Do you know whether chunking also happens vertically @erikvansebille? wondering whether `data.shape[1]` would be the equivalent of `grid.zdim`. Note that `data` is just `Field.data`
> 

Yep, I think you're right that `data.shape[1]` will always be zdim; although it could be that we have some cases were `len(data)=3`, and `data[0]` is then _either_ depth or time. In that case, it wouldn't work... 

This was the implementation in earlier Parcels versions, but I can't remember if we harmonised it for `len(data)` to always be `4`. That would certainly make the logic easier!

Anyways, I'd give it a try, if I were you. The passing/failing of unit tests will give you an idea whether it works or not...

VeckoTheGecko:
Note for review: ba1bc6c622a0c318db5cd8bf481218039ac2a105 includes testing that shows the refactor of the 3d interpolation is equivalent to the previous version for all combinations of interp methods and grid types.

VeckoTheGecko:
I'm quite aware that this is getting to be a big PR. To be expected for such a big refactor, but I think it would be good to continue in other PRs (for vector interp and indexing) to keep this reviewable.

tests/test_interpolation.py
```diff
@@ -0,0 +1,124 @@
...
+    @pytest.mark.parametrize(
+        "func, eta, xsi, expected",
+        [
+            # pytest.param(interpolation._nearest_3d, ...),
+            # pytest.param(interpolation._cgrid_velocity_3d, ...),
+            # pytest.param(interpolation._linear_invdist_land_tracer_3d, ...),
+            # pytest.param(interpolation._linear_3d, ...),
+            # pytest.param(interpolation._tracer_3d, ...),
+        ],
+    )
```
VeckoTheGecko, line 93: Thoughts on test cases @erikvansebille ?

VeckoTheGecko:
I added an extra commit in the history that shows 100% equivalence of the refactor. See updated https://github.com/OceanParcels/Parcels/pull/1816#issuecomment-2596309397

parcels/_index_search.py
```diff
@@ -0,0 +1,116 @@
...
+            assert tt >= 0 and tt <= 1, "Vertical s grid is being wrongly interpolated in time"
+            depth_vector = dv2[0, :] * (1 - tt) + dv2[1, :] * tt
+    else:
+        depth_vector = (
+            (1 - xsi) * (1 - eta) * grid.depth[:, yi, xi]
+            + xsi * (1 - eta) * grid.depth[:, yi, xi + 1]
+            + xsi * eta * grid.depth[:, yi + 1, xi + 1]
+            + (1 - xsi) * eta * grid.depth[:, yi + 1, xi]
+        )
+    z = np.float32(z)  # type: ignore # TODO: remove type ignore once we migrate to float64
```
erikvansebille, line 93: I'm surprised by this statement here. I know it doesn't come from this PR, but why only cast to float32 in the s-case (and not the z-case)? In general, `z` can be either float32 or float64, depending on the `lonlatdepth_dtype`

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+@register_2d_interpolator("nearest")
+def _nearest_2d(ctx: InterpolationContext2D) -> float:
+    xii = ctx.xi if ctx.xsi <= 0.5 else ctx.xi + 1
+    yii = ctx.yi if ctx.eta <= 0.5 else ctx.yi + 1
+    return ctx.data[ctx.ti, yii, xii]
+
+
+def _unit_square_to_target(*, eta: float, xsi: float, data: np.ndarray, yi: int, xi: int) -> float:
```
erikvansebille, line 116: I find this a bit confusing function title. What is 'the target' refer to? Why no `_interp_on_unit_square()`?

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+    """Interpolation on a unit square. See Delandmeter and Sebille 2019."""
```
erikvansebille, line 117:
```suggestion
    """Interpolation on a unit square. See Delandmeter and Van Sebille (2019), 10.5194/gmd-12-3571-2019."""
```
Good practice to add dois to references

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+        + xsi * eta * data[yi + 1, xi + 1]
+        + (1 - xsi) * eta * data[yi + 1, xi]
+    )
+
+
+@register_2d_interpolator("linear")
+@register_2d_interpolator("bgrid_velocity")
+@register_2d_interpolator("partialslip")
+@register_2d_interpolator("freeslip")
+def _linear_2d(ctx: InterpolationContext2D) -> float:
```
erikvansebille, line 130: what does the `ctx` abbreviation here stand for?

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+    xsi = ctx.xsi
+    eta = ctx.eta
+    data = ctx.data
+    yi = ctx.yi
+    xi = ctx.xi
+    ti = ctx.ti
+    val = _unit_square_to_target(eta=eta, xsi=xsi, data=data[ti, :, :], yi=yi, xi=xi)
+    return val
```
erikvansebille, line 138: Why can't you do something like this
```suggestion
    return _unit_square_to_target(eta=ctx.eta, xsi=ctx.xsi, data=ctx.data[ctx.ti, :, :], yi=ctx.yi, xi=ctx.xi)
```
That would make the function quite a bit simpler/shorter

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+@register_2d_interpolator("linear_invdist_land_tracer")
+def _linear_invdist_land_tracer_2d(ctx: InterpolationContext2D) -> float:
+    xsi = ctx.xsi
+    eta = ctx.eta
+    data = ctx.data
+    yi = ctx.yi
+    xi = ctx.xi
+    ti = ctx.ti
```
erikvansebille, line 148: Also here, why can't we refer to `ctx.*` directly in the function?

parcels/_interpolation.py
```diff
@@ -0,0 +1,291 @@
...
+    return (1 - zeta) * f0 + zeta * f1
+
+
+@register_3d_interpolator("linear")
+@register_3d_interpolator("partialslip")
+@register_3d_interpolator("freeslip")
+def _linear_3d(ctx: InterpolationContext3D) -> float:
+    zeta = ctx.zeta
+    eta = ctx.eta
+    xsi = ctx.xsi
```
erikvansebille, line 264: And here, why not use `ctx.*` directly?

erikvansebille (commented):
First bit of reviews (of the three main new files). Rest to come after the weekend

parcels/_interpolation.py
VeckoTheGecko: Good suggestion. The naming on a lot of these methods were a bit uncertain

parcels/_interpolation.py
VeckoTheGecko: Context

parcels/_interpolation.py
VeckoTheGecko: Thanks, that slipped by. During the refactor this function was bigger (so extracting the vars was nicer comparent to repeatedly `ctx`ing), but definitely agree with your suggestion here

parcels/_interpolation.py
VeckoTheGecko: We can, but I was concerned it would impact readability since this is a larger function

parcels/_interpolation.py
VeckoTheGecko: Can do!

parcels/_index_search.py
VeckoTheGecko: No idea :/

parcels/_interpolation.py
```diff
@@ -0,0 +1,282 @@
...
+    eta: float
+    xsi: float
+    ti: int
+    yi: int
+    xi: int
+
+
+@dataclass
+class InterpolationContext3D:
+    """Information provided by Parcels during 3D spatial interpolation. See Delandmeter, P. and van Sebille, E (2019) for more info.
```
erikvansebille, line 40:
```suggestion
    """Information provided by Parcels during 3D spatial interpolation. See Delandmeter and Van Sebille (2019), 10.5194/gmd-12-3571-2019 for more info.
```

parcels/_index_search.py
erikvansebille: OK, then I propose we remove it

parcels/_interpolation.py
erikvansebille: OK I see. Is fine then

parcels/_index_search.py
erikvansebille: Then I propose we remove it
```suggestion
```

parcels/field.py
erikvansebille, line 1: Should we not also move the `_search_indices()`, `_search_indices_curvilinear()` and `_search_indices_rectilinear()` methods to the `_index_search.py` file? Why are they still in `field.py`?

In general, I see that even in this PR, `field.py` still contains a lot of methods/functions that don't specifically need to be here. That would really clean up the field.py file?

E.g. `VectorField.dist`, `VectorField._is_land2D()` and `VectorField.jacobian` can go to an `interpolation_utils.py` file (mimicking what is in `include` folder for C)?

And the spatial interpolation for VectorFields can also go to `_interpolation.py`?

parcels/tools/statuscodes.py
```diff
@@ -29,62 +32,50 @@ class StatusCode:
...
+    def __init__(self, time, field=None, msg=""):
         if field is not None and field.grid.time_origin and time is not None:
             time = field.grid.time_origin.fulltime(time)
         message = f"{field.name if field else 'Field'} sampled outside time domain at time {time}."
-        if msg == "allow_time_extrapoltion":
-            message += " Try setting allow_time_extrapolation to True"
-        elif msg == "show_time":
-            message += " Try explicitly providing a 'show_time'"
+        if msg == "show_time":
+            message += " Try explicitly providing a 'show_time'."
```
erikvansebille, line 76: We don't have show_time anymore (was part of the old plotting we got rid off years ago, so this is not needed
```suggestion
```

erikvansebille (commented):
More comments, now on all files except for the tests/* files

parcels/field.py
VeckoTheGecko: Agreed - I have pushed these changes. I haven't made any changes to VectorField, I will make those in another PR

parcels/field.py
```diff
@@ -953,508 +960,79 @@ def set_depth_from_field(self, field):
...
+        return f(ctx)
 
     @deprecated_made_private  # TODO: Remove 6 months after v3.1.0
-    def interpolator3D(self, *args, **kwargs):
-        return self._interpolator3D(*args, **kwargs)
+    def interpolator3D(self, *_):
+        raise NotImplementedError
 
     def _interpolator3D(self, ti, z, y, x, time, particle=None):
+        """Impelement 3D interpolation with coordinate transformations as seen in Delandmeter, P. and van Sebille, E (2019)."""
```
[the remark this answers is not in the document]
VeckoTheGecko: Yes - I double checked and got the last of them, thanks

tests/test_interpolation.py
```diff
@@ -0,0 +1,127 @@
...
+    @pytest.mark.parametrize(
+        "func, eta, xsi, expected",
+        [
+            pytest.param(interpolation._nearest_2d, 0.49, 0.49, 3.0, id="nearest_2d-1"),
+            pytest.param(interpolation._nearest_2d, 0.49, 0.51, 4.0, id="nearest_2d-2"),
+            pytest.param(interpolation._nearest_2d, 0.51, 0.49, 5.0, id="nearest_2d-3"),
+            pytest.param(interpolation._nearest_2d, 0.51, 0.51, 6.0, id="nearest_2d-4"),
+            pytest.param(interpolation._tracer_2d, None, None, 6.0, id="tracer_2d"),
+            # pytest.param(interpolation._linear_2d, ...),
+            # pytest.param(interpolation._linear_invdist_land_tracer_2d, ...),
```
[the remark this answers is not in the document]
VeckoTheGecko: > I think it's crucial that we test _all_ interpolation methods with these simple unit tests

Agreed

> But I must say I don't really understand how this test-function is now set up. Why a class?

Admittedly classes aren't used much in Pytest beyond a way of grouping together tests (even then - not used often). I thought it would be a straight-forward way to group the tests that are testing on ti,zi,yi,xi = 0,1,1,1 on the data_2d and data_3d datasets.



> And indeed, I would not test just one point, but a few edge cases too.

Agreed - those can be functions outside of this class. Also testing of the land interp method

parcels/_interpolation.py
```diff
@@ -0,0 +1,282 @@
...
+    )
+
+
+@register_2d_interpolator("linear")
+@register_2d_interpolator("bgrid_velocity")
+@register_2d_interpolator("partialslip")
+@register_2d_interpolator("freeslip")
+def _linear_2d(ctx: InterpolationContext2D) -> float:
+    val = _interp_on_unit_square(eta=ctx.eta, xsi=ctx.xsi, data=ctx.data[ctx.ti, :, :], yi=ctx.yi, xi=ctx.xi)
+    return val
```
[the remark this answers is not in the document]
VeckoTheGecko: Yep, can do

tests/test_interpolation.py
```diff
@@ -0,0 +1,127 @@
...
+        ],
+    )
+    def test_2d(self, data_2d, func, eta, xsi, expected):
+        ctx = interpolation.InterpolationContext2D(data_2d, eta, xsi, self.ti, self.yi, self.xi)
+        assert func(ctx) == expected
```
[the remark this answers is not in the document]
VeckoTheGecko: I'm not sure about how to go about comparing to jit without going all the way back up to constructing the `Field` class, and then testing the interpolation on that. Which I guess wouldn't be too difficult to do.

I guess that would be the only way of doing it? I assume that the JIT functions don't follow the same structure as the new scipy interpolation functions?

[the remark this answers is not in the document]
erikvansebille: Yes, I fear that the only way to test again JIT is going through the `pset.execute()`. So make a FieldSet, a large ParticleSet, and then call a Field-evaoluation (`particle.u = fieldset.U[particle.time, particle.depth, particle.lat, particle.lon]`) in a custom kernel and assert whether all particle.u are the same value as the new Scipy interpolation.

Not the cleanest code and I can imagine you're slightly disappointed to require all these extra classes in this test-function (defeats the purpose of unit tests somewhat), but by far the most robust validation that our new Scipy works. When/if we move away from JIT, we can remove all that code ;-)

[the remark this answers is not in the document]
VeckoTheGecko: > Not the cleanest code and I can imagine you're slightly disappointed to require all these extra classes in this test-function

Yes, but at the end of the day its important to have proper validation and clean code can come gradually as we refactor parts of the codebase and improve :)

tests/test_interpolation.py
```diff
@@ -0,0 +1,164 @@
...
+    def test_interpolator2(ctx: interpolation.InterpolationContext3D):
+        assert ctx.data.shape[1] == zdim
+        # The array z dimension is the same as the fieldset z dimension
+        return 0
+
+    fieldset.U[0.5, 0.5, 0.5, 0.5]
+
+
+@pytest.mark.parametrize("interp_method", ["linear", "freeslip", "nearest", "cgrid_velocity"])
+def test_scipy_vs_jit(interp_method):
```
erikvansebille, line 132: I added this unit test to check if interpolation/integration of Scipy and JIT give the same results (to within a small tolerance). Note that the "cgrid_velocity" interpolation_method currently breaks on my local machine; the other three interpolation methods pass. So it seems that there is a bug somewhere in the "cgrid_velocity"? Perhaps good to first also check this function on current `master`, though

@VeckoTheGecko, feel free to improve/update this unit test!

VeckoTheGecko: > So it seems that there is a bug somewhere in the "cgrid_velocity"?

I don't think so. From what I can tell `runtime=4e-3, dt=1e-3` are too small timesteps which leads to a loss in precision. The error goes away with `runtime=4, dt=1` which I think would be more reasonable anyway?

erikvansebille: Well, the test passes with these larger runtime because the `pset` is empty at the end of the execute then. On such a small domain, the velocities are simply too large to keep particles inside the domain long enough. They all get OutofBounds and thus deleted, leaving an empty `pset` (and hence no failing asserts)

Dividing all the velocities by 1000 reintroduces the assert error (see d6f13ad)

parcels/_index_search.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def search_indices_vertical_z(grid: Grid, gridindexingtype: GridIndexingType, z: float):
+    if grid.depth[-1] > grid.depth[0]:
+        if z < grid.depth[0]:
+            # Since MOM5 is indexed at cell bottom, allow z at depth[0] - dz where dz = (depth[1] - depth[0])
+            if gridindexingtype == "mom5" and z > 2 * grid.depth[0] - grid.depth[1]:
+                return (-1, z / grid.depth[0])
+            else:
+                _raise_field_out_of_bound_surface_error(z, 0, 0)
```
erikvansebille, line 33: It's not ideal that this function passes `y=0` and `x=0`. I understand we don't know what `x` and `y` are at this moment, but printing them as zeros can be misleading to users. Would it not be better to print `not available` or `unknown` or so? Or is there no way to somehow figure out what `x` and `y` are?

parcels/_index_search.py
```diff
@@ -0,0 +1,336 @@
...
+)
+
+from .grid import GridType
+
+if TYPE_CHECKING:
+    from .field import Field
+    from .grid import Grid
+
+
+def search_indices_vertical_z(grid: Grid, gridindexingtype: GridIndexingType, z: float):
```
erikvansebille, line 26: Would these methods not be private? Do we expect users to call `search_indices_vertical_z` directly?

VeckoTheGecko: The methods are in a private module `_index_search.py`

erikvansebille: Ah ok, so if the module is private then by default all methods are private too. That actually makes sense, now that I think about it ;-)

parcels/_index_search.py
VeckoTheGecko: The previous code was also putting them to be 0, so this was just a continuation of that. Easy fix to set `x` and `y` to be `None`!

erikvansebille: OK, and could we then print these `None`s as 'unknown' in the actual warning message to the user? Would then make more sense to them?

tests/test_interpolation.py
VeckoTheGecko: https://github.com/OceanParcels/Parcels/pull/1834

VeckoTheGecko:
Cleaned up now. We can either merge this first, or https://github.com/OceanParcels/Parcels/pull/1834. Whichever one comes first, I'll rebase the other

VeckoTheGecko:
Pushed a small change to the Patch unit test commit (now cc87fcd)

erikvansebille (approved):

PR #1821, opened 2025-01-10
Move Field tests to separate file

VeckoTheGecko:
Just think its clearer to have `test_fieldset.py` not include tests specifically for `Field`

### DIFF
--- a/tests/test_advection.py
+++ b/tests/test_advection.py
@@ -73,7 +73,6 @@ def test_advection_zonal(lon, lat, depth, mode):
     }
     dimensions = {"lon": lon, "lat": lat}
     fieldset2D = FieldSet.from_data(data2D, dimensions, mesh="spherical", transpose=True)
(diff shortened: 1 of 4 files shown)


tests/test_advection.py
```diff
@@ -73,7 +73,6 @@ def test_advection_zonal(lon, lat, depth, mode):
     }
     dimensions = {"lon": lon, "lat": lat}
     fieldset2D = FieldSet.from_data(data2D, dimensions, mesh="spherical", transpose=True)
-    assert fieldset2D.U._creation_log == "from_data"
```
erikvansebille, line 76: Why not test for this creation_log here? Does it hamper development to perform this test?

erikvansebille (approved):
Good idea to separate out, indeed

tests/test_advection.py
VeckoTheGecko: I moved it to `test_fieldset_from_data()`, thought it was better suited there

PR #1852, opened 2025-02-11
Add PR template

VeckoTheGecko:
This PR template will be the default in the text box when creating a pull request. I thought a simple checklist would be helpful (notably the choosing the correct base branch reminder).

```md
<!-- Feel free to remove list items that are not relevant for your changes. -->

- [ ] Chose the correct base branch (`main` for v3 changes, `v4-dev` for v4 changes)
- [ ] Fixes #
- [ ] Added tests
- [ ] Added documentation

```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,6 @@
+<!-- Feel free to remove list items that are not relevant for your changes. -->
+
+- [ ] Chose the correct base branch (`main` for v3 changes, `v4-dev` for v4 changes)
+- [ ] Fixes #
+- [ ] Added tests
+- [ ] Added documentation


erikvansebille (approved):
Good idea! Is it also an idea to add a checkbox whether this is a PR that should be highlighted in the release notes? Is there an easy workflow then to find the PRs in the list that have this checked? Or is that over-engineering?

VeckoTheGecko:
> Is there an easy workflow then to find the PRs in the list that have this checked?

We can have a label for items that are "important changes" that we want to have in their own section in the release notes. This is supported by the configuration Github uses for generating the release notes.

> Or is that over-engineering?

I think so. I think having the mental overhead on a PR level would, in total, be more than curating the notes during release. I think this would really be helpful in larger projects with many contributors.

PR #1860, opened 2025-02-13
JIT cleanup

### LINKED ISSUE #1847: Remove JIT
Parcels v4 will not support Just-In-Time (JIT) compilation. This means that the `JITParticle` class will be removed from the codebase. This decision was made for the following reasons:

1. We want to leverage the power of `xarray` and `uxarray` for data handling and interpolation. These libraries are not compatible with the JIT compilation in v3.
2. We want to make the codebase more maintainable and easier to understand. The JIT compilation pre-v4 adds complexity to the codebase and makes it harder to debug and maintain.
3. We have quite a few features in Parcels (also v3) that only work in Scipy mode (particle-particle interaction, particle-field interaction, etc.).
4. We want users to write more flexible/complex kernels. JIT doesn't support calling functions, or using methods from `numpy` or `scipy`, while this is possible in Scipy mode.

Essentially, the only advantage of JIT was its speed. But now that the ecosystem for just-in-time compilation with python has matured in the last 10 years, we want to leverage other packages and methods (`cython`, `numba`, `jax`?) and Python internals for speed-up.

Furthermore, we think we have some good ideas how to speed up Parcels itself without JIT compilation, such as relying more on vectorized operations.

In short, we think that the disadvantages of JIT in Parcels v3 outweigh the advantages, and we want to make Parcels v4 a more modern and maintainable codebase.

In our development of v4, we will first focus on making the codebase more modular and easier to extend. Once we have a working codebase, we will release this as `v4-alpha`. After that, we will start working on performance improvements.

To keep the review process manageable, I have created a set of PRs where I removed parts of the JIT infrastructure. They are branched off each other, in order:
1. #1853 
2. #1855 
3. #1857 
4. #1858 
5. #1859 
6. #1860

erikvansebille:
This PR bundles some last assorted cleanups of JIT and its references in the code, in preparation for Parcels v4

Fixes #1847

### DIFF
--- a/tests/test_grids.py
+++ b/tests/test_grids.py
@@ -255,7 +255,7 @@ def test_rectilinear_s_grids_advect1():
     for i in range(depth_g0.shape[0]):
         for k in range(depth_g0.shape[2]):
             depth_g0[i, :, k] = bath[i] * k / (depth_g0.shape[2] - 1)
-    depth_g0 = depth_g0.transpose()  # we don't change it on purpose, to check if the transpose op if fixed in jit
+    depth_g0 = depth_g0.transpose()
 
     grid = RectilinearSGrid(lon_g0, lat_g0, depth=depth_g0)
 
--- a/.github/ci/min-core-deps.yml
+++ b/.github/ci/min-core-deps.yml
@@ -8,7 +8,6 @@ dependencies:
   # Run ci/min_deps_check.py to verify that this file respects the policy.
   - python=3.10
   - cftime=1.6
-  - cgen=2020.1
   - dask=2022.8
   - matplotlib-base=3.5
   # netcdf follows a 1.major.minor[.patch] convention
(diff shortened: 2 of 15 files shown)


VeckoTheGecko:
We can drop `cgen` as dependency, and remove references across the codebase

VeckoTheGecko:
@erikvansebille I want to push some edits related to this. Should I create a new PR or push to this one?

tests/test_grids.py
```diff
@@ -255,7 +255,7 @@ def bath_func(lon):
     for i in range(depth_g0.shape[0]):
         for k in range(depth_g0.shape[2]):
             depth_g0[i, :, k] = bath[i] * k / (depth_g0.shape[2] - 1)
-    depth_g0 = depth_g0.transpose()  # we don't change it on purpose, to check if the transpose op if fixed in jit
+    depth_g0 = depth_g0.transpose()
```
VeckoTheGecko, line 258: Briefly looking at this I'm not sure what this line of code does, or what the comment is about. Any idea? Perhaps we can improve the comment?

erikvansebille:
Any idea why Windows CI is failing now that `cgen` has been dropped from the environment files, @VeckoTheGecko?

VeckoTheGecko:
@erikvansebille I think its unrelated, there is some flakiness with the windows run where it fails on configuration of the pagefile. Safe to merge

PR #1862, opened 2025-02-14
Adding more background to v4 development on the v4-docpage

erikvansebille:
I extended the description on the v4-docpage a bit to better explain the goals, process and timeline for Parcels v4 development


docs/v4/index.md
VeckoTheGecko, line 1: Would it be beneficial to mention alignment with xarray/the Pangeo ecosystem here? (as a point 5) I think it would be interesting for Pangeo users/users who were aware of the previously eager `from_xarray()` methods in v3 :)

Otherwise looks good to me

VeckoTheGecko (approved):

PR #1865, opened 2025-02-17
Adding link to v4 documentation on stable docs too

erikvansebille:
This PR adds a link to the v4 documentation (on `latest`) to the `stable` docs-page too; for better visibility of our v4-development efforts

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -16,6 +16,7 @@ If you need more help with Parcels, try the `Discussions page on GitHub <https:/
    :hidden:
 
    Home <self>
+   v4 <https://docs.oceanparcels.org/en/latest/v4>
    Installation <installation>
    Tutorials & Documentation <documentation/index>
    API reference <reference>


docs/index.rst
```diff
@@ -16,6 +16,7 @@ If you need more help with Parcels, try the `Discussions page on GitHub <https:/
    :hidden:
 
    Home <self>
+   v4 <https://docs.oceanparcels.org/en/latest/v4>
```
erikvansebille, line 19: v4 is now an external link (with associated symbol); not sure if we can easily change this?

VeckoTheGecko (approved):
LGTM. I think this is good, but just want to point out that there are other ways we can surface this (instead or in addition to) if you have other preferences.

- Button/[admonition](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html) on the index page
- Banner announcement at the top of the website (which we use as a [warning in the v4 docs](https://docs.oceanparcels.org/en/latest/))

VeckoTheGecko:
OH. I just realised that this wouldn't surface in `stable` with the current configuration of the website (since stable points to the latest _release_). I'll see if I can update that in the RTD config, though nothing immediately jumps out...

Confirming that:
- `latest` -> `v4-dev`
- `stable` -> `main`

would be ideal?

erikvansebille:
Yep, those two would be ideal

PR #1871, opened 2025-02-19
Also storing the output from test_scipy_vs_jit()

### LINKED ISSUE #1869: Scipy vs JIT interpolation test for v4
#1816 introduced `test_scipy_vs_jit()` to compare interpolation methods between scipy and JIT mode. This was removed in the JIT cleanup for v4, but it would be good to add a test on the v4-dev branch which has these values from JIT saved as raw data to be compared against. This would be helpful so that we don't have any regressions on our interpolation schemes.

erikvansebille:
So that we can use it to confirm no regressions on the interpolation schemes in v4 (see also #1869)

- [x] Fixes #1869


erikvansebille:
I've now uploaded the `zarr` files, but see that these are quite big. Perhaps store the data in another format (netcdf?)

VeckoTheGecko:
> I've now uploaded the `zarr` files, but see that these are quite big. Perhaps store the data in another format (netcdf?)

Big in terms of line count, sure, but that's only because these attr files aren't binary so it shows the line count. In terms of raw size of the repo after the fact, I don't think there's much of a difference zarr/netcdf.

I think we can just go ahead with this - small files anyway.

VeckoTheGecko (approved):

Issue #1842, opened 2025-02-06
Combine particle.xi, particle.yi, particle.zi, particle.ti into one flattened gid_index


erikvansebille:
During discussions about unstructured grid support for Parcels v4, we realised that it is failrly trivial to combine the particle index positions (`particle.xi`, `particle.yi`, `particle.zi` and `particle.ti`) into one `particle.grid_index`(?), which can then be unraveled with modulo and division operators based on `xdim`, `ydim`, `zdim` and `tdim`.  

The advantage would be that the memory footprint of the `ParticleSet` would be less, because it would save three `Variables` of N particles.

fluidnumerics-joe:
Does this duplicate (or is this duplicated by #1874 ?

erikvansebille:
Yes indeed, I wrote this Issue quickly during our hackathon; it's the same as #1874

VeckoTheGecko:
Duplicate of #1874 . Let's continue discussion there

PR #1878, opened 2025-02-24
1874 ravel unravel index

### LINKED ISSUE #1874: Create a Field.unravel() method to map from particle gridindex to ti, zi, yi, xi

fluidnumericsJoe:
- [x] Fixes #1874 and #1842
- [x] Added tests - See notes

## Notes
* `ParticleSet` now uses indices `ei,ti` in place of `zi,yi,xi,ti`. The time index is kept separate since it appears that the initialization for `ti` is a bit different than for spatial indices; I wasn't convinced that combining spatial and temporal indexing wouldn't have unforeseen consequences.
* `Field.ravel_index` and `Field.unravel_index` have been added to map between `ei` and `(zi,yi,xi)`
* Relevant `_index_search*` methods have been updated to ravel and unravel indices where needed.
* Tests have been updated to remove references to `(zi,yi,xi)` and instead rely on `unravel_index` where needed.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -22,12 +22,11 @@ from parcels._typing import (
     GridIndexingType,
     InterpMethod,
     Mesh,
-    TimePeriodic,
     VectorType,
     assert_valid_gridindexingtype,
     assert_valid_interp_method,
 )
-from parcels.tools._helpers import default_repr, field_repr, timedelta_to_float
+from parcels.tools._helpers import default_repr, field_repr
 from parcels.tools.converters import (
     TimeConverter,
     UnitConverter,
@@ -155,10 +154,6 @@ class Field:
     allow_time_extrapolation : bool
         boolean whether to allow for extrapolation in time
         (i.e. beyond the last available time snapshot)
-    time_periodic : bool, float or datetime.timedelta
-        To loop periodically over the time component of the Field. It is set to either False or the length of the period (either float in seconds or datetime.timedelta object).
-        The last value of the time series can be provided (which is the same as the initial one) or not (Default: False)
-        This flag overrides the allow_time_extrapolation and sets it to False
     chunkdims_name_map : str, optional
         Gives a name map to the FieldFileBuffer that declared a mapping between chunksize name, NetCDF dimension and Parcels dimension;
         required only if currently incompatible OCM field is loaded and chunking is used by 'chunksize' (which is the default)
@@ -174,7 +169,6 @@ class Field:
     """
 
     allow_time_extrapolation: bool
-    time_periodic: TimePeriodic
     _cast_data_dtype: type[np.float32] | type[np.float64]
 
     def __init__(
@@ -196,7 +190,6 @@ class Field:
         time_origin: TimeConverter | None = None,
         interp_method: InterpMethod = "linear",
         allow_time_extrapolation: bool | None = None,
-        time_periodic: TimePeriodic = False,
         gridindexingtype: GridIndexingType = "nemo",
         to_write: bool = False,
         **kwargs,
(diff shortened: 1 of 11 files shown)


fluidnumerics-joe:
It appears there's a good number of notebooks to go through and update..

parcels/_index_search.py
```diff
@@ -218,17 +218,14 @@ def _search_indices_rectilinear(
...
+        particle.ei[field.igrid] = field.ravel_index(xi, yi, zi)
 
     return (zeta, eta, xsi, zi, yi, xi)
 
 
 def _search_indices_curvilinear(field: Field, time, z, y, x, ti=-1, particle=None, search2D=False):
     if particle:
-        xi = particle.xi[field.igrid]
-        yi = particle.yi[field.igrid]
+        zi, yi, xi = field.unravel_index(particle.ei[field.igrid])
```
erikvansebille, line 228: Would it be possible to get the `field.igrid` here from the `field` that calls the function? so instead implement
```suggestion
        zi, yi, xi = field.unravel_index(particle.ei)
```

erikvansebille (commented):
Nice, clean implementation @fluidnumerics-joe! A few comments/suggestions below

erikvansebille:
> It appears there's a good number of notebooks to go through and update..

There was only one notebook that failed; I fixed it in https://github.com/OceanParcels/Parcels/pull/1878/commits/7c5531dcaff790bbf0c53562044a3082564f3134

parcels/field.py
```diff
@@ -1313,6 +1313,48 @@ def computeTimeChunk(self, data, tindex):
...
+
+        Returns
+        -------
+        int
+            flat index
+        """
+        return xi + self.grid.xdim * (yi + self.grid.ydim * zi)
+
+    def unravel_index(self, ei):
+        """Return the xi, yi, zi indices for a given flat index.
```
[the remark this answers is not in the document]
fluidnumerics-joe: Made this change in https://github.com/OceanParcels/Parcels/pull/1878/commits/4e7d26f7c54721a1948666e35f68a86aa56bf848

parcels/_index_search.py
fluidnumerics-joe: It's possible, but I'm not sure how this kernel in `test/test_particlefile.py` would work.. 
https://github.com/OceanParcels/Parcels/pull/1878/files#diff-14e032bcf23d770966c77a78ba9016961b70b8c5a1928e7022717708031d0268R276

parcels/field.py
```diff
@@ -1313,6 +1313,48 @@ def computeTimeChunk(self, data, tindex):
         self.filebuffers[tindex] = filebuffer
         return data
 
+    def ravel_index(self, xi, yi, zi):
+        """Return the flat index of the given grid points.
```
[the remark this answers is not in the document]
erikvansebille: The `_ti` in computeTimeChunk is completely separate from the `particle.ti` (and `Grid._ti` will hopefully soon go when we ditch computeTimeChunk!). Th particle.ti is used in the Interpolator
https://github.com/OceanParcels/Parcels/blob/00959f6d57bd0baa617d1cb97bf4571cfa7236ff/parcels/_interpolation.py#L10-L26

parcels/_index_search.py
fluidnumerics-joe: Perfect. Got this squared a way in https://github.com/OceanParcels/Parcels/pull/1878/commits/42770dc3d00398e663a0cc378429c50060cb1423
Had to adjust the `test_multi_structured_grids` test assertions as well, but this should be good to go now.

erikvansebille (approved):
Looks all good now, but we may want to also merge in #1881 (adding `ti` to the ravelled index too)

Issue #1844, opened 2025-02-10
Create an inventory of features to keep and what to drop


erikvansebille:

erikvansebille:
> > * `time_periodic` in `FieldSet`
> 
> just noting that this would extend to `Field.time_periodic` and would result in `Grid.period` being dropped as well. Quite keen for this to go - should I work on this [@erikvansebille](https://github.com/erikvansebille) ?

Yes, feel free to take this up. Should(?) be a relatively simple feature to remove

VeckoTheGecko:
> * `postIterationCallbacks` functions in `pset.execute()`

This was introduced in #764 . It looks like its main usage was as a hook for the `test_from_netcdf_memory_containment` test, would that be correct @erikvansebille or is there some usecase I'm missing?

erikvansebille:
> > * `postIterationCallbacks` functions in `pset.execute()`
> 
> This was introduced in [#764](https://github.com/OceanParcels/Parcels/pull/764) . It looks like its main usage was as a hook for the `test_from_netcdf_memory_containment` test, would that be correct [@erikvansebille](https://github.com/erikvansebille) or is there some usecase I'm missing?

Yep indeed, this was introduced by @CKehl to analyse memory use, for his study published at https://doi.org/10.1016/j.cageo.2023.105322. Keeping it has the advantage that we canmonitor memory usage, but we don't do that now so I'm not sure how useful it is...

VeckoTheGecko:
I think we should also remove the `deferred_load = False` option. This will be handled down the line by xarray, and we default to `deferred_load = False` in (all?) the execution flow of Parcels.

Removing this will help with refactoring.

erikvansebille:
> I think we should also remove `deferred_load` (opting for the `deferred_load = False` path in the code). This will be handled down the line by xarray, and we default to `deferred_load = True` in (all?) the execution flow of Parcels.
> 
> Removing this will help with refactoring.

Yes, I agree. `deferred_load=True` is only for very large (>10GB) hydrodynamic datasets, and we don't test these in our suite. So we can remove `deferred_load` without crippling the CI

VeckoTheGecko:
Ah yes. I agree. Let's drop `deferred_load=True` (not `deferred_load=False`). That would be much clearer.

VeckoTheGecko:
This is an old issue - and everything here is captured in the migration guide/other issues. Closing...

PR #1880, opened 2025-02-25
Remove `time_periodic`

VeckoTheGecko:
Changes:  
- update notebook markdown cell (didn't rerun, they will be rerun down the line anyway)
- remove `Field.time_periodic`, corresponding method arguments, docstrings
- remove `time_periodic` method arguments from `FieldSet`, update docstrings
- Flag tests explicitly testing time periodic functionality with `v4remove` and `xfail`
- Update other tests
- remove `Grid.periods` (and update `_computeTimeChunk`)

- xref #1844

### DIFF
--- a/docs/v4/TODO.md
+++ b/docs/v4/TODO.md
@@ -5,3 +5,4 @@ List of tasks that are important to do before the release of version 4 (but can'
 - [ ] Make migration guide for v3 to v4
 - [ ] Just prior to release: Update conda feedstock recipe dependencies (remove cgen and compiler dependencies). Make sure that recipe is up-to-date.
 - [ ] Revamp the oceanparcels.org landing page, and perhaps also consider new logo/branding?
+- [ ] Look into xarray and whether users can create periodic datasets without increasing the size of the original dataset (i.e., no compromise alternative to `time_periodic` param in v3). Update docs accordingly.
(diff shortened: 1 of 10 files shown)


parcels/grid.py
erikvansebille, line 1: I think that `Grid._add_last_periodic_data_timestep` can also be removed?

parcels/tools/_helpers.py
```diff
@@ -74,7 +74,6 @@ def field_repr(field: Field) -> str:
     name            : {field.name!r}
     grid            : {field.grid!r}
     extrapolate time: {field.allow_time_extrapolation!r}
-    time_periodic   : {field.time_periodic!r}
```
erikvansebille, line 77: There are still a lot of notebook outputs that show this time_periodic. But perhaps we can better wait until the end of the v4-development to update/rerun all the notebooks?

erikvansebille (commented):
Looks good! Just two small comments

parcels/tools/_helpers.py
VeckoTheGecko: yep, that was what I was thinking as well

parcels/grid.py
VeckoTheGecko: good catch! Done

docs/v4/TODO.md
```diff
@@ -5,3 +5,4 @@ List of tasks that are important to do before the release of version 4 (but can'
 - [ ] Make migration guide for v3 to v4
 - [ ] Just prior to release: Update conda feedstock recipe dependencies (remove cgen and compiler dependencies). Make sure that recipe is up-to-date.
 - [ ] Revamp the oceanparcels.org landing page, and perhaps also consider new logo/branding?
+- [ ] Look into xarray and whether users can create periodic datasets without increasing the size of the original dataset (i.e., no compromise alternative to `time_periodic` param in v3). Update docs accordingly.
```
erikvansebille, line 8:
```suggestion
- [ ] Look into xarray and whether users can create periodic datasets without increasing the size of the original dataset (i.e., no compromise alternative to `time_periodic` param in v3). Update docs accordingly.
- [ ] Rerun all the tutorials so that their output is in line with new v4 print statements etc
```

erikvansebille (approved):

PR #1886, opened 2025-02-26
Fixes error raised when using multiple files for fields with no time dimension

andrew-s28:
- [x] Adds additional fix for #1831

This is a small change that changes a `ValueError` (raised when initializing a field which does not have a time dimension from multiple files, e.g., `Cs_w` used in `FieldSet.from_croco()`) into a `FieldSetWarning`. This enables the support for such fields implemented in #1835.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -686,7 +686,7 @@ class Field:
         kwargs["data_full_zdim"] = data_full_zdim
 
         if len(data_filenames) > 1 and "time" not in dimensions and timestamps is None:
-            warnings.warn("Multiple files given but no time dimension specified", FieldSetWarning, stacklevel=2)
+            warnings.warn("Multiple files given but no time dimension specified. See https://github.com/OceanParcels/Parcels/issues/1831 for more info.", FieldSetWarning, stacklevel=2)
 
         if grid is None:
             # Concatenate time variable to determine overall dimension
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -686,7 +686,11 @@ class Field:
         kwargs["data_full_zdim"] = data_full_zdim
 
         if len(data_filenames) > 1 and "time" not in dimensions and timestamps is None:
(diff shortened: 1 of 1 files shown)


parcels/field.py
```diff
@@ -686,7 +686,7 @@ def from_netcdf(
         kwargs["data_full_zdim"] = data_full_zdim
 
         if len(data_filenames) > 1 and "time" not in dimensions and timestamps is None:
-            raise RuntimeError("Multiple files given but no time dimension specified")
+            warnings.warn("Multiple files given but no time dimension specified", FieldSetWarning, stacklevel=2)
```
VeckoTheGecko, line 689:
```suggestion
            warnings.warn("Multiple files given but no time dimension specified. See https://github.com/OceanParcels/Parcels/issues/1831 for more info.", FieldSetWarning, stacklevel=2)
```

VeckoTheGecko (approved):
Thanks for the pr! Just a small edit - I think due to the nature of this issue it would be good to also link to it in the error message

parcels/field.py
VeckoTheGecko: Actually, on second thought, perhaps linkint to the issue would be confusing - I'm not sure in which scenarios people would be expected to see the error message ....

andrew-s28: Yes, I'm not sure linking to the issue is necessary here. Prior to the implementation of the CROCO model support, which relies on parameters like `Cs_w` and `H` that don't change in time and was the original source of issues, the error would have caught someone using multiple velocity files (perhaps each intended to each be a different time), but without a time dimension in the files themselves, which does seems like it would be a rare edge case.

A potential alternative that would leave that error in place would be to add on to the if statement a condition that excludes only the fields names (`self.name`?) used by the CROCO advection (`H`, `Cs_w`, and `zeta`). But I thought just keeping it simple and swapping to a warning is also fine, since this is all an admittedly niche use case.

PR #1888, opened 2025-02-27
Removing the calculation of periods also from _time_index

erikvansebille:
This PR continues the removal of `time_periodic` support in #1880, by also removing the calculation of periods in `Field._time_index()`


VeckoTheGecko (approved):
Good catch - didn't see this in my first pass

PR #1892, opened 2025-02-27
Removing explicit field chunking

erikvansebille:
This PR removes the support for `chunking` in the `FieldSet`. As we transition to native xarray/dask to provide Field data, the Parcels implementation is not needed anymore?

- [ ] Fixes #
- [ ] Added tests
- [ ] Added documentation


VeckoTheGecko:
> As we transition to native xarray/dask to provide Field data, the Parcels implementation is not needed anymore?

Yes, I think this is correct. The chunks parameter can be specified when reading in with Xarray ([Parallel Computing with Dask | Chunking and performance](https://docs.xarray.dev/en/stable/user-guide/dask.html#chunking-and-performance))

docs/examples/example_mitgcm.py
```diff
@@ -41,7 +41,6 @@ def periodicBC(particle, fieldset, time):  # pragma: no cover
         str(path),
         pset,
         outputdt=timedelta(days=1),
-        chunks=(len(pset), 1),
```
VeckoTheGecko, line 44: I assume this was unintentional? Chunking on the particlefile would still be needed.

VeckoTheGecko (approved):
Looks good. I want to push some edits to this, @erikvansebille are you still actively working on this PR? If not, I can PR into this PR

docs/examples/example_mitgcm.py
erikvansebille: Yep good point; that can be reverted!

VeckoTheGecko: bfb1f62

PR #1895, opened 2025-02-27
Moving the time interpolation down to within _Interpolator functions

erikvansebille:
Parcels previously would run an interpolation on time index `ti` and one on `ti+1`, and then at the end linearly interpolate the results.

In this PR, we change that to do the time-interpolation closer to the `InterpolationContext`. We need to discuss whether this is the right place now, or whether we want to push it even into the interpolators themselves. The latter would open possible support also non-linear time-interpolation

Hence, this is more a proof of concept than a full-fledged PR to be merged...


parcels/tools/_helpers.py
```diff
@@ -152,3 +152,8 @@ def timedelta_to_float(dt: float | timedelta | np.timedelta64) -> float:
     if isinstance(dt, np.timedelta64):
         return float(dt / np.timedelta64(1, "s"))
     return float(dt)
+
+
+def calculate_next_ti(ti, tau, tdim):
+    """Check if the time is beyond the last time in the field"""
+    return tau > np.finfo(float).eps and ti < tdim - 1
```
VeckoTheGecko, line 159: Could we rename to `should_calculate_next_ti` to be clearer?

And refactor to

```python
tau_is_significant = tau > np.finfo(float).eps
return tau_is_significant and ti < tdim - 1
```

tests/test_grids.py
```diff
@@ -455,7 +455,8 @@ def sampleVel(particle, fieldset, time):  # pragma: no cover
     assert (pset[0].meridional - 1) < 1e-6
 
 
-@pytest.mark.parametrize("vert_mode", ["zlev", "slev1", "slev2"])
+@pytest.mark.v4alpha
+@pytest.mark.parametrize("vert_mode", ["zlev"])  # , "slev1", "slev2"])  # v4 not supported?
```
VeckoTheGecko, line 459: Could we clarify the intention here in the comment? (and add `# TODO v4: `)

parcels/tools/_helpers.py
```diff
@@ -152,3 +152,8 @@ def timedelta_to_float(dt: float | timedelta | np.timedelta64) -> float:
     if isinstance(dt, np.timedelta64):
         return float(dt / np.timedelta64(1, "s"))
     return float(dt)
+
+
+def calculate_next_ti(ti, tau, tdim):
```
VeckoTheGecko, line 157:
```suggestion
def calculate_next_ti(ti, tau, tdim) -> bool:
```

parcels/field.py
```diff
@@ -600,17 +600,21 @@ def set_scaling_factor(self, factor):
...
-            return _search_indices_rectilinear(self, time, z, y, x, ti, particle=particle, search2D=search2D)
+            (zeta, eta, xsi, zi, yi, xi) = _search_indices_rectilinear(
+                self, time, z, y, x, ti, particle=particle, search2D=search2D
+            )
         else:
-            return _search_indices_curvilinear(self, time, z, y, x, ti, particle=particle, search2D=search2D)
+            (zeta, eta, xsi, zi, yi, xi) = _search_indices_curvilinear(
+                self, time, z, y, x, ti, particle=particle, search2D=search2D
+            )
+        return (tau, zeta, eta, xsi, ti, zi, yi, xi)
```
VeckoTheGecko, line 614: I think


```python
        if self.grid._gtype in [GridType.RectilinearSGrid, GridType.RectilinearZGrid]:
            spatial_component = _search_indices_rectilinear(
                self, time, z, y, x, ti, particle=particle, search2D=search2D
            )
        else:
            spatial_component = _search_indices_curvilinear(
                self, time, z, y, x, ti, particle=particle, search2D=search2D
            )
        return tau, *spatial_component
```
Would be a bit less repetitive? Though users would need to look at `_search_indices_*` to understand what's actually being returned, which I think is fine.

parcels/field.py
```diff
@@ -664,13 +673,24 @@ def _time_index(self, time):
...
+        elif ti == len(self.grid.time) - 1:
+            tau = 1
         else:
-            return time_index.argmin() - 1 if time_index.any() else 0
+            tau = (
+                (time - self.grid.time[ti]) / (self.grid.time[ti + 1] - self.grid.time[ti])
+                if self.grid.time[ti] != self.grid.time[ti + 1]
+                else 0
+            )
+        return tau, ti
```
VeckoTheGecko, line 693: Could we either:
- Update the docstring and function name for `_time_index` reflecting it also returns tau, or
- Split the tau calculation into a `_get_tau(grid, ti)` function

?

(slight preference for the latter)

parcels/field.py
```diff
@@ -887,59 +897,68 @@ def spatial_c_grid_interpolation2D(self, ti, z, y, x, time, particle=None, apply
...
-            (-(1 - eta) * U - (1 - xsi) * V) * py[0]
-            + ((1 - eta) * U - xsi * V) * py[1]
-            + (eta * U + xsi * V) * py[2]
-            + (-eta * U + (1 - xsi) * V) * py[3]
-        ) / jac
-        if isinstance(u, da.core.Array):
-            u = u.compute()
-            v = v.compute()
+
+        def _calc_UV(ti, yi, xi):
```
VeckoTheGecko, line 901: Should we move this function definition out of `spatial_c_grid_interpolation2D`?

parcels/field.py
```diff
@@ -1283,66 +1310,23 @@ def spatial_slip_interpolation(self, ti, z, y, x, time, particle=None, applyConv
...
                 v = self.V.units.to_target(v, z, y, x)
-            if "3D" in self.vector_type:
-                w = self.W.eval(time, z, y, x, particle=particle, applyConversion=False)
-                if applyConversion:
-                    w = self.W.units.to_target(w, z, y, x)
-                return (u, v, w)
-            else:
-                return (u, v)
+        elif self.U.interp_method == "cgrid_velocity":
+            tau, ti = self.U._time_index(time)
```
VeckoTheGecko, line 1323: tau and ti arean't used here

parcels/field.py
```diff
@@ -1283,66 +1310,23 @@ def spatial_slip_interpolation(self, ti, z, y, x, time, particle=None, applyConv
...
     def eval(self, time, z, y, x, particle=None, applyConversion=True):
+        if self.U.interp_method in ["partialslip", "freeslip"]:
+            return self.slip_interpolation(time, z, y, x, particle=particle, applyConversion=applyConversion)
+
         if self.U.interp_method not in ["cgrid_velocity", "partialslip", "freeslip"]:
             u = self.U.eval(time, z, y, x, particle=particle, applyConversion=False)
             v = self.V.eval(time, z, y, x, particle=particle, applyConversion=False)
             if applyConversion:
                 u = self.U.units.to_target(u, z, y, x)
                 v = self.V.units.to_target(v, z, y, x)
```
VeckoTheGecko, line 1321: Could you elaborate on the deletion of the else block here?

VeckoTheGecko (commented):
I think it may be a good idea with these interpolation functions (where all the arguments are numeric) to force keyword only arguments. i.e.,

```diff
- def x(a, b, c):
+ def x(*, a, b, c):
```

```
>>> def x(*, a, b, c): pass
... 
>>> x(1,2,3)
  File "<stdin>", line 1, in <module>
TypeError: x() takes 0 positional arguments but 3 were given
```

This forces users to use keyword arguments, preventing errors like accidentally switching yi, and xi in the function call which can be easily missed.

```py
_get_data_temporalinterp(ti=ti, yi=yi + j, xi=xi + i)
```

It can become a bit repetitive when calling them, but I think it's worth it for the added safety? Thoughts @erikvansebille?

parcels/field.py
erikvansebille: Hmm, note that it also injects `ti` halfway, so I'm not in favour of this. I'd prefer the clarity over the reduced repetitiveness...

parcels/field.py
erikvansebille: I prefer the first option, as the other (spatial) search_indices functions also return both the index and the relative position; so a `search_time_index` should return the same `(int, float)` type, I feel

parcels/field.py
erikvansebille: It's really only need for `spatial_c_grid_interpolation2D`, so I'd prefer to keep it here (for now)

parcels/field.py
erikvansebille: The 2D vs 3D is now dealt with in the interpolation methods themselves, so is not needed here anymore

parcels/tools/_helpers.py
VeckoTheGecko: @erikvansebille  I was just thinking wouldn't `np.isequal(tau, 0.0)` have the save effect?

parcels/field.py
VeckoTheGecko: Ah! Didn't see that :)

parcels/field.py
```diff
@@ -1223,66 +1250,22 @@ def spatial_slip_interpolation(self, ti, z, y, x, time, particle=None, applyConv
             return u, v
 
     def eval(self, time, z, y, x, particle=None, applyConversion=True):
+        if self.U.interp_method in ["partialslip", "freeslip"]:
+            return self.slip_interpolation(time, z, y, x, particle=particle, applyConversion=applyConversion)
+
         if self.U.interp_method not in ["cgrid_velocity", "partialslip", "freeslip"]:
```
VeckoTheGecko, line 1256:
```suggestion
        if self.U.interp_method not in ["cgrid_velocity"]:
```

parcels/field.py
VeckoTheGecko: Ok, sounds good

parcels/tools/_helpers.py
erikvansebille: Sure, but we would then need to do something like `tau_is_significant = not np.isequal(tau, 0.0)`; would that be more readable, with the `not`?

VeckoTheGecko: I think `return not np.isequal(tau, 0) and ti<tdim-1` is good.

The reason I suggested before to refactor was since `tau > np.finfo(float).eps` wasn't immediately clear what was happening, but I think `np.isequal` is good

parcels/tools/_helpers.py
```diff
@@ -152,3 +152,9 @@ def timedelta_to_float(dt: float | timedelta | np.timedelta64) -> float:
     if isinstance(dt, np.timedelta64):
         return float(dt / np.timedelta64(1, "s"))
     return float(dt)
+
+
+def should_calculate_next_ti(ti: int, tau: float, tdim: int) -> np.bool:
```
VeckoTheGecko, line 157: Eh. The mypy error is because this function either returns a `np.bool` (because of `tau > np.finfo(float).eps`) or a normal bool (due to the `ti < tdim - 1` calculation).

Maybe let's just remove the `-> np.bool` so that it isn't type checked...

VeckoTheGecko (approved):

PR #1896, opened 2025-02-27
fieldfilebuffer.py cleanup

VeckoTheGecko:
Remove obsolete DeferredNetcdfFileBuffer, _FileBuffer classes and unused attributes.

I also realised while doing this that fieldfilebuffer is really where a lot of the magic happens when it comes to reading in from netcdf files and coercing field input into a unified format parcels can understand. Good for me to know - I think there can be some improvements made here :)

- [x] Chose the correct base branch (ie, `swap_space_and_time_interpolation_order`)

🚧 merge after #1892
mypy fixed in next PR

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -43,7 +43,6 @@ from parcels.tools.warnings import FieldSetWarning, _deprecated_param_netcdf_dec
 
 from ._index_search import _search_indices_curvilinear, _search_indices_rectilinear
 from .fieldfilebuffer import (
-    DeferredNetcdfFileBuffer,
     NetcdfFileBuffer,
 )
 from .grid import Grid, GridType, _calc_cell_areas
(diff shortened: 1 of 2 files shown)


erikvansebille (approved):

erikvansebille:
> Good for me to know - I think there can be some improvements made here :)

Yes, I think that the entire `fieldfilebuffer.py` file can be removed at some point as we transition to native `xarray`. But better to do it in small steps

A next step could be to also remove the support for the `indices` dictionary to subset the domain? This would also be felt with by `xarray`/`dask` in v4?

VeckoTheGecko:
> as we transition to native `xarray`. But better to do it in small steps

I think there is quite some code in this file that deals with coercing the data into a unified format. Though I think that this code should be replaced with functions that consume and produce dask arrays (or produce and consume xarray objects without triggering computation).



> A next step could be to also remove the support for the `indices` dictionary to subset the domain? This would also be felt with by `xarray`/`dask` in v4?

Yes, I think so. I'm currently working on a few things - but I think this can do with further clearing

PR #1905, opened 2025-03-03
Remove `FieldSet.compute_on_defer`

VeckoTheGecko:
This is undocumented functionality in Parcels providing computation for deferred loading.

- xref #1844

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -52,7 +52,6 @@ class FieldSet:
             for name, field in fields.items():
                 self.add_field(field, name)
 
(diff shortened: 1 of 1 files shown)


erikvansebille (approved):
Looks good! This is exactly the type of feature that can be handled directly by xarray in v4

I was wondering why there is no unit-test, but see now that we removed that earlier already: https://github.com/OceanParcels/Parcels/blob/76e3233f325fab730562eacf5e6a34c03098d12b/tests/test_fieldset.py#L907-L924

PR #1913, opened 2025-03-05
Add Pixi for development

### LINKED ISSUE #1912: Add Pixi support for Parcels
> [pixi](https://github.com/prefix-dev/pixi) is a fast software package manager built on top of the existing conda ecosystem. Spins up development environments quickly on Windows, macOS and Linux.
> 
> Automatic lockfiles produce reproducible environments across operating systems (without Docker!).
> 
> pixi supports Python, R, C/C++, Rust, Ruby, and many other languages.

Pixi is a modern alternative to Conda with the following benefits:
- Lock files (make it easier to find out exact versions of packages that are being used)
- Task runner (make it easier to run tasks specific to the codebase, like running tests, testing notebooks, building docs, etc).
- Faster

VeckoTheGecko:
Adding Pixi for development (haven't made any changes to our CI setup - that will continue as is). Only downside with this is that there is now duplication between `environment.yml` and `pyproject.toml`

@erikvansebille :
- would you mind trying the new dev setup instructions to make sure its clear?
- thoughts on the `[tool.pixi.tasks]` section? (should we add any more here that would be useful for v4 dev?)

- Fixes #1912 
- [x] Added documentation

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 .git_archival.txt  export-subst
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,8 @@ dist/parcels*.egg
 parcels/_version_setup.py
 .pytest_cache
 .coverage
+
+# pixi environments
+.pixi
+*.egg-info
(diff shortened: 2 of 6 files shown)


erikvansebille:
Yep, installation was smooth and worked without issues!

erikvansebille (approved):

PR #1917, opened 2025-03-06
Remove `indices` keyword from FieldSet and Field

VeckoTheGecko:
Remove `indices` keyword from FieldSet and Field, and cleanup of `fieldfilebuffer.py`. I was encountering some difficulty removing the `indices` keyword in `fieldfilebuffer.py`. Feel free to have a look @erikvansebille , but I propose we leave for another PR.

This PR also drops CROCO support for the timebeing.  We will re-evaluate further in development how to re-implement it

- xref #1844


parcels/field.py
```diff
@@ -444,17 +438,7 @@ def from_netcdf(
...
-            if len(indices[ind]) == 0:
-                raise RuntimeError(f"Indices for {ind} can not be empty")
-            assert np.min(indices[ind]) >= 0, (
-                "Negative indices are currently not allowed in Parcels. "
-                + "This is related to the non-increasing dimension it could generate "
-                + "if the domain goes from lon[-4] to lon[6] for example. "
-                + "Please raise an issue on https://github.com/OceanParcels/parcels/issues "
-                + "if you would need such feature implemented."
-            )
+        indices = {}  # TODO Nick: Cleanup
```
erikvansebille, line 441: Will you stil clean this up in this PR? Or in a next one?

parcels/field.py
```diff
@@ -557,7 +538,6 @@ def from_netcdf(
             allow_time_extrapolation = False if "time" in dimensions else True
 
         kwargs["dimensions"] = dimensions.copy()
-        kwargs["indices"] = indices
         kwargs["netcdf_engine"] = netcdf_engine
```
erikvansebille, line 541: Also remove netcdf_engine here then?

parcels/fieldfilebuffer.py
```diff
@@ -48,79 +43,55 @@ def close(self):
...
-                if self.nolonlatindices:
-                    return np.array(depth[self.indices["depth"], :, :])
-                else:
-                    return np.array(depth[self.indices["depth"], self.indices["lat"], self.indices["lon"]])
-            elif len(depth.shape) == 4:
-                if self.nolonlatindices:
-                    return np.array(depth[:, self.indices["depth"], :, :])
-                else:
-                    return np.array(depth[:, self.indices["depth"], self.indices["lat"], self.indices["lon"]])
+            self.data_full_zdim = self.zdim
```
erikvansebille, line 92: This difference between fulL_zdim and dim should also be removed at some point. This should be handled by array and the interpolation methods. Or don't we need to worry too much because at some point the whole fieldfilebuffer.py will be gone - as we use array for all this?

parcels/fieldset.py
```diff
@@ -409,21 +403,18 @@ def from_netcdf(
                 for dim, p in paths.items():
                     paths[dim] = cls._parse_wildcards(p, filenames, var)
 
-            # Use dimensions[var] and indices[var] if either of them is a dict of dicts
+            # Use dimensions[var] if its a dict of dicts
```
erikvansebille, line 406:
```suggestion
            # Use dimensions[var] if it's a dict of dicts
```

parcels/field.py
VeckoTheGecko: Forgot to remove this comment. Really struggled to clean up any further than here :/

parcels/field.py
VeckoTheGecko: Will be done in #1921

parcels/fieldfilebuffer.py
VeckoTheGecko: > Or don't we need to worry too much because at some point the whole fieldfilebuffer.py will be gone - as we use array for all this?

Indeed I think we don't need to worry about this for now. Down the line I can see a lot of this transformation being lazy methods on top of xarray objects

parcels/fieldset.py
VeckoTheGecko: done in 9badc73e2c5392e9fd440eb0ef121d1aca26bbbb

parcels/field.py
VeckoTheGecko: done in 0f89a3d044a33c07ba5fcd241a52f8d077aa7674 (separate PR to avoid lots of rebasing)

PR #1920, opened 2025-03-06
Remove timestamps keyword from Field and FieldSet

### LINKED ISSUE #1918: Remove `timestamps` keyword from Field and FieldSet
I think this responsibility can be pushed up to xarray as well. Where users load in data, assign a time dimension, combine into one xarray dataset, and then pass into Parcels.

VeckoTheGecko:
This will be handled natively by xarray.

- Fixes #1918

🚧 Merge after #1917


docs/examples/example_globcurrent.py
```diff
@@ -51,15 +55,15 @@ def set_globcurrent_fieldset(
     "use_xarray", [True, pytest.param(False, marks=pytest.mark.xfail)]
 )
 def test_globcurrent_fieldset(use_xarray):
-    fieldset = set_globcurrent_fieldset(use_xarray=use_xarray)
```
erikvansebille, line 54: why is this not tested anymore now?

docs/v4/TODO.md
```diff
@@ -5,5 +5,8 @@ List of tasks that are important to do before the release of version 4 (but can'
 - [ ] Make migration guide for v3 to v4
 - [ ] Just prior to release: Update conda feedstock recipe dependencies (remove cgen and compiler dependencies). Make sure that recipe is up-to-date.
 - [ ] Revamp the oceanparcels.org landing page, and perhaps also consider new logo/branding?
-- [ ] Look into xarray and whether users can create periodic datasets without increasing the size of the original dataset (i.e., no compromise alternative to `time_periodic` param in v3). Update docs accordingly.
 - [ ] Rerun all the tutorials so that their output is in line with new v4 print statements etc
+- [ ] Remove `tutorial_timestamps.ipynb` tutorial (GH1918)
```
erikvansebille, line 9: Why not do this now already?

tests/test_field.py
```diff
@@ -30,7 +30,15 @@ def test_field_from_netcdf_variables():
...
-@pytest.mark.parametrize("with_timestamps", [True, False])
+@pytest.mark.parametrize(
+    "with_timestamps",
+    [
+        pytest.param(
+            True, marks=[pytest.mark.v4remove, pytest.mark.xfail(reason="GH1918. This test should be rewritten.")]
+        ),
+        False,
+    ],
+)
```
erikvansebille, line 41: I would just remove the test altogether; it's not relevant anymore

docs/v4/TODO.md
VeckoTheGecko: Though just typing that out that can be done by looking at the v3 docs..... Ok, let's remove :)

tests/test_field.py
VeckoTheGecko: I was a bit thrown by the name `test_field_from_netcdf` so I thought the intention here was more general than just the timestamps. I don't mind either way - should I still remove?

docs/examples/example_globcurrent.py
VeckoTheGecko: The globcurrent dataset has some vars with `attrs["Unit"]` instead of `attrs["units"]` hindering cf decoding. This is very easy to patch if we use xarray (done in this PR for this test case), but that forces `set_globcurrent_fieldset()` to have to use xarray.

tests/test_field.py
erikvansebille: Ah I see. Yes, let’s leave the test in for now then. We could remove the (failing) `with_timestamps` argument to keep the code somewhat cleaner, as we don’t expect timestamps to return later?

PR #1924, opened 2025-03-10
Removing scaling_factors

erikvansebille:
Removing the `set_scaling_factor` method from FieldSet, as this will be dealt with in xarray in v4

- [x] Chose the correct base branch (`remove_fieldset_from_pop`)


VeckoTheGecko (changes requested):
Just made #1925 - should we remove `check_fieldsets_in_kernels` from `kernel.py` and `interationkernel.py` for the timebeing?

Otherwise LGTM

PR #1937, opened 2025-03-12
Moving the _search_time_index function to _index_search.py

erikvansebille:
This PR moves the `Field._search_time_index` method to its own function in `_index_search.py`, where I feel it would fit better 

### DIFF
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -450,9 +450,11 @@ def test_fieldset_write(tmp_zarrfile):
     fieldset.U.to_write = True
 
     def UpdateU(particle, fieldset, time):  # pragma: no cover
+        from parcels._index_search import _search_time_index
+
         tmp1, tmp2 = fieldset.UV[particle]
         _, yi, xi = fieldset.U.unravel_index(particle.ei)
-        _, ti = fieldset.U._search_time_index(time)
+        _, ti = _search_time_index(fieldset.U.grid, time)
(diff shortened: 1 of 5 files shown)


erikvansebille:
@VeckoTheGecko, can you help with fixing the mypy errors? Not sure what is going wrong (and why it wasn't flagged when the same function was still part of field.py...)

tests/test_fieldset.py
```diff
@@ -450,9 +450,11 @@ def test_fieldset_write(tmp_zarrfile):
     fieldset.U.to_write = True
 
     def UpdateU(particle, fieldset, time):  # pragma: no cover
+        from parcels._index_search import _search_time_index
+
```
VeckoTheGecko, line 454: Could we move this to the top of the file? (unless there was a specific reason (e..g, circular imports) to put it here?)

VeckoTheGecko (changes requested):
Small comment, otherwise LGTM! Will fix mypy

tests/test_fieldset.py
erikvansebille: I wanted to test if an import statement _inside_ a Kernel also worked. But if you strongly prefer to put it at the top then I can do

VeckoTheGecko: Oh! This is a kernel, of course. ok, that's fine

VeckoTheGecko (approved):

Issue #1883, opened 2025-02-26
Add support for particleset.fieldset as a uxarray object


fluidnumericsJoe:
Make changes to the particleset to handle the scenario when the `fieldset` is a `uxarray` object.

fluidnumerics-joe:
I've started really re-working this in the https://github.com/OceanParcels/Parcels/tree/feature/uxarray_xarray_fields branch. 

At the moment, most of the work is being done in the [xfield.py](https://github.com/OceanParcels/Parcels/blob/feature/uxarray_xarray_fields/parcels/xfield.py) module. 

This module defines the `XField` and `XVectorField` classes. `XField` is meant to be a thin wrapper around the `xarray.DataArray` or `uxarray.UxDataArray` classes that enforces a uniform schema (ensuring appropriate `dims`, `coords`, `attributes`, and attribute values) and permits dynamic definition of the interpolation scheme as a `Callable` function with a validated signature (matching the signature of the provided `XField._interp_template` function`). When no interpolation function is provided, the default is to use `XField._interp_template`, which simply returns zeros for the evaluation. 

Additionally, I'm working on preserving some of the functionality from the original `Field` class; specifically, I've retained the `Field.igrid` and `Field.allow_time_extrapolation` attributes in addition to a pointer to the parent `Fieldset` object. With the `XVectorField` class, we will be able to create `UV` attributes in the `XFieldset` class to keep much of the same programming patterns that currently exist in Parcels modules that leverage the new `XFieldset`.

After reviewing xarray and uxarray documentation and playing around with creating sample xarray and uxarray datasets for the stommel gyre configuration, it became clear that leveraging field and fieldset as thin wrappers around the `(Ux)DataArray` and `(Ux)Dataset` classes (respectively) is quite a major rewrite of the `Field` and `Fieldset` classes.  Ultimately, yes, `XField` and `XFieldset` will be renamed ultimately back to `Field` and `Fieldset`.

fluidnumerics-joe:
>  I can imagine this changing somewhat as development progresses

I agree. This has been slowly evolving as I'm working on the implementation and the stommel examples.

erikvansebille:
[...]

PR #1942, opened 2025-03-25
Removing fieldtype argument from Field constructor

erikvansebille:
Removing `fieldtype` argument from `Field` constructor, as it's more pythonic to attach UnitConverters directly to a field?

See also https://github.com/OceanParcels/Parcels/issues/1883#issuecomment-2748744348

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -167,10 +167,12 @@ class Field:
                 time_origin = TimeConverter(0)
             self._grid = Grid.create_grid(lon, lat, depth, time, time_origin=time_origin, mesh=mesh)
         self.igrid = -1
-        if self.grid.mesh == "spherical" and self.name in unitconverters_map.keys():
-            self.units = unitconverters_map[self.name]
-        else:
-            self.units = UnitConverter()
+        self.units = UnitConverter()
+        if self.grid.mesh == "spherical":
+            try:
+                self.units = unitconverters_map[self.name]
+            except KeyError:
+                pass
         if isinstance(interp_method, dict):
             if self.name in interp_method:
                 self.interp_method = interp_method[self.name]


parcels/field.py
```diff
@@ -170,13 +167,10 @@ def __init__(
...
-        self.fieldtype = self.name if fieldtype is None else fieldtype
-        if self.grid.mesh == "flat" or (self.fieldtype not in unitconverters_map.keys()):
-            self.units = UnitConverter()
-        elif self.grid.mesh == "spherical":
-            self.units = unitconverters_map[self.fieldtype]
+        if self.grid.mesh == "spherical" and self.name in unitconverters_map.keys():
+            self.units = unitconverters_map[self.name]
         else:
-            raise ValueError("Unsupported mesh type. Choose either: 'spherical' or 'flat'")
+            self.units = UnitConverter()
```
VeckoTheGecko, line 173: I think this reads a bit better
```py
self.units = UnitConverter()
if self.grid.mesh == "spherical":
    try:
        self.units = unitconverters_map[self.name]
    except KeyError:
        pass
```

VeckoTheGecko (changes requested):
Minor edits, otherwise LGTM.

Should we be concerned with the naming of 
```
UWind.units = parcels.tools.converters.GeographicPolar()
VWind.units = parcels.tools.converters.Geographic()
```
? Previously it might have been easier to identify that `VWind` would be of field type `V`, whereas now its probably not as clear to the user that it needs a `Geographic` converter and not a `GeographicPolar` converter. Perhaps this would just be something to keep an eye on moving forward, that we can revisit the naming/how to specify these unit converters.

VeckoTheGecko:
I also added a setter for `Field.units`

VeckoTheGecko (approved):

Issue #1663, opened 2024-08-20
Refactoring of FieldSet field store


VeckoTheGecko:
https://github.com/OceanParcels/Parcels/blob/595b7a94cdf97182df53c88fa99cf05646f07346/parcels/fieldset.py#L162-L197

I don't like the current implementation of `.add_field()` calling `set_attr(self, name, field)`. Programmatically storing these fields messes with static typecheckers, and it would be clearer if this was handled another way (either a list of fields, a dictionary, or something else). We can set [`__getattr__`](https://stackoverflow.com/questions/4295678/understanding-the-difference-between-getattr-and-getattribute) so users can still do `fieldset.U`

erikvansebille:
Agree, this can be made _much_ cleaner. Note though that this could also be touched upon / impacted by the unstructured grids development?

PR #1946, opened 2025-03-31
Feature/uxarray xarray fields

### LINKED ISSUE #1883: Add support for particleset.fieldset as a uxarray object

fluidnumerics-joe:
- [ X ] Added tests
- [ X ] Resolves #1883


v4-tests/test_uxarray_fieldset.py
```diff
@@ -0,0 +1,64 @@
...
+    FieldSet,
+    ParticleSet,
+    Particle,
+    UXPiecewiseConstantFace,
+    UXPiecewiseLinearNode,
+)
+import os
+
+# Get path of this script
+V4_TEST_DATA = f"{os.path.dirname(__file__)}/test_data"
```
VeckoTheGecko, line 14: How will we be handling test data in v4 of Parcels? At the moment we download it via `parcels/tools/exampledata_utils.py`, but perhaps it would be better to:

- Generate the data (i.e., xarray datasets) on the fly in the idealised cases
- Download the data (from some hosting provider) for realistic cases. Perhaps using [`pooch`](https://pypi.org/project/pooch/) rather than our current custom downloading mechanism.

This will also mean that we can remove `parcels/data` (and any other committed data files) which I think would be good.

Let's continue this discussion in an issue :)

VeckoTheGecko: #1950

VeckoTheGecko (changes requested):
Could we have the tests folder as `tests/v4` rather than `v4-tests` . At the moment the tests aren't being collected (hence aren't run in CI)

```
pytest --collect-only | grep v4-tests
```

v4-tests/test_uxarray_fieldset.py
fluidnumerics-joe: This is a good question. I liked @willirath 's suggestion when we were all together of using data posted to zenodo for the tests. Some of the tests (e.g. Stommel) can be generated on-the-fly with some helper functions. Removing data (nc files, etc) would be awesome; lighten things up.

fluidnumerics-joe:
> (also, would you be able to run `pre-commit run --all-files` so that the hooks are run? I think there is some drift from our QAQC tools in this PR)

Yea, sorry about that. Got a bit sloppy here. Pushing this up now.

fluidnumerics-joe:
Prettier is failing locally with 
```
prettier requires at least version 14 of Node, please upgrade
```
but 
```
$ node --version
v22.14.0
```
I forced my way past this for now, since the other checks are all passing, but perhaps this is an issue with my conda environment setup... python dependency woes.

VeckoTheGecko:
@fluidnumerics-joe what about `nodejs --version`? I think its probably due to that (https://github.com/pre-commit/pre-commit/issues/2787#issuecomment-1447584355)

fluidnumerics-joe:
> @fluidnumerics-joe what about `nodejs --version`? I think its probably due to that ([pre-commit/pre-commit#2787 (comment)](https://github.com/pre-commit/pre-commit/issues/2787#issuecomment-1447584355))

```
$ nodejs --version
v12.22.9
```

Good call

v4-tests/test_uxarray_fieldset.py
fluidnumerics-joe: We'll need to update the `tests/v4` to use the FESOM channel data that was just merged in at https://github.com/OceanParcels/parcels-data/pull/1

fluidnumerics-joe:
@VeckoTheGecko - This is ready for review and for fixing the test data ingestion for the `tests/v4`

VeckoTheGecko:
> fixing the test data ingestion for the `tests/v4`

Done

Can do a full review tomorrow

tests/test_fieldset.py
VeckoTheGecko, line 1: All the v4remove and v4alpha tests also need to be marked with `@pytest.mark.xfail` so that Pytest knows not to run them. I can do this

VeckoTheGecko (commented):
Just dropping a partial review now since a lot of thoughts have come up. Some small review comments for the timebeing.

I think we need to answer an important question moving forward:

How are `lat`,`lon`,`depth` handled? Previously these were numpy arrays anywhere from dimension 1-4 depending on the geometry of the grid itself. This became a maintenance burden since every part of the code that needed these values also needed to include logic to deal with it for each different type of grid. With the addition of unstructured grids if we go the same route it will just further complicate things. We need to find a good abstraction that we can use to bridge the gap between datasets and interpolation, so that we can tuck this complexity into a module and provide a clean API to the rest of the codebase (I can start investigating what sort of stuff is used internally in xarray and scipy for interpolation - perhaps there is inspiration we can take from those packages). For reference, @fluidnumerics-joe what is the structure of `lon`,`lat`,`depth` in the case of unstructured grids?

I don't know if this PR is the right place to answer this, or whether these questions should block this PR. Another option would be to create new issues to highlight these concerns. Next week I'm also hardly in the office (conference for 2 days, and then shortened week due to easter). @fluidnumerics-joe  I assume this might be a blocker on your end? Not sure what other development you're working on.

v4-tests/test_uxarray_fieldset.py
fluidnumerics-joe: Resolved in https://github.com/OceanParcels/Parcels/pull/1946/commits/4bf5096b64674c3249dee5b26b5e7b1570555d06

parcels/field.py
```diff
@@ -76,142 +66,169 @@ def _deal_with_errors(error, key, vector_type: VectorType):
...
-        lat=None,
-        depth=None,
-        time=None,
-        grid=None,
-        mesh: Mesh = "flat",
-        time_origin: TimeConverter | None = None,
-        interp_method: InterpMethod = "linear",
+        name: str,
+        data: xr.DataArray | ux.UxDataArray,
+        grid: ux.Grid | None = None,  # To do : Once parcels.Grid class is added, allow for it to be passed here
```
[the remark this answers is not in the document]
fluidnumerics-joe: Got it. I'll make these changes

fluidnumerics-joe:
> Just dropping a partial review now since a lot of thoughts have come up. Some small review comments for the timebeing.
> 
> I think we need to answer an important question moving forward:
> 
> How are `lat`,`lon`,`depth` handled? Previously these were numpy arrays anywhere from dimension 1-4 depending on the geometry of the grid itself. This became a maintenance burden since every part of the code that needed these values also needed to include logic to deal with it for each different type of grid. With the addition of unstructured grids if we go the same route it will just further complicate things. IMO we need to find a good abstraction that we can use to bridge the gap between datasets and interpolation, so that we can tuck this complexity into a module and provide a clean API to the rest of the codebase (I can start investigating what sort of stuff is used internally in xarray and scipy for interpolation - perhaps there is inspiration we can take from those packages). For reference, @fluidnumerics-joe what is the structure of `lon`,`lat`,`depth` in the case of unstructured grids?
>

In an unstructured grid, we think about a mesh as a collection of *vertices* (often called *nodes*). *Elements*  are formed by creating convex shapes by joining sets of nodes; for FESOM and ICON, elements are just extruded triangular cells.  In FESOM and ICON, the mesh is unstructured in 2-D (lat,lon) but extruded in the depth direction. Because of this the unstructured mesh description is 2-D; the third depth dimension is handled similarly to structured grid models.

In this context, the top and bottom *faces* of each element is related to three vertices (where the vertices are defined by lat and lon only). *Edges* are just lines that connect two vertices; you can also relate an element to the edges that make up its boundary. Depending on the discretization, a field can be located on a `face` center, `edge` center, or on a `node`.

For a `uxarray.UxDataArray` or `uxarray.UxDataset` there are the following coordinates
* `edge_lat`, `edge_lon` dimensioned `(n_edge)` - These give the latitude and longitude at edge centers 
* `face_lat`, `face_lon` dimensioned `(n_face)` - These give the latitude and longitude at face centers 
* `node_lat`, `node_lon` dimensioned `(n_node)` - These give the latitude and longitude on nodes/vertices
* `nz1` dimensioned `(nz1)` - This is a depth coordinate registered to the (3-D) element center vertically
* `nz` dimensioned `(nz)` - This is a depth coordinate registered to the (3-D) element faces vertically

The `uxarray.UxDataArray.uxgrid` or `uxarray.UxDataset.uxgrid` attribute provides the connectivity tables that relate
* Edges to nodes - `edge_node_connectivity`
* Faces to edges - `face_edge_connectivity` 
* Faces to nodes - `face_node_connectivity`
* Faces to faces (useful for neighbor lookups) - `face_face_connectivity`

There's no hope of having a single memory layout that would efficiently be used for rectinlinear structured, curvilinear structured, and unstructured lat, lon, and depth. In the end, the primary workhorse of the `Field` class is the `eval` method; under the hood this primarily houses a `search` method to find the nearest grid cell for local interpolation and then the interpolation method. By having the `interp_method` as a callable, we can maintain a suite of built-in interpolators that can be used for the various mesh types we plan to support. Although the `search` methods are not a callable attribute of the `Field` class, it's apparent we need a suite of `search` methods appropriate for each mesh type as well. 

Keeping these bits underneath the call stack of the `Field.eval` method hides the complexity of evaluating a field at a point in space and time from everything in `Fieldset` and higher. A big difference between us and xarray and scipy is that we are trying to expose the ability to also provide your own interpolator. At this level, the end user needs to be keenly aware of the data available for the field they plan to tie their interpolator to. We can't be expected to hide that (nor should we).


> I don't know if this PR is the right place to answer this, or whether these questions should block this PR. Another option would be to create new issues to highlight these concerns. Next week I'm also hardly in the office (conference for 2 days, and then shortened week due to easter). @fluidnumerics-joe I assume this might be a blocker on your end? Not sure what other development you're working on.

Not really a blocker right now. I have built-in interpolators for face registered and node registered data, which is sufficient for a FESOM and ICON mesh. The `Field.eval` method currently leverages a spatial hash search or nearest neighbor search (depending on the presence of an optional `ei` input). In terms of letting this block the PR, I'd push for getting this into v4-dev sooner so that we can all work from another common point with this level of xarray/uxarray integration.

VeckoTheGecko:
Thank you for the in depth response here @fluidnumerics-joe and for the overview of how unstructured grids are handled here! 

> A big difference between us and xarray and scipy is that we are trying to expose the ability to also provide your own interpolator.

Yes. I have seen quite some stalled discussions in the xarray repo about custom interpolation.

---
Ok, going to give everything a once over now before approving just to clarify some things and get familiar.

parcels/fieldset.py
```diff
@@ -23,127 +17,103 @@ class FieldSet:
...
+    during execution for interpolation.  Each `UXDataArray.attributes` field dictionary
+    must have:
+      * "location" key set to "face", "node", or "edge" to define which pairing of points a field is associated with.
+      * "mesh" key to define which parent model the fields are associated with (e.g. "fesom_mesh", "icon_mesh")
+
     """
 
-    def __init__(self, U: Field | None, V: Field | None, fields=None):
-        self.gridset = GridSet()
+    def __init__(self, datasets: list[xr.Dataset | ux.UxDataset]):
```
VeckoTheGecko, line 45: Hmm. Perhaps it would be better to do `__init__(self, fields: list[Field])`? It feels a bit construed passing in datasets and the constructing the fields within the init. Thoughts?

parcels/fieldset.py
```diff
@@ -174,7 +144,8 @@ def add_field(self, field: Field, name: str | None = None):
             raise RuntimeError(f"FieldSet already has a Field with name '{name}'")
         else:
             setattr(self, name, field)
-            self.gridset.add_grid(field)
+            self._gridset_size += 1
+            self._fieldnames.append(name)
```
VeckoTheGecko, line 148: Not 100% sure if it would be as simple as `_gridset_size += 1` since fields could share a grid? at the moment this is tied to `len(self._fieldnames)` so maybe remove `_gridset_size` and update the `gridset_size` property to return `len(self._fieldnames)` then later we could handle duplicate grids if needed?

parcels/fieldset.py
```diff
@@ -194,21 +165,48 @@ def add_constant_field(self, name: str, value: float, mesh: Mesh = "flat"):
                correction for zonal velocity U near the poles.
             2. flat: No conversion, lat/lon are assumed to be in m.
         """
-        self.add_field(Field(name, value, lon=0, lat=0, mesh=mesh))
+        time = 0.0
+        values = np.zeros((1, 1, 1, 1), dtype=np.float32) + value
+        data = xr.DataArray(
```
VeckoTheGecko, line 170:
```suggestion
        values = np.full((1, 1, 1, 1), value)
        data = xr.DataArray(
```
And update the docstring for `value : float` to remove 32 bit requirement?

parcels/fieldset.py
```diff
@@ -274,33 +237,26 @@ def _parse_wildcards(cls, paths, filenames, var):
         return paths
```
VeckoTheGecko, line 237: `_parse_wildcards` is legacy code and can be removed

parcels/fieldset.py
```diff
@@ -274,33 +237,26 @@ def _parse_wildcards(cls, paths, filenames, var):
...
 
     @classmethod
-    def from_netcdf(
+    def from_data(
```
VeckoTheGecko, line 240: Do you think we could get rid of `from_data` altogether @fluidnumerics-joe ? Or do you see a usecase for it that would be difficult to achieve via getting the user xarray datasets themselves?

VeckoTheGecko (commented):
Ok. Managed to look through in full now

parcels/field.py
```diff
@@ -37,20 +27,20 @@
...
+__all__ = ["Field", "GridType", "VectorField"]
 
-    from parcels.fieldset import FieldSet
 
-__all__ = ["Field", "VectorField"]
+class GridType(IntEnum):
+    RectilinearZGrid = 0
+    RectilinearSGrid = 1
+    CurvilinearZGrid = 2
+    CurvilinearSGrid = 3
```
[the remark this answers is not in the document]
fluidnumerics-joe: My thinking is that `grid.py` and `gridset.py` in their current form are going away since the grid is intimately a part of uxarray objects. Are you migrating this definition to `v4/grid.py` ?

parcels/field.py
```diff
@@ -76,142 +66,169 @@ def _deal_with_errors(error, key, vector_type: VectorType):
...
+    * For a C-Grid, the "location" setting for a field has the following interpretation:
+        * "node" ~> the field is associated with the vorticity points (node_lat, node_lon)
+        * "face" ~> the field is associated with the tracer points (face_lat, face_lon)
+        * "x_edge" ~> the field is associated with the u-velocity points (face_lat, node_lon)
+        * "y_edge" ~> the field is associated with the v-velocity points (node_lat, face_lon)
+
+    When using a uxarray.UxDataArray object,
+    * The uxarray.UxDataArray.UxGrid object must have the "Conventions" attribute set to "UGRID-1.0"
+      and the uxarray.UxDataArray object must comply with the UGRID conventions.
+      See https://ugrid-conventions.github.io/ugrid-conventions/ for more information.
```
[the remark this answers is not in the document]
fluidnumerics-joe: I'll add a `TODO Nick` here

parcels/fieldset.py
fluidnumerics-joe: If we change this to list[Field], then the end user would have to create fields from `xr.DataArray` or `ux.UxDataArray` objects. The odd bit here is that when calling `open_dataset` or `open_mfdataset` you get a `xr.DataSet` or `ux.UxDataSet` object, not a `DataArray`. Changing this to a `list[Field]`, IMO, would put more boilerplate on the end user.

parcels/fieldset.py
fluidnumerics-joe: Made the change in https://github.com/OceanParcels/Parcels/pull/1946/commits/81793d980c6adee0c310374be4ae33e750b79886

parcels/fieldset.py
fluidnumerics-joe: Resolved in https://github.com/OceanParcels/Parcels/pull/1946/commits/9d3a7e28907ec56428f9e9efffeee35941a3890c

parcels/fieldset.py
fluidnumerics-joe: I think it would be ok to do away with `from_data` entirely, so long as we put together docs on how to construct an xarray dataset that "plays nice" with parcels. What do you think @erikvansebille ?

erikvansebille: Yep, I agree that we don't need a `from_data` anymore in v4.

VeckoTheGecko: > so long as we put together docs on how to construct an xarray dataset that "plays nice" with parcels

yes, 100%. I think it would be good to even add tooling for validating datasets are compliant (both in metadata and in variable naming) and giving informative error messages

parcels/field.py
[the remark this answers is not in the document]
VeckoTheGecko: Yes, they'll be going away. Ok, let's leave this here and `grid.py` can be cleaned up in another PR

parcels/fieldset.py
VeckoTheGecko: TLDR; do you want me to PR into this with what I have in mind?

### `FieldSet` state

Looking through the code a bit more, I think we should be careful with what state is stored on the FieldSet object and work closer to the conceptual definition of a Fieldset when it comes to the data that we are storing.

A FieldSet is a collection of Fields, which themselves have data and a grid. This "collection of Fields" (e.g., dictionary) should really be containing Field objects.

Currently here we store `.datasets: list[xr.Dataset | ux.UxDataset]`, as well as use `setattr(self, name, field)` to store the actual Field objects - the Field objects being probably the more useful one as it contains more information. Its a bit of doubling up since `.datasets` can be derived from the `Fields`. Using `setattr(self, name, field)` is a bit of an anti-pattern because:

- It makes it difficult to iterate through the fields when necessary (often iterating through `self.__dict__` and doing `isinstance()` checks to filter down to just the FIelds)
- Static analysis tools struggle with it (since its a dynamic definition, so is only determined at runtime)

I think if we update the class so that the FieldSet contains `fields: dict[str, Field | VectorField] I think that would be nice and we can avoid all the `setattr` stuff (which has been on my wishlist (old issue #1663) since it isn't needed). We can modify `__getattr__` to then still allow for `fieldset.U` (i.e., it will look at the `fields` object for the `U`).

### `__init__`
I think that initialisers should only have two responsibilities:
- validate that the input is correct/compatible with Parcels (i.e., format, naming convention, etc)
- store the data

This results in an initialiser that is quite thin, and a signature that is very closely related to the state stored on the object (xarray is a good example for this). This flexibility is nice since all code paths will go through the initialiser.

> would put more boilerplate on the end user

I don't think so. I imagine users will mostly call classmethods to construct their FieldSets. Down the line we can also update the `__init__` to be more ergonomic (e.g., allow users so give a dict of `fieldname: (data, grid, ...)` rather than having to construct the Field objects themselves - similar to how xarray does it).

Maybe we can do `from_datasets` as a classmethod for the timebeing to bridge this gap @fluidnumerics-joe ? So we can update the initialiser and then make the current calls from `FieldSet(` to `FieldSet.from_datasets(`

---

Just a reminder I'm at a conference Tue/Wed so can be slow to respond

parcels/fieldset.py
fluidnumerics-joe: I'll get `from_data` removed here

parcels/fieldset.py
VeckoTheGecko: Yep, definitely, let's ship this and get it into another PR - that works well

parcels/fieldset.py
fluidnumerics-joe: Resolved in [0af7de9](https://github.com/OceanParcels/Parcels/pull/1946/commits/0af7de934fdd73f9bf149d1fcbafdb6942f0af9c)

PR #1954, opened 2025-04-03
Update dark-theme docs

erikvansebille:
Now that we have a nice new dark theme for ocean parcels.org (https://github.com/OceanParcels/oceanparcels_website/pull/141), it's time to also update the docs-page. This PR updates the logos and updates the homepage animation to have transparent background

I'm basing this PR on v3, but it should probably also be merged into v4?


VeckoTheGecko (approved):
Looks good

> I'm basing this PR on v3, but it should probably also be merged into v4?

~~Should we just merge straight into v4 instead of into both? I'm sure more visual tweaks will come up, and I think focusing on v4 would be simpler for these :)~~ Just realised that the logos are included here. Yes, in that case let's update both branches. Just when merging choose "squash merge" so that its easy to apply the changes to the other branch

erikvansebille:
> Should we just merge straight into v4 instead of into both? I'm sure more visual tweaks will come up, and I think focusing on v4 would be simpler for these :)

I agree that v4 should indeed be the most important target. I tried to change but then also get a lot of other commits... How do I most easily rebase(?)

But I also think it's too nice to not also update for v3. Can't we just cherry-pick the commits in this PR and also merge them in `main` or `v4` too?

erikvansebille:
There's still something strange going on where the transparent background of the homepage animation is not rendered correctly in dark mode. Not sure if this is a markdown issue or a sphinx issue (it also happens when I sphinx-build locally)? I can't really find a solution online....

VeckoTheGecko:
> can't really find a solution online....

Looks to be https://github.com/pydata/pydata-sphinx-theme/issues/918 ?

If you check out https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/light-dark.html#images-and-content-that-work-in-both-themes ?

We might need to update our version of the theme. Will check

PR #1966, opened 2025-04-08
New grid object for structured grids

VeckoTheGecko:
Currently making this PR for visibility and so that we can align our approaches between #1946 and this.

Note that this PR only adds the new Grid object (along with a GridAdapter object) but doesn't _use_ them in the larger codebase. That will be for future development.

Just to clarify, I'm not 100% sure if `GridAdapter` will make it into the final v4 and can't even particularly comment on its usefulness due to the change of state/grid manipulation done by the rest of the codebase. It might be better to bring the rest of the codebase to work with the new Grid object.

- [] Fixes #
- [x] Added tests
- [x] Added documentation

cc @fluidnumerics-joe


fluidnumerics-joe:
@VeckoTheGecko - Do we really want to maintain a duplicate of code already provided by `xgcm` ? ie, why wouldn't we want to just use `xgcm` as a dependency ?

VeckoTheGecko:
> @VeckoTheGecko - Do we really want to maintain a duplicate of code already provided by `xgcm` ? ie, why wouldn't we want to just use `xgcm` as a dependency ?

Oh yes, I should have mentioned this in the issue body. I think we should vendor as opposed to adding it as a dependency for the following reasons:

1. xgcm is quite a specific tool (i.e., enabling grid aware calculus on ocean grids). We only want the "grid aware" part, which (a) is not particularly part of the public API, and (b) is a relatively small part of their codebase . Depending on xgcm would couple us to their development, and relying on their private API would be potentially unstable. Also they're looking to get rid of the Axis object from the public API https://github.com/xgcm/xgcm/issues/405 , which I think further would make things difficult for us.
- xgcm doesn't look to have active maintainers ([source](https://discourse.pangeo.io/t/status-of-xgcm/4986/5)). Their last release was in 2022 (they have an intiative for a new release though https://github.com/xgcm/xgcm/issues/670). The project is still stable though from what I can tell

fluidnumerics-joe:
> > @VeckoTheGecko - Do we really want to maintain a duplicate of code already provided by `xgcm` ? ie, why wouldn't we want to just use `xgcm` as a dependency ?
> 
> Oh yes, I should have mentioned this in the issue body. I think we should vendor as opposed to adding it as a dependency for the following reasons:
> 
> 1. xgcm is quite a specific tool (i.e., enabling grid aware calculus on ocean grids). We only want the "grid aware" part, which (a) is not particularly part of the public API, and (b) is a relatively small part of their codebase . Depending on xgcm would couple us to their development, and relying on their private API would be potentially unstable. Also they're looking to get rid of the Axis object from the public API [Lets get rid of `Axis` xgcm/xgcm#405](https://github.com/xgcm/xgcm/issues/405) , which I think further would make things difficult for us.
> 
> * xgcm doesn't look to have active maintainers ([source](https://discourse.pangeo.io/t/status-of-xgcm/4986/5)). Their last release was in 2022 (they have an intiative for a new release though [Time to issue a new release xgcm/xgcm#670](https://github.com/xgcm/xgcm/issues/670)). The project is still stable though from what I can tell

All good reasons.

VeckoTheGecko:
For context as well https://github.com/xgcm/xgcm/discussions/658

VeckoTheGecko:
I think that this is ready for review. Post-merge we can work out how to integrate this into Field, and decide if GridAdapter is needed after all

parcels/v4/comodo.py
erikvansebille, line 1: What does the word 'comodo' stand for? It's a bit non-descriptive to me

tests/v4/grid_datasets.py
erikvansebille, line 1: Would/should these then go to the _datasets.py sub package of #1972?

VeckoTheGecko: Yes, will consolidate this now that we have that merged

parcels/v4/comodo.py
VeckoTheGecko: XGCM relies on the COMODO conventions, which they link to [here](https://web.archive.org/web/20160417032300/http://pycomodo.forge.imag.fr/norm.html).... I'm not sure the history here and whether COMODO or pycomodo even exists anymore - the website is only available via web archive. Good enough for now until something better comes along?

erikvansebille (approved):

fluidnumerics-joe (approved):

jbusecke:
> > @VeckoTheGecko - Do we really want to maintain a duplicate of code already provided by `xgcm` ? ie, why wouldn't we want to just use `xgcm` as a dependency ?
> 
> Oh yes, I should have mentioned this in the issue body. I think we should vendor as opposed to adding it as a dependency for the following reasons:
> 
Also they're looking to get rid of the Axis object from the public API [Lets get rid of `Axis` xgcm/xgcm#405](https://github.com/xgcm/xgcm/issues/405) , which I think further would make things difficult for us.
> 
> 
>     * xgcm doesn't look to have active maintainers ([source](https://discourse.pangeo.io/t/status-of-xgcm/4986/5)). Their last release was in 2022 (they have an intiative for a new release though [Time to issue a new release xgcm/xgcm#670](https://github.com/xgcm/xgcm/issues/670)). The project is still stable though from what I can tell

Hey @VeckoTheGecko sorry for the late comment (we just released xgcm 0.9.0 and I saw the link). I know we chatted about this personally, but I also wanted to state this publicly. 

>     1. xgcm is quite a specific tool (i.e., enabling grid aware calculus on ocean grids). We only want the "grid aware" part, which (a) is not particularly part of the public API, and (b) is a relatively small part of their codebase . Depending on xgcm would couple us to their development, and relying on their private API would be potentially unstable. 

I would be curious what parts of the API you are using here. I do not consider the Axis/Grid as private to be honest. We abandoned the idea of dropping axis, so I do not think that will be a risk for you folks here. I also think that the grid-awareness (and associated metadata parsing) is really the 'core product' of xgcm, whereas the calculus could at this point be entirely coupled out into one or many different packages that supply grid_ufuncs. 

As you know my time is limited at the moment but xgcm is not dead. Might be worth checking out the new release too!

Issue #1965, opened 2025-04-08
Package structure


VeckoTheGecko:
As we move into version 4, I would like to see an update to the structure of the parcels package to something along the lines of:

```
parcels
├── __init__.py
├── _core
│   ├── __init__.py
│   ├── field.py
│   ├── fieldset.py
│   ├── grid.py
│   ├── interpolation_utils.py
│   ├── utils.py
│   └── ...
├── tutorial.py # or example.py, a namespace for us to put utilities like downloading datasets that are specific to the tutorial notebooks
├── _loggers.py
├── warnings.py
├── errors.py
├── _version.py
├── _repr.py
├── _compat.py
├── _typing.py
├── _datasets # namespace to put utilities for generating xarray datasets for use in testing or tutorials
│   ├── __init__.py
│   └── ...
└── ...
```

I propose implementing this final structure _at the end_ of version 4 development so that we don't interfere with the development process with merge conflicts due to many renamed files and moved functions.

This structure has the following benefits:
- **Improved organisation**: Delineates which parts of the package are core to Parcels (grids, fields, particles, particlesets, writers, interpolation helpers), and which are more "stand alone" supporting code (user utilities, warnings, reprs, compatability items, dataset generation).
- **Allows us to craft our namespaces**: Currently our `__init__.py` files have many `from ... import *` statements meaning that sometimes things can creep into the public API on the `parcels.` namespace by accident just when added to a file. This also has the side-effect of running into circular imports. By promoting mostly empty `__init__.py` files, and by selectively importing in `parcels/__init__.py` we can be selective and have more control over what we expose to the user under the `parcels.` namespace.
  - **Makes the codebase "private by default"**: With this new structure (namely having `_core`), items coming into the public API becomes a conscious decision.

PR #1972, opened 2025-04-09
Add `parcels._datasets` subpackage and developer documentation

VeckoTheGecko:
In line with #1965, I have added a `parcels._datasets` subpackage. See the developer documentation in `parcels/_datasets/__init__.py` for more info.

I think that this really helps us with testing, user documentation (for the advanced users, who want to see exactly all the datasets that we support), and maintenance (since in our tests we can just rely on these example datasets, rather than having to create them from scratch as we currently do quite a bit).

This will will need to be surfaced in some way in the documentation down the line - I'll make an issue for that now.

### DIFF
--- a/parcels/_datasets/__init__.py
+++ b/parcels/_datasets/__init__.py
@@ -1,0 +1,46 @@
+"""
+Datasets compatible with Parcels.
+
+This subpackage uses xarray to generate *idealised* structured and unstructured hydrodynamical datasets that are compatible with Parcels. The goals are three-fold:
+
+1. To provide users with documentation for the types of datasets they can expect Parcels to work with.
+2. To supply our tutorials with hydrodynamical datasets.
+3. To offer developers datasets for use in test cases.
+
+Note that this subpackage is part of the private API for Parcels. Users should not rely directly on the functions defined within this module. Instead, if you want to generate your own datasets, copy the functions from this module into your own code.
+
+Developers, note that only idealised datasets that are (a) quick to generate, and (b) only use dependencies already shipped with Parcels, should be added to this subpackage. Real world datasets should be added to the `parcels-data` repository. No data files should be added to this subpackage.
+
+Parcels Dataset Philosophy
+-------------------------
+
+When adding example datasets, there may be a tension between wanting to add a specific example or wanting to add machinery to generate completely arbitrary datasets (e.g., with different grid resolutions, with different ranges, with different datetimes etc.). There are trade-offs to both approaches:
+
+Working with specific hardcoded examples:
+
+* Pros
+    * the example is stable and self-contained
+    * easy to see exactly what the dataset is, there little to no dependency on other functions defined in the same module
+      * datasets don't "break" due to changes in other functions (e.g., grid edges becoming out of sync with grid centres)
+* Cons
+    * inflexible for use in tests where you want to test a large range of datasets, or you want to test a specific resolution
+
+Working with generated datasets is the opposite of all the above.
+
+Most of the time we only want a single dataset. For example, for use in a tutorial, or for testing a specific feature of Parcels - such as (in the case of structured grids) checking that the grid from a certain vendor is correctly parsed, or checking that indexing is correctly picked up. As such, one should often opt for hardcoded datasets. These are more stable and easier to see exactly what the dataset is. We may have specific examples that become the default "go to" dataset for testing when we don't care about the specific details of the dataset.
+
+Sometimes we may want to test Parcels against a whole range of datasets varying in a certain way - to ensure Parcels works as expected for subspace of possible datasets. For these, we should add machinery to create generated datasets.
+
+Structure
+--------
+
+This subpackage is broken down into structured and unstructured parts. Each of these have common submodules:
+
+* ``providers`` -> hardcoded datasets with the intention of mimicking datasets from a certain provider
+* ``generic`` -> hardcoded datasets that are generic, and not tied to a certain provider (instead focusing on the fundamental properties of the dataset)
+* ``generated`` -> functions to generate datasets with varying properties
+* ``utils`` -> any utility functions necessary related to either generating or validating datasets
+
+There may be extra submodules than the ones listed above.
+
+"""
--- a/parcels/_datasets/__init__.py
+++ b/parcels/_datasets/__init__.py
@@ -20,7 +20,7 @@ Working with specific hardcoded examples:
 
 * Pros
     * the example is stable and self-contained
-    * easy to see exactly what the dataset is, there little to no dependency on other functions defined in the same module
+    * easy to see exactly what the dataset is, there is little to no dependency on other functions defined in the same module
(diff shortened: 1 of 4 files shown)


VeckoTheGecko:
@erikvansebille could you checkout the dev documentation and let me know what you think?

VeckoTheGecko:
ci failure is unrelated - not sure why it flaked. As good as green

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+Most of the time we only want a single dataset. For example, for use in a tutorial, or for testing a specific feature of Parcels - such as (in the case of structured grids) checking that the grid from a certain vendor is correctly parsed, or checking that indexing is correctly picked up. As such, one should often opt for hardcoded datasets. These are more stable and easier to see exactly what the dataset is. We may have specific examples that become the default "go to" dataset for testing when we don't care about the specific details of the dataset.
+
+Sometimes we may want to test Parcels against a whole range of datasets varying in a certain way - to ensure Parcels works as expected for subspace of possible datasets. For these, we should add machinery to create generated datasets.
+
+Structure
+--------
+
+This subpackage is broken down into structured and unstructured parts. Each of these have common submodules:
+
+* ``providers`` -> hardcoded datasets with the intention of mimicking datasets from a certain provider
```
VeckoTheGecko, line 39: Perhaps another word than `providers`? I'm meaning more NEMO/Mitgcm/CROCO/ROMS/mom5 etc ...

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+This subpackage uses xarray to generate *idealised* structured and unstructured hydrodynamical datasets that are compatible with Parcels. The goals are three-fold:
+
+1. To provide users with documentation for the types of datasets they can expect Parcels to work with.
+2. To supply our tutorials with hydrodynamical datasets.
+3. To offer developers datasets for use in test cases.
+
+Note that this subpackage is part of the private API for Parcels. Users should not rely directly on the functions defined within this module. Instead, if you want to generate your own datasets, copy the functions from this module into your own code.
+
+Developers, note that only idealised datasets that are (a) quick to generate, and (b) only use dependencies already shipped with Parcels, should be added to this subpackage. Real world datasets should be added to the `parcels-data` repository. No data files should be added to this subpackage.
```
erikvansebille, line 12:
```suggestion
Developers, note that you should only add functions that create idealised datasets to this subpackage if they are (a) quick to generate, and (b) only use dependencies already shipped with Parcels. No data files should be added to this subpackage. Real world data files should be added to the `OceanParcels/parcels-data` repository on GitHub. 
```

parcels/_datasets/__init__.py
erikvansebille, line 1: Perhaps rename to `dataset_generators` or `dataset_functions` to make extra clear that this is not the place for data files, but only generator functions?

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+Parcels Dataset Philosophy
+-------------------------
+
+When adding example datasets, there may be a tension between wanting to add a specific example or wanting to add machinery to generate completely arbitrary datasets (e.g., with different grid resolutions, with different ranges, with different datetimes etc.). There are trade-offs to both approaches:
```
erikvansebille, line 17: 'completely arbitrary' is a somewhat strange wording here. They are per definition not arbitrary. Do you mean e.g. 'highly complex' or 'very versatile'? Later on you call them 'generic'; perhaps use that word throughout?

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+Working with specific hardcoded examples:
+
+* Pros
+    * the example is stable and self-contained
+    * easy to see exactly what the dataset is, there little to no dependency on other functions defined in the same module
```
erikvansebille, line 23:
```suggestion
    * easy to see exactly what the dataset is, there is little to no dependency on other functions defined in the same module
```

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+Note that this subpackage is part of the private API for Parcels. Users should not rely directly on the functions defined within this module. Instead, if you want to generate your own datasets, copy the functions from this module into your own code.
+
+Developers, note that only idealised datasets that are (a) quick to generate, and (b) only use dependencies already shipped with Parcels, should be added to this subpackage. Real world datasets should be added to the `parcels-data` repository. No data files should be added to this subpackage.
+
+Parcels Dataset Philosophy
+-------------------------
+
+When adding example datasets, there may be a tension between wanting to add a specific example or wanting to add machinery to generate completely arbitrary datasets (e.g., with different grid resolutions, with different ranges, with different datetimes etc.). There are trade-offs to both approaches:
+
+Working with specific hardcoded examples:
```
erikvansebille, line 19: If you call them 'hardcodes examples' here, I would use the same wording in the paragraph above. So that users understand the link between the two

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+* Pros
+    * the example is stable and self-contained
+    * easy to see exactly what the dataset is, there little to no dependency on other functions defined in the same module
+      * datasets don't "break" due to changes in other functions (e.g., grid edges becoming out of sync with grid centres)
+* Cons
+    * inflexible for use in tests where you want to test a large range of datasets, or you want to test a specific resolution
+
+Working with generated datasets is the opposite of all the above.
+
+Most of the time we only want a single dataset. For example, for use in a tutorial, or for testing a specific feature of Parcels - such as (in the case of structured grids) checking that the grid from a certain vendor is correctly parsed, or checking that indexing is correctly picked up. As such, one should often opt for hardcoded datasets. These are more stable and easier to see exactly what the dataset is. We may have specific examples that become the default "go to" dataset for testing when we don't care about the specific details of the dataset.
```
erikvansebille, line 30: Is the word 'vendor' the best choice here? On the ocean parcels.org website, we call them " output from (ocean) circulation models." Use that same terminology throughout here?

Also, the repeated 'specific' in th least sentence is a bit confusing. Reword?

parcels/_datasets/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+Sometimes we may want to test Parcels against a whole range of datasets varying in a certain way - to ensure Parcels works as expected for subspace of possible datasets. For these, we should add machinery to create generated datasets.
```
erikvansebille, line 32:
```suggestion
Sometimes we may want to test Parcels against a whole range of datasets varying in a certain way - to ensure Parcels works as expected for subsets or combinations of possible datasets. For these, we should add machinery to create generated datasets.
```

parcels/_datasets/__init__.py
erikvansebille: `circulationmodels`? See also the comment above about aligning with the text on oceanparcels.org

PR #1978, opened 2025-04-17
Updating Parcels pitch in docs

erikvansebille:
As we move towards v4, I feel it's time to say goodbye to the tongue-in-cheek bacronym for Parcels. We don't use it anymore on the revamped oceanparcels.org website, and when we remove it here in the readme it will be gone from the codebase too.

I also aligned the pitch with that on the oceanparcels website


docs/index.rst
```diff
@@ -2,7 +2,7 @@ Parcels documentation
 =====================
 
 
-Welcome to the documentation of Parcels. **Parcels** (Probably A Really Computationally Efficient Lagrangian Simulator) is a set of Python classes and methods to create customisable particle tracking simulations using output from Ocean Circulation models. Parcels can be used to track passive and active particulates such as water, plankton, `plastic <http://www.topios.org/>`_ and `fish <https://github.com/Jacketless/IKAMOANA>`_.
+Welcome to the documentation of Parcels. **Parcels** provides a set of Python classes and methods to create customisable particle tracking simulations using gridded output from (ocean) circulation models. Parcels can be used to track passive and active particulates such as water, plankton, `plastic <http://www.topios.org/>`_ and `fish <https://github.com/Jacketless/IKAMOANA>`_.
```
VeckoTheGecko, line 5: > set of Python classes and methods to create

A bit off-tangent for this PR, but thoughts on "modelling framework"?

VeckoTheGecko: > because they typically think of a modelling framework as a black-box code

I see. From a software engineering/data science POV hearing "framework" I think of something like Pandas (dataframes) and Sklearn (fit-predict) APIs that offer a lot of flexibility but provide a structured way of thinking about the problem.

> So I'd prefer to keep as is
Sounds good. Our main target demographic is oceanographers anyway :)

VeckoTheGecko (approved):

PR #1980, opened 2025-04-21
Patch/uxarray

fluidnumerics-joe:
- [ X ] Chose the correct base branch (`main` for v3 changes, `v4-dev` for v4 changes)

This PR brings in a couple patches uncovered during testing/development of examples with the new v4 field and fieldset.

PR #1983, opened 2025-04-22
CI: Run `tests/v4` separately from whole test suite and disable MacOS and Windows tests

VeckoTheGecko:
- Fixes None

Changes:
- Add test to matrix so that `tests/v4` is also run separately by itself
  - This check makes it easier to parse whether a set of changes has broken other functionality that has been added in v4 (which is more pressing than test failures from old v3 tests)
- Disable tests on Windows and Mac
  - These will be introduced again at a later stage when they're of interest again. This is mainly to reduce CI wait times and resource usage.


VeckoTheGecko:
@fluidnumerics-joe  just updated this so it can be merged before #1980 . Red is due to some things in #1980. Should we go ahead and merge so you can rebase?

PR #1986, opened 2025-04-24
Add Field tests and type guarding in Field init

VeckoTheGecko:
This PR does some cleanup in the `field.py` file, adds type guarding in the Field initialiser, and adds some tests for, laying some of the prep work for structured grid compatability in the Field object.

- [x] Added tests

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -13,6 +13,7 @@ from uxarray.grid.neighbors import _barycentric_coordinates
 from parcels._typing import (
     Mesh,
     VectorType,
+    assert_valid_mesh,
 )
 from parcels.tools._helpers import default_repr, field_repr
 from parcels.tools.converters import (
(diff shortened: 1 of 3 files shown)


parcels/field.py
```diff
@@ -631,47 +653,58 @@ def __getitem__(self, key):
...
+            "Field is missing a 'Conventions' attribute in the field's metadata. "
             "This attribute is required for uxarray.UxDataArray objects."
         )
     if grid.attrs["Conventions"] != "UGRID-1.0":
         raise ValueError(
-            f"Field {name} has a 'Conventions' attribute that is not 'UGRID-1.0'. "
+            "Field has a 'Conventions' attribute that is not 'UGRID-1.0'. "
             "This attribute is required for uxarray.UxDataArray objects."
             "See https://ugrid-conventions.github.io/ugrid-conventions/ for more information."
         )
```
VeckoTheGecko, line 697: I think its nice removing `name` from here as it's only used in the error message itself. The context can be added via `add_note` (introduced in 3.11) as the message propagates up

fluidnumerics-joe (approved):

Issue #1988, opened 2025-04-25
[v4 Discussion] Remove allow_time_extrapolation


fluidnumerics-joe:
I'd like to propose removing the `Field.allow_time_extrapolation` attribute (in v4) and push in favor of following a set of assumed semantics

* If the length to the time dimension is 1, the field is assumed constant in time, and a parcels simulation should be allowed to integrate forward in time as long as it likes
* If the length of the time dimension is greater than 1, then the simulation time should only be permissible within the time levels defined in the `Field` object.

Given that I'm still learning some of the many internal features in parcels that I'm certain exist to enable important use cases, I can't gather why time extrapolation would be valid except in the case of having a constant-in-time field.

erikvansebille:
Ah good point; you may indeed be correct that the only meaningful use of `allow_time_extrapolation` is when the Field has time dimension 1. And then, it should be the default (because interpolation is impossible), so even in that case we don't need it as an option

In fact, `allow_time_extrapolation` has always been a bit dangerous for users, because they will not realise when their simulation extends beyond the length of the dataset if they run with `True`. Since the velocity fields are essentially frozen-in-time in 'extrapolation' mode, resulting trajectories can be wrong. 
 
So yes, I think we can remove the `allow_time_extrapolation` attribute in v4!

VeckoTheGecko:
Regarding implementation details, I propose that a simulation can only be run on the overlapping window of all fields.

```mermaid
gantt
    title Field Time Periods and Overlap
    dateFormat  YYYY-MM-DD HH:mm
    section Field
    Field 1      :p1, 2024-06-10 09:00, 2h
    Field 2      :p2, 2024-06-10 10:00, 2h
    Field 3      :p3, 2024-06-10 10:30, 2h
    section Overlap
    Overlap (1,2,3):overlap, 2024-06-10 10:30, 0.5h

```

- All particles must be initialised within this window
- The runtime should not exceed this window (and start time should be within the window)

Note that the usecase of users defining Fields and conditionally checking fields based on time domains and current clock wouldn't be supported (e.g., a user can't have a field for each month of a year, and then access those fields using heaps of if conditions in the kernel) - this would need to be done via Xarray to create a matching dataset

erikvansebille:
Yes, I agree that the particles can only be run in the overlapping window. That is also the intent of these lines in the current v3 `main`:
[...]

PR #1989, opened 2025-04-28
Cleaning up Field state 

VeckoTheGecko:
This PR cleans up the state on the `Field` object (building on #1946)

This PR implements the following changes:

- remove `_location`
- remove `_vertical_location`
- remove `_spatialhash`
- remove `_gtype`
- remove `_lonlat_minmax`
- add `gridadapter`
  - provides the Field with access to the grid adapter where its needed. eventually we'll get rid of this class entirely
- update `GridAdapter` to use composition
- add `_core.utils` subpackage to put helper functions etc
- update Field init to use overloads

<details><summary>For context, here is the discussion I had with @fluidnumerics-joe </summary>
<p>

> ### Needed state on Field
> 
> Field:
> - `name` -> needed
> - `data` -> needed
> - `grid` -> needed
> - `_interp_method` -> needed
> - `igrid` -> needed
> - `_parent_mesh` -> Remove. gives indication of the model its coming from. Unstructured specific? Can be determined from the attrs dict.
> - `_location` -> Tells us if face, node, edge registered (uxarray). Remove in favor of `data.attrs['location']`
> - `_vertical_location` -> remove in favor of `data.attrs....`
> - `_mesh_type` -> can be removed down the line
> - `units` -> Keep for now. Can likely be removed down the line. Nick thinks that this can just be removed from the codebase to a large extent where `_mesh_type` is picked up from attrs (based on units of `lon`/`lat`), and units for the variables are picked up. If units are degrees, a dask operation can be done to convert (Nick to investigate where unit converters are usd in the codebase).
> - `allow_time_extrapolation` -> needed? Let's remove entirely? [Joe posted issue](https://github.com/OceanParcels/Parcels/issues/1988)
>     - if its a single timeslice, then we can extrapolate the time
>     - otherwise, just use the provided time domain. Users (if they want extrapolation, or periodic looping of their simulation) can use xarray methods to create their dataset of interest (`xr.concat` etc.). Q:*Does `xr.concat` actually increase the size of the dataset if you concat multiple of the same dataset? Or does it just create pointers to the same location?*
>     -  
> - `_spatialhash` -> get rid of (opt for`data.uxgrid.spatial_hash is None` etc.... and calculate on the fly)
> - `_gtype` -> via GridAdapter (if need be)
> - `_lonlat_minmax` -> via GridAdapter (if need be)

</p>
</details> 

- [x] Chose the correct base branch (`field-testing` which will go into `v4-dev`)
- [x] Fixes None
- [x] Added tests

### DIFF
--- a/parcels/_core/utils/unstructured.py
+++ b/parcels/_core/utils/unstructured.py
@@ -1,0 +1,28 @@
+from collections.abc import Hashable
+
+DIM_TO_VERTICAL_LOCATION_MAP = {
+    "nz1": "center",
+    "nz": "face",
+}
+
+
+def get_vertical_location_from_dims(dims: tuple[Hashable, ...]):
+    """
+    Determine the vertical location of the field based on the uxarray.UxDataArray object variables.
+
+    Only used for unstructured grids.
+    """
+    vertical_dims_in_data = set(dims) & set(DIM_TO_VERTICAL_LOCATION_MAP.keys())
+
+    if len(vertical_dims_in_data) != 1:
+        raise ValueError(
+            f"Expected exactly one vertical dimension ({set(DIM_TO_VERTICAL_LOCATION_MAP.keys())}) in the data, got {vertical_dims_in_data}"
+        )
+
+    return DIM_TO_VERTICAL_LOCATION_MAP[vertical_dims_in_data.pop()]
+
+
+def get_vertical_dim_name_from_location(location: str):
+    """Determine the vertical location of the field based on the uxarray.UxGrid object variables."""
+    location_to_dim_map = {v: k for k, v in DIM_TO_VERTICAL_LOCATION_MAP.items()}
+    return location_to_dim_map[location]
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -29,6 +29,7 @@ from parcels.tools.statuscodes import (
     _raise_field_out_of_bound_error,
 )
 from parcels.v4.grid import Grid
+from parcels.v4.gridadapter import GridAdapter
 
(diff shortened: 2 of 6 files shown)


parcels/_core/utils/unstructured.py
VeckoTheGecko, line 1: Perhaps a bit over-engineered for what is essentially a few if statements... Thoughts @fluidnumerics-joe ? Is this useful?

VeckoTheGecko (commented):
One thing I'm a bit concerned about is managing state that is specific to the (un)structured Fields and their implementations within Parcels. It would be good to keep `Field` "grid agnostic", meaning grid specific logic is managed within the grid object[^1]. The problem with this is that, for unstructured grids, we can't just add data to `uxarray.Grid` since we don't have control over that class, so we might need to go for some other mechanism[^2].

@fluidnumerics-joe is there any state that you think you'll need to save for unstructured grids in Parcels (i.e., that doesn't make sense to put in `uxarray` upstream?). I haven't changed the `n_face` property of `Field` yet - would be great to know your thoughts on that.

[^1]: This helps with abstraction: making the code easier to read and no need for a lot of "if" statements to define variables in one case for unstructured or another for unstructured 
[^2]: This could either be caching (e.g., `lrucache`) for items not specific to a field object, or by adding an object `unstructured_state` to the `Field` class [^3]
[^3]: I love footnotes

parcels/_core/utils/unstructured.py
fluidnumerics-joe: Not over-engineered at all. It can only be so simple.

fluidnumerics-joe (approved):

PR #1994, opened 2025-04-29
List of fields fixes

fluidnumerics-joe:

This PR  cleans up init for list of fields and fix v4 test errors. It brings in changes to `add_field` and `add_vector_field` to avoid using the `setattr` method. Instead, we opt to extend the dictionary for the fields (which map `{field.name : field}`). The `add_vector_field` method optionally adds the `VectorField` object (if it's not already present) and each of its components by calling `add_field`.

`tests/v4` is now updated to use the new list of fields. 

@VeckoTheGecko - there are still a few `TODO: Nick` items listed in here to keep things safe and tidy.

### DIFF
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -117,10 +117,7 @@ class FieldSet:
         if name in self.fields:
             raise ValueError(f"FieldSet already has a Field with name '{name}'")
 
-        if hasattr(self, name):  # check if Field with same name already exists when adding new Field
-            raise RuntimeError(f"FieldSet already has a Field with name '{name}'")
-        else:
-            self.fields[name] = field
+        self.fields[name] = field
 
     def add_constant_field(self, name: str, value, mesh: Mesh = "flat"):
         """Wrapper function to add a Field that is constant in space,
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -1,7 +1,6 @@
 import functools
 
 import numpy as np
-import uxarray as ux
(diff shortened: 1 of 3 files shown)


VeckoTheGecko (commented):
@fluidnumerics-joe feel free to re-request a review once the changes we discussed are pushed :)

fluidnumerics-joe:
@VeckoTheGecko - I've opted to skip the tests involving `ParticleSet`. There's quite a few changes that need to come in for `ParticleSet.__init__` and `ParticleSet.execute` related to `dimrange`, `time_interval`, and how `ei` is initialized to get this to work. It's a fair amount of work that deserves it's own PR at this point.

fluidnumerics-joe:
@VeckoTheGecko - Ready to review again

parcels/fieldset.py
```diff
@@ -140,8 +120,7 @@ def add_field(self, field: Field, name: str | None = None):
         if hasattr(self, name):  # check if Field with same name already exists when adding new Field
             raise RuntimeError(f"FieldSet already has a Field with name '{name}'")
         else:
-            setattr(self, name, field)
```
VeckoTheGecko, line 143: This `if hasattr...` block can be removed (duplication of check above), and the else statement can be de-indented

VeckoTheGecko (changes requested):
Small change, but once that's fixed LGTM!

VeckoTheGecko (approved):

PR #1996, opened 2025-04-30
Remove global_statics and unused fixture

VeckoTheGecko:
Global_statics was mainly used in versions of parcels with c compilation

- [x] Fixes None


erikvansebille (approved):
Good catch; these could indeed have been deleted when I removed JIT support

PR #1998, opened 2025-05-02
Add tests for new FieldSet object

VeckoTheGecko:
- [x] Fixes none
- [x] Added tests

Changes:
- testing:
  - Add U and V fields to example datasets
  - Add fieldset tests (and stubs for future tests)
- cleanup
  - Remove reference to `_parent_mesh`
  - Split `reprs` into a separate file
- code
  - Add check for field type in `FieldSet.add_field`
  - Add `fieldset.constants` (avoids call to `setattr`)
  - Add `FieldSet.__getattr__`
  - Fix `FieldSet.add_constant_field`

### DIFF
--- a/parcels/_reprs.py
+++ b/parcels/_reprs.py
@@ -1,0 +1,84 @@
+"""Parcels reprs"""
+
+from __future__ import annotations
+
+import textwrap
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from parcels import Field, FieldSet, ParticleSet
+
+
+def field_repr(field: Field) -> str:
+    """Return a pretty repr for Field"""
+    out = f"""<{type(field).__name__}>
+    name            : {field.name!r}
+    data            : {field.data!r}
+    extrapolate time: {field.allow_time_extrapolation!r}
+"""
+    return textwrap.dedent(out).strip()
+
+
+def _format_list_items_multiline(items: list[str], level: int = 1) -> str:
+    """Given a list of strings, formats them across multiple lines.
+
+    Uses indentation levels of 4 spaces provided by ``level``.
+
+    Example
+    -------
+    >>> output = _format_list_items_multiline(["item1", "item2", "item3"], 4)
+    >>> f"my_items: {output}"
+    my_items: [
+        item1,
+        item2,
+        item3,
+    ]
+    """
+    if len(items) == 0:
+        return "[]"
+
+    assert level >= 1, "Indentation level >=1 supported"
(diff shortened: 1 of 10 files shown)


parcels/_reprs.py
VeckoTheGecko, line 1: Just split this out from `parcels.tools._helpers` so that its easier to find

fluidnumerics-joe:
Taking a look right after the elphe meeting this morning

parcels/fieldset.py
```diff
@@ -41,10 +42,21 @@ class FieldSet:
     """
 
     def __init__(self, fields: list[Field | VectorField]):
-        # TODO Nick : Enforce fields to be list of Field or VectorField objects
+        for field in fields:
+            if not isinstance(field, (Field, VectorField)):
+                raise ValueError(f"Expected `field` to be a Field or VectorField object. Got {field}")
```
fluidnumerics-joe, line 47: Would we want this to say `Got {type(field)}`, consistent with what is used down in  line 128 ?

parcels/fieldset.py
```diff
@@ -137,21 +149,25 @@ def add_constant_field(self, name: str, value, mesh: Mesh = "flat"):
...
+            coords={
+                "ZG": (["ZG"], np.arange(1), {"axis": "Z"}),
+                "YG": (["YG"], np.arange(1), {"axis": "Y"}),
+                "XG": (["XG"], np.arange(1), {"axis": "X"}),
+                "lon": (["XG"], np.arange(1), {"axis": "X"}),
+                "lat": (["YG"], np.arange(1), {"axis": "Y"}),
+                "depth": (["ZG"], np.arange(1), {"axis": "Z"}),
+            },
         )
+        grid = Grid(da)
```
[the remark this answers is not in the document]
fluidnumerics-joe: Agreed. Staying consistent with the idea that a field is associated with a grid is good

fluidnumerics-joe (changes requested):
Just a couple minor changes

parcels/fieldset.py
VeckoTheGecko: I don't think so. Having `Got {field}` is nice since that would surface the type and the name of the field (which is important for the for loop)

VeckoTheGecko:
> Just a couple minor changes

I've responded now :)

> Any more tests we should have in here @fluidnumerics-joe ? (even if its just stubs `def test_my_thing():...` so that we don't forget them later)

Let me know if anything comes to mind.

Good to merge?

PR #1999, opened 2025-05-06
Remove allow_time_extrapolation attribute on Field

### LINKED ISSUE #1988: [v4 Discussion] Remove allow_time_extrapolation

fluidnumerics-joe:
When a xarray or uxarray dataarray does not have a time dimension, one is added to the Field.data dataarray. When a field.data datarray has a time dimension that is length 1, we assume that extrapolation is allowed meaning that we just keep using the same single time-slice for that field.

- [ X ] Fixes #1988

### DIFF
--- a/parcels/_index_search.py
+++ b/parcels/_index_search.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
     # from .grid import Grid
 
 
-def _search_time_index(field: Field, time: datetime, allow_time_extrapolation=True):
+def _search_time_index(field: Field, time: datetime):
     """Find and return the index and relative coordinate in the time array associated with a given time.
 
     Parameters
@@ -37,8 +37,9 @@ def _search_time_index(field: Field, time: datetime, allow_time_extrapolation=Tr
     Note that we normalize to either the first or the last index
     if the sampled value is outside the time value range.
     """
-    if not allow_time_extrapolation and (time < field.data.time[0] or time > field.data.time[-1]):
+    if (len(field.data.time)) > 1 and (time < field.data.time[0] or time > field.data.time[-1]):
         _raise_time_extrapolation_error(time, field=None)
+
     time_index = field.data.time <= time
 
     if time_index.all():
--- a/parcels/_index_search.py
+++ b/parcels/_index_search.py
@@ -37,7 +37,7 @@ def _search_time_index(field: Field, time: datetime):
     Note that we normalize to either the first or the last index
     if the sampled value is outside the time value range.
     """
-    if (len(field.data.time)) > 1 and (time < field.data.time[0] or time > field.data.time[-1]):
+    if field.time_interval is not None and time in field.time_interval:
         _raise_time_extrapolation_error(time, field=None)
 
     time_index = field.data.time <= time
--- a/parcels/_index_search.py
+++ b/parcels/_index_search.py
@@ -37,7 +37,10 @@ def _search_time_index(field: Field, time: datetime):
     Note that we normalize to either the first or the last index
     if the sampled value is outside the time value range.
     """
-    if field.time_interval is not None and time in field.time_interval:
+    if field.time_interval is None:
+        return 0
+
+    if time in field.time_interval:
         _raise_time_extrapolation_error(time, field=None)
 
     time_index = field.data.time <= time
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -149,7 +149,6 @@ class Field:
         grid: ux.Grid | Grid,
         mesh_type: Mesh = "flat",
         interp_method: Callable | None = None,
-        allow_time_extrapolation: bool | None = None,
     ):
         if not isinstance(data, (ux.UxDataArray, xr.DataArray)):
             raise ValueError(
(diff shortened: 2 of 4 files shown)


parcels/field.py
```diff
@@ -204,10 +203,12 @@ def __init__(
...
-        if allow_time_extrapolation is None:
-            self.allow_time_extrapolation = True if len(getattr(self.data, "time", [])) == 1 else False
-        else:
-            self.allow_time_extrapolation = allow_time_extrapolation
+        # Check if time is in self.data.dims
+        if "time" not in self.data.dims:
+            # Add time dimension of length 1 if not present
+            # While the choice of actual date is arbitrary, it should be a datetime object
+            # Here, we use the current time
+            self.data = self.data.expand_dims({"time": [datetime.now()]})
```
erikvansebille, line 211: Is this not a bit dangerous? Because it means that the result of this operation is different every time a user runs it. That can become a debugging headache. Better to set a neutral value? E.g. 1 Jan 2000? Or (as an inside-joke) 27-Sept-2015, the day of the first Parcels commit?

parcels/particleset.py
```diff
@@ -1130,13 +1130,13 @@ def _warn_particle_times_outside_fieldset_time_bounds(release_times: np.ndarray,
     if np.any(release_times):
         if np.any(release_times < time[0]):
             warnings.warn(
-                "Some particles are set to be released before the fieldset's first time and allow_time_extrapolation is set to False.",
+                "Some particles are set to be released before the fieldset's first time and the fields are not constant in time.",
```
erikvansebille, line 1133: Is this still a correct warning? Should it not also check for the length of `time`?

parcels/_index_search.py
```diff
@@ -37,8 +37,9 @@ def _search_time_index(field: Field, time: datetime, allow_time_extrapolation=Tr
     Note that we normalize to either the first or the last index
     if the sampled value is outside the time value range.
     """
-    if not allow_time_extrapolation and (time < field.data.time[0] or time > field.data.time[-1]):
+    if (len(field.data.time)) > 1 and (time < field.data.time[0] or time > field.data.time[-1]):
         _raise_time_extrapolation_error(time, field=None)
+
```
VeckoTheGecko, line 42: I think it would be easier if we do something like `if Field.time_interval and time in Field.time_interval:`. I just put up #2000 which does the groundwork for this

parcels/particleset.py
VeckoTheGecko:
```suggestion
                "Some particles are set to be released outside the FieldSet's executable time domain.",
```

parcels/particleset.py
```diff
@@ -161,7 +161,7 @@ def ArrayClass_init(self, *args, **kwargs):
             raise NotImplementedError("If fieldset.time_origin is not a date, time of a particle must be a double")
         time = np.array([self.time_origin.reltime(t) if _convert_to_reltime(t) else t for t in time])
         assert lon.size == time.size, "time and positions (lon, lat, depth) do not have the same lengths."
-        if isinstance(fieldset.U, Field) and (not fieldset.U.allow_time_extrapolation):
+        if isinstance(fieldset.U, Field) and (len(fieldset.U.data.time) > 1):
```
VeckoTheGecko, line 164: This would just be:

```suggestion
        if fieldset.time_interval:
```

And then we can update `_warn_particle_times_outside_fieldset_time_bounds` to take `release_times: np.array, time_interval: TimeInterval`

VeckoTheGecko (changes requested):

VeckoTheGecko:
I think it would be easier to rely on the `time_interval` tooling that was introduced earlier (and is also suggested now in #2000 for `Field`). This would mean a constant field can be detected by `Field.time_interval is None` (which I think is nicer than expanding the dims and pontentially triggering computation)/

parcels/_index_search.py
fluidnumerics-joe: Made the change in https://github.com/OceanParcels/Parcels/pull/1999/commits/14ef097f8c45a7d50fbdba25b3b7d61806e0288b

parcels/particleset.py
fluidnumerics-joe: Changes made in https://github.com/OceanParcels/Parcels/pull/1999/commits/2afafa0740efdc42e19ac66575b898f1a60ac721

parcels/field.py
fluidnumerics-joe: Change made in https://github.com/OceanParcels/Parcels/pull/1999/commits/80eba8860e7ce7d76918068c92cde64121308ce9

parcels/field.py
```diff
@@ -209,10 +208,9 @@ def __init__(
         else:
             raise ValueError("Unsupported mesh type in data array attributes. Choose either: 'spherical' or 'flat'")
 
-        if allow_time_extrapolation is None:
-            self.allow_time_extrapolation = True if len(getattr(self.data, "time", [])) == 1 else False
-        else:
-            self.allow_time_extrapolation = allow_time_extrapolation
+        # Check if time is in self.data.dims
```
VeckoTheGecko, line 211: comment isn't needed I don't think
```suggestion
```

VeckoTheGecko (approved):
Looks good! Minor feedback

PR #2002, opened 2025-05-07
Assert that Fields in FieldSet have the same calendar for their time domain

VeckoTheGecko:
- [x] Fixes None (related to #1988)
- [x] Added tests
- [x] Added documentation

For a given FieldSet, all Fields need to define time in the same way. It doesn't make sense to run a simulation where one Field is defined using a 365 day calendar while another is defined on a 360 day calendar as Parcels wouldn't know what to do with the mismatched days.

This PR:
- adds to the FieldSet initialiser and `add_field` method assertions that the fields are on the same calendar (and also adds tests for this)
- adds `VectorField.time_interval` (which should have need done in #2000)

It is the users responsibility that they provide xarray datasets that have the same calendars. If their data doesn't have the same calendar, they can use xarray to modify them in place before providing them to Parcels.

### DIFF
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -166,7 +166,13 @@ class Field:
         self.name = name
         self.data = data
         self.grid = grid
-        self.time_interval = get_time_interval(data)
+        try:
+            self.time_interval = get_time_interval(data)
+        except ValueError as e:
+            e.add_note(
+                f"Error getting time interval for field {name!r}. Are you sure that the time dimension on the xarray dataset is stored as datetime or cftime datetime objects?"
+            )
+            raise e
 
         # For compatibility with parts of the codebase that rely on v3 definition of Grid.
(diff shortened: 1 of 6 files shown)


parcels/field.py
```diff
@@ -166,7 +166,13 @@ def __init__(
...
         self.data = data
         self.grid = grid
-        self.time_interval = get_time_interval(data)
+        try:
+            self.time_interval = get_time_interval(data)
+        except ValueError as e:
+            e.add_note(
+                f"Error getting time interval for field {name!r}. Are you sure that the time dimension on the xarray dataset is stored as datetime or cftime datetime objects?"
+            )
+            raise e
```
VeckoTheGecko, line 175: nicer error message ;)

erikvansebille (approved):

PR #2028, opened 2025-06-04
Delete v3 tests that are handled by xarray

VeckoTheGecko:
Now that we rely on Xarray, quite some tests are no longer relevant. This PR goes through these tests examining if there are parts worth keeping, and deleting the rest.

- [x] Chose the correct base branch (`v4-dev`<-`gridadapter-bfix`<-`HEAD`)


VeckoTheGecko:
@fluidnumerics-joe feel free to poke through the v3 tests and flag anything else. Will polish this PR tmrw

fluidnumerics-joe (approved):

PR #2032, opened 2025-06-09
2031 uxgrid vertical search

### LINKED ISSUE #2031: Vertical Search for unstructured Grids
An interesting, subtle issue with the UxArray.uxgrid has turned up and impacts our move to pushing `search` underneath the `XGrid` and `UXGrid` classes. The UxArray.uxgrid class does not contain information about the vertical grid at all; instead it holds information about the lateral unstructured grid only.

```
tests/v4/test_uxarray_fieldset.py::test_fesom2_square_delaunay_uniform_z_coordinate_eval <uxarray.Grid>
Original Grid Type: regional_delaunay
Grid Dimensions:
  * n_node: 400
  * n_face: 721
  * n_max_face_nodes: 3
Grid Coordinates (Spherical):
  * node_lon: (400,)
  * node_lat: (400,)
Grid Coordinates (Cartesian):
Grid Connectivity Variables:
  * face_node_connectivity: (721, 3)
Grid Descriptor Variables:
```

The vertical grid is housed in the DataArray as a dimension and coordinate. There are typically two options for the vertical grid dimension that are relevant, depending on whether the Field is centered on vertical layers or placed at vertical layer faces. 

Within our `UxGrid` object, I'd propose that we require a vertical dimension to be passed in during construction. We should assume that the vertical dimension is the vertical layer faces (in uxarray this is the `nz` dimension); the layer centers can be easily calculated on the fly. This would enable vertical searching

fluidnumerics-joe:
- [x] Fixes #2031
- [x] Added tests

This PR introduces changes that enable vertical search for unstructured Grids. This currently assumes that z-layer coordinates are used with no partial or shaved cells.

For the vertical interpolation, we now test interpolation of fields that are uniform in space and time and those that vary linearly with depth.

Additionally, fixes are introduced in the `application_kernels/interpolation.py` that resolve previous test failures for `test_fesom2_square_delaunay_uniform_z_coordinate_eval`


parcels/uxgrid.py
```diff
@@ -15,34 +15,56 @@ class UxGrid(BaseGrid):
...
+    def __init__(self, grid: ux.grid.Grid, z: np.ndarray) -> UxGrid:
+        """
+        Initializes the UxGrid with a uxarray grid and vertical coordinate array.
+
+        Parameters
+        ----------
+        grid : ux.grid.Grid
+            The uxarray grid object containing the unstructured grid data.
+        z : np.ndarray
+            A 1D array of vertical coordinates (depths) corresponding to the vertical position of layer faces of the grid
```
VeckoTheGecko, line 27: Perhaps this should be updated to an `xr.DataArray` object so that labels are maintained (motivated by https://github.com/OceanParcels/Parcels/issues/2031#issuecomment-2958323090)

VeckoTheGecko: Also, depending on the result of the comment we should update this `__init__`

e.g., "Yes, uxarray allows nz to be spatially and temporally varying. For now though, let's only consider the case where its non spatially/temporally varying, and leave that open for future development."

In that case it would be good to update the `init` to ensure the time and face ID dims aren't in `nz` so that if people try to provide spatially varying nz they get a `NotImplementedError`.

VeckoTheGecko (changes requested):

parcels/uxgrid.py
fluidnumerics-joe: Yea, I think a `DataArray` would be better. I think it'd be ok to initially support only fixed in space and time vertical coordinate. This will work for flat bottomed topography and should be sufficient for demonstration. After getting through baroclinic gyre examples, it'd be good to work with the FESOM2 folks to add topography to that simulation to test out `nz` as a function of `faceID` (and get it implemented of course).

I'll get this done this morning

VeckoTheGecko (approved):

fluidnumerics-joe:
On the unit test that is failing. when I run the full suite of `v4` tests, it fails. When I call it directly, it does not fail. I'm looking into this

fluidnumerics-joe:
> On the unit test that is failing. when I run the full suite of `v4` tests, it fails. When I call it directly, it does not fail. I'm looking into this

In the new test I incorporated, updating the data set to modify values changes the expected values for tests called after it. Looking into a workaround here

VeckoTheGecko:
Oh! Looks like we might have some test pollution https://www.youtube.com/watch?v=w5O4zTusyJ0

fluidnumerics-joe:
Indeed this was test pollution. For one of the tests, I was overwriting the values stored by the pressure and vertical velocity; tests that were run later on that also used the same dataset ended up failing assertions due to the change in the underlying data. 

Solution here was to do a deep copy of the dataset for the test where pressure and vertical velocity are modified, rather than using the dataset provided by our `_datasets/unstructured/generic.py` directly. All should be good now

PR #2040, opened 2025-06-18
Patch `time_interval` for all constant fields

VeckoTheGecko:
The case with all constant fields was not considered before.
- [x] Fixes None
- [x] Tests added


VeckoTheGecko:
Added a test and patched this @fluidnumerics-joe - ready for your branch to be re-synced with `v4-dev`!

Issue #2034, opened 2025-06-11
v4 ParticleSet feature discussion


fluidnumericsJoe:
The `ParticleSet.execute` method advances all particles in the `ParticleSet` through time by repeatedly applying a given kernel function (e.g., `AdvectionRK4`) until a specified endtime or runtime is reached. It also handles:

* Releasing new particles (repeat)
* Writing particle output
* Running interaction kernels (if provided)
* Invoking custom kernels
* Advancing the fieldset time index appropriately

Currently, the time step size `dt` can be `float`, `timedelta`, or `np.timedelta64` and can be positive (forward in time) or negative (backward in time)

The duration of the execution in "model-time" is set either by `endtime` or  `runtime` (both cannot be provided at the same time). The `endtime` is the model time while `runtime` is the amount of time in the model that passes relative to the time at the start of `ParticleSet.execute`

File output frequency is controlled by `outputdt`. While there is a standard file output for the particle set, users can optionally write Field data to file.

Particle-particle interactions can be handled by setting the `pyfunc_inter` argument.

While I suspect we will want all of these features in v4 of Parcels, I want to get a discussion started around what features we want to *start* with so that we can incrementally incorporate some of the changes we've made for time management and evaluation in fields and fieldsets.

I would propose that we initially focus on (pre v4alpha)
* shifting all time data to be `timedelta` or `datetime` objects
* invoking existing advection and time stepping kernels
* managing file IO for particle sets

This would cover a good deal of use cases and would allow us to quickly get some demonstrations together for using the new `UxGrid` and `XGrid` classes with `Field` and `FieldSet`. 

Is the list of features above complete ? Are there other items we want to consider for `v4alpha` ?

fluidnumerics-joe:
I think this all makes sense. I'll start on a first cut with

* shifting all time data to be timedelta or datetime objects
* invoking arbitrary kernels using existing kernels.py framework (with tests focusing on various combinations of time integrators and interpolation kernels)
* no particle interaction kernels (will open an issue for this when a PR for this issue is created)
* managing file IO for particle sets
* using only `endtime` (type matches the `time` dimension type from the `FieldSet`)

VeckoTheGecko:
> * no particle interaction kernels (will open an issue for this when a PR for this issue is created)

[...]

Issue #2045, opened 2025-06-20
Field sampled outside time domain


miniufo:
I have a flow `fieldset`:

```python
<xarray.Dataset> Size: 403GB
Dimensions:  (x: 8192, y: 8192, time: 501)
Coordinates:
  * x        (x) float32 33kB -3.142 -3.141 -3.14 -3.139 ... 3.139 3.14 3.141
  * y        (y) float32 33kB -3.142 -3.141 -3.14 -3.139 ... 3.139 3.14 3.141
  * time     (time) float32 2kB 0.0 0.001 0.002 0.003 ... 0.497 0.498 0.499 0.5
Data variables:
    u        (time, y, x) float32 134GB ...
    v        (time, y, x) float32 134GB ...
    vor      (time, y, x) float32 134GB ...
```

and `ParticleFile`:
```python
pfile = pset.ParticleFile(name='/Data/2DTurbulence/8192/testOffline', outputdt=0.001)
pset.execute(
    parcels.AdvectionRK4,
    endtime=0.5,
    dt=0.0001,
    output_file=pfile,
)
```

This raise the error:

```python
INFO: Output files are stored in /Data/2DTurbulence/8192/testOffline.zarr.
  0%|▍                                                                                           | 0.0021000000000000003/0.5 [01:12<4:11:59, 30366.34s/it]
---------------------------------------------------------------------------
TimeExtrapolationError                    Traceback (most recent call last)
Cell In[5], line 1
----> 1 pset.execute(
      2     parcels.AdvectionRK4,
      3     endtime=0.5,
      4     dt=0.0001,
      5     output_file=pfile,
      6 )

File [~/miniconda3/lib/python3.13/site-packages/parcels/particleset.py:1180](http://192.168.1.105:8888/home/qianyk/miniconda3/lib/python3.13/site-packages/parcels/particleset.py#line=1179), in ParticleSet.execute(self, pyfunc, pyfunc_inter, endtime, runtime, dt, output_file, verbose_progress, postIterationCallbacks, callbackdt, delete_cfiles)
   1178 if self._interaction_kernel is None:
   1179     if not skip_kernel:
-> 1180         res = self._kernel.execute(self, endtime=next_time, dt=dt)
   1181         if res == StatusCode.StopAllExecution:
   1182             return StatusCode.StopAllExecution

[...]

PR #2060, opened 2025-06-30
Minimal v4 particleset

fluidnumericsJoe:
- [x] Added tests

This pull request provides a minimal implementation of the `ParticleSet` for execution as discussed in #2034 . 

## Highlight of changes
* `repeatdt` logic is removed
* Dynamic instantiation of the `_pclass` type is removed for now; this currently means that `particle.ei` (encoded indices) are not stored and available for reuse. There's some discussion to be had around how this is incorporated cleanly and this will ultimately be brought in through a future PR.
* Particle-Particle interaction is removed in this version. My understanding that that interacting particles will be brought in once the issues around non-interacting particles settle down.
* Time handling is done through datetime and timedelta objects
* A new test is added for `pset.execute` with the unstructured stommel gyre generic dataset. 
* Notebook tutorial for using Parcels with unstructured Stommel gyre is updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,21 +18,16 @@ defaults:
 
 jobs:
   unit-test:
-    name: "py${{ matrix.python-version }} | ${{ matrix.os }} | unit tests (${{ matrix.test-subfolder }})"
+    name: "py${{ matrix.python-version }} | ${{ matrix.os }} | unit tests"
     runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu] #, mac, windows] # TODO v4: Re-enable windows and mac
-        test-subfolder: ["/"]
         python-version: ["3.12"]
         include:
           - os: ubuntu
             python-version: "3.11"
-            test-subfolder: "/"
-          - os: ubuntu
-            python-version: "3.12"
-            test-subfolder: "/v4"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -43,7 +38,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Unit test
         run: |
-          coverage run -m pytest -v -s --html=${{ matrix.os }}_${{ matrix.python-version }}_unit_test_report.html --self-contained-html tests${{ matrix.test-subfolder }}
+          coverage run -m pytest -v -s --html=${{ matrix.os }}_${{ matrix.python-version }}_unit_test_report.html --self-contained-html
           coverage xml
       - name: Codecov
         uses: codecov/codecov-action@v5.3.1
@@ -58,6 +53,8 @@ jobs:
           name: Unittest report ${{ matrix.os }}-${{ matrix.python-version }}
           path: ${{ matrix.os }}_${{ matrix.python-version }}_unit_test_report.html
   integration-test:
+    # TODO v4: Re-enable the workflow once development has stabilized and we want to run integration tests again
+    if: false
     name: "py${{ matrix.python-version }} | ${{ matrix.os }} | integration tests"
     runs-on: ${{ matrix.os }}-latest
     strategy:
@@ -111,6 +108,8 @@ jobs:
           pattern: "* report *"
   typechecking:
     name: mypy
+    # TODO v4: Enable typechecking again
+    if: false
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/parcels/_core/utils/time.py
+++ b/parcels/_core/utils/time.py
@@ -6,10 +6,10 @@ from typing import TYPE_CHECKING, TypeVar
 import cftime
 import numpy as np
 
-T = TypeVar("T", datetime, cftime.datetime)
-
 if TYPE_CHECKING:
-    from parcels._typing import DatetimeLike
+    from parcels._typing import TimeLike
+
+T = TypeVar("T", bound="TimeLike")
 
 
 class TimeInterval:
@@ -92,7 +92,7 @@ def is_compatible(
 
 
 def get_datetime_type_calendar(
-    example_datetime: DatetimeLike,
+    example_datetime: TimeLike,
 ) -> tuple[type, str | None]:
     """Get the type and calendar of a datetime object.
 
--- a/parcels/_core/utils/unstructured.py
+++ b/parcels/_core/utils/unstructured.py
@@ -1,12 +1,10 @@
-from collections.abc import Hashable
-
 DIM_TO_VERTICAL_LOCATION_MAP = {
     "nz1": "center",
     "nz": "face",
 }
 
 
-def get_vertical_location_from_dims(dims: tuple[Hashable, ...]):
+def get_vertical_location_from_dims(dims: tuple[str, ...]):
     """
     Determine the vertical location of the field based on the uxarray.UxDataArray object variables.
 
--- a/parcels/_index_search.py
+++ b/parcels/_index_search.py
@@ -356,7 +356,7 @@ def _search_indices_curvilinear_2d(
 
 
 ## TODO :  Still need to implement the search_indices_curvilinear
-def _search_indices_curvilinear(field: Field, time, z, y, x, ti, particle=None, search2D=False):
+def _search_indices_curvilinear(field, time, z, y, x, ti, particle=None, search2D=False):
     if particle:
         zi, yi, xi = field.unravel_index(particle.ei)
     else:
--- a/parcels/_typing.py
+++ b/parcels/_typing.py
@@ -34,7 +34,7 @@ Mesh = Literal["spherical", "flat"]  # corresponds with `mesh`
 VectorType = Literal["3D", "3DSigma", "2D"] | None  # corresponds with `vector_type`
 GridIndexingType = Literal["pop", "mom5", "mitgcm", "nemo", "croco"]  # corresponds with `gridindexingtype`
 NetcdfEngine = Literal["netcdf4", "xarray", "scipy"]
-DatetimeLike = datetime | cftime_datetime | np.datetime64
+TimeLike = datetime | cftime_datetime | np.datetime64
 
 KernelFunction = Callable[..., None]
 
--- a/parcels/fieldset.py
+++ b/parcels/fieldset.py
@@ -15,7 +15,7 @@ from parcels.field import Field, VectorField
 from parcels.xgrid import XGrid
 
 if TYPE_CHECKING:
-    from parcels._typing import DatetimeLike
+    from parcels._typing import TimeLike
     from parcels.basegrid import BaseGrid
 __all__ = ["FieldSet"]
 
@@ -57,7 +57,7 @@ class FieldSet:
         assert_compatible_calendars(fields)
 
         self.fields = {f.name: f for f in fields}
-        self.constants = {}
+        self.constants: dict[str, float] = {}
 
     def __getattr__(self, name):
         """Get the field by name. If the field is not found, check if it's a constant."""
@@ -236,7 +236,7 @@ class CalendarError(Exception):  # TODO: Move to a parcels errors module
     """Exception raised when the calendar of a field is not compatible with the rest of the Fields. The user should ensure that they only add fields to a FieldSet that have compatible CFtime calendars."""
 
 
-def assert_compatible_calendars(fields: Iterable[Field]):
+def assert_compatible_calendars(fields: Iterable[Field | VectorField]):
     time_intervals = [f.time_interval for f in fields if f.time_interval is not None]
 
     if len(time_intervals) == 0:  # All time intervals are none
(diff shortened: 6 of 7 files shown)


parcels/particleset.py
```diff
@@ -862,15 +709,11 @@ def set_variable_write_status(self, var, write_status):
 
     def execute(
         self,
+        endtime: timedelta | datetime,
+        dt: np.float64 | np.float32 | timedelta,
         pyfunc=AdvectionRK4,
-        pyfunc_inter=None,
-        endtime=None,
-        runtime: float | timedelta | np.timedelta64 | None = None,
-        dt: float | timedelta | np.timedelta64 = 1.0,
```
VeckoTheGecko, line 869: Would we be able to update this to the following?
```python
    def execute(
        self,
        pyfunc=AdvectionRK4,
        *,
        endtime: timedelta | datetime | None = None,
        runtime = None, # Add a `if runtime is not None: raise NotImplementedError` in the code below
        dt: np.float64 | np.float32 | timedelta | None = None,
        output_file=None,
        verbose_progress=True,
    ):
```

This:
1. keeps the relative order of the args somewhat
2. forces everything after `*` to be a keyword only argument
3. flags intention for near-future dev that `runtime` will need to be implemented

I think (1) is important since users in v3 would be tripped up a bit by the position of `pyfunc` otherwise

VeckoTheGecko (commented):
Did a quick first pass - can do a more in depth review tomorrow

parcels/particleset.py
```diff
@@ -89,65 +77,28 @@ def __init__(
...
 
         if isinstance(pid_orig, (type(None), bool)):
             pid_orig = np.arange(lon.size)
 
         if depth is None:
-            mindepth = self.fieldset.dimrange("depth")[0]
+            mindepth = 0
+            for field in self.fieldset.fields.values():
+                if field.grid.depth is not None:
+                    mindepth = min(mindepth, field.grid.depth[0])
```
erikvansebille, line 101: Perhaps not an issue we want to fix now, but for meaningful usecases (“release particles at the surface when no depth is provided”) this assumes that depth is positive downward. What if depth is positive upward? Perhaps make a small TODO note?
```suggestion
                    mindepth = min(mindepth, field.grid.depth[0]). # TODO also support depth-positive-upward grids
```

parcels/particleset.py
```diff
@@ -160,6 +111,7 @@ def ArrayClass_init(self, *args, **kwargs):
             time = np.array([np.datetime64(t) for t in time])
         if time.size > 0 and isinstance(time[0], np.timedelta64) and not self.time_origin:
             raise NotImplementedError("If fieldset.time_origin is not a date, time of a particle must be a double")
+
```
erikvansebille, line 114: The error message in the line above is not true anymore, right? Particle time cannot be a double?

parcels/particleset.py
```diff
@@ -464,8 +334,6 @@ def from_list(
             Optional list of initial depth values for particles. Default is 0m
         time :
             Optional list of start time values for particles. Default is fieldset.U.time[0]
-        repeatdt :
```
erikvansebille, line 467: Not sure it’s useful to also update the docstring already. We’ll have to put it back later anyways. OK to keep the change, but next time simply don’t touch public API docstrings until we’ve settled on API, especially when temporarily disabling functionality?

parcels/particleset.py
```diff
@@ -883,26 +726,17 @@ def execute(
...
-            End time for the timestepping loop.
-            It is either a datetime object or a positive double. (Default value = None)
-        runtime :
-            Length of the timestepping loop. Use instead of endtime.
-            It is either a timedelta object or a positive double. (Default value = None)
-        dt :
+        endtime (datetime.datetime or np.timedelta64): :
+            End time for the timestepping loop. If a timedelta is provided, it is interpreted as the total simulation time.
+            If a datetime is provided, it is interpreted as the end time of the simulation.
+        dt (timedelta):
```
erikvansebille, line 732: Or float?

parcels/particleset.py
```diff
@@ -919,197 +753,82 @@ def execute(
...
                 self._kernel = self.Kernel(pyfunc)
+
         if output_file:
             output_file.metadata["parcels_kernels"] = self._kernel.name
 
-        # Set up the interaction kernel(s) if not set and given.
-        if self._interaction_kernel is None and pyfunc_inter is not None:
-            if isinstance(pyfunc_inter, InteractionKernel):
-                self._interaction_kernel = pyfunc_inter
+        # The fieldset time intervale defines the extent of time that is allowed to be
```
erikvansebille, line 760:
```suggestion
        # The fieldset time interval defines the extent of time that is allowed to be
```

parcels/particleset.py
```diff
@@ -919,197 +753,82 @@ def execute(
...
+        # simulated. If `fieldset.time_interval` is not None, it will be used to determine the endtime (the min of endtime or fieldset.time_interval[1]).
+        # If `fieldset.time_interval` is None, the endtime will be determined by the
+        # `endtime` parameter or the fieldset's time dimension.
+        # Time parameters for the main for loop are converted to floats, since the interpolation kernels expect float objects for time
```
erikvansebille, line 764: Ah, this is an interesting inning. Should we change that too, in the long run? Keeping timedelta64 throughout may avoid round-off issues like in #2045

parcels/particleset.py
```diff
@@ -919,197 +753,82 @@ def execute(
...
             # Check if we can fast-forward to the next time needed for the particles
-            if dt > 0:
-                skip_kernel = True if min(self.time) > (time + dt) else False
-            else:
-                skip_kernel = True if max(self.time) < (time + dt) else False
-
-            time_at_startofloop = time
-
-            next_input = self.fieldset.computeTimeChunk(time, dt)
+            # if dt > 0:
```
erikvansebille, line 811: Is this a debug comment, or intentionally left in?

parcels/particleset.py
VeckoTheGecko: > What if depth is positive upward?

In both instances is the 0 point the ocean surface? In which case, I think it would be beneficial of us to just choose one and assert in the grid initialiser (that way we can make an assumption one way or the other in the rest of the code - and also it's easy to make this change before initialisation using xarray (either manually by users, or automatically by our classmethods that would flow into here).

tests/v4/test_particleset.py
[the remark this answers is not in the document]
fluidnumerics-joe: AFAIK, this should work. I'll add a test

fluidnumerics-joe:
Here's what I've gathered from the comments above
* The `endtime` input must match the same type as the `time` dimension in the fields of the underlying `fieldset`. From what I can tell, the most straightforward way to get the type of the `time` dimension is to get the type of one of the `time_interval` endpoints. 
* We need to bring back `runtime`, which is required to be a `timedelta` object. 
* We need to enforce either `endtime` or `runtime` being set. I would argue that if there is a valid time_interval, we could default to using the valid time_timerval
* All time handling should remain as `datetime` or `timedelta` objects; in either case, it should be consistent with the `time` dimension type coming from the underlying dataset.


Some open questions
* What are the data types for the time dimension for idealized simulations that don't leverage a calendar system ? Do they actually come in as `timedelta` or something else? I guess I'll find out later today with this MITgcm example I'll send across in #2053

parcels/particleset.py
fluidnumerics-joe: The time handling in the pset.execute function has been fully converted to datetime or timedelta objects and I've brought back in the `runtime` argument, per the other discussions. As a result, this comment is no longer entirely relevant

parcels/particleset.py
fluidnumerics-joe: You're right. I'll get this on my next round of edits

fluidnumerics-joe:
I believe I've addressed a good deal of the issues that were raised. To help keep this PR moving, if it looks like I've resolved your issue, please resolve the comment. Otherwise, I will assume it is not resolved and work on it in my next round of edits.

parcels/particleset.py
erikvansebille: Agree, let's resolve this conversation for now

VeckoTheGecko:
> It appears we may need to coerce `time` to be a dimension and coordinate with knowledge of the parent simulation timestep, if reading NetCDF output straight from MITgcm.

I think we should just not support int time dimensions. If a user wants to use MITgcm runs without a calendar, they have to update the time dimension to timedelta or datetime/cftime on the xarray dataset level before passing to Parcels. We can have an explainable error message denoting the problem.

parcels/particleset.py
VeckoTheGecko: Perhaps just revert the parameter list to how it was and leaving unused variables in there (syntax highlighting in vscode will show them as unused, which is fine for now).

I think at some point we need to start an issue to fully discuss how the execute loop of v4 looks different to v3 (perhaps the conclusion is v3 execute functions the same as v4 execute - but even then it would be good to see both sides). As part of that discussion, we can wrap in elements from https://github.com/OceanParcels/Parcels/pull/2060#discussion_r2175402689

VeckoTheGecko: > Perhaps just revert the parameter list to how it was and leaving unused variables in there

Actually, I think something like:

```python
if repeatdt is not None:
     raise NotImplementedError
```
at the top of pset execute would be good. That way any tests are guaranteed not to pass through those variables, and allows us time to fully scope in #2065

VeckoTheGecko:
To summarise:

- Docstrings/param list/NotImplementedError https://github.com/OceanParcels/Parcels/pull/2060#discussion_r2179569675
- runtime/endtime (maybe just mark one as NotImplemented - or you can go with an implementation if its straightforward) https://github.com/OceanParcels/Parcels/pull/2060#discussion_r2179575230
- xfail outputfile test https://github.com/OceanParcels/Parcels/pull/2060#issuecomment-3026962817

After that I think we're good to merge.

parcels/particleset.py
```diff
@@ -919,197 +753,82 @@ def execute(
...
+        # `endtime` parameter or the fieldset's time dimension.
+        # Time parameters for the main for loop are converted to floats, since the interpolation kernels expect float objects for time
+        # The initial time (in float point) representation is t0=0.0 and time is interpreted as relative to the start of the time interval
+        fieldset_timeinterval = self.fieldset.time_interval
+
+        if fieldset_timeinterval is None:
+            if isinstance(endtime, datetime):
+                raise NotImplementedError(
+                    "If fieldset.time_interval is None, endtime must be a timedelta not a datetime"
+                )
```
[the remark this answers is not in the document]
fluidnumerics-joe: Right now, what I've done is to allow for either `runtime` or `endtime` to be sent in. The `runtime` is assumed to be a `timedelta`. `endtime` data type is either `timedelta` or `datetime` but must match the time datatype from the fieldset (as deduced from the `time_interval.left` data type. There's a few cases that are handled.

* If `runtime` and `endtime` are both provided an error is thrown
* If `runtime` is provided and `endtime` is not provided, the `start_time` (local variable used for controlling the forward stepping loop) is set to `timedelta(seconds=0)` and the `end_time` (local variable used for determining the last `time` value in the forward stepping loop) is set to `runtime`
* If `runtime` is not provided and `endtime` is provided, the `endtime` is checked to match the type of the `time_interval.left`. The `start_time` is set to `time_interval.left` and the `end_time` is set to the minimum of `endtime` or `time_interval.right`
* If `runtime` is not provided, `endtime` is provided, and time_interval is `None` then an error is thrown.

The way I've implemented it surely is not the cleanest and tidiest, but it works. I'd say further cleanup should come in a following PR.

parcels/particleset.py
fluidnumerics-joe: Resolved in https://github.com/OceanParcels/Parcels/pull/2060/commits/a55c8520fd273d90ec9e86c16bfbdfbdbc264d01

parcels/particleset.py
fluidnumerics-joe: Fixed in https://github.com/OceanParcels/Parcels/pull/2060/commits/173411398775b91762126e8a6b2fc017812efd52

fluidnumerics-joe:
I have a couple issues in the v4 tests that are failing and I'll take care of resolving these in the morning.

erikvansebille (commented):
While testing my own run on this branch, I found some small bugs that I've either fixed or(see comments) or propose simple changes to below

VeckoTheGecko:
> Well, this is an odd thing to error out on : https://github.com/OceanParcels/Parcels/actions/runs/16056110569/job/45310801757?pr=2060#step:4:779

`data` looks to be a numpy array and not an xarray dataarray - hence doesn't have a `.values` attribute - could that be it?

fluidnumerics-joe:
The `field` that is passed into the interpolator `data` attribute is a uxarray.dataarray. This hasn't changed at all in the generic dataset. an it's suddenly failing?

fluidnumerics-joe:
Also, removing the `.values` fails other tests, since (again). `field.data` is a DataArray

VeckoTheGecko:
@fluidnumerics-joe the problem is here:

https://github.com/OceanParcels/Parcels/blob/d544d4c66d63287644efa83fe9fd6a16d63022d6/parcels/kernel.py#L320

This looks to have been introduced in fe264240754ee2ba0a67e9d30c9f538fb6f41d5f to fix #867 . The kernel side of the code is something that we haven't touched yet.... perhaps its time we start looking at it

fluidnumerics-joe:
@VeckoTheGecko - good find. That resolved the issues on the v4 tests. Updating this branch and will re-request review

PR #2066, opened 2025-07-02
Update UxGrid search API to match BaseGrid

### LINKED ISSUE #2048: A clear API for handling grid axes, barycentric coordinates, index searching, and ravelling.
Terminology[^1]:
- **Parcels Grid**: The logical grid representation (via XGrid or UxGrid) used by Parcels to determine particle position. In structured grids, this corresponds to the grid of F-points (see #2037). Time is not part of the grid abstraction.
- **Axis (Axes)**: The spatial dimensions of the Parcels Grid. For structured grids: X, Y, Z; for unstructured: Z and FACE. These relate closely to the underlying data layout (e.g., in the structured case, only a small correction about left/right grid indexing is required to get to the underlying data representation). Indexing a particle involves locating it along each axis.
- **Index (Indices)**: The cell position of a particle along a given axis. Combined indices define the full grid cell a particle occupies.
- **Barycentric Coordinates**: The coordinates defining a particle’s position within a grid cell. In structured grids, this is a single coordinate per axis (e.g., xsi); in unstructured, this can be a mix (e.g., a single coordinate for zeta, and n-coordinates for an n-gon that the particle resides in - 3 for triangular cells).
- **Encoded Indices (ei)**: A compact, single-integer encoding of a particle’s full grid indices.

---

Locating a particle’s cell and barycentric coordinates within a Parcels Grid is essential for interpolation. This functionality should reside in the Grid class, exposing a clear API that the Field class can use for interpolation—ensuring separation of concerns, so Field need not manage grid geometry.

In the structured case, Parcels must support grids with varying axis dimensionality:
- 3D:
  - Options: 2D lat + 2D lon + 1D depth (curvilinear), or 1D lat + 1D lon + 1D depth (rectilinear)
  - Use case: 3D U/V fields
- 2D:
  - Options: 2D lat + 2D lon, or 1D lat/lon + 1D depth
  - Use case: Wave flumes, windage
- 1D: Single 1D axis (lat/lon/depth)
- 0D: Constant value everywhere (constant field)

Data may exist on any subset of the grid axes (e.g., U may be on X, Y, Z; depth-averaged U on X, Y). During interpolation, the particle’s full position is used, but only relevant axis indices (e.g., xi, yi) are needed to access the data.

## Proposed API

```python
>>> x, y, z = ..., ..., ... # float float float
>>> grid = ...
```

### `.search(z, y, x)`

```python
>>> grid.search(z, y, x)
# returns:
# {
#   axis: (index, bcoords), ...
# }

# if grid is 3D structured
{"X": (xi, xsi), "Y": (yi, eta), "Z": (zi, zeta)}

# if grid is 2D structured (no depth)
{"X": (xi, xsi), "Y": (yi, eta)}

# if grid is 1D structured (depth)
{"Z": (zi, zeta)}

# if grid is unstructured
{"Z": (zi, zeta), "FACE": (fi, bcoords)}
# e.g., {"Z": (2, 0.5), "FACE": (200, np.array(0.2, 0.3, 0.5))}
```

### `ravel_index(axis_indices)`

```python
>>> axis_indices = {"X": xi, "Y": yi, "Z": zi} # 3d structured grid
>>> axis_indices = {"X": xi, "Y": yi} # 2d structured grid
>>> axis_indices = {"Z": zi} # 1d structured grid (depth)
>>> axis_indices = {"Z": zi, "FACE": fi} # unstructured grid
>>> grid.ravel_index(axis_indices)
# returns ei
```
<details><summary>Old API for reference</summary>
<p>

## Old approach

```python
# XGrid class
def unravel_index(self, ei):
    """
    Converts a single encoded index back into a Z, Y, and X indices.

    Parameters
    ----------
    ei : int
        Encoded index to be unraveled.

    Returns
    -------
    zi : int
        Vertical index.
    yi : int
        Y index.
    xi : int
        X index.
    """

def ravel_index(self, zi, yi, xi):
    ... # inverse of unravel_index

# UxGrid class
def unravel_index(self, ei):
    """
    Converts a single encoded index back into a vertical index and face index.

    Parameters
    ----------
    ei : int
        Encoded index to be unraveled.

    Returns
    -------
    zi : int
        Vertical index.
    fi : int
        Face index.
    """

def ravel_index(self, zi, fi):
    ... # inverse of unravel_index

# applies to both XGrid and UxGrid
def search(self, z: float, y: float, x: float, ei=None, search2D: bool = False):
    """
    Perform a spatial (and optionally vertical) search to locate the grid element
    that contains a given point (x, y, z).

    This method delegates to grid-type-specific logic (e.g., structured or unstructured)
    to determine the appropriate indices and interpolation coordinates for evaluating a field.

    Parameters
    ----------
    z : float
        Vertical coordinate of the query point. If `search2D=True`, this may be ignored.
    y : float
        Latitude or vertical index, depending on grid type and projection.
    x : float
        Longitude or horizontal index, depending on grid type and projection.
    ei : int, optional
        A previously computed encoded index (e.g., raveled face or cell index). If provided,
        the search will first attempt to validate and reuse it before falling back to
        a global or local search strategy.
    search2D : bool, default=False
        If True, perform only a 2D search (x, y), ignoring the vertical component z.

    Returns
    -------
    bcoords : np.ndarray or tuple
        Interpolation weights or barycentric coordinates within the containing cell/face.
        The interpretation of `bcoords` depends on the grid type.
    ei : int
        Encoded index of the identified grid cell or face. This value can be cached for
        future lookups to accelerate repeated searches.

    Raises
    ------
    FieldOutOfBoundError
        Raised when the queried point lies outside the bounds of the grid.
    NotImplementedError
        Raised if the search method is not implemented for the current grid type.
    """
```

</p>
</details> 

## Benefits of this approach

- Clarifies the return values of .search() and .ravel_index() based on the Grid’s Axes.
  - For 2D grids, only relevant axes are returned - e.g., no need to return zi, zeta = (0, 0). This simplifies interpolator logic and removes the need for grid dimensionality checks in the interpolator.
  - Removes the need to coerce arrays to 4D, superseding #2047.
- Unifies the Grid API, stabilizing BaseGrid behavior across subclasses (XGrid, UxGrid) and enabling future extensions.
- Clarifies the link between Axes and barycentric coordinates.
- Provides full particle position on the grid via the new API (including the zeta value).
- Improves maintainability, as all grid logic stays within the Grid class.

## Additional considerations

- Caching ei may be unnecessary for 0D or 1D grids, where lookups are already fast.
- Unclear role of search2D parameter. Is this legacy or is it still needed?
- Sigma layer support: Future users must provide F-points for depth (i.e., a 3D depth array based on X and Y via bathymetry). This would require users to broadcast the sigma ratio with their bathymetry before passing to Parcels (which I think is a good direction for us to go in when we get around to it).
- Particles still live in z,y,x space (just this might not be relevant when indexing on some grids). For surface advection we can have z=0

---

Keen to hear your thoughts (on concept, implementation or terminology) here.

[^1]: Once finalised, I think we would benefit from documenting this terminology (internally and in documentation for those interested in writing interpolators).

VeckoTheGecko:
Continuing the work in #2037 to fix #2048, this PR defines the axes for UxGrid as `["Z", "FACE"]`. See #2048 for the full context

Changes:
- Remove search2D from UxGrid.search
- Update UxGrid ravelling to match BaseGrid API
- Update Field.eval and VectorField eval to match new grid search API (also update interpolation template to remove `ei, bcoords` and replace witth `position`)

Open questions:
- In terms of naming, should we go with "FACE" or "CELL"? Is there convention here?
- Do you feel that there should be more test cases for ravelling/unravelling on an unstructured grid @fluidnumerics-joe ? (kinda what is done for structured grids at the moment?). If yes, I don't think its a high priority - though stubs would be helpful so we don't forget

- [x] Fixes #2048
- Added tests - None (yet)

### DIFF
--- a/parcels/uxgrid.py
+++ b/parcels/uxgrid.py
@@ -11,7 +11,7 @@ from parcels.xgrid import _search_1d_array
 
 from .basegrid import BaseGrid
 
-_UXGRID_AXES = Literal["Z", "CELL"]
+_UXGRID_AXES = Literal["Z", "FACE"]
 
 
 class UxGrid(BaseGrid):
@@ -83,7 +83,7 @@ class UxGrid(BaseGrid):
         if fi == -1:
             raise FieldOutOfBoundError(z, y, x)
 
-        return {"Z": (zi, zeta), "CELL": (fi, bcoords[0])}
+        return {"Z": (zi, zeta), "FACE": (fi, bcoords[0])}
 
     def _get_barycentric_coordinates(self, y, x, fi):
         """Checks if a point is inside a given face id on a UxGrid."""
@@ -103,9 +103,9 @@ class UxGrid(BaseGrid):
         return bcoord, err
 
     def ravel_index(self, axis_indices: dict[_UXGRID_AXES, int]):
-        return axis_indices["CELL"] + self.uxgrid.n_face * axis_indices["Z"]
+        return axis_indices["FACE"] + self.uxgrid.n_face * axis_indices["Z"]
 
     def unravel_index(self, ei) -> dict[_UXGRID_AXES, int]:
         zi = ei // self.uxgrid.n_face
         fi = ei % self.uxgrid.n_face
-        return {"Z": zi, "CELL": fi}
+        return {"Z": zi, "FACE": fi}
--- a/parcels/uxgrid.py
+++ b/parcels/uxgrid.py
@@ -7,6 +7,7 @@ import uxarray as ux
 from uxarray.grid.neighbors import _barycentric_coordinates
 
 from parcels.field import FieldOutOfBoundError  # Adjust import as necessary
+from parcels.xgrid import _search_1d_array
 
 from .basegrid import BaseGrid
 
@@ -62,18 +63,7 @@ class UxGrid(BaseGrid):
                 return bcoords, fid
             return None, None
 
-        def find_vertical_index() -> int:
-            nz = self.z.shape[0]
-            if nz == 1:
-                return 0
-            zf = self.z.values
-            # Return zi such that zf[zi] <= z < zf[zi+1]
-            zi = np.searchsorted(zf, z, side="right") - 1  # Search assumes that z is positive and increasing with i
-            if zi < 0 or zi >= nz - 1:
-                raise FieldOutOfBoundError(z, y, x)
-            return zi
-
-        zi = find_vertical_index()  # Find the vertical cell center nearest to z
+        zi, zeta = _search_1d_array(self.z.values, z)
 
         if ei is not None:
             _, fi = self.unravel_index(ei)
@@ -93,7 +83,7 @@ class UxGrid(BaseGrid):
         if fi == -1:
             raise FieldOutOfBoundError(z, y, x)
 
-        return bcoords[0], self.ravel_index(zi, fi[0])
+        return {"Z": (zi, zeta), "CELL": (fi, bcoords[0])}
 
     def _get_barycentric_coordinates(self, y, x, fi):
         """Checks if a point is inside a given face id on a UxGrid."""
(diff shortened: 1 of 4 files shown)


parcels/uxgrid.py
```diff
@@ -113,40 +102,10 @@ def _get_barycentric_coordinates(self, y, x, fi):
...
-        Returns
-        -------
-        int
-            Encoded index combining the face index and vertical index.
-        """
-        return fi + self.uxgrid.n_face * zi
-
-    def unravel_index(self, ei):
-        """
-        Converts a single encoded index back into a vertical index and face index.
```
VeckoTheGecko, line 136: Docstring here isn't needed - handled in the BaseGrid class

parcels/xgrid.py
```diff
@@ -15,7 +15,6 @@
 
 _XGCM_AXIS_DIRECTION = Literal["X", "Y", "Z", "T"]
 _XGCM_AXIS_POSITION = Literal["center", "left", "right", "inner", "outer"]
-_AXIS_DIRECTION = Literal["X", "Y", "Z"]
 _XGCM_AXES = Mapping[_XGCM_AXIS_DIRECTION, xgcm.Axis]
```
VeckoTheGecko, line 18: This was a duplicate definition - we already had _XGRID_AXES

fluidnumerics-joe:
> In terms of naming, should we go with "FACE" or "CELL"? Is there convention here?
For lateral indexing we should use "FACE" for face indices and "NODE" for node/vertex indices . A cell or element refers to the full 3-d object that you get when a face is extruded vertically between two vertical layers.

WRT testing ravel/unravel, we have a few tests that check evaluation of a field from our generic data sets; under the hood, this exercises both raveling and unraveling. Are you thinking of tests that just exercise ravel/unravel ?

VeckoTheGecko:
> Are you thinking of tests that just exercise ravel/unravel ?

Yes. In the xgrid case I have a test that checks that roundtrips ravel and unravel and checks that it generates unique ei's . Maybe that's overkill, I can also see that POV

PR #2068, opened 2025-07-03
Removing grid.py and time_origin for v4

erikvansebille:
As time will become a `np.datetime64`/`np.timedelta64` object, no need anymore for `time_origin` and `TimeConverter`. This PR removes them. 

It also removes the old grid.py


parcels/particlefile.py
```diff
@@ -173,9 +169,8 @@ def _create_variables_attribute_dict(self):
             "lat": {"long_name": "", "standard_name": "latitude", "units": "degrees_north", "axis": "Y"},
         }
 
-        if self.time_origin.calendar is not None:
-            attrs["time"]["units"] = "seconds since " + str(self.time_origin)
-            attrs["time"]["calendar"] = _set_calendar(self.time_origin.calendar)
+        attrs["time"]["units"] = "seconds since "  # TODO fix units
+        attrs["time"]["calendar"] = "None"  # TODO fix calendar
```
VeckoTheGecko, line 173: This manual setting of attributes won't be necessary if we make sure to store in the output time dim as the same time variable used in the execution loop

erikvansebille: Agree, but we can change that when we overhaul `particlefile.py`

VeckoTheGecko:
image
🤩

VeckoTheGecko (approved):

Issue #2078, opened 2025-07-09
Using our own timedelta class to support float*timedelta64 multiplication in Kernels?


erikvansebille:
I want to get your quick thoughts on a stumbling block I've hit with the particledata-xarray implementation (implementing #1822). 

I've made `particle.dt` a `numpy.timedelta64` object, but this means that we can't do something like 
```python
lon += u * particle.dt
```
in a Kernel (or anywhere else) because multiplying `u` (a `float`) with `particle.dt` (a `timedelta64`) results in a `TypeError`...

Three possible solutions I see

1. Don't use `np.timedelta64` for `dt`, but instead keep them as `float`. This however means that we'll probably lose all the advantages of `particle.time` being a `np.datetime64` too; because `particle.time + particle.dt` won't work anymore
2. (on-thy-fly) convert any mention of `particle.dt` in a Kernel to a `float`; but that means a codeconverter (again...), which we wanted to get rid of in v4
3. Create our own `parcels.timedelta64` class that inherits(?) from `np.timedelta64` but also adds multiplications with a `float` (and then returns a `float`). Something like
```python
class timedelta64:
    def __init__(self, value):
        self.value = np.timedelta64(value)

    def __mul__(self, other):
        # Returns float seconds
        return (self.value / np.timedelta64(1, "s")) * other

    def __rmul__(self, other):
        return self.__mul__(other)
```

What do you think, @VeckoTheGecko? Do you have a preference? Or do you see another solution?

erikvansebille:
@VeckoTheGecko and I just had a f2f discussion, and we decided to (at least temporarily) implement a fourth option:

4. Keep datetime64/timedelta64 in the pset.execute and kernel loop, but use floats for time and dt inside the kernel. That would ensure backward compatibility of v3-kernels, at the cost of users not being able to use datetime64 objects in their kernels

erikvansebille:
Another complication is that we may want to use `particle.time` in its original `datetime64`/`timedelta64` format in the Interpolators (see #2081), so that option 4 above (cusing float iniside the Kernels) may not be a good choice after all.

For the moment, I've simply converted dt to a float inside the Advection Kernels in be87206a941b5e314c2fb7dd0ea9e7ebd4ce77c2, using 
```python
dt = particle.dt / np.timedelta64(1, "s")
```

This pragmatic solution allows us to continue developing the particles data class (#2079) and Interpolators (#2081) while we chew on the Issue discussed here

PR #2079, opened 2025-07-10
Xarray dataset for particledata

### LINKED ISSUE #1822: Refactor ParticleData internals to use a pd.DataFrame
Conceptually the internals of ParticleData (i.e., dictionary of arrays, which all needs ot be the same length) is very similar to that of a dataframe. Migrating this across would be a boost to maintainability, and likely also performance.

Pandas is already a dependency of xarray, hence there isn't cost to adding it as an explicit dependency.

erikvansebille:
This PR swaps the custom ParticleData class to an `xarray.Dataset`. 

Note that it also changed the time-handling in `pet.execute()` to fully use `numpy.datetime64/numpy.timedelta64`. Also note that this PR will _not_ change the `ParticleFile` class, which will be a next PR

Other aspects that still need to be done
- [x] implement/test `particle.delete`
- [x] clean up `Particle` class
- [x] add support for custom `Variables`
- [ ] fix the particles as time delta issue (#2078)
- [x] bring in unit tests from test_kernel_language
- [x] implement `particle.ei` into the DataSet
- [x] Support negative `dt` for backward-in-time simulations
- [ ] Implement Type-annotations and -checking in new functions and methods

- [x] Fixes #1822
- [x] Added tests


parcels/particle.py
```diff
@@ -110,54 +109,28 @@ class Particle:
...
+            object.__setattr__(self, name, value)
+        else:
+            self._data[name][self._index] = value
+
+    def delete(self):
+        """Signal the particle for deletion."""
+        self.state = StatusCode.Delete
 
     @classmethod
     def add_variable(cls, var, *args, **kwargs):
```
VeckoTheGecko, line 136: I don't think the `add_variable` and `add_variables` methods would work anymore

parcels/particleset.py
```diff
@@ -141,27 +135,60 @@ def __init__(
...
+            },
         )
+        # add extra fields from the custom Particle class
+        for v in pclass.__dict__.values():
+            if isinstance(v, Variable):
+                if isinstance(v.initial, attrgetter):
+                    initial = v.initial(self).values
+                else:
+                    initial = v.initial * np.ones(len(trajectory_ids), dtype=v.dtype)
+                self._data[v.name] = (["trajectory"], initial)
```
VeckoTheGecko, line 167: I think that this dataset generation should be wrapped up into a method on the Particle class (so that this part of the code does not have to worry about custom particle classes).

parcels/particleset.py
```diff
@@ -172,41 +199,33 @@ def __getattr__(self, name):
...
         """
-        for v in self.particledata.ptype.variables:
-            if v.name == name:
-                return getattr(self.particledata, name)
+        if name in self._data:
+            return self._data[name]
         if name in self.__dict__ and name[0] != "_":
             return self.__dict__[name]
         else:
             return False
```
VeckoTheGecko, line 207: I think that this can be reduced to: `return self._data[name]` (see other comment)

parcels/particleset.py
```diff
@@ -172,41 +199,33 @@ def __getattr__(self, name):
...
 
     @property
     def size(self):
-        # ==== to change at some point - len and size are different things ==== #
-        return len(self.particledata)
-
-    @property
-    def pclass(self):
-        return self._pclass
+        return len(self._data["trajectory"])
```
VeckoTheGecko, line 222:
```suggestion
        return len(self)
```

tests/v4/test_kernel.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Testing for merging of two Kernels with the same variable declared
+    def MoveEast1(particle, fieldset, time):  # pragma: no cover
+        add_lon = 0.2
+        particle_dlon += add_lon  # noqa
+
+    def MoveEast2(particle, fieldset, time):  # pragma: no cover
+        particle_dlon += add_lon  # noqa
+
+    pset.execute([MoveEast1, MoveEast2], runtime=np.timedelta64(2, "s"))
```
VeckoTheGecko, line 36: is dt implicitly 1 second?

tests/v4/test_kernel.py
VeckoTheGecko, line 1: some of these tests have been migrated from v3 - can we delete the v3 ones then? (atm they're doubled up)

VeckoTheGecko (changes requested):
A few things that we need to work out at a conceptual level:

- `Particle` is a bit overloaded in responsibility at the moment. It acts both as a thin wrapper (hooking into the underlying `_data` via the index in order to make modifications), as well as handles* adding variables to create custom particles.
  - This overloading of responsibility is a problem because the former is concerned about the particle itself in a loop, while the latter is about types of particles (which is declared before the particleset execution loop, and is then used to actually generate the underlying particle data class).
  - I think these should be split apart - not sure on naming though. Perhaps `ParticleAccessor` and `DefaultParticle`
- Not sure if the usage of `._index` in the `Particle` at the moment is robust to the deleting of particles 


*this functionality looks to be untested at the moment, and the implementations of `add_variable` aren't yet updated

tests/v4/test_kernel.py
erikvansebille: Yep, default is `dt=np.timedelta64(1, "s")`. That is actually (also) tested here

parcels/application_kernels/interpolation.py
```diff
@@ -32,7 +32,7 @@ def UXPiecewiseConstantFace(
     This interpolation method is appropriate for fields that are
     face registered, such as u,v in FESOM.
     """
-    return field.data.values[ti, position["Z"][0], position["FACE"][0]]
+    return field.data.values[ti, position["Z"][0], position["FACE"][0]][0]
```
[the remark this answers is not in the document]
erikvansebille: No, I mentioned this in person before. The UXGrid interpolation was returning a list of(of length-1), rather than a float. That's why we need the extra [0] here. @fluidnumerics-joe, can you confirm?

parcels/particleset.py
erikvansebille: Fixed in d0c5537c0abd410f12dbc9063e50a12a437cd526

parcels/particle.py
```diff
@@ -110,54 +109,28 @@ class Particle:
...
+        self._index = index
+
+    def __getattr__(self, name):
+        if name in ["_data", "_index"]:
+            return object.__getattribute__(self, name)
+        _data = object.__getattribute__(self, "_data")
+        if name in _data:
+            return _data[name].values[self._index]
+        else:
+            return False
```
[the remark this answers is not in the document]
erikvansebille: We need the part below to avoid circular imports on init (when `._data` is set to `data`), but the rest can be simplified indeed
```
if name in ["_data", "_index"]:
            return object.__getattribute__(self, name)
```

Fixed this in bb41d4213890c9449bc50df18fbadeefa902fc51, and also added a test

parcels/application_kernels/interpolation.py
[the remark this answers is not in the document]
fluidnumerics-joe: That's a bit odd as a length one array, I believe, would've failed some of the existing tests. But here we are. I'll keep an eye on this change as development progresses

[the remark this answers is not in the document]
VeckoTheGecko: Found the problem - will be resolved in #2087

Issue #2054, opened 2025-06-30
Testing and support for 2D fields in X-Z and Y-Z, and 1D fields in X, Y, or Z


VeckoTheGecko:
#2048 introduced an API that allows the grid to naturally extend towards structured grids of different dimensionality (2D where there isn't a X or isn't a Y direction, or 1D grids).

Having testing for these grids would be really beneficial so that we can:
- a) ensure native support for this - and so that our assertions on initialisation are compatible 
- b) document how these grids can be constructed

erikvansebille:
Good idea! I'd be happy to work on this in the coming days. Where should I put these test cases?

VeckoTheGecko:
Awesome 🤩



> Where should I put these test cases?

Below `test_invalid_lon_lat` in `tests/v4/test_xgrid.py`

VeckoTheGecko:
cc @jimena-medinarubio

PR #2080, opened 2025-07-10
Add `axes` and `get_axis_dim` to BaseGrid and other cleanup

### LINKED ISSUE #2056: Adding `axes` and `get_axis_dim` to BaseGrid
In essene, ravelling and unravelling needs to know two things:

- Which axes are part of the grid
- How to get the dimensionality of an axis (i.e., how many observations/cell edges are there along a dimension)

Once that is known (and now that #2048 unifies the API for ravel/unravel across grids) the algorithm for ravelling and unravelling grids is the same regardless of grid.

I propose we make the following changes to `BaseGrid`:
- add a `grid.axes` cached property that needs to be defined on the subclasses that returns a `tuple[GridAxis]` (where `GridAxis` is a string) listing the axes names for the grid. Implementation lies on the subclasses.
- add a `get_axis_dim(axis: GridAxes) -> int` function that gives the dimensionality along a dimension (returning `None` for values not in `grid.axes`). Implementation lies on the subclasses. An lru_cache can be added to this method.
- move the ravel/unravel logic from the subclasses to the 
- remove xdim/ydim/zdim/tdim from the XGrid class, as they can be replaced with `grid.get_axis_dim("X")` etc. 
  - these attributes were at first only for backwards compatibility with v3

VeckoTheGecko:
Changes:
- Added `axes` and `get_axis_dim()` to the BaseGrid class with corresponding implementations in the UxGrid and XGrid classes. Also moved ravel/unravel logic to the BaseGrid class (fixes #2056)
- Removed `lon, lat, depth` from the Field class
  - These attributes didn't really make sense in the unstructured context (especially now we've settled on a different API), and also weren't used in the codebase. It makes more sense for these to live on the grids anyway where needed.
- Add uxarray to pyproject.toml dependencies (25c589bcb4355cf72c1f07151233c4f8391c83c0)
- Other cleanup (see commit messages)

Open questions:
- @fluidnumerics-joe, the implementation of `Field.depth` differed from `UxGrid.depth` - with the former using a helper function `get_vertical_location_from_dims`. Is this function still needed, or can we remove it?

- [x] Fixes #2056
- [x] Added tests
- [x] Added documentation


parcels/basegrid.py
```diff
@@ -134,4 +136,106 @@ def unravel_index(self, ei: int) -> dict[str, int]:
...
+        -------
+        list[str]
+            List of axis names, e.g. ["Z", "Y", "X"] for a 3D structured grid or ["Z", "FACE"] for an unstructured grid.
+        """
+        ...
+
+    @abstractmethod
+    def get_axis_dim(self, axis: str) -> int:
+        """
+        Return the dimensionality (number of cells/edges) along a specific axis.
```
VeckoTheGecko, line 162:
```suggestion
        Return the dimensionality (number of cells/faces) along a specific axis.
```

I think that this should be cells/faces and not 'edges'. This is what we ravel against - so edges don't really apply (since particles can't occupy that)

parcels/basegrid.py
```diff
@@ -134,4 +136,106 @@ def unravel_index(self, ei: int) -> dict[str, int]:
...
+        >>> ravel(dims, indices)
+        9
+        # Calculation: 0 * (3 * 4) + 2 * (4) + 1 = 0 + 8 + 1 = 9
+    """
+    strides = np.cumprod(dims[::-1])[::-1]
+    ei = 0
+    for i in range(len(dims) - 1):
+        ei += indices[i] * strides[i + 1]
+
+    return ei + indices[-1]
```
VeckoTheGecko, line 241: Admittedly feels a bit like a low priority change (since the code before was easy to read before anyway). I was thinking thought that this helps for #2054, makes it a bit easier for people to add their own grids should we want to support that in future, and consolidates the concept in one place.

tests/v4/test_xgrid.py
```diff
@@ -85,30 +90,6 @@ def test_invalid_lon_lat():
         XGrid(xgcm.Grid(ds, periodic=False))
 
 
-def test_xgrid_ravel_unravel_index():
```
VeckoTheGecko, line 88: Now tested on the BaseGrid instead (pretty much the exact same test - except generalised)

parcels/basegrid.py
fluidnumerics-joe: It's kind of nice if we don't need to implement this for each extension of BaseGrid.. But knowing we can optionally override in a type extension is good. If this doesn't fail existing tests, IMO, its good.

tests/v4/test_uxarray_fieldset.py
```diff
@@ -77,25 +77,25 @@ def uvw_fesom_channel(ds_fesom_channel) -> VectorField:
...
     pset = ParticleSet(fieldset, pclass=Particle)
     assert pset.fieldset == fieldset
 
 
 def test_set_interp_methods(ds_fesom_channel, uv_fesom_channel):
     fieldset = FieldSet([uv_fesom_channel, uv_fesom_channel.U, uv_fesom_channel.V])
     # Check that the fieldset has the expected properties
-    assert (fieldset.U == ds_fesom_channel.U).all()
-    assert (fieldset.V == ds_fesom_channel.V).all()
+    assert (fieldset.U.data == ds_fesom_channel.U).all()
```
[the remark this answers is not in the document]
erikvansebille: OK, I see and is fine then. From the user perspective I liked the old implementation of `Field.__getattr__` (what you removed https://github.com/OceanParcels/Parcels/commit/389373002b6cdfad86d40b6f6e24b864d8ddf1d6), though, because the Field is then the same as the DataArray. But if you prefer not to do that, I'll accept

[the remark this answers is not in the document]
VeckoTheGecko: Does this approach differ from what was happening in v3 @erikvansebille ? Was `Field.__getattr__` defined then?

PR #2090, opened 2025-07-17
Reverting back to float for time and dt inside pset.execute

erikvansebille:
Reverting to using `float64` for `time` and `dt` in `pset.execute()`, because it might increase performance a lot (to be explored). See also https://github.com/OceanParcels/parcels-benchmarks/pull/1

- [ ] Fixes #
- [ ] Added tests
- [ ] Added documentation


erikvansebille:
Turns out that `datetime64` and `timedelta64` computations are _not_ the bottleneck in the kernel-loop (see https://github.com/OceanParcels/parcels-benchmarks/pull/1#issuecomment-3087697033), so closing this PR

erikvansebille:
We're not pursuing this anymore; we're also getting good performance with numpy.datetime64

PR #2104, opened 2025-07-22
Swap out spatialhash query for get_faces_containing_point

### LINKED ISSUE #2103: Test get_faces_containing_point for unstructured grid search
The uxarray team has added the `get_faces_containing_point` function that has been suggested for us to try it out in Parcels.

Currently, the UXArray spatial hashing API does not support crossing the antimeridian and fails for elements that are colinear in lat/lon space. Using `get_faces_containing_point` is expected to enable global grid searching (crossing the anti-meridian). It was suggested in a [UXArray PR where spatial hashing is being updated](https://github.com/UXARRAY/uxarray/pull/1273)

Leveraging the `get_faces_containing_point` function would remove the need for spatial hash initialization (saving some memory) and would replace the `uxgrid.spatial_hash.query` call. We would need to implement locally our own barycentric coordinate calculation. 

In the spirit of robustness, I would opt for using 3-d cartesian coordinates (on the unit sphere) to compute barycentric coordinates. This would help us cope with global grids and deal with elements that may be colinear in lat-lon (e.g. near polar caps). 

For performance, I want to offer barycentric coordinates calculated using lat/lon. This cuts down on the number of computations, but is only suitable when users are running regional grids that do not cross the antimeridian or are away from poles.

fluidnumerics-joe:
- [X] Chose the correct base branch (`main` for v3 changes, `v4-dev` for v4 changes)
- [X] Fixes #2103

This PR brings in `uxarray.get_faces_containing_point` for the initial and backup search method to find the face a particle resides in. Additionally, to cope with particles crossing the antimeridian or being in cells that are colinear in lat/lon space, I've added a fallback option for calculating barycentric coordinate in cartesian (x,y,z) space (on a unit sphere). I've added a generic dataset that crosses the antimeridian with tests for particle positions within elements that cross the antimeridian.

### DIFF
--- a/parcels/_datasets/unstructured/generic.py
+++ b/parcels/_datasets/unstructured/generic.py
@@ -208,7 +208,109 @@ def _fesom2_square_delaunay_uniform_z_coordinate():
     return ux.UxDataset({"U": u, "V": v, "W": w, "p": p}, uxgrid=uxgrid)
 
 
+def _fesom2_square_delaunay_antimeridian():
+    """
+    Delaunay grid that crosses the antimeridian with uniform z-coordinate, mimicking a FESOM2 dataset.
+    This dataset consists of a square domain with closed boundaries, where the grid is generated using Delaunay triangulation.
+    The bottom topography is flat and uniform, and the vertical grid spacing is constant with 10 layers spanning [0,1000.0]
+    The lateral velocity field components are non-zero constant, and the vertical velocity component is zero.
+    The pressure field is constant.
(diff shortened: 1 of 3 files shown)


fluidnumerics-joe:
Ok, I've added the cartesian barycentric coordinates as a backup to the latlon calculation. I'll just add a test with a generic dataset that crosses the antimeridian. This *should* exercise this new feature. Hopefully just one more commit to get this wrapped up.

Issue #2107, opened 2025-07-24
Remove under the hood combining of kernels into single function


VeckoTheGecko:
_Part of #2106_

In v3 (and previously in Parcels) concatenated kernels are actually (under the hood) combined into a single function[^1]. This leads to behaviour like the following:

```python
# test_kernel_execution.py
@pytest.mark.parametrize("mode", ["scipy"])
def test_kernel_var_defining(mode):
    fieldset = FieldSet.from_data(
        {"U": [[0, 1], [2, 3]], "V": np.ones((2, 2))}, {"lon": [0, 2], "lat": [0, 2]}, mesh="flat"
    )

    def DefineA(particle, fieldset, time):
        a = 2

    def Update_dlon(particle, fieldset, time):
        particle_dlon += a

    kernels = [DefineA, Update_dlon]

    """
    Final generated kernel is effectively
        # Setcoords
        particle_dlon = 0  # noqa
        particle_dlat = 0  # noqa
        particle_ddepth = 0  # noqa
        particle.lon = particle.lon_nextloop
        particle.lat = particle.lat_nextloop
        particle.depth = particle.depth_nextloop
        particle.time = particle.time_nextloop

        # DefineA
        a = 2

        # Update_dlon
        particle_dlon += a

        # Updatecoords
        particle.lon_nextloop = particle.lon + particle_dlon
        particle.lat_nextloop = particle.lat + particle_dlat
        particle.depth_nextloop = particle.depth + particle_ddepth
        particle.time_nextloop = particle.time + particle.dt
    """

    pset = ParticleSet(fieldset, pclass=ptype[mode], lon=0, lat=0)
    pset.execute(kernels, endtime=1, dt=1)

    # if we reverse the kernels, the variable a would not be defined
    pset = ParticleSet(fieldset, pclass=ptype[mode], lon=0, lat=0)
    pset.execute(kernels[::-1], endtime=1, dt=1)  # Errors :/
```

This bespoke behaviour deviates from a lot of assumptions people have about Python code, which negatively affects:
- Static analysers
  - Ruff reports errors (which needs to be manually silenced via `noqa` comments):
    - Local variable `particle_dlon` is assigned to but never used (F841)
    - Undefined name `particle_dlon` (F821)
- Users
  - They need to get familiar with this behaviour
  - They encounter less informative error messages (since the code being run is AST (i.e., not defined at a file location), there is no line number where the error occured `<ast>:2: UnboundLocalError`)
[...]

Issue #2084, opened 2025-07-16
Rethink Kernel loop in v4


erikvansebille:
Parcels v3 introduced a revamped Kernel-loop for vectorial/commutable summing (see [tutorial](https://docs.oceanparcels.org/en/latest/examples/tutorial_kernelloop.html)). However, it's a bit clunky because
1. It requires a copy of the particle position (`lon_nextloop`, `lat_nextloop`, etc), straining memory
2. The final position of the particle is by default not written
3. Particle position (lon, lat, depth) is treated differently from other Particle variables
4. The API may be confusing to users: why do they have to update `particle_dlon` rather than `particle.lon`?

Hence, it may be good to revisit the implementation of the Kernel loop in v4, keeping the advantages like vectorial summing but avoiding the downsides listed above

erikvansebille:
https://github.com/OceanParcels/Parcels/pull/2123#discussion_r2241638088 could be a way to implemented this (to be tested)

erikvansebille:
Fixed in #2207 and #2333

PR #2123, opened 2025-07-29
Reworking kernel internels

VeckoTheGecko:
This PR is part of a significant changeset as outlined in #2107 . As there are a lot of related parts, its a bit of a larger PR.

This PR looks to:
- Remove dead code.
- Provide a single execution path for compiling kernels via function objects (an execution path that is compatible with numba).
  - Currently its possible to compile kernels from function objects, as well as dynamically by accessing the user context and providing code strings/ast .
- Run kernels sequentially instead of combining them together.

Changes:
- Removal of BaseKernel class
- Remove `particle.delete()` (this is a breaking change)
- Store `particle_dlon`, `particle_dlat`, `particle_ddepth` (this is a breaking change)
- Remove `Kernel.funccode` , `Kernel.py_ast`, `Kernel.funcvars` from internals and init
- Remove `Kernel.funcname`  from init
- _see commit log for additional minor changes_

Future work:
- decouple kernel init from fieldset and particle type (fieldset and particle type is only needed at compilation time)
- remove `_positionupdate_kernels_added` state variable (easier to do this another way)

Breaking changes are recorded in `v3to4-breaking-changes.md` so that we have an overview during development (the purpose of this doc currently is only for breaking changes we intend to propogate to v4 of Parcels so that we can use this as a basis for the migration guide. For things that are currently breaking changes or things where its uncertain where we'll land, I propose we use the issue tracker instead).

I have one test failure in this branch on `tests/v4/test_advection.py::test_moving_eddy[RK45-1e-05]` - still need to investigate.


parcels/application_kernels/advectiondiffusion.py
```diff
@@ -24,7 +26,7 @@ def AdvectionDiffusionM1(particle, fieldset, time):  # pragma: no cover
     The Wiener increment `dW` is normally distributed with zero
     mean and a standard deviation of sqrt(dt).
     """
-    dt = particle.dt / np.timedelta64(1, "s")  # noqa TODO improve API for converting dt to seconds
```
erikvansebille, line 27: Don’t we want to keep the reminder note that we may want to consider the API for dt?

parcels/kernel.py
```diff
@@ -88,119 +39,92 @@ class Kernel(BaseKernel):
...
+
     def add_positionupdate_kernels(self):
         # Adding kernels that set and update the coordinate changes
         def Setcoords(particle, fieldset, time):  # pragma: no cover
             import numpy as np  # noqa
 
-            particle_dlon = 0  # noqa
-            particle_dlat = 0  # noqa
-            particle_ddepth = 0  # noqa
+            particle.dlon = 0
```
erikvansebille, line 113: So now we have three location variables per direction for each particle, increasing the memory footprint of the pset. Is there no way to keep using only two?

parcels/kernel.py
```diff
@@ -294,15 +209,11 @@ def from_list(cls, fieldset, ptype, pyfunc_list, *args, **kwargs):
             Additional keyword arguments passed to first kernel during construction.
         """
         if not isinstance(pyfunc_list, list):
-            raise TypeError(f"Argument function_list should be a list of functions. Got {type(pyfunc_list)}")
-        if len(pyfunc_list) == 0:
-            raise ValueError("Argument function_list should have at least one function.")
+            raise TypeError(f"Argument `pyfunc_list` should be a list of functions. Got {type(pyfunc_list)}")
         if not all([isinstance(f, types.FunctionType) for f in pyfunc_list]):
-            raise ValueError("Argument function_lst should be a list of functions.")
+            raise ValueError("Argument `pyfunc_list` should be a list of functions.")
```
erikvansebille, line 214: Why is this a ValueError and two lines above a TypeError?

parcels/kernel.py
```diff
@@ -390,7 +301,13 @@ def evaluate_particle(self, p, endtime):
             except KeyError:
                 if sign_dt * (endtime - p.time_nextloop) <= p.dt:
                     p.dt = sign_dt * (endtime - p.time_nextloop)
-            res = self._pyfunc(p, self._fieldset, p.time_nextloop)
+            res = None
+            for f in self._pyfuncs:
+                res_tmp = f(p, self._fieldset, p.time_nextloop)
+                if res_tmp is not None:  # TODO v4: Remove once all kernels return StatusCode
+                    res = res_tmp
```
erikvansebille, line 308: In the JIT version of v3, we kept the “highest” statuscode, with the idea that this was the most important

tests/v4/test_particleset_execute.py
```diff
@@ -154,8 +154,9 @@ def test_uxstommelgyre_pset_execute(verbose_progress):
         runtime=np.timedelta64(10, "m"),
         dt=np.timedelta64(60, "s"),
         pyfunc=AdvectionEE,
-        verbose_progress=verbose_progress,
     )
+    assert utils.hash_float_array([p.lon for p in pset]) == 1165396086
```
erikvansebille, line 158: I wondered, it this sensitive to computer platform because we hash _all_ digits? We’re working with floats, after all. Why don’t we round it first, so that the hash becomes more like np.isclose?

v3to4-breaking-changes.md
```diff
@@ -0,0 +1,4 @@
+Kernels:
+
+- `particle.delete()` is no longer valid. Have to do `particle.state = StatusCode.Delete`
+- Sharing state between kernels must be done via the particle data (as now the kernels are not combined under the hood).
```
erikvansebille, line 4:
```suggestion
- Sharing state between kernels must be done via the particle data (as now the kernels are not combined under the hood).
- dt is a np.timedelta64 object

```

parcels/kernel.py
erikvansebille: Can we not do
1. initialise `particle.dlon` = 0
2. At the start of each kennel loop, `particle.lon += particle.dlon`
3. Then run the kernels?

This way, we could get rid of `particle.lon_nextloop`? Or am I missing something?

parcels/application_kernels/advectiondiffusion.py
VeckoTheGecko: Yes, I've added them back.

They were automatically removed by [Ruff](https://docs.astral.sh/ruff/rules/unused-noqa/) (if it was formatted like `# noqa # my comment`, Ruff would have kept `# my comment`).

parcels/kernel.py
VeckoTheGecko: I think that we should do this in a separate PR. I saw that change as being separate, and it would be good to test it properly (i.e., once we have particle file output so that we can compare).

erikvansebille: OK, fully agree!

erikvansebille: This effort would then be part of solving #2084

parcels/kernel.py
VeckoTheGecko: > Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).
> - [Python docs](https://docs.python.org/3/library/exceptions.html)

The first is a TypeError since we expect a list, the second is a ValueError since we get a list, but the value of the list is wrong (i.e., it contains non-functions). Honestly the line between `ValueError` and `TypeError` is a bit blurry here, and generally I don't think that the distinction really matters. Some packages I see reach for `ValueError` when one can argue that `TypeError` is needed.

erikvansebille: OK, thanks for enlightening me. All good then

erikvansebille (approved):

parcels/kernel.py
VeckoTheGecko: What you're refering to is in JIT mode during interpolation. This is different from the particle status code that is returned after each kernel (which is what this part of the code is handling).

Leaving as is.

PR #2133, opened 2025-08-05
Fix PolyTEOS10-bsq kernel

michaeldenes:
- [ ] Fixes #
- [ ] Added tests
- [ ] Added documentation

I’ve come across a bug in the PolyTEOS10 kernel used to calculate the density of seawater from the temperature and salinity fields. The kernel is based on equation (13) from https://www.sciencedirect.com/science/article/pii/S1463500315000566 where the density is determined from the summation of a vertical reference profile and a residual function (or density anomaly). Previously, the kernel only computed the density anomaly and was missing the vertical reference profile. At 0 m depth, the vertical reference profile component is 0, but is non-zero away from the surface.

This small pull-request updates the kernel to include the vertical reference profile component of the equation.


erikvansebille (approved):

erikvansebille:
Ready to merge this in, @michaeldenes?

Issue #2004, opened 2025-05-08
Example datasets for different (ocean) circulation models


VeckoTheGecko:

VeckoTheGecko:
> analytical functions...This is a bit more useful for validating interpolation methods.

Agreed that we'd need analytical functions for validating interpolation methods.

> I would vote for using analytical functions rather than random data in the fields

How straight forward would this be? (I imagine the grid information would need to be accounted for here? This would be quite involved dataset generation no?)

Rather than modifying the raw datasets which are shared across the project, would we be able to either (a) provide specific examples in a `datasets.structured.analytical`, or (b) do this by patching the data?

Patching:
```py
from parcels._datasets.structured.generic import datasets as datasets_structured
ds = datasets_structured["some_dataset"]   # dataset has random data
ds["U"] = patch_with_analytical_field(ds["U"], equation="f = x*y*z") # dataset has same structure/attrs/etc, but now with analytical data
```

I think that this:
- is more extensible (i.e., can expand to non-linear functions)
- leaves the original datasets unchanged
  - I imagine in a lot of tests we don't care about the actual data. For these test cases, the `generic` module as is makes it quite easy to see the structure of the different datasets and including lots of code for analytical field generation would muddy things here I think.

(b) is probably much more work, might be easier to go with (a)

VeckoTheGecko:
> What would be a use case for random data ?
Oh yes, no particular usecase for random data - just wanted _something_ in there.


> In the short term, I would say constant data would be more useful than random (if we can't interpolate constant data correctly, all bets are off).

Agreed. What do you have in mind? 
- replace `np.random` with `np.ones`? 
- more spatially varying so you can do interesting interpolation?
  - if this, perhaps create a new dataset as opposed to modifying existing? that way the existing datasets can stay a bit simpler

VeckoTheGecko:
In response to https://github.com/OceanParcels/Parcels/issues/2004#issuecomment-2912516985

> * Curvilinear cartesian grid
> * Spherical grid rectilinear
> * Spherical grid curvilinear

Yes, I think this would be good to put in `datasets.structured.generic` since they aren't specific to a certain model.

> * Examples where time domain is managed with the `calendar` package in MITgcm

Do you know if/how this differs from cftime?

VeckoTheGecko:
Hi @JamiePringle , thanks for the offer. The intention of this issue is actually just to match the _structure_ of the datasets (rather than have actual datasets).

[...]

PR #2135, opened 2025-08-06
More xgrid generated datasets

erikvansebille:
This PR moves the RadialRotation, MovingEddy, DecayingMovingEddy, StommelGyre and Peninsula flow fields to `parcels/_datasets/structured/generated.py`. It also adds unit tests to `test_advection.py` for each of these generated flow fields

- [x] Addresses #2004
- [x] Added tests

### DIFF
--- a/parcels/application_kernels/TEOSseawaterdensity.py
+++ b/parcels/application_kernels/TEOSseawaterdensity.py
@@ -6,7 +6,7 @@ __all__ = ["PolyTEOS10_bsq"]
 
 
 def PolyTEOS10_bsq(particle, fieldset, time):  # pragma: no cover
-    """Calculates density based on the polyTEOS10-bsq algorithm from Appendix A.2 of
+    """Calculates density based on the polyTEOS10-bsq algorithm from Appendix A.1 and A.2 of
     https://www.sciencedirect.com/science/article/pii/S1463500315000566
     requires fieldset.abs_salinity and fieldset.cons_temperature Fields in the fieldset
     and a particle.density Variable in the ParticleSet
@@ -27,10 +27,20 @@ def PolyTEOS10_bsq(particle, fieldset, time):  # pragma: no cover
     SA = fieldset.abs_salinity[time, particle.depth, particle.lat, particle.lon]
     CT = fieldset.cons_temperature[time, particle.depth, particle.lat, particle.lon]
 
-    SAu = 40 * 35.16504 / 35
-    CTu = 40
+    SAu = 40.0 * 35.16504 / 35.0
+    CTu = 40.0
     Zu = 1e4
-    deltaS = 32
+    deltaS = 32.0
+
+    zz = -Z / Zu
+    R00 = 4.6494977072e01
+    R01 = -5.2099962525e00
+    R02 = 2.2601900708e-01
+    R03 = 6.4326772569e-02
+    R04 = 1.5616995503e-02
+    R05 = -1.7243708991e-03
+    r0 = (((((R05 * zz + R04) * zz + R03) * zz + R02) * zz + R01) * zz + R00) * zz
+
     R000 = 8.0189615746e02
     R100 = 8.6672408165e02
     R200 = -1.7864682637e03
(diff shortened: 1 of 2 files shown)


tests/v4/test_particleset.py
```diff
@@ -54,6 +54,19 @@ def test_pset_with_pids(fieldset, offset, npart=100):
...
+    vars = ["sample_var", "sample_var2"]
+    for var in vars:
+        MyParticleLoop = MyParticleLoop.add_variable(Variable(var))
+
+    MyParticleList = Particle.add_variable([Variable(var) for var in vars])
+
+    for var in vars:
+        assert hasattr(MyParticleLoop, var), f"Variable {var} not added to MyParticleLoop"
+        assert hasattr(MyParticleList, var), f"Variable {var} not added to MyParticleList"
+
```
VeckoTheGecko, line 68: I'm not quite sure what this test is?

parcels/application_kernels/advection.py
```diff
@@ -115,9 +115,15 @@ def AdvectionRK45(particle, fieldset, time):  # pragma: no cover
...
-    )  # TODO: improve API for converting dt to seconds
+    dt = particle.next_dt / np.timedelta64(1, "s")  # TODO: improve API for converting dt to seconds
+    if dt > fieldset.RK45_max_dt:
+        dt = fieldset.RK45_max_dt
+        particle.next_dt = fieldset.RK45_max_dt * np.timedelta64(1, "s")
+    if dt < fieldset.RK45_min_dt:
+        particle.next_dt = fieldset.RK45_min_dt * np.timedelta64(1, "s")
+        return StatusCode.Repeat
+    particle.dt = particle.next_dt
+
```
VeckoTheGecko, line 126: Is this a change in behaviour from v3, or is it something that was in v3 and is being re-implemented in v4? Perhaps worth noting in the body of the PR description?

parcels/xgrid.py
```diff
@@ -271,7 +271,10 @@ def _gtype(self):
     def search(self, z, y, x, ei=None):
         ds = self.xgcm_grid._ds
 
-        zi, zeta = _search_1d_array(ds.depth.values, z)
+        if hasattr(ds, "depth"):
```
VeckoTheGecko, line 274: I think this would be better than a `hasattr` check
```suggestion
        if "Z" in self.grid.axes:
```

parcels/fieldset.py
```diff
@@ -164,7 +164,8 @@ def add_constant(self, name, value):
         """
         if name in self.constants:
             raise ValueError(f"FieldSet already has a constant with name '{name}'")
-
+        if isinstance(value, np.timedelta64):
+            value = value / np.timedelta64(1, "s")  # Convert to seconds
         self.constants[name] = np.float32(value)
```
VeckoTheGecko, line 169: I think that it would be better for us to error out when provided non-numeric data, than it would be to convert types under the hood (as then users might expect a type but receive another)

VeckoTheGecko (changes requested):

parcels/_datasets/structured/generated.py
```diff
@@ -18,3 +20,267 @@ def simple_UV_dataset(dims=(360, 2, 30, 4), maxdepth=1, mesh_type="spherical"):
...
+            "YG": (["YG"], np.arange(ydim), {"axis": "Y", "c_grid_axis_shift": -0.5}),
+            "XC": (["XC"], np.arange(xdim) + 0.5, {"axis": "X"}),
+            "XG": (["XG"], np.arange(xdim), {"axis": "X", "c_grid_axis_shift": -0.5}),
+            "lat": (["YG"], lat, {"axis": "Y", "c_grid_axis_shift": 0.5}),
+            "lon": (["XG"], lon, {"axis": "X", "c_grid_axis_shift": -0.5}),
+        },
+    )
+
+
+def moving_eddy_dataset(xdim=2, ydim=2):  # TODO check if this also works with xdim=1, ydim=1
```
VeckoTheGecko, line 64: Not sure what this TODO comment means - would a dim = 1 dataset be useful?

parcels/application_kernels/advection.py
erikvansebille: Yes, this was a (small) bug in RK45 in v3 that I fixed now

parcels/_datasets/structured/generated.py
erikvansebille: Yes, because the field has no spatial variability, it would also make sense on xdim=1. But v4-dev currently doesn't support it yet. #2122 does support it (I think), so that's why I left the TODO here

PR #2155, opened 2025-08-20
Moving the `mesh` information to the `(U)XGrid` class

### LINKED ISSUE #2154: Clean up the `mesh_type` property of the Grid
In the current `v4-dev`, the `_mesh_type` is a property of the `Field`, but we mostly need access to it in the `Grid`. A Field requires a `UnitConverter`, but may not need the mesh information directly. 

Hence, we should probably clean up the code so that
1. A `Grid` has a `_mesh_type`
2. The `UnitConverter` of a `Field` is determined based on both the `_mesh_type` of its `grid` and and its name (or another way to find out if a field contains velocity or diffusion values)

erikvansebille:
This PR cleans up the different ways in which `mesh` type information ("spherical" or "flat") was used in v4. It is now a property of the `(U)XGrid` class

- [x] Fixes #2154 
- [x] Added tests


VeckoTheGecko (approved):
Agreed that this does clean things up significantly.

Do we want to rename `mesh` to `mesh_type`? In v3 we used `mesh`, in v4-dev we were inconsistent (`mesh` and `mesh_type`). I agree we should just chose one, but I don't personally find `mesh` to be confusing naming (especially since we have type annotations), and the renaming of the public variable seems more confusing for users than keeping the original. This is a nitpick - if you feel strongly about `mesh_type` go for it.

erikvansebille:
> Agreed that this does clean things up significantly.
> 
> Do we want to rename `mesh` to `mesh_type`? In v3 we used `mesh`, in v4-dev we were inconsistent (`mesh` and `mesh_type`). I agree we should just chose one, but I don't personally find `mesh` to be confusing naming (especially since we have type annotations), and the renaming of the public variable seems more confusing for users than keeping the original. This is a nitpick - if you feel strongly about `mesh_type` go for it.

I see your point about trying to keep the API as similar as possible. Let's indeed then go for `mesh`. I'll update the code

PR #2156, opened 2025-08-22
Adding the round_and_hash_float_array also to v3

erikvansebille:
This PR adds the `tests/utils.round_and_hash_float_array()` function also to v3, so we can easily compare unit test results between v3 and v4.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,5 +1,6 @@
 """General helper functions and utilies for test suite."""
 
+import struct
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


VeckoTheGecko (approved):
LGTM. Thought if we want to 

> unit test results between v3 and v4.

we of course also need to identify tests that we want to use for this comparison. The test suites have diverged quite a bit - but I imagine you already have some tests in mind @erikvansebille ? Also, we have the saved outputs introduced in ffb2fcab6036b8d1450a0d36b2a678407058aa0b and used in `tests/v4/test_interpolation.py` - but I assume you want to expand the comparison to other parts in the codebase?

PR #2158, opened 2025-08-23
Feature/morton hashing

### LINKED ISSUE #2144: Spatial hashing and curvilinear search performance improvements
With closure of #2132 and initial benchmarking from @erikvansebille in https://github.com/OceanParcels/parcels-benchmarks/pull/5, it appears there are a few areas of improvement that are possible for the spatial hashing implementation

* Hash table creation currently takes around 132 seconds for the MOI benchmark. This requires large particle simulations to make the cost of initialization worth it. We need to look into ways to speed this up.

* The `_search_curvilinear_2d` function currently has a bug that causes only the first face index to be captured on output of the spatial hash query. Fixing this bug should have significant improvements as we have found that the subsequent search iterates just as much as the previous implementation (in fact, with spatial hashing, we should be able to immediatley compute computational coordinates upon finding the face a particle resides in).

fluidnumericsJoe:
- [ X ] Fixes #2144 

## Overview
This PR brings in a fairly substantial change to the spatial hashing. In experimenting with the spatial hashing in the MOi benchmark, we found that the spatial hashing spent a large amount of time in initialization. Further, the search times were not providing an improvement over the naive implementation for searching in curvilinear grids. This was found to be related to a few key issues
1. The chosen hash function that relates hash keys to a short list of faces/cells in the curvilinear grid produced long searches in areas where the curvilinear grid approached the poles (e.g. over canada and siberia in the tripolar grid)
2. The list of lists structure for the hash table required a large number of append operations (within nested for loops) during hash table creation. Attempts to pre-allocate and use a CSR format for the hash table are hampered by point (1); a large number of overlaps of curvilinear cells with the hash grid resulted in large allocations that are prohibitive
3. Querying was a challenge to vectorize with the list of lists hash table and computation of barycentric coordinates for particle in cell checking was found to be expensive

This PR addresses these issues by changing the hashing strategy. The "hash function" is a function that takes in a position on the grid and returns an integer value that has some correlation with spatial locality. In other words the hash function of two nearby points in space, should return integer values that are close to each other in value. 

## Morton encoding for hash table construction
Here, we opt to use [Morton encoding](https://en.wikipedia.org/wiki/Z-order_curve) of three floating point values to compute the hash of a position. In spherical meshes, the latitude and longitude points are converted to `x,y,z` locations on the unit sphere. On "flat" meshes, `z` is fixed to zero and `x` and `y` are taken as is, with no conversions. The hash table is constructed by relating the `(j,i)` indices of the curvilinear grid to the morton code of the centroid of each curvilinear cell. The hash table is quick to calculate and is simply stored as a compressed sparse row (CSR) matrix and sorted in order of increasing morton number. This last step (sorting) adds some cost to the initialization but saves us considerably in the querying stage since `np.searchsorted` can be used (instead of `np.search`).

  A comparison of the construction time for the MOi benchmark is shown below for v4-dev and the new proposed morton encoding spatial hash function.

Construction time (s) vs  Version

## Morton encoding for particle position queries
With the hash table as a CSR matrix, it's a bit easier to vectorize queries over a large number of particles. Essentially, the morton code of all particle positions are calculated and searched (using `np.searchsorted`) against the hash table. This returns, for each particle, a short list of `(j,i)` indices to check for each particle. We then perform distance minimization between each particle and it list of curvilinear cells where the distance is measured from the curvilinear cell centroid. This avoids the need for barycentric coordinate calculation. At the end, the method returns, as before, a list of `(j,i)` indices, one for each particle position, where the particle is found. 

Below is a figure comparing the original v4-dev naive search, with the previous spatial hashing search, and the new morton spatial hashing for the MOi benchmark across a range of particle counts. Note that that new morton hashing search times include the spatial hash reconstruction, which is minimal (~2s on system76 Lemur with NVMe drives)

chart (1)

The above figure shows improved scaling in the morton hash based search functionality. However, the 'v4 search indices' and 'v4 hash table' runtimes were generated on @erikvansebille 's work-station and the 'v4 morton' runtimes were on mine, making direct comparison challenging.

Below shows the spatial hashing search (what's currently on v4-dev) and the propose "v4 morton " implementation (x-axis is log scale).

v4-dev (spatial hashing) and v4 morton

### DIFF
--- a/parcels/spatialhash.py
+++ b/parcels/spatialhash.py
@@ -368,11 +368,26 @@ def _encode_morton3d(x, y, z, xmin, xmax, ymin, ymax, zmin, zmax):
     Quantize (x, y, z) to 10 bits each (0..1023), dilate the bits so there are
     two zeros between successive bits, and interleave them into a 3D Morton code.
 
+    Parameters
+    ----------
+    x, y, z : array_like
+        Input coordinates to encode. Can be scalars or arrays (broadcasting applies).
+    xmin, xmax : float
+        Minimum and maximum bounds for x coordinate.
+    ymin, ymax : float
+        Minimum and maximum bounds for y coordinate.
+    zmin, zmax : float
+        Minimum and maximum bounds for z coordinate.
+
+    Returns
+    -------
+    code : ndarray, dtype=uint32
+        The resulting Morton codes, same shape as the broadcasted input coordinates.
+
     Notes
     -----
     - Works with scalars or NumPy arrays (broadcasting applies).
-    - Output is up to 30 bits; we return np.uint32 (or np.uint64 if you prefer).
-    - Requires `part1by2` defined as in your previous snippet.
+    - Output is up to 30 bits returned as uint32.
     """
     # Convert inputs to ndarray for consistent dtype/ufunc behavior.
     x = np.asarray(x)


parcels/spatialhash.py
```diff
@@ -30,171 +30,234 @@ def __init__(
...
+            self._zc = np.mean(_zbound, axis=-1)
+
+        else:
+            # Boundaries of the hash grid are the bounding box of the source grid
+            self._xmin = self._source_grid.lon.min()
+            self._xmax = self._source_grid.lon.max()
+            self._ymin = self._source_grid.lat.min()
+            self._ymax = self._source_grid.lat.max()
+            self._zmin = 0.0
+            self._zmax = 0.0
```
erikvansebille, line 83: Do we even need `z`, if we are 2D and it appears to be zero everywhere? Or do I miss something? 
Because if this is to generalise later to 3dimensional search; I don't think we need that. I don't expect us to do any 3D searches, given that depth is always perpendicular to lon/lat in the grid (and thus can be searched in 1D)

parcels/spatialhash.py
```diff
@@ -242,69 +305,107 @@ def _barycentric_coordinates(nodes, point, min_area=1e-8):
...
-    The lon and lat arrays are assumed to be 2D arrays of points with dimensions (n_y, n_x).
-    The area is computed using the Shoelace formula.
-    This method is only used during hashgrid construction to determine the size of the hash cells.
+def _latlon_rad_to_xyz(lat, lon):
+    """Converts Spherical latitude and longitude coordinates into Cartesian x,
+    y, z coordinates.
+    """
+    x = np.cos(lon) * np.cos(lat)
+    y = np.sin(lon) * np.cos(lat)
+    z = np.sin(lat)
```
erikvansebille, line 314: Are you sure this is the right formula? Eqs 4-6 on https://mathworld.wolfram.com/SphericalCoordinates.html have some of the `sin` and `cos` swapped?

parcels/spatialhash.py
```diff
@@ -242,69 +305,107 @@ def _barycentric_coordinates(nodes, point, min_area=1e-8):
...
+def _encode_morton3d(x, y, z, xmin, xmax, ymin, ymax, zmin, zmax):
+    """
+    Quantize (x, y, z) to 10 bits each (0..1023), dilate the bits so there are
+    two zeros between successive bits, and interleave them into a 3D Morton code.
+
+    Notes
+    -----
+    - Works with scalars or NumPy arrays (broadcasting applies).
+    - Output is up to 30 bits; we return np.uint32 (or np.uint64 if you prefer).
+    - Requires `part1by2` defined as in your previous snippet.
```
erikvansebille, line 374: Whose previous snippet does 'your' refer to here? 

If you copied codes from somewhere else, perhaps add a reference link?

parcels/spatialhash.py
```diff
@@ -242,69 +305,107 @@ def _barycentric_coordinates(nodes, point, min_area=1e-8):
...
+    # --- 1) Normalize each coordinate to [0, 1] over its bounding box. ---
+    # Compute denominators once (avoid division by zero if bounds equal).
+    dx = xmax - xmin
+    dy = ymax - ymin
+    dz = zmax - zmin
+
+    # Normalize to [0,1]; if a range is degenerate, map to 0 to avoid NaN/inf.
+    xn = np.where(dx != 0, (x - xmin) / dx, 0.0)
+    yn = np.where(dy != 0, (y - ymin) / dy, 0.0)
+    zn = np.where(dz != 0, (z - zmin) / dz, 0.0)
```
erikvansebille, line 390: These still give warnings `RuntimeWarning: invalid value encountered in divide` which can be fixed by the following lines

```suggestion
    with np.errstate(invalid="ignore"):
        xn = np.where(dx != 0, (x - xmin) / dx, 0.0)
        yn = np.where(dy != 0, (y - ymin) / dy, 0.0)
        zn = np.where(dz != 0, (z - zmin) / dz, 0.0)
```

parcels/spatialhash.py
```diff
@@ -242,69 +305,107 @@ def _barycentric_coordinates(nodes, point, min_area=1e-8):
...
+    dy3 = _dilate_bits(yq).astype(np.uint64)
+    dz3 = _dilate_bits(zq).astype(np.uint64)
+
+    # --- 4) Interleave the dilated bits into a single Morton code. ---
+    # Bit layout (from LSB upward): x0,y0,z0, x1,y1,z1, ..., x9,y9,z9
+    # We shift z's bits by 2, y's by 1, x stays at 0, then OR them together.
+    # Cast to a wide type before shifting/OR to be safe when arrays are used.
+    code = (dz3 << 2) | (dy3 << 1) | dx3
+
+    # If you want a compact type, it fits in 30 bits; uint32 is enough.
```
erikvansebille, line 410: Again, who is 'you' here? Rephrase more in line with the rest of Parcels comments?

```suggestion
    # Since our compact type fits in 30 bits, uint32 is enough.
```

erikvansebille (commented):
This is fantastic, @fluidnumerics-joe! The performance improvements are amazing!! For 1M particles, I now get a 2 seconds time, instead of the 600 seconds with the spatial hash map 🤩. 

What a great leap towards a fast v4!

Is this implementation also useful for unstructured grids?

parcels/spatialhash.py
fluidnumerics-joe: I wrote the original single point methods and used copilot to help vectorize everything. It must've left this comment  in the docstring

parcels/spatialhash.py
erikvansebille: OK clear. Perhaps good to add this a comment that this is needed for `mesh="flat"`?
```suggestion
            # setting min and max below is needed for mesh="flat"
            self._zmin = 0.0
            self._zmax = 0.0
```

erikvansebille:
> I want to follow up on this after this pr. I suspect yes. 

Yes, fully agree that this should be a separate PR

erikvansebille (approved):

PR #2189, opened 2025-09-08
DOC: Add page for v4-migration guide

VeckoTheGecko:
This PR renames the breaking changes file to be the migration guide so that its visible in the documentation website.

Once development has stabilised (or in the week leading up to the parcels event) we can flesh this out with code examples.

- [x] Fixes none


docs/v4/index.md
```diff
@@ -22,4 +22,5 @@ api
 nojit
 TODO
 Parcels v4 Project Board <https://github.com/orgs/OceanParcels/projects/5>
+Parcels v4 migration guide <../community/v4-migration>
```
VeckoTheGecko, line 25: Linking it as well from this page, but once v4 is released this whole "v4" section will be deleted (we probably might want to do other restructuring as well, to be decided/discussed later)

Issue #2143, opened 2025-08-13
Refactor particleset attributes and particle data for vectorized kernels


erikvansebille:
Following from #2122, it would be good to rethink the API and refactor the code for particleset data. See below the original comment by @veckothegecko
> Hmm, I assume that this change was needed since we are passing the particleset all the way to the kernels now? I think we should at some point pass the particle data itself down - as I touched on f2f, there is a distinction between the particle data and the ParticleSet (e.g., it doesn't make sense to call pset.execute() in a kernel). By passing the particledata down to the kernel it makes this clear. This also means that we don't silently limit users variable names (currently they can't call their variable names `execute` or anything that is already a method/attr on ParticleSet).

Happy for that to be a future issue.

Perhaps:
- Make ParticleData its own class (dict of arrays under the hood, but allowing `.attr` notation to get and set values)
  - Only have private variables/methods/names on this dataset (i.e., `_data`). All public attribute access will be assumed to correspond with particle data -> this results in no shadowing
  - Disallow names starting with `_` in `parcels/particle.py@Variable` as this could collide with PariticleData
- Pass ParticleData down to the kernels
- Update `pset._data` to be `pset.data` so that its on our public API.
  - @erikvansebille can we then remove the `__getattr__` and `__setattr__` methods on the `pset`? I don't think its too prohibitive to do `pset.data.lon` if users want to access their lon values, or would you prefer to still have the `pset.lon`? (a getattr? or also a setattr?)

Also, thoughts on explicitly disallowing `object.__setattr__(self, name, value)` in that last else clause? I know that its to maintain the default behaviour for Python objects, but I don't see the benefit in supporting it here - and since it works outside of our model it means that that data would fall through the cracks if we write particle data (and also means that if people typo in their kernel it will be a silent bug)

_Originally posted by @VeckoTheGecko in https://github.com/OceanParcels/Parcels/pull/2122#discussion_r2263531738_

VeckoTheGecko:
> we now pass a list of KernelParticles to the Kernels, instead of the entire ParticleSet

*not a list of `KernelParticle` instances, but a `KernelParticle` instance where the indices is a boolean mask (instead of single indices - 1, 2, 3 etc. - as it was before)

[...]

PR #2199, opened 2025-09-10
Updating Argo tutorial to v4

### LINKED ISSUE #2192: Create a vectorised version of the Argo floats kernel
To explore how vectorised Kernels would work with kernels with more complex logic, it would be nice to migrate the Argo floats kernel tutorial from https://docs.oceanparcels.org/en/latest/examples/tutorial_Argofloats.html

This may also mean making a decision where to store tutorial notebooks for v4 and turning on the CI for these

erikvansebille:
This PR converts the Argo tutorial from v3 to v4. 

It still has a few (breaking) issues that will need to be resolved:
- [ ] [BREAKING] The output zarr file can't be parsed by cftime. Getting an error `ValueError: Failed to decode variable 'time': unable to decode time units 'seconds since 2002-01-01T00:00:00' with "calendar 'standard'". Try opening your dataset with decode_times=False or installing cftime if it is not installed` which is strange because `seconds since 2002-01-01T00:00:00` seems a perfectly fine time unit?
- [x] [BREAKING] The logic in the `ArgoVerticalMovement()` kernel doesn't work. Particle properties, including `cycle_phase`, are not updated as expected. This likely has to do with the current implementation of `KernelParticle`, which is rather ad-hoc
- [x] After loading the `GlobCurrent_example_data` as a DataSet, we have to recalculate the `ds["time"]`, otherwise we get a `ValueError: Expected right to be a np.timedelta64, datetime, cftime.datetime, or np.datetime64. Got <class 'numpy.float64'>. Error getting time interval for field 'U'. Are you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects?`
- [ ] Performance can be improved. Running the notebook without an explicit `ds.load()` is extremely slow (> 15 minutes) compared to the ~70 seconds when the full DataSet is loaded in memory. 
- [ ] The particle drift_age and other calendar-based properties can't yet be converted to `np.timedelta` because that gives an error in output_file writing (`KeyError: dtype('<m8[s]')` in `attrs["time"].update(_get_calendar_and_units(time_interval))`)
- [x] Wrap some of the lines for creating the Fieldset from the DataSet into a method (`FieldSet.from_GlobCurrent` or something?)
- [ ] CI has to be turned on for this notebook so that it gets tested/run

UPDATE: I now swapped to Copernicus Marine data; which solved a few of the major issues above

- [x] Fixes #2192 
- [x] Added tests
- [x] Added documentation


parcels/particleset.py
```diff
@@ -538,7 +538,7 @@ def execute(
 
         # Set up pbar
         if output_file:
-            logger.info(f"Output files are stored in {output_file.store}.")
```
VeckoTheGecko, line 541: This needs to be rethought to work with all zarr stores. Agreed we should print path if there is one thought

erikvansebille: Done in d1cb597eae187d8c8316e997001e5cfa682e10dc

erikvansebille: Even better! Now done in 5237ccce9262db4274509a9411b76363fc13eb17

erikvansebille:
The CI for this PR fails because of the unit test introduced in dcdd2d15ecac52655db3f938f26b106aa3538d43. 

This is intentional though (until we fix #2143), so perhaps we should move that dcdd2d15ecac52655db3f938f26b106aa3538d43 commit to another branch and merge this one? 

@VeckoTheGecko, can you help out if you agree?

VeckoTheGecko (approved):
Looks good. Note that we are yet to re-enable notebooks in CI - let's do that in a future PR (it becomes much easier to do in CI and locally once #2205 is fixed)

Issue #2175, opened 2025-09-03
Morton hashing still requires while loop in some cases in Curvilinear Grid


fluidnumerics-joe:
In discussion with @erikvansebille , there is desire to remove the `while` loop in the curvilinear grid cell search immediately following the spatial hash query. In the query method, I decided to relax the condition to match the particle's morton encoded position with an exact morton code from source grid as this strict matching caused a number of tests to fail. The thinking is that if a particle position has a "nearby" morton encoding then it will be spatially close to a cell centroid with a morton encoding in close proximity.

However, relaxing strict matching prevents us from safely removing the `while` loop in the curvilinear grid cell search. It sounds like we need to spend some time reviewing quantization and bit interleaving strategies to see if we can enforce exact morton code matching on particle positions.

erikvansebille:
I just uploaded a minimal example (i.e. a breaking case) in #2177. It's one of the unit tests: tests/v4/test_index_search.py::test_grid_indexing_fpoints

PR #2211, opened 2025-09-14
Use element indices `particle.ei` in curvilinear search

erikvansebille:
While the Morton encoding for the curvilinear search (#2175)  is very fast, it turns out to be even faster to use the element indices (`particle.ei`) in the search.

Using the benchmark script in [OceanParcels/parcels-benchmarks@`9a39630` (#5)](https://github.com/OceanParcels/parcels-benchmarks/pull/5/commits/9a39630a6c03f12f445a842726daec17a85e186c), we get the following speed-up:

```
Running 4_999_696 particles with parcels v4
Execution time 1: 44 seconds
Execution time 2: 1 seconds
```

where execution 1 is with the Morton search (since `particle.ei = None`) and execution 2 is with the code in this PR to use the `particle.ei`. 

Note that this is the most optimal case where particles don't move between searches, but that is actually very often the case because
1. A Kernel loop requires multiple Field interpolations at the same location
2. Most particles move less than one grid cell between Kernel loops

### DIFF
--- a/parcels/_index_search.py
+++ b/parcels/_index_search.py
@@ -75,20 +75,18 @@ def curvilinear_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray
 
 
 def _search_indices_curvilinear_2d(
-    grid: XGrid, y: np.ndarray, x: np.ndarray, yi_guess: np.ndarray | None = None, xi_guess: np.ndarray | None = None
+    grid: XGrid, y: np.ndarray, x: np.ndarray, yi: np.ndarray | None = None, xi: np.ndarray | None = None
 ):
-    yi_guess = np.array(yi_guess)
-    xi_guess = np.array(xi_guess)
-    xi = np.full(len(x), GRID_SEARCH_ERROR, dtype=np.int32)
-    yi = np.full(len(y), GRID_SEARCH_ERROR, dtype=np.int32)
-    if np.any(xi_guess):
+    if np.any(xi):
         # If an initial guess is provided, we first perform a point in cell check for all guessed indices
-        is_in_cell, coords = curvilinear_point_in_cell(grid, y, x, yi_guess, xi_guess)
+        is_in_cell, coords = curvilinear_point_in_cell(grid, y, x, yi, xi)
         y_check = y[is_in_cell == 0]
         x_check = x[is_in_cell == 0]
         zero_indices = np.where(is_in_cell == 0)[0]
     else:
         # Otherwise, we need to check all points
+        yi = np.full(len(y), GRID_SEARCH_ERROR, dtype=np.int32)
+        xi = np.full(len(x), GRID_SEARCH_ERROR, dtype=np.int32)
         y_check = y
         x_check = x
         coords = -1.0 * np.ones((len(y), 2), dtype=np.float32)
--- a/parcels/field.py
+++ b/parcels/field.py
@@ -212,13 +212,14 @@ class Field:
         conversion to the result. Note that we defer to
         scipy.interpolate to perform spatial interpolation.
         """
-        # if particle is None:
-        _ei = None
-        # else:
-        #    _ei = particle.ei[self.igrid]
+        if particle is None:
+            _ei = None
+        else:
+            _ei = particle.ei[:, self.igrid]
 
         tau, ti = _search_time_index(self, time)
         position = self.grid.search(z, y, x, ei=_ei)
+        _update_particles_ei(particle, position, self)
         _update_particle_states_position(particle, position)
 
         value = self._interp_method(self, ti, position, tau, time, z, y, x)
@@ -251,6 +252,7 @@ class VectorField:
         self.V = V
         self.W = W
         self.grid = U.grid
+        self.igrid = U.igrid
 
         if W is None:
             _assert_same_time_interval((U, V))
@@ -294,13 +296,14 @@ class VectorField:
         conversion to the result. Note that we defer to
         scipy.interpolate to perform spatial interpolation.
         """
-        # if particle is None:
-        _ei = None
-        # else:
-        #    _ei = particle.ei[self.igrid]
+        if particle is None:
+            _ei = None
+        else:
+            _ei = particle.ei[:, self.igrid]
 
         tau, ti = _search_time_index(self.U, time)
         position = self.grid.search(z, y, x, ei=_ei)
+        _update_particles_ei(particle, position, self)
         _update_particle_states_position(particle, position)
 
         if self._vector_interp_method is None:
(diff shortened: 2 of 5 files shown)


erikvansebille:
@fluidnumerics-joe, there's a failing unit test in the uxgrid part: 
```
FAILED tests/v4/test_particleset_execute.py::test_uxstommelgyre_pset_execute - KeyError: 'FACE'
```
with detailed error
```
parcels/particleset.py:543: in execute
    self._kernel.execute(self, endtime=next_time, dt=dt)
parcels/kernel.py:261: in execute
    f(pset[evaluate_particles], self._fieldset)
parcels/application_kernels/advection.py:103: in AdvectionEE
    (u1, v1) = fieldset.UV[particles]
               ^^^^^^^^^^^^^^^^^^^^^^
parcels/field.py:338: in __getitem__
    return self.eval(key.time, key.depth, key.lat, key.lon, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
parcels/field.py:305: in eval
    position = self.grid.search(z, y, x, ei=_ei)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
parcels/uxgrid.py:91: in search
    bcoords = try_face(fi)
              ^^^^^^^^^^^^
parcels/uxgrid.py:77: in try_face
    bcoords, err = self._get_barycentric_coordinates_latlon(y, x, fid)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
parcels/uxgrid.py:117: in _get_barycentric_coordinates_latlon
    n_nodes = self.uxgrid.n_nodes_per_face[fi].to_numpy()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/parcels/lib/python3.13/site-packages/xarray/core/dataarray.py:902: in __getitem__
    return self._getitem_coord(key)
           ^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/parcels/lib/python3.13/site-packages/xarray/core/dataarray.py:896: in _getitem_coord
    _, key, var = _get_virtual_variable(self._coords, key, dim_sizes)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Probably has to do with `basegrid.ravel` and `.unravel` now expecting an ndarray (since they are vectorized)

Can you see if you can fix this for uxgrid?

tests/v4/test_particleset_execute.py
```diff
@@ -400,8 +400,8 @@ def test_uxstommelgyre_pset_execute():
         dt=np.timedelta64(60, "s"),
         pyfunc=AdvectionEE,
     )
-    assert utils.round_and_hash_float_array([p.lon for p in pset]) == 1165396086
-    assert utils.round_and_hash_float_array([p.lat for p in pset]) == 1142124776
+    assert utils.round_and_hash_float_array([p.lon for p in pset]) == 1165397121
+    assert utils.round_and_hash_float_array([p.lat for p in pset]) == 1142123780
```
VeckoTheGecko, line 404: Does it make sense that the output changes alongside this change?

parcels/field.py
```diff
@@ -251,6 +252,7 @@ def __init__(
         self.V = V
         self.W = W
         self.grid = U.grid
+        self.igrid = U.igrid
```
VeckoTheGecko, line 255: `igrid` is not being updated in the code - so all the indexing is done on the last grid (perhaps leading to the change in hash that we saw elsewhere?) Should we resolve that in another PR?

Changes required:
```
FieldSet.__init__
   - make the field addition use `add_field` instead of plainly constructing the 
     dict of fields
FieldSet.add_field()
   - update fieldset.fields
   - update fieldset.gridset (also make this a normal variable instead of a property)
   - update field.igrid indexing the grid position in the updated gridset

```

tests/v4/test_basegrid.py
```diff
@@ -35,7 +35,7 @@ def get_axis_dim(self, axis: str) -> int:
 def test_basegrid_ravel_unravel_index(grid):
     axes = grid.axes
     dimensionalities = (grid.get_axis_dim(axis) for axis in axes)
-    all_possible_axis_indices = itertools.product(*[range(dim) for dim in dimensionalities])
+    all_possible_axis_indices = itertools.product(*[np.arange(dim)[:, np.newaxis] for dim in dimensionalities])
```
VeckoTheGecko, line 38: Added 9826c9ae644c4541ca18d2dc9c1f63c51a2f189e (since many of the original changes weren't required)

VeckoTheGecko (approved):
Looks good! Some minor clarifications but otherwise LGTM

parcels/field.py
erikvansebille: Ah, I didn't realise that `grid` still required some plumbing. Now created #2221 to track progress

tests/v4/test_basegrid.py
erikvansebille: Hmm, I intentionally made these changes because the `position` dictionary now has lists of indices, rather than individual integers. 

So I very intentionally made the change to test_basegrid.py; to check whether an input like `TestGrid({"Z": 10, "Y": 20, "X": 30})` would work. I'd prefer to keep testing that particular set-up therefore

parcels/_index_search.py
```diff
@@ -75,20 +75,18 @@ def curvilinear_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray
 
 
 def _search_indices_curvilinear_2d(
-    grid: XGrid, y: np.ndarray, x: np.ndarray, yi_guess: np.ndarray | None = None, xi_guess: np.ndarray | None = None
+    grid: XGrid, y: np.ndarray, x: np.ndarray, yi: np.ndarray | None = None, xi: np.ndarray | None = None
```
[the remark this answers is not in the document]
VeckoTheGecko: `xi_previous`?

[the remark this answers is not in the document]
erikvansebille: Could be an option.. But why not simply `xi`? If we don't call it `xi`, then the first thing we'll have to do in the method is `xi = xi_previous`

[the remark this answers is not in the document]
VeckoTheGecko: > * Particle positions - which we call `lon`, `lat` and `depth` (even though they could also be on flat grid, let's call them always the same)
> * Grid locations - which I propose we consistently call `i`, `j` and `k` following e.g. Nemo convention. These may or may not be in the x, y and z directions, depending on whether the grid is rotated.
> 
> To avoid this confusion, perhaps best to avoid the symbols/variables `x` and `y` as much as we can throughout the codebase?

I think this sounds good, and I think its imporant to track. Just opened #2223

tests/v4/test_particleset_execute.py
VeckoTheGecko: All good. I don't think that this is something worth worrying about

PR #2218, opened 2025-09-15
Feature/uxgrid morton hashing

fluidnumericsJoe:


fluidnumerics-joe:
I was able to track down the bad elements in the antimeridian test case that were causing an insanely large number of hash table entries ( O(10^6) ). The boundary mask needed to be corrected. I also adjusted this `_fesom2_square_delaunay_antimeridian` domain boundary to be a bit smaller to keep the memory footprint manageable for unit testing. This one should be good to go, pending a review.

PR #2220, opened 2025-09-16
Make xgcm a dependency and remove vendored code

### LINKED ISSUE #2193: Make xgcm a dependency and remove vendored code
https://github.com/OceanParcels/Parcels/pull/1966#issuecomment-3208495531 , and my conversations with @jbusecke highlight that we can make xgcm a normal dependency without having to vendor code.

Currently we do access `xgcm.Grid._ds` within our code - so that is the only remaining concern I have. We need access to the grid values as well as the staggered grid information.

We can store a copy of the `ds` outside of the Xgcm grid class so that we aren't relying on 3rd party private API - I think that's the best way forward.

VeckoTheGecko:
Fixes #2193


erikvansebille (approved):
Good that we can use xgcm as a dependency now!

PR #2238, opened 2025-09-25
Update from_copernicusmarine ingestion

### LINKED ISSUE #2236: `FieldSet.from_copernicusmarine()` and non-advection kernels
cc @erikvansebille 

I've started experimenting with the new `from_copernicusmarine()` functionality in v4-dev (#2230) and realised that in its current implementation it does not support non-advection kernels. 

To elaborate, `from_copernicusmarine()`detects U and V variables. However, in use cases such as a [CTD_BGC instrument](https://github.com/OceanParcels/virtualship/blob/43626e657f827f0e9341e8c76674e3dcc33c6192/src/virtualship/instruments/ctd_bgc.py#L161) in VirtualShip there are no advection kernels. This raises the potential issue that `from_copernicusmarine()` is assuming that all users will be advecting their particles. Whereas it's possible that this won't always be the case?

@VeckoTheGecko 's suggestion is that the method is renamed to something like `from_copernicusmarine_currents()` to be explicit about what this specific function does. 

Alternatively, is there scope to handle non-advection scenarios in `from_copernicusmarine()` or a new method? I'd be happy to help out with this but also appreciate there could be more pressing jobs in the v4 pipeline, especially considering this could be solved for now by simply renaming the method.

### LINKED ISSUE #2237: Update `.from_copernicusmarine` to parse vertical velocities
> Just realise: we should also add `W` parsing

I'll deal with this separately next. Merging for now so people can test with horizontal velocities

_Originally posted by @VeckoTheGecko in https://github.com/OceanParcels/Parcels/issues/2230#issuecomment-3328327226_

VeckoTheGecko:
- Update `from_copernicusmarine` to parse in W fields from CF naming - fixes #2237

- Update `from_copernicusmarine` allow no currents (no log messages are emitted here) - fixes #2236

Note that if there is a W the UV field _isn't_ added - is this in line with v3 @erikvansebille ?


VeckoTheGecko:
> ~I want to do that now separately alongside a bunch of other runtime warnings we're getting~ Actually, I can do it here - wont take long

I think we only see these in the tests actually - let's merge this

PR #2259, opened 2025-09-30
Fix upload of Conda package to prefix

VeckoTheGecko:
Take 2

Continuing #2258


VeckoTheGecko:
@erikvansebille @fluidnumerics-joe v4 alpha of Parcels has been released
image

(you might need to push to your respective notebook repos for Binder to invalidate the cache and build a new image)

PR #2265, opened 2025-10-01
Link github account creation

willirath:
This should make it as easy as possible to create a Github account.


VeckoTheGecko:
Can you update the head branch to v4-dev?

PR #2282, opened 2025-10-02
Insert integration time in the progress bar

### LINKED ISSUE #2149: Use datetime64/timedelta64 for progressbar
The current implementation of the tqdm progress bar in `v4-dev` converts the time to seconds since start-time. 

https://github.com/OceanParcels/Parcels/blob/add8158bd4ef6bd0c4055e7677f8f2455e1b1fe2/parcels/particleset.py#L564-L565

https://github.com/OceanParcels/Parcels/blob/add8158bd4ef6bd0c4055e7677f8f2455e1b1fe2/parcels/particleset.py#L587-L588

It would be neat to have the progress displayed as actual time units (so without the conversion)

nilodna:
Updating progress bar description to show the `integration time. fixes #2149

### DIFF
--- a/parcels/_core/particleset.py
+++ b/parcels/_core/particleset.py
@@ -566,6 +566,7 @@ class ParticleSet:
                         next_output += outputdt
 
             if verbose_progress:
+                pbar.set_description("Integration time: " + str(time))
                 pbar.update((next_time - time) / np.timedelta64(1, "s"))
 
             time = next_time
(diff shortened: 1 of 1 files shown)


VeckoTheGecko (commented):
Just wondering why there is the need for the isinstance checks here as well as the second statement?

Is the second statement because this doesn't show on the first loop?

PR #2284, opened 2025-10-02
Bugfix/2260 vectorize pwlinear node

### LINKED ISSUE #2260: Vectorize the UXPiecewiseLinearNode interpolator
When working with multiple particles and the UXPiecewiseLinearNode interpolator, we currently fail with the following message :

```
--> 660 fzk = np.sum(field.data.values[ti, k, node_ids] * bcoords, axis=-1)
    661 fzkp1 = np.sum(field.data.values[ti, k + 1, node_ids] * bcoords, axis=-1)
    663 # Then, do piecewise linear interpolation in the vertical direction

IndexError: shape mismatch: indexing arrays could not be broadcast together with shapes (100,) (100,) (100,3)
```

This issue can be minimally reproduced with 2 particles and a field that uses the UXPiecewiseLinearNode interpolator

fluidnumericsJoe:
This PR fixes #2260 and adds a test that reproduces the issue reported #2260


tests/test_particleset_execute.py
```diff
@@ -479,6 +479,50 @@ def test_uxstommelgyre_pset_execute():
...
+        interp_method=UXPiecewiseLinearNode,
+    )
+    P = Field(
+        name="P",
+        data=ds.p,
+        grid=grid,
+        interp_method=UXPiecewiseConstantFace,
+    )
+    UVW = VectorField(name="UVW", U=U, V=V, W=W)
+    fieldset = FieldSet([UVW, UVW.U, UVW.V, UVW.W, P])
```
VeckoTheGecko, line 510: Should we consider creating a `fieldset_unstructured` fixture?

PR #2293, opened 2025-10-02
Add open binder and download link pre-amble (remove zip tooling)

### LINKED ISSUE #2206: Add download/open in Binder pre-amble for notebooks
If you want local copies of the notebooks, you can currently download them as a zip of them from our tutorials homepage https://docs.oceanparcels.org/en/latest/documentation/index.html . However:

- its unusual that someone would want to download all the notebooks (they probably just want to look at one of particular interest)
- this creating of the zip creates unnecessary complexity in the build process

---

A better option would be to implement something similar to what xarray has on their gallery page - an autogenerated pre-amble ([specified in `conf.py`](https://github.com/pydata/xarray/blob/699d8957ec174f118108005aeb6ba99c1920167a/doc/conf.py#L102-L110)) of

Image

Here these are just links (no added complexity to the build process) and has the added benefit of providing a link directly to open in Binder and try the code.

---
Needed changes:
- Add pre-amble (Binder link, and "download from GitHub" link)
  - Note: the link should not just point to main - but should depend on the version being built. This prevents downloading the latest notebook when an old version of docs is being looked at. Create a function `_get_github_ref_for_build()` which returns a GitHub reference/tag/branch name to be interpolated into the string (if "unknown" - e.g., running local - use `main`). Inspecting `READTHEDOCS_VERSION_TYPE` and `READTHEDOCS_VERSION_NAME` environment variables ([docs](https://docs.readthedocs.com/platform/stable/reference/environment-variables.html)) will help here.
- Remove zip building workflow

nilodna:
Add at the top of all jupyter notebook examples page a button to binder or a direct link to the source code fixes #2206 

EDIT VeckoTheGecko: This PR also removes the tooling for zipping up notebook examples, which is no longer needed.


erikvansebille:
Thanks for this work, @nilodna. Nice! 

Do you know if it's possible to add a bit larger vspace between the binder badge and the title of the Notebook? They're very close together now (see screenshot below)
Screenshot 2025-10-03 at 07 34 07

nilodna:
The code is a HTML snippet, so it should be easy @erikvansebille. I can do this today during the presentations.

docs/conf.py
```diff
@@ -384,6 +384,23 @@ def linkcode_resolve(domain, info):
 nbsphinx_execute = "never"
 # -- Options for LaTeX output ---------------------------------------------
 
+# -- Detect branch name -----------------------------------
+# Works on GitHub Actions, ReadTheDocs, and local fallback
+BRANCH = (
+    os.environ.get("GITHUB_REF_NAME")  # GitHub Actions
```
VeckoTheGecko, line 390: I'm not sure that this is reliable actually - since a lot of contributions come from forks.

Also the link it generates atm is https://github.com/OceanParcels/Parcels/blob/2293/docs/examples/tutorial_unitconverters.ipynb (using the PR number instead). Perhaps we can just remove this?

docs/conf.py
```diff
@@ -384,6 +384,23 @@ def linkcode_resolve(domain, info):
 nbsphinx_execute = "never"
 # -- Options for LaTeX output ---------------------------------------------
 
+# -- Detect branch name -----------------------------------
+# Works on GitHub Actions, ReadTheDocs, and local fallback
```
VeckoTheGecko, line 388: We can remove these comments I think, code is readable as is

VeckoTheGecko (changes requested):

docs/conf.py
nilodna: I got your point, @VeckoTheGecko. I'm still trying to figure out how this works on Github (locally is fine, targeting the documentation to the main version).

VeckoTheGecko: Let's chat f2f in the break :)

docs/conf.py
```diff
@@ -384,6 +384,23 @@ def linkcode_resolve(domain, info):
...
+    os.environ.get("GITHUB_REF_NAME")  # GitHub Actions
+    or os.environ.get("READTHEDOCS_VERSION")  # ReadTheDocs
+    or "main"  # fallback
+)
+
+nbsphinx_prolog = f"""
+.. raw:: html
+
+    <a href="https://mybinder.org/v2/gh/OceanParcels/Parcels/{BRANCH}?urlpath=lab/tree/docs/{{{{  env.doc2path(env.docname, base=None)  }}}}" target="_blank"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg"></a>
+    or view it <a href="https://github.com/OceanParcels/Parcels/blob/{BRANCH}/docs/{{{{  env.doc2path(env.docname, base=None)  }}}}" target="_blank">on GitHub</a>
```
VeckoTheGecko, line 399:
```suggestion
    Run this notebook in the cloud <a href="https://mybinder.org/v2/gh/OceanParcels/Parcels/{BRANCH}?urlpath=lab/tree/docs/{{{{  env.doc2path(env.docname, base=None)  }}}}" target="_blank"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg"></a>
    , or view it <a href="https://github.com/OceanParcels/Parcels/blob/{BRANCH}/docs/{{{{  env.doc2path(env.docname, base=None)  }}}}" target="_blank">on GitHub</a>
```

For those that don't know what binder is ;)

VeckoTheGecko (approved):
Final screenshot:

Image

Thanks @nilodna ! This is great 🚀

PR #2296, opened 2025-10-03
Update issue templates with "needs-triage" label

VeckoTheGecko:
This flags all issues that come in as "needs-triage" - this is to help us be more dilligent with issue labelling as things come in (making it an intentional step to remove the needs triage label, and add other labels to issues).

Thoughts @erikvansebille ?


.github/ISSUE_TEMPLATE/02_bug.yaml
```diff
@@ -1,6 +1,6 @@
 name: "🐛 Bug Report"
 description: Describe a bug in the parcels codebase. For support related to your specific use case, please post in Q&A in our Discussions tab.
-labels: ["bug"]
+labels: ["bug", "needs-triage"]
```
erikvansebille, line 3: Do we need to explain what we mean with "triage"? Not sure all our users would know the concept.

PR #2297, opened 2025-10-03
Update logo from OceanParcels to Parcels in documentation

VeckoTheGecko:
Followup on https://github.com/OceanParcels/oceanparcels_website/pull/215 for the docs


VeckoTheGecko:
## Screenshot as is

**desktop**

image

**ipad**

image


## Screenshot with 100px buffer on all sides:

**desktop**
image



**ipad**

image


---

I think with the 100px buffer looks better

Also I think we should buffer the logo on the oceanparcels website

**desktop before buffer**

image

erikvansebille (approved):

VeckoTheGecko:
> Also I think we should buffer the logo on the oceanparcels website (cc @erikvansebille - I realised there are a bunch of other assets, svg etc - so perhaps you can add a rough buffer on your end?)

Should we put this in an issue in the parcels-code.org repo so that we can track?

VeckoTheGecko:
i realised that the old logo was an svg, and I realised I put a PNG as the logo instead. If the resolution of the image looks grainy we can switch back to svg

PR #2301, opened 2025-10-04
Test against minimum dependencies, remove old deps

### LINKED ISSUE #2286: Test against minimum versions for dependencies
From #2205

> We don't have much control over the versions of dependencies installed in this environment, with it just defaulting to the latest versions (we don't test against minimum supported versions of other core packages (xarray, zarr, uxarray). This means that our lower pinning in our environment.yml file as well as our feedstock is likely out of date).

It would be good to fix the lower pinning of our dependencies and test against this in CI. Note that we follow [SPEC0](https://scientific-python.org/specs/spec-0000/) in terms of the packages that we support.

This would look like the following in `pixi.toml`, with some comments re. implementation (no need to have those comments in the final thing):
```toml
[environments]
test-minimal = { features = ["test", "minimal"], solve-group = "test" }

[dependencies]
# versions here will be manually walked up according to SPEC0 when they're noticed to be out of date
# or will be bumped to make sure that we have new features only available in newer releases.
# Once these are bumped, the minimum should also be bumped (but that will be clear anyway - 
# since the pixi lock process will informatively fail anyway)
numpy = ">=1.9.1"
xarray = ">=0.10.8" 

[feature.minimum.dependencies]
# here we pin to the lowest versions for our default dependencies
python = "==3.11"
numpy = "==1.9.1"
xarray = "==0.10.8" 

```

Also need to update CI matrix with the `test minimal` dep

---

If you want to dip your feet into Pixi @andrew-s28 , this would be a good issue ;)

### LINKED ISSUE #2292: Remove SciPy as a core dependency?
Based on a helpful comment left in [pixi.toml](https://github.com/OceanParcels/Parcels/blob/v4-dev/pixi.toml#L40) and a quick search of the code base, I don't think that Parcels relies on SciPy anywhere internally anymore. 

The only place I could find `scipy` imported is in the docs (specifically in [tutorial_peninsula_AvsCgrid.ipynb](https://github.com/OceanParcels/Parcels/blob/v4-dev/docs/examples/tutorial_peninsula_AvsCgrid.ipynb)), so perhaps it makes sense to just include SciPy as a docs dependency (taking advantage of pixi for this 😄).

Somewhat related to #2286.

### LINKED ISSUE #2304: Remove sklearn as core dependency
Similar to https://github.com/OceanParcels/Parcels/issues/2292, I don't think we depend anywhere on sklearn in v4 of Parcels. It would be good for us to remove it from the dendencies.

andrew-s28:
Fixes #2286, fixes #2292, fixes #2304

Finding the minimum dependencies that pass the tests was a little bit tricky. It turns out that, since the dependencies aren't pinned for the Python 3.11 tests, anything below NumPy 2.3.3 isn't currently being tested. 

When I tried to use NumPy 1.x (even the most recent 1.26.4), the tests fail with 109 errors. Most of these seem to be caused by changes in how NumPy handles casting array values, which I think changed with NumPy 2.x (but I couldn't find a super defnitive PR over in the NumPy repo). However, even at NumPy 2.0.x the tests fail with 4 errors, still caused by some minor casting issues. Finally, using NumPy 2.1.0, all the tests pass. 

Pinning NumPy to 2.1.0 is above the minimum shown in [SPEC0](https://scientific-python.org/specs/spec-0000/), and forces several other dependencies to have more recent minimum versions. Currently I have this as a draft PR to determine how best to handle this. For example, if these are the minimum versions that pass tests, should the minimums in the core dependencies section also be updated? Should the tests be modified?

Minimum reproducible examples are shown in the drop downs below. Environment details for `test-py311` and `test-minimum` can be found in the attached files (or by running `pixi list -e test-py311` and `pixi list -e test-minimum`).

[test-minimum.txt](https://github.com/user-attachments/files/22691404/test-minimum.txt)
[test-py311.txt](https://github.com/user-attachments/files/22691405/test-py311.txt)

<details>
  <summary>Minimum Reproducible Example</summary>

    # test_dates.py
  
    import numpy as np
    from cftime import datetime as cftime_datetime
  
    cf = cftime_datetime(1900, 1, 1, calendar="365_day")
    td = np.timedelta64(86400, "s")
  
    print(all(np.array([td]) >= 0))  # errors for NumPy<=1.26.4
    print(cf + td)  # errors for NumPy>=2.0.0,<2.1.0

  I used `uv` to run the test script, which was easier for quick demonstration purposes:

    $ uv run --with 'numpy==1.26.4,cftime' test_dates.py
    $ uv run --with 'numpy==2.0.2,cftime' test_dates.py

  You should see them each fail on one of the two print lines.
</details>
<details>
  <summary>Running Test Suite with Different NumPy Versions</summary>
  In order to reproduce the tests failing, you can replace all of the dependencies under the <code>[feature.minimum.dependencies]</code> section with either <code>numpy = "==1.26.4"</code> or <code>numpy = "==2.0.2"</code>, then run the tests as usual:

    $ pixi run -e test-minimum tests

</details>

EDIT VeckoTheGecko: Using `fixes #issue_number` so that issues are automatically closed on merge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,7 @@ src/parcels/_version_setup.py
 *.egg-info
 
 # Ignore pixi.lock file for this project. The con of 22k lines of noise it adds to diffs is not worth
-# the benefit of perfectly reproducible environments for all developers (and all the tooling that would
+# the minor benefit of perfectly reproducible environments for all developers (and all the tooling that would
 # be required to support that - see https://github.com/pydata/xarray/issues/10732#issuecomment-3327780806
 # for more details)
 pixi.lock


VeckoTheGecko:
> Pinning NumPy to 2.1.0 is above the minimum shown in [SPEC0](https://scientific-python.org/specs/spec-0000/)

Let's pin numpy >2.1.0 and also have the other packages be more recent.

Let's scope this PR to not touch code, and just focus on the environments and CI. We can discuss/explore policy or adherance to SPEC0 in another issues/PRs (+ I would like some time to discuss with other maintainers to see how they handle SPEC0 :) ).

Can we do the following as well in this PR?:
- #2304 (just made this)
- I think you accidentally deleted a notebook here, if we can revert that
- Can you update the `[dependencies]` section to have `>={{version as listed in minimum environment}}`
- Can you update `.github/ci/recipe.yaml` to match

andrew-s28:
@VeckoTheGecko Sounds good, thanks for your feedback.

The notebook was removed as per your comment in #2292, since this notebook was the only place remaining in Parcels with a SciPy dependency.

Can do on the rest!

VeckoTheGecko:
> The notebook was removed as per your comment in #2292, since this notebook was the only place remaining in Parcels with a SciPy dependency.

Ah woops! I meant completely remove the dependency rather than moving it to the notebook environment :)) . If you could leave the notebook that would be great

andrew-s28:
@VeckoTheGecko I think I've addressed all the comments now and things may be ready to go - let me know if there's anything missing!

Issue #2216, opened 2025-09-15
Restructure Parcels to use `src` layout


VeckoTheGecko:
xref #1965 

---

> We can avoid this completely by using a `src` layout. If we're restructuring soon anyway, might as well do that. 

 _Originally posted by @VeckoTheGecko in [#2214](https://github.com/OceanParcels/Parcels/issues/2214#issuecomment-3292613866)_

PR #2305, opened 2025-10-06
Update pixi.toml to install Parcels from source

### LINKED ISSUE #2272: Does pixi default env have an editable install of parcels?
### Parcels version

v4-dev

### Description

Curenltly, the pixi default env does not appear to have a working (editable or other) parcels package installed.

### Code sample

```shell
$ git clone https://github.com/OceanParcels/Parcels.git
$ cd Parcels
$ pixi install
$ pixi shell
(Paarcels) $ cd /tmp
(Paarcels) $ python -c "import parcels"
```

```text
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'parcels'
```

VeckoTheGecko:
Fixes #2272

Before:
```sh
pixi run -e test-latest 'cd tests && python -c "import parcels"'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'parcels'
```

After:
```
pixi run -e test-latest 'cd tests && python -c "import parcels"'

```

This enables #2216 - how should we time that @erikvansebille to reduce disruption? After #2303 is merged?

### DIFF
--- a/.github/workflows/cache-pixi-lock.yml
+++ b/.github/workflows/cache-pixi-lock.yml
@@ -29,7 +29,7 @@ jobs:
       - uses: prefix-dev/setup-pixi@v0.9.0
         if: ${{ !steps.restore.outputs.cache-hit }}
         with:
-          pixi-version: v0.49.0
+          pixi-version: v0.56.0
           run-install: false
       - name: Run pixi lock
         if: ${{ !steps.restore.outputs.cache-hit }}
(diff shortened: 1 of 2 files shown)


erikvansebille:
Looks good to me. Why should we wait until #2303? Is this change not completely perpendicular to that PR?

VeckoTheGecko:
This change is, but #2216 isn't - I was wondering about that

erikvansebille:
Maybe I don't understand what #2216 then is. Is it 'simply' a rename of the `parcels` directory to `src`? 

In that case, let's indeed do it after #2303. That PR is ready, except for the failing assert on the UX hash, which fails in many more tests (right?)

PR #2312, opened 2025-10-07
Assert that field names are not valid Python attributes

### LINKED ISSUE #2101: Assert that field names have to valid Python attributes
This enables us to consistently use attribute access such as `fieldset.U`

- use `assert_valid_python_varname` function already in `particle.py` (move out to `_core.utils.common`)
- Add call to this function in `VectorField`, and `Field` init, and into `Fieldset.add_constant`. Try and except, catching and rewriting the error message
- Add tests

nilodna:
Fixes #2101

### DIFF
--- a/parcels/_core/utils/string.py
+++ b/parcels/_core/utils/string.py
@@ -1,0 +1,6 @@
+from keyword import iskeyword, kwlist
+
+def _assert_valid_python_varname(name):
+    if name.isidentifier() and not iskeyword(name):
+        return
+    raise ValueError(f"Received invalid Python variable name {name!r}. Avoid using the following names: {", ".join(kwlist)}")
--- a/parcels/_core/field.py
+++ b/parcels/_core/field.py
@@ -19,6 +19,7 @@ from parcels._core.statuscodes import (
     StatusCode,
 )
 from parcels._core.utils.time import TimeInterval
+from parcels._core.utils.string import _assert_valid_python_varname
 from parcels._core.uxgrid import UxGrid
 from parcels._core.xgrid import XGrid, _transpose_xfield_data_to_tzyx
 from parcels._reprs import default_repr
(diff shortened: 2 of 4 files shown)


parcels/_core/utils/string.py
```diff
@@ -0,0 +1,9 @@
+from keyword import iskeyword, kwlist
+
+
+def _assert_valid_python_varname(name):
+    if name.isidentifier() and not iskeyword(name):
+        return
+    raise ValueError(
+        f"Received invalid Python variable name {name!r}. Avoid using the following names: {', '.join(kwlist)}"
+    )
```
VeckoTheGecko, line 9: I think including the kwlist error message in the output everytime is perhaps excessive - try the following?

```suggestion
def _assert_valid_python_varname(name):
    if name.isidentifier() and not iskeyword(name):
        return
    msg = f"Received invalid Python variable name {name!r}. "
    if name in kwlist:
        msg += f"HINT: Avoid using the following names: {', '.join(kwlist)}"
    raise ValueError(msg)
```

parcels/_core/utils/string.py
```diff
@@ -0,0 +1,9 @@
+from keyword import iskeyword, kwlist
+
+
+def _assert_valid_python_varname(name):
+    if name.isidentifier() and not iskeyword(name):
```
VeckoTheGecko, line 5: Could we move the string check into here as well to avoid duplication?


```suggestion
    if not isinstance(name, str):
        raise ValueError(f"Expected `name` to be a string, got {type(name)}.")
    if name.isidentifier() and not iskeyword(name):
```

VeckoTheGecko (changes requested):
Thanks for the PR! Comments below. Does the test (`test_variable_invalid_init`) at 

https://github.com/Parcels-code/Parcels/blob/3a05d61da96f0abb030bf5103888f14936c758e1/tests/test_particle.py#L30-L32

as well as the test `test_fieldset_add_constant` in `test_fieldset.py` show how we can make a test `test_fieldset_add_constant_invalid_name` just below for testing this?

Also, you can edit the test in `test_field_init_param_types` after the line 

```
with pytest.raises(ValueError, match="Expected `name` to be a string"):
```

Hope that helps!

src/parcels/_core/utils/string.py
```diff
@@ -0,0 +1,9 @@
+from keyword import iskeyword, kwlist
+
+
+def _assert_valid_python_varname(name):
+    if name.isidentifier() and not iskeyword(name):
```
nilodna: Sure! I'll do that.

Issue #2314, opened 2025-10-08
Add AdvectionRK2 and AdvectionRK2_3D


erikvansebille:
The 'default' Advection kernel in Parcels has long been a fourth-order Runge-Kutta integrator; but e.g. [Nordam and Duran (2020)](https://gmd.copernicus.org/articles/13/5935/2020/) have shown that this high order is often overkill in terms of accuracy; especially when users also add diffusion or mix the Advection with other, lower-order custom kernels.

In the Parcels event tutorial on [Making custom kernels](https://github.com/Parcels-code/10year-anniversary-session2), @michaeldenes showed nicely that a second-order Runge-Kutta method is often nearly as accurate, while requiring only half the amount of Field interpolations (two instead of four). 

<img width="816" height="310" alt="Image" src="https://github.com/user-attachments/assets/94ec4e05-598b-438b-88c3-d73d76edae8a" />

So I propose that we add an RK2 method to the default kernels, and use that for most of the tutorials for v4. 

@michaeldenes, could you add your code, and add it as tests in e.g. 
https://github.com/Parcels-code/Parcels/blob/deee3909d1b0159dc33a1b8ee2c80f5f32dc8f52/tests/test_advection.py#L258-L269
and similar test functions below?

erikvansebille:
Done in #2315

PR #2315, opened 2025-10-08
Adding in the Runge-Kutta 2 advection schemes

michaeldenes:
I've implemented the RK2 and RK2_3D schemes for v4. I've also added the two schemes to (some of) the unit tests, but I've not touched `test_nemo_3D_curvilinear_fieldset`.

This will solve #2314

### DIFF
--- a/src/parcels/kernels/__init__.py
+++ b/src/parcels/kernels/__init__.py
@@ -1,6 +1,8 @@
 from .advection import (
     AdvectionAnalytical,
     AdvectionEE,
+    AdvectionRK2,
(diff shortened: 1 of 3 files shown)


src/parcels/kernels/__init__.py
erikvansebille, line 1: @VeckoTheGecko, do we really need to add new Kernels at three locations? (two locations here in `__init__.py` and also at the top of `advection.py`). Can there not be less duplication?

erikvansebille (approved):

PR #2313, opened 2025-10-08
Renaming OceanParcels to Parcels-code throughout parcels repo

erikvansebille:
This PR renaming OceanParcels to Parcels-code throughout the parcels repo, and also removes the last references to the backcronym (Probably A Really Computationally Efficient Lagrangian Simulator) that we don't use anymore

PR #2316, opened 2025-10-08
[v3] Binder pre amble (port of #2293)

VeckoTheGecko:
Porting #2293 to v3 of Parcels (fixing the repo name in the binder links and link resolve)

Since we have all the redirects for the websites and the repos - I don't think its worth porting #2313 across to this repo as the only difference for users will be branding.


VeckoTheGecko:
Ready to review @erikvansebille  (note the binder links will only work post-merge due to how RTD builds for PRs)

erikvansebille:
Hmm, when I open one of the binders I get the error below. Is that expected?
Screenshot 2025-10-08 at 14 42 08

Issue #2309, opened 2025-10-07
Importing from .kernels and .interpolators namespaces


reint-fischer:
As of #2247 kernels shipped with parcels are no longer available to import using `from parcels import AdvectionRK4`, but instead using `from parcels.kernels import AdvectionRK4`. As discussed with @erikvansebille and @VeckoTheGecko, we like this as a feature, making it explicit for the users, and can do the same for interpolators. On the other hand it would add a step to the migration from v3 to v4.

@VeckoTheGecko you also wrote some style guidelines in #2247, but I was unclear whether you had finished these bullets or I am not understanding them correctly:

- > When writing package code, we import directly from the file (i.e., prefer from parcels._core.field import Field over from parcels import Field). This is to make clear the import path

- > When writing testing code, prefer to import from the parcels namespace for public API (of course, import from

- > The parcels and parcels/kernels are the only __init__.py files we need to currate. We don't need to specify more __init__.py files 

If we agree this should be the syntax for users, I can start writing this into the notebooks I will work on, edit this in any tests, and include it in the migration guide.

VeckoTheGecko:
> If we agree this should be the syntax for users

No, this is only for developers. Users only use the public API. Or am I misunderstanding?

I think it would be good to track those style conventions in dev-notes through

reint-fischer:
I guess that was my original question: right now `from parcels import AdvectionRK4` breaks, with `from parcels.kernels import AdvectionRK4` working.

PR #2317, opened 2025-10-08
Add Field sampling tutorial v4

reint-fischer:
I have rewritten the Field sampling tutorial to work with v4 and added it back into the documentation. To do so I needed to change the FieldSet from Peninsula.nc to the Copernicus Marine dataset, which also contains a temperature field. I have also adapted the tutorials to import the kernels using `parcels.kernels` as per #2309.

There was some unexpected behavior from the delayed particle release, with not all initial values being sampled, which I plan to look into in conjunction with the repeatdt tutorial.


VeckoTheGecko (approved):
looks good :)

PR #2319, opened 2025-10-08
Add method to import from fesom2

fluidnumericsJoe:
This PR adds a method to create a fieldset from a ` UXArray.UxDataset`. This new `fieldset.from_fesom2` method follows similar implementation patterns in comparison to `from_copernicusmarine`. Standard variable names for the velocity fields are renamed to `U`, `V`, and `W`, and the presence (or lack of) of `W` is used to determine whether to create `UV` or `UVW` vector field.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
         types: [text]
         files: \.(json|ipynb)$
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.2
+    rev: v0.14.4
     hooks:
       - id: ruff
         name: ruff lint (.py)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,10 @@
+## GitHub Interaction Guidelines
+
+- **NEVER impersonate the user on GitHub**, always sign off with something like
+  "[This is Claude Code on behalf of Jane Doe]"
+- Never create issues nor pull requests on the GitHub repository unless
+  explicitly instructed
+- Never post "update" messages, progress reports, or explanatory comments on
+  GitHub issues/PRs unless specifically instructed
+- When creating commits, always include a co-authorship trailer:
+  `Co-authored-by: Claude <noreply@anthropic.com>`


src/parcels/_core/field.py
```diff
@@ -483,3 +487,34 @@ def _get_positions(field: Field, time, z, y, x, particles, _ei) -> tuple[dict, d
...
+    supported_uxinterp_mapping = {
+        # (nz1, nface): face-center laterally, layer centers vertically — piecewise constant
+        "nz1,n_face": UXPiecewiseConstantFace,
+        # (nz, nnode): node/corner laterally, layer interfaces vertically — barycentric lateral & linear vertical
+        "nz,n_node": UXPiecewiseLinearNode,  # <- replace with your real interpolator
+    }
+    # Extract only spatial dimensions, neglecting time
+    da_spatial_dims = tuple(d for d in da.dims if d not in ("time",))
+    if len(da_spatial_dims) < 2:
+        return None
```
erikvansebille: Should this not give an error, rather than no interpolator at all? Or at least a warning?

src/parcels/_core/field.py
```diff
@@ -483,3 +487,34 @@ def _get_positions(field: Field, time, z, y, x, particles, _ei) -> tuple[dict, d
     _update_particles_ei(particles, grid_positions, field)
     _update_particle_states_position(particles, grid_positions)
     return particle_positions, grid_positions
+
+
+def _select_uxinterpolator(da: ux.UxDataArray):
```
erikvansebille: I can also imagine this function to live in FieldSet, where we parse all the metadata and find out whether the dataset is CF-compliant

src/parcels/_core/fieldset.py
```diff
@@ -257,6 +259,47 @@ def from_copernicusmarine(ds: xr.Dataset):
...
+        if not all(dim in ds_dims for dim in ["time", "nz", "nz1"]):
+            raise ValueError(
+                f"Dataset missing one of the required dimensions 'time', 'nz', or 'nz1'. Found dimensions {ds_dims}"
+            )
+        grid = UxGrid(ds.uxgrid, z=ds.coords["nz"])
+        ds = _discover_fesom2_U_and_V(ds)
+
+        fields = {}
+        if "U" in ds.data_vars and "V" in ds.data_vars:
+            # TO DO: Assign interpolator based on dimensions
```
erikvansebille: Indeed, I would set the Interpolator here (rather than in Field)

fluidnumerics-joe: Removed this comment in https://github.com/Parcels-code/Parcels/pull/2319/commits/1261d2390c054b03a7f61f2cd3cc5b0c19ff033e

src/parcels/_core/field.py
fluidnumerics-joe: This is now set in `fieldset.py` and the default interpolator selection in `field.py` has been removed. This also required updating all tests to explicitly pass the `interp_method`.

fluidnumerics-joe:
@VeckoTheGecko - I've noticed that the tests for mac environments hangs in the queue forever.. Is this something we need to wait on before merging or is there something wrong ?

PR #2333, opened 2025-10-14
Removing time_nextloop from Kernel loop

### LINKED ISSUE #2331: output time incorrect
### Parcels version

v4-dev

### Description

When I initialise a ParticleSet at the start of the FieldSet.time_interval, with a ParticleSet.dt of 5 minutes and an outputdt of 10 minutes, and a runtime of 30 minutes (dtype=np.timedelta64), the simulation is run for 25 minutes and the output is written at 5, 15, and 25 minutes from the start of the simulation. I would expect it to run for 30 minutes, with output written at 0, 10, 20, and 30 minutes.

I believe part of the issue is in the time loop of `particleset.execute()`, in line 565:
```
output_file.write(self, next_output)
```
[In v3.1.4 the equivalent](https://github.com/Parcels-code/Parcels/blob/fb88390d2bc9e2d45cd3056c601c7acaa9afcdac/parcels/particleset.py#L1218C25-L1218C69) is:

```
output_file.write(self, time_at_startofloop)
```
As I understand it this means that outputs are no longer written for the initial timestep. I have not familiarized myself enough with the full new time loop to say whether changing `next_output` would completely solve this issue. This would also not resolve the runtime being 25 instead of 30 minutes I assume.

### Code sample

```python
# Particle locations and initial time
npart = 10  # number of particles to be released
lon = 32 * np.ones(npart)
lat = np.linspace(-32.5, -30.5, npart, dtype=np.float32)
time = np.repeat(
    ds.time.values[0], npart
)  # release all particles at the start time of the fieldset
depth = np.repeat(ds.depth.values[0], npart)

pset = parcels.ParticleSet(
    fieldset=fieldset, pclass=SampleParticle, lon=lon, lat=lat, time=time, depth=depth
)

output_file = parcels.ParticleFile("SampleTemp.zarr", outputdt=np.timedelta64(10,'m'))

pset.execute(
    [parcels.kernels.AdvectionRK4],
    runtime=np.timedelta64(30,'m'),
    dt=np.timedelta64(5,'m'),
    output_file=output_file,
)

Particle_data = xr.open_zarr("SampleTemp.zarr", decode_times=False)

Particle_data.time[0,:].values
```

```text
array([ 300,  900, 1500], dtype=uint64)
```

reint-fischer:
This PR fixes the writing of variables in the Kernel loop, and thereby fixes #2331

### DIFF
--- a/src/parcels/_core/field.py
+++ b/src/parcels/_core/field.py
@@ -316,19 +316,17 @@ class VectorField:
                 w = self.W._interp_method(self.W, ti, position, tau, time, z, y, x)
             else:
                 w = 0.0
-
-            if applyConversion:
-                u = self.U.units.to_target(u, z, y, x)
-                v = self.V.units.to_target(v, z, y, x)
-
         else:
-            (u, v, w) = self._vector_interp_method(self, ti, position, tau, time, z, y, x, applyConversion)
+            (u, v, w) = self._vector_interp_method(self, ti, position, tau, time, z, y, x)
+
+        if applyConversion:
+            u = self.U.units.to_target(u, z, y, x)
+            v = self.V.units.to_target(v, z, y, x)
+            if "3D" in self.vector_type:
+                w = self.W.units.to_target(w, z, y, x)
 
         for vel in (u, v, w):
             _update_particle_states_interp_value(particles, vel)
-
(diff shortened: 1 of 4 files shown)


VeckoTheGecko (approved):
LGTM

PR #2348, opened 2025-10-24
Fixes/tests from virtualship development at EDITO hackathon

erikvansebille:
This PR has some extra unit tests, fixes and migration guide updates that were found during the EDITO hackathon work on VirtualShip. Most important ones

- Give an error when `Grid.depth` is not monotonically increasing
- Fix breaking index_search for particles that were exactly on the right edge

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build conda package
-        uses: prefix-dev/rattler-build-action@v0.2.19
+        uses: prefix-dev/rattler-build-action@v0.2.34
         with:
           recipe-path: .github/ci/recipe.yaml
 


src/parcels/_core/index_search.py
```diff
@@ -44,7 +44,7 @@ def _search_1d_array(
     # TODO v4: We probably rework this to deal with 0D arrays before this point (as we already know field dimensionality)
     if len(arr) < 2:
         return np.zeros(shape=x.shape, dtype=np.int32), np.zeros_like(x)
-    index = np.searchsorted(arr, x, side="right") - 1
+    index = np.clip(np.searchsorted(arr, x, side="right") - 1, 0, len(arr) - 2)
```
VeckoTheGecko: I'm not 100% sure why this clipping is introduced. Is this related to out of bound searches?

PR #2363, opened 2025-11-06
Support for 1D fields without lon and/or lat

erikvansebille:
This PR adds support for Fields that don't have longitude and/or latitude dimensions. A relevant example would be 1-dimensional depth profiles. 

This PR contributes to #2054


VeckoTheGecko (approved):
Looks good

PR #2364, opened 2025-11-07
Raise error when creating Field with dimensions without axis

erikvansebille:
This PR raises an error when users create a Field based on a grid with dimensions without axis. I found on multiple occasions, that without this error, the assert just below breaks:
https://github.com/Parcels-code/Parcels/blob/419ce76609c8ca8cc933821061a77454922a6fc6/src/parcels/_core/xgrid.py#L65-L67

One example is the unit test added to this PR
https://github.com/Parcels-code/Parcels/blob/419ce76609c8ca8cc933821061a77454922a6fc6/tests/test_xgrid.py#L138-L142

Not sure though if this is the best place for this error message. Some other, previously passing tests, now fail.

@VeckoTheGecko, perhaps you could weigh in where this error would be most appropriate ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu, mac, windows]
+        os: [ubuntu, windows]
         pixi-environment: [test-latest]
         include:
           - os: ubuntu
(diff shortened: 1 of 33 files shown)


VeckoTheGecko:
> Not sure though if this is the best place for this error message. Some other, previously passing tests, now fail.
> 
> @VeckoTheGecko, perhaps you could weigh in where this error would be most appropriate ?

I think it would be great to have in a dedicated function - I've pushed a patch baa16c7 . Thoughts?

Issue #2382, opened 2025-11-17
Output bug with time_interval = None


reint-fischer:
### Parcels version

v4-dev

### Description

When running a simulation with the NEMO curvilinear data, which is constant in time, writing to an output_file breaks the simulation. This is because a number of variables (`start_time`, `time_interval.left`) create unexpected behavior when being `== 0 ` or `== None`.

I started fixing this, but ran into another bug, which is related to an issue discussed in #2378:
> For example, an alternative is to set the `fieldset.time_interval` for field sets that don't have any time dimensions to `TimeInterval(np.datetime64('-9999-01-01'), np.datetime64('9999-12-31'))`, instead of `None`, as it is now:
> 
> https://github.com/Parcels-code/Parcels/blob/a760d04d46838250d8b68ff814a1c5ed135aa090/src/parcels/_core/fieldset.py#L88-L89
> 
> I couldn't get that easily to work, but could make an effort if you think that setting an 'infinite' TimeInterval is a cleaner solution

### Code sample

```python
data_folder = parcels.download_example_dataset("NemoCurvilinear_data")
files = data_folder.glob("*.nc4")
ds_fields = xr.open_mfdataset(files, combine="nested", data_vars="minimal", coords="minimal", compat="override")
ds_fields = (
    ds_fields.isel(time_counter=0, drop=True)
    .isel(time=0, drop=True)
    .isel(z_a=0, drop=True)
    .rename({"glamf": "lon", "gphif": "lat", "z": "depth"})
)

# TODO: replace manual fieldset creation with FieldSet.from_nemo() once available 
import xgcm
xgcm_grid = xgcm.Grid(
    ds_fields,
    coords={
        "X": {"left": "x"},
        "Y": {"left": "y"},
    },
    periodic=False,
    autoparse_metadata=False,
)
grid = parcels.XGrid(xgcm_grid, mesh="spherical")

U = parcels.Field("U", ds_fields["U"], grid, interp_method=parcels.interpolators.XLinear)
V = parcels.Field("V", ds_fields["V"], grid, interp_method=parcels.interpolators.XLinear)
U.units = parcels.GeographicPolar()
V.units = parcels.GeographicPolar()  # U and V need GeographicPolar for  C-Grid interpolation to work correctly
UV = parcels.VectorField("UV", U, V, vector_interp_method=parcels.interpolators.CGrid_Velocity)
fieldset = parcels.FieldSet([U, V, UV])

npart = 20
[...]

PR #2388, opened 2025-11-17
Fix writing for Fieldsets without time_interval

erikvansebille:
This fixes the bug described in #2382

### DIFF
--- a/src/parcels/_core/field.py
+++ b/src/parcels/_core/field.py
@@ -21,7 +21,7 @@ from parcels._core.statuscodes import (
 from parcels._core.utils.string import _assert_str_and_python_varname
 from parcels._core.utils.time import TimeInterval
 from parcels._core.uxgrid import UxGrid
-from parcels._core.xgrid import XGrid, _transpose_xfield_data_to_tzyx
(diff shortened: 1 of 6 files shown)


src/parcels/_core/particlefile.py
```diff
@@ -309,7 +312,12 @@ def _convert_particle_data_time_to_float_seconds(particle_data, time_interval):
     #! Important that this is a shallow copy, so that updates to this propogate back to the original data
     particle_data = particle_data.copy()
 
-    particle_data["time"] = ((particle_data["time"] - time_interval.left) / np.timedelta64(1, "s")).astype(np.float64)
+    if time_interval is None:
+        particle_data["time"] = (particle_data["time"] / np.timedelta64(1, "s")).astype(np.float64)
```
VeckoTheGecko: What is `particle_data["time"]` here? A timedelta object or a datetime object? (and if datetime, is that wrt. the beginning of the simuation?)

VeckoTheGecko (commented):
Could we add tests?

VeckoTheGecko (approved):

PR #2391, opened 2025-11-18
Adjust longitudes for antimeridian cells to account for periodicity

fluidnumericsJoe:
Resolves #2383


reint-fischer:
Note that the error thrown before was at `lon=180.2` which is larger than `grid.lon.max=180`, but having fixed this still leads to an error:
```
FAILED tests/test_advection.py::test_nemo_curvilinear_fieldset - parcels._core.statuscodes.GridSearchingError: Grid searching failed at (z=[0.], lat=[79.6822], lon=[179.7905])
```

fluidnumerics-joe:
Copy that. I'll get something more robust in place

PR #2395, opened 2025-11-19
v4 Reference API using autoapi

### LINKED ISSUE #2269: Set up API Reference for version 4
Version 3 of Parcels had an API reference that was manually currated. Now that we have a new package structure and a curated public API , I that we can have this API automatically generated - and is something to be explored (likely via the `sphinx-autoapi` package).

This reduces our maintenance burden meaning that as soon as something is public it gets documented in our API referece. Private functions (i.e., items with an underscore, or not listed in `__all__` should not appear)

Here is a rough patch

```diff
diff --git a/docs/conf.py b/docs/conf.py
index e45d4218..67d0cb45 100755
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,11 +40,13 @@ extensions = [
     "sphinx.ext.linkcode",
     "sphinx.ext.mathjax",
     "sphinx.ext.napoleon",
+    "autoapi.extension",
     "myst_parser",
     "nbsphinx",
     "numpydoc",
     "sphinxcontrib.mermaid",
 ]
+autoapi_dirs = ["../parcels"]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
diff --git a/docs/reference.rst b/docs/reference.rst
deleted file mode 100644
index caecbd39..00000000
--- a/docs/reference.rst
+++ /dev/null
@@ -1,14 +0,0 @@
-Parcels API
-===================================
-
-.. toctree::
-  :maxdepth: 2
-
-  Fieldsets and Fields <reference/fields>
-  Particles and particle sets <reference/particles>
-  Particlefile <reference/particlefile>
-  Predefined Kernels <reference/predefined_kernels>
-  User-defined Kernels <reference/userdefined_kernels>
-  Particle-particle interaction <reference/interaction>
-  Gridsets and grids <reference/grids>
-  Miscellaneous <reference/misc>
diff --git a/docs/reference/fields.rst b/docs/reference/fields.rst
deleted file mode 100644
index 9a26548f..00000000
--- a/docs/reference/fields.rst
+++ /dev/null
@@ -1,16 +0,0 @@
-Fieldsets and Fields
-====================
-
-parcels.fieldset module
------------------------
-
-.. automodule:: parcels.fieldset
-    :members:
-    :show-inheritance:
-
-parcels.field module
---------------------
-
-.. automodule:: parcels.field
-    :members:
-    :show-inheritance:
diff --git a/docs/reference/grids.rst b/docs/reference/grids.rst
deleted file mode 100644
index e8d47a11..00000000
--- a/docs/reference/grids.rst
+++ /dev/null
@@ -1,9 +0,0 @@
-Gridsets and grids
-==================
-
-parcels.grid module
--------------------
-
-.. automodule:: parcels.grid
-    :members:
-    :show-inheritance:
diff --git a/docs/reference/interaction.rst b/docs/reference/interaction.rst
deleted file mode 100644
index e9cd525d..00000000
--- a/docs/reference/interaction.rst
+++ /dev/null
@@ -1,16 +0,0 @@
-Particle-field interaction
-==========================
-
-parcels.interaction module
---------------------------
-
-.. automodule:: parcels.interaction
-    :members:
-    :show-inheritance: yes
-
-parcels.interaction.neighborsearch module
------------------------------------------
-
-.. automodule:: parcels.interaction.neighborsearch
-    :members:
-    :show-inheritance: yes
diff --git a/docs/reference/misc.rst b/docs/reference/misc.rst
deleted file mode 100644
index 0e379d46..00000000
--- a/docs/reference/misc.rst
+++ /dev/null
@@ -1,36 +0,0 @@
-Miscellaneous
-=============
-
-parcels.tools.statuscodes module
---------------------------------
-
-.. automodule:: parcels.tools.statuscodes
-    :members:
-    :show-inheritance:
-
-parcels.tools.converters module
--------------------------------
-
-.. automodule:: parcels.tools.converters
-    :members:
-    :show-inheritance:
-
-parcels.tools.loggers module
-----------------------------
-
-.. automodule:: parcels.tools.loggers
-    :members:
-    :undoc-members:
-
-parcels.tools.warnings module
-
-.. automodule:: parcels.tools.warnings
-    :members:
-    :undoc-members:
-
-parcels.tools.exampledata_utils module
---------------------------------------
-
-.. automodule:: parcels.tools.exampledata_utils
-    :members:
-    :undoc-members:
diff --git a/docs/reference/particlefile.rst b/docs/reference/particlefile.rst
deleted file mode 100644
index ef0104fb..00000000
--- a/docs/reference/particlefile.rst
+++ /dev/null
@@ -1,9 +0,0 @@
-Particlefile
-============
-
-parcels.particlefile module
---------------------------------------------
-
-.. automodule:: parcels.particlefile
-    :members:
-    :show-inheritance: yes
diff --git a/docs/reference/particles.rst b/docs/reference/particles.rst
deleted file mode 100644
index c5dd8aa6..00000000
--- a/docs/reference/particles.rst
+++ /dev/null
@@ -1,16 +0,0 @@
-Particles and particle sets
-===========================
-
-parcels.particle module
------------------------
-
-.. automodule:: parcels.particle
-    :members:
-    :show-inheritance:
-
-parcels.particleset module
---------------------------
-
-.. automodule:: parcels.particleset
-    :members:
-    :show-inheritance:
diff --git a/docs/reference/predefined_kernels.rst b/docs/reference/predefined_kernels.rst
deleted file mode 100644
index c168a435..00000000
--- a/docs/reference/predefined_kernels.rst
+++ /dev/null
@@ -1,33 +0,0 @@
-Predefined Kernels
-===================
-
-parcels.application_kernels.advection module
---------------------------------------------
-
-.. automodule:: parcels.application_kernels.advection
-    :members:
-    :show-inheritance:
-
-parcels.application_kernels.advectiondiffusion module
------------------------------------------------------
-
-
-.. automodule:: parcels.application_kernels.advectiondiffusion
-    :members:
-    :show-inheritance:
-
-parcels.application_kernels.EOSseawaterproperties module
---------------------------------------------------------
-
-
-.. automodule:: parcels.application_kernels.EOSseawaterproperties
-    :members:
-    :show-inheritance:
-
-parcels.application_kernels.TEOSseawaterdensity module
-------------------------------------------------------
-
-
-.. automodule:: parcels.application_kernels.TEOSseawaterdensity
-    :members:
-    :show-inheritance:
diff --git a/docs/reference/userdefined_kernels.rst b/docs/reference/userdefined_kernels.rst
deleted file mode 100644
index 78f39d77..00000000
--- a/docs/reference/userdefined_kernels.rst
+++ /dev/null
@@ -1,9 +0,0 @@
-User-defined Kernels
-====================
-
-parcels.kernel module
---------------------------------
-
-.. automodule:: parcels.kernel
-    :members:
-    :undoc-members:
diff --git a/pixi.toml b/pixi.toml
index 58f5a6a5..86afd4ec 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -70,6 +70,7 @@ ipython = "*"
 sphinx = "*"
 pandoc = "*"
 pydata-sphinx-theme = "*"
+sphinx-autoapi = "*"
 sphinx-autobuild = "*"
 myst-parser = "*"
 sphinxcontrib-mermaid = "*"

```

reint-fischer:
Fixes #2269 

I have implemented [sphinx-autoapi](https://sphinx-autoapi.readthedocs.io/en/latest/index.html) to generate the reference API based on the names made public by `__all__` in __init__.py. This means that curating which classes and methods are included happens fully in this list, and no custom list is necessary (as was necessary in reference.rst, see #2269).

This also means however, that we have less control over the order and presentation done by `autoapi`. The classes, exceptions, and attributes are all documented on one page now, and their order is determined by the order in which they are imported, which is fixed by alphabetically by `isort`. It is also not possible to control the order of the autosummary at the top of the page, which is sorted [by group](https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_member_order). From what I can see, autoapi is still quite a small project, and most other projects have `autodoc` style API references like we had before.

I think these complications are challenging enough to suggest returning to our manually curated API by reverting #2270 . @VeckoTheGecko, and @erikvansebille , would you take a look at [the docs build](https://parcels--2395.org.readthedocs.build/en/2395/reference/parcels/index.html#) and tell me what you think?

### DIFF
--- a/src/parcels/_core/basegrid.py
+++ b/src/parcels/_core/basegrid.py
@@ -22,7 +22,7 @@ class GridType(IntEnum):
 
 
 class BaseGrid(ABC):
-    """Define a base class for parcels X and Ux grids with common methods and properties"""
+    """Base class for parcels.XGrid and parcels.UxGrid defining common methods and properties"""
 
     @abstractmethod
     def search(self, z: float, y: float, x: float, ei=None) -> dict[str, tuple[int, float | np.ndarray]]:
--- a/docs/_autoapi_templates/python/module.rst
+++ b/docs/_autoapi_templates/python/module.rst
@@ -65,23 +65,6 @@ Exceptions
 
 
             {% endif %}
-            {% set visible_subpackages = obj.subpackages|selectattr("display")|list %}
-            {% set visible_submodules = obj.submodules|selectattr("display")|list %}
-            {% set visible_submodules = (visible_subpackages + visible_submodules)|sort %}
-            {% if visible_submodules %}
-Submodules
-----------
-
-.. toctree::
-   :caption: Submodules
-   :maxdepth: 1
-
-            {% for submodule in visible_submodules %}
-   {{ submodule.include_path }}
-            {% endfor %}
-
-
-         {% endif %}
             {% set visible_attributes = visible_children|selectattr("type", "equalto", "data")|list %}
             {% if visible_attributes %}
                {% if "attribute" in own_page_types or "show-module-summary" in autoapi_options %}
@@ -129,6 +112,18 @@ Functions
                {% endif %}
 
 
+            {% endif %}
+            {% set visible_subpackages = obj.subpackages|selectattr("display")|list %}
+            {% set visible_submodules = obj.submodules|selectattr("display")|list %}
+            {% set visible_submodules = (visible_subpackages + visible_submodules)|sort %}
+            {% if visible_submodules %}
+.. toctree::
+   :caption: Submodules
+   :maxdepth: 1
+
+            {% for submodule in visible_submodules %}
+   {{ submodule.include_path }}
+            {% endfor %}
             {% endif %}
             {% set this_page_children = visible_children|rejectattr("type", "in", own_page_types)|list %}
             {% if this_page_children %}


pyproject.toml
```diff
@@ -131,6 +125,17 @@ ignore = [
...
+# [tool.ruff.lint.per-file-ignores]
+# # Ignore docstring rules everywhere except for the stable files (particleset.py, interpolators.py).
+# "!src/parcels/{_core/particleset,interpolators}.py" = [
+#   # Missing docstring in public class
+#   "D101",
+#   # Missing docstring in public method
+#   "D102",
+#   # Missing docstring in public function
+#   "D103",
+# ]
```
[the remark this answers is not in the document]
reint-fischer: I have edited and activated this, and it now checks for all the files with public classes that docstrings are present, which I think is a minimal requirement. It does mean that the list of files is hardcoded, which means that new files may not be tracked. This `ruff.lint` can be easily extended to include rules as we update the docstrings (e.g. checking returns :https://docs.astral.sh/ruff/rules/docstring-missing-returns/#docstring-missing-returns-doc201)

VeckoTheGecko:
Actually, according to PEP257 it seems like there is some support

> String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.

so you can do

```python
Particle = get_default_particle(np.float32)
"""The default Particle used in Parcels simulations."""
```
image

reint-fischer:
> I don't think it makes sense to make `parcels.Particle` a class - since its an instance of a ParticleClass, not a different class.
> 
> Is the main concern here that there isn't a docstring for the attribute?

To clarify, I guess my concerns were twofold:
1. Since all the other core parcels components are classes (and listed as such in parcels.__init__.py), they all show up at the top of the reference page (and autoapi designates them to have their own page). Although I don’t think it’s necessary for parcels.Particle to be documented on a separate page, it does feel out of place as an attribute with `AllParcelsErrorCodes` and `logger`, rather than with the core classes or as a part of `parcels.ParticleClass`. My concern is that users will be confused by `parcels.Particle` being listed as an “Attribute” if they even find it in the first place.
2. The lack of a docstring made it even more confusing to understand what the role of `parcels.Particle` is as an “Attribute", so it’s great you found this clean solution!

> so you can do
> 
> ```python
> Particle = get_default_particle(np.float32)
> """The default Particle used in Parcels simulations.”””
> ```

docs/_autoapi_templates/python/class.rst
erikvansebille: Where does this file come from? If it is taken from another source, then acknowledge that at the top in a comment?

src/parcels/_core/basegrid.py
```diff
@@ -20,6 +22,8 @@ class GridType(IntEnum):
 
 
 class BaseGrid(ABC):
+    """Docstring for Basegrid class"""
```
erikvansebille: This needs to be expanded? Especially since it's the first one on the list of Classes, it looks a bit strange ("meta"?)

erikvansebille (commented):
Looks good! Just a few comments below

PR #2400, opened 2025-11-21
Adding v4 Kernel loop explanation

reint-fischer:
This PR builds on #2333, which updated the explanation of the Kernel loop. In my view, the Kernel loop tutorial should really be more of an explanation/background reading guide, which is why I have made it into a `.md` file. I have updated the example code, including adding boolean logic for working with StatusCodes. Should we test these last Kernels somewhere?

### DIFF
--- a/docs/getting_started/tutorial_quickstart.md
+++ b/docs/getting_started/tutorial_quickstart.md
@@ -99,7 +99,7 @@ TODO: link to a list of included kernels
 ```
 
 ```{code-cell}
-kernels = [parcels.kernels.AdvectionEE]
+kernels = [parcels.kernels.AdvectionRK2]
 ```
 
 ## Prepare output: `ParticleFile`
--- a/src/parcels/kernels/advection.py
+++ b/src/parcels/kernels/advection.py
@@ -18,101 +18,120 @@ __all__ = [
 ]
 
 
+def _constrain_dt_to_within_time_interval(time_interval, time, dt):
+    """Helper function to make sure dt does not go outside time_interval.
+
+    This is especially relevant for higher-order RK methods (RK2, RK4, RK45),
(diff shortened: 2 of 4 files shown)


docs/user_guide/examples/explanation_kernelloop.md
```diff
@@ -0,0 +1,194 @@
...
+
+# The Parcels Kernel loop
+
+On this page we discuss Parcels' execution loop, and what happens under the hood when you combine multiple Kernels.
+
+This is probably not very relevant when you only use the built-in Advection kernels, but can be important when you are writing and combining your own Kernels!
+
+## Background
+
+When you run a Parcels simulation (i.e. a call to `pset.execute()`), the Kernel loop is the main part of the code that is executed. This part of the code loops through all particles and executes the Kernels that are defined for each particle.
```
erikvansebille:
```suggestion
When you run a Parcels simulation (i.e. a call to `pset.execute()`), the Kernel loop is the main part of the code that is executed. This part of the code loops through time and executes the Kernels for all particle.
```

docs/user_guide/examples/explanation_kernelloop.md
```diff
@@ -0,0 +1,194 @@
...
+```
+def KeepInOcean(particles, fieldset):
+    through_surface = particles.state == StatusCode.ErrorThroughSurface
+    particles[through_surface].dz = 0.0
+    particles[through_surface].state = StatusCode.Success
+```
+
+Kernel functions such as the ones above can then be added to the list of kernels in `pset.execute()`.
+
+Note that these Kernels that control what to do with `particles.state` should typically be added at the _end_ of the Kernel list, because otherwise later Kernels may overwrite the `particles.state` or the `particles.dlon` variables.
```
erikvansebille: Make this a note (with a box around it)?

Issue #2380, opened 2025-11-17
Issue with uneven `dt` and RK advection


erikvansebille:
### Parcels version

v4dev

### Description

in v4-dev, `particles.dt` is currently a `np.timedelta64` and these use integer operations under the hood. 

This means that setting an uneven dt (e.g. `dt=np.timedelta64(1, "s")`) can lead to errors in AdvectionRK2 and AdvectionRK4, where fields are sampled at `particles.time + 0.5 * particles.dt`

This could be another reason to change the API for `dt` (see also #2130) and replace `np.datetime64` `np.timedelta64` for `time` and `dt` to a `float`?

### Code sample

```python
def AdvectionRK4(particles, fieldset):  # pragma: no cover
    """Advection of particles using fourth-order Runge-Kutta integration."""

    assert np.all((particles.dt * 0.5) * 2 == particles.dt)

   ...  # Rest of code here (and also for other RK kernels)
```

```text
=========================== short test summary info ============================
FAILED tests/test_advection.py::test_advection_3D_outofbounds[True-up] - AssertionError
FAILED tests/test_advection.py::test_advection_3D_outofbounds[True-down] - AssertionError
FAILED tests/test_advection.py::test_advection_3D_outofbounds[False-up] - AssertionError
FAILED tests/test_advection.py::test_advection_3D_outofbounds[False-down] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[None-0.2--0.3] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[None-0.2-u1] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[None-v1--0.3] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[None-v1-u1] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[-0.2-0.2--0.3] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[-0.2-0.2-u1] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[-0.2-v1--0.3] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[-0.2-v1-u1] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[w2-0.2--0.3] - AssertionError
FAILED tests/test_advection.py::test_length1dimensions[w2-0.2-u1] - AssertionError
[...]

Issue #2130, opened 2025-08-04
Improve dt API for kernels


erikvansebille:
Now that `particle.dt` is a `np.timedelta64` object, the kernels also need access to `dt` as a float, mainly for computing distances via multiplication with velocities. 

An option (instead of having to manually convert `dt` in the kernels, which is what's now in v4-dev) is to have two types of `dt`; one in `np.timedelta64` and one in `float`. 

However, we would have to be very carful that the two version of `dt` stay consistent, if users change only one!

erikvansebille:
Following up from the [v4 tutorial at the Parcels 10 year anniversary event](https://github.com/OceanParcels/10year-anniversary-session4), it's clear that users find the need to convert `particle.dt` from a `np.timedelta64` to a `float` inside Kernels confusing and cumbersome; especially because Field interpolation at e.g. `particle.time + particle.dt` (required for higher-order advection schemes) does need a `dt` as a `np.timedelta64`.

I see at least a few options:
 
1. Keep the API as is now in v4 (with a line `dt = particles.dt / np.timedelta64(1, 's')` at the top of each kernel
2. Add a `particle.dt_as_float` property/method that converts timedelta to float. Users would then still need to think when to use particle.dt` and when to use `particle.dt_as_float`.
3. Use `floats` for `particle.dt` _and_ `particle.time` inside the Kernel. This is how v3 also works, so an advantage is that it doesn't require any changes from the users. This would also require updating the time search.

After discussions over beer with @willirath and @fluidnumerics-joe, I now think that option 3 is perhaps the best. There are two disadvantages to option 3:
a. We can't use date times and time deltas anymore inside kernels
b. It may be confusing to users that they have to provide dt as a time delta, but that in the Kernel it's a float

However, these disadvantages may be less impactful than the disadvantages of options 1 and 2 above

PR #2405, opened 2025-11-25
Changing `time` and `dt` inside kernel loop to float

erikvansebille:
As discussed in #2130, using `np.timedelta64` and `np.datetime64` as core data structured for `particles.dt` and `particles.time` brings some major issues:
1. Users will have to convert from timedelta to float inside each kernel, and use the float for multiplying with velocities
2. There is a fundamental issue with uneven `timedelta64` `dt`, see #2380

This PR therefore implements a revert to the v3 approach, where `time` and `dt` are `floats` throughout the Kernel. This also allows (later) integration with `pint` for unit-aware computations in the Kernel

During development, I also came across a bug in RK45 (fixed in 18a45909504fe98b03afa442854e1c2444144555), which needs a `particles.next_dt` variable in order to be able to _increase_ dt in the Kernel loop (otherwise, displacements are computed with the old dt, but the time-update is with the new dt).

### DIFF
--- a/src/parcels/_core/utils/time.py
+++ b/src/parcels/_core/utils/time.py
@@ -56,7 +56,7 @@ class TimeInterval:
     def __contains__(self, item: T) -> bool:
         return self.left <= item <= self.right
 
-    def is_all_time_in_interval(self, time):
+    def is_all_time_in_interval(self, time: float):
         item = np.atleast_1d(time)
         return (0 <= item).all() and (item <= self.time_length_as_flt).all()
 


src/parcels/_core/particleset.py
```diff
@@ -632,19 +618,29 @@ def _get_simulation_start_and_end_times(
...
+
     return start_time, endtime
 
 
 def _get_start_time(first_release_time, time_interval, sign_dt, runtime):
     if time_interval is None:
-        time_interval = TimeInterval(left=np.timedelta64(0, "s"), right=runtime)
+        time_interval = TimeInterval(np.timedelta64(0, "s"), right=np.timedelta64(int(runtime * 1e9), "ns"))
 
     if sign_dt == 1:
```
VeckoTheGecko: Separate to this PR, something I want to work on is typechecking in Parcels (I think in cases like this it would make it a lot easier to reason around what the types are for these different parameters)

tests/test_advection.py
```diff
@@ -416,13 +425,14 @@ def test_peninsula_fieldset(kernel, rtol, grid_type):
...
-    if kernel == AdvectionRK45:
-        fieldset.add_constant("RK45_tol", rtol)
-
     SampleParticle = Particle.add_variable(
         [Variable("p", initial=0.0, dtype=np.float32), Variable("p_start", initial=0.0, dtype=np.float32)]
     )
 
+    if kernel == AdvectionRK45:
+        fieldset.add_constant("RK45_tol", rtol)
+        SampleParticle = SampleParticle.add_variable(Variable("next_dt"))
```
VeckoTheGecko: as above

tests/test_advection.py
```diff
@@ -328,8 +333,11 @@ def test_decaying_moving_eddy(kernel, rtol):
     if kernel == AdvectionRK45:
         fieldset.add_constant("RK45_tol", rtol)
         fieldset.add_constant("RK45_min_dt", 10 * 60)
+        MyParticle = Particle.add_variable(Variable("next_dt"))
```
VeckoTheGecko: as below

tests/test_advection.py
```diff
@@ -287,8 +287,13 @@ def test_moving_eddy(kernel, rtol):
 
     if kernel == AdvectionRK45:
         fieldset.add_constant("RK45_tol", rtol)
+        MyParticle = Particle.add_variable(Variable("next_dt"))
+    else:
+        MyParticle = Particle
```
VeckoTheGecko: as below

VeckoTheGecko (commented):
This is the rest of my review feedback.

src/parcels/_core/particleset.py
erikvansebille: Yep, I agree that more typechecking would be good

src/parcels/_core/utils/time.py
```diff
@@ -44,13 +44,19 @@ def __init__(self, left: T, right: T) -> None:
...
+        else:
+            self.time_length_as_flt = right - left
 
     def __contains__(self, item: T) -> bool:
         return self.left <= item <= self.right
 
     def is_all_time_in_interval(self, time):
         item = np.atleast_1d(time)
-        return (self.left <= item).all() and (item <= self.right).all()
+        return (0 <= item).all() and (item <= self.time_length_as_flt).all()
```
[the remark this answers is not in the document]
erikvansebille: Yep, expanded in 2b166b8798cd5d2bb1b9238e299042874a6f1e18

tests/test_advection.py
erikvansebille: Implemented in ef12d26119a0babe61f61f005587fca641c39a3f

src/parcels/_core/utils/time.py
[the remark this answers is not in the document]
VeckoTheGecko: I think the type annotation is sufficient, renaming `is_all_time_in_interval` to `is_all_time_in_flt_interval` seems a bit excessive

PR #2394, opened 2025-11-18
Add Interpolation tutorial v4

reint-fischer:
I have rewritten the interpolation tutorial to work in v4. To do this we needed to fix the `XLinearInvdistLandTracer` interpolator (in #2389).

### TODO
- [x] List all `parcels.interpolators.*`
- [x] Discuss writing your own interpolator 
- [ ] ~~Fix C-grid Tracer interpolation~~

PR #2421, opened 2025-12-05
Uxinterpolation tutorial v4

fluidnumericsJoe:
This PR builds on #2394 and adds a section for unstructured grid interpolation.

In working on this tutorial, I've found that the waschpress point implementation was not robust; in some cases, particles were erroneously detected inside faces they were not inside of. I've corrected this by implementing strictly barycentric coordinates for 3-sided cells with a much simpler algorithm. With this implementation, I've also enforced a check to ensure that `ux.grid` objects have at most 3 nodes per face (triangle only meshes). I think this is fine, since the ICON and FESOM2 cases we're currently prioritizing only use triangular meshes. Nonetheless, I'll open a discussion on Zulip to gauge interest in supporting  n-gon meshes.

Additionally, when using the "spherical" mesh type, I've found that the spherical coordinate transformation on particle (lat,lon) positions produces cartesian locations that are not in the plane spanned by the element faces. This allows the sum of the barycentric coordinates to be greater than one. To fix this, I've added a step in the point-in-cell check to remove the component of the cartesian location that is normal to the element plane when performing point in cell checks. This resolves issues with false negatives on spherical type meshes.

### DIFF
--- a/tests/test_uxarray_fieldset.py
+++ b/tests/test_uxarray_fieldset.py
@@ -123,25 +123,25 @@ def test_fesom2_square_delaunay_uniform_z_coordinate_eval():
     fieldset = FieldSet([UVW, P, UVW.U, UVW.V, UVW.W])
 
     assert np.isclose(
-        fieldset.U.eval(time=ds.time[0].values, z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
+        fieldset.U.eval(time=[0.0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
         1.0,
         rtol=1e-3,
         atol=1e-6,
     )
     assert np.isclose(
-        fieldset.V.eval(time=ds.time[0].values, z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
+        fieldset.V.eval(time=[0.0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
         1.0,
         rtol=1e-3,
         atol=1e-6,
     )
     assert np.isclose(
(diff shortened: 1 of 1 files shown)


tests/test_uxarray_fieldset.py
```diff
@@ -115,17 +115,37 @@ def test_fesom2_square_delaunay_uniform_z_coordinate_eval():
...
-    P = Field(name="p", data=ds.p, grid=grid, interp_method=UXPiecewiseLinearNode)
+    P = Field(name="p", data=ds.p, grid=grid, interp_method=UxPiecewiseLinearNode)
     fieldset = FieldSet([UVW, P, UVW.U, UVW.V, UVW.W])
 
-    assert fieldset.U.eval(time=[0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False) == 1.0
-    assert fieldset.V.eval(time=[0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False) == 1.0
-    assert fieldset.W.eval(time=[0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False) == 0.0
-    assert fieldset.p.eval(time=[0], z=[1.0], y=[30.0], x=[30.0], applyConversion=False) == 1.0
+    assert np.isclose(
+        fieldset.U.eval(time=ds.time[0].values, z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
```
erikvansebille: Now that time is a float (as of #2405),  this needs to change


```suggestion
        fieldset.U.eval(time=[0.], z=[1.0], y=[30.0], x=[30.0], applyConversion=False),
```

Issue #2396, opened 2025-11-19
Particle-particle interaction in v4


erikvansebille:
Now that kernels have access to _all_ particles, it might be much simpler to do particle-particle interaction in Parcels v4. It would be good to explore how feasible this feature is, and (if it is feasible) update [the v3 tutorial](https://docs.parcels-code.org/en/latest/examples/tutorial_interaction.html) to showcase this feature

PR #2426, opened 2025-12-10
Making particle-particle interaction tutorial v4-compatible

erikvansebille:
This PR shows how to use 'simple'/standard kernels for particle-particle interaction in v4. 

Now that Kernels have access to _all_ particles, it is fairly straightforward to write both the `Pull` and the `Merge` kernel in vectorised form (barring the KernelParticle bug in #2143).

This means that we don't need all the machinery for InteractionParticles, so that can all be removed from the codebase

This PR implements #2396

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows]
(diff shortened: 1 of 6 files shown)


VeckoTheGecko (approved):
Looks great!

Issue #2003, opened 2025-05-08
Supporting various circulation models and Parcels' internal representation of data


VeckoTheGecko:
*This issue is a mix of discussion on internal representation of data, and tasks required for supporting reading data from various circulation models.*

### Internal representation of data

We've decided on working internally with Xarray datasets, now an important matter to discuss is what these datasets actually look like internally[^3] (i.e., dimension ordering, how is depth defined, ...). (1) Are we working with the datasets in the form close to the model output? (i.e., close to the raw NetCDF files?) Or (2) are we working with datasets that match a certain internal representation defined by us?

@erikvansebille  I recall from our group meetings that you were leaning towards (1), saying something along the lines of "allowing users to write interpolator methods while using the knowledge about their model" (please, if I don't fully understand your viewpoint or am missing something, comment below).

I am personally leaning towards (2) for the following reasons:
- Its safer
    - At the point of intialisation we can throw an informative error message if the dataset doesn't match our internal representation. This is easier to debug than runtime errors from failed index searching.
- It makes our code simpler and more performant
    - All our indexing methods can make assumptions according to our internal model (e.g., that the depth array is increasing)
    - We need to do less checking before we do something
- Makes testing easier as from FieldSet all the way through, we only need to test things in our internal data model (e.g., no need to test some edge case introduced only when model X is running `pset.execute()`)
- It's all lazy anyway
    - Any dataset transformation done to get to this "internal state" would be done lazily via Dask
- I feel that we can define an internal representation within parcels that encompasses all models that we want to support. *Admittedly, I'm not 100% sure on whether my feeling here is well informed - I'm not intimately familiar with the model outputs we work with.*
- We already have some assumptions about internal representation with assuming that the data is ordered [tdim, zdim, ydim, xdim]
- Allows interpolation methods to be portable between Fields (and allows us to ship them in Parcels)

A downside of (2) is (as you point out) that those writing interpolation methods need to understand how data is handled internally in Parcels as opposed to their model - that is something we would need good documentation for to support those writing interpolation methods.

### Tasks required for supporting reading data from various circulation models

Sub-issues:
- [ ] Example datasets for different (ocean) circulation models
- [ ] (2) Decide on an internal representation of data within parcels (or decide on how we are going to define search methods in a flexible way to work with different datasets)
    - [DOC] Document this internal representation in the v4 docs (useful for those writing interpolators etc.)
[...]

PR #2432, opened 2025-12-12
Add `FieldSet.from_sgrid_conventions()`

VeckoTheGecko:
Towards #2003 

Changes: 
- See commits and descriptions

The testing as it stands for `from_conventions` is just testing init on the datasets provided.

I think there's room for future improvements (which I've put as comments in the body)

This acts as the base for future methods (`from_nemo` etc) for ingesting structured data. Here is what I have in mind:

```python
def from_nemo(field_data: dict[str, xr.Dataset], ds_mesh: xr.Dataset) -> xr.Dataset:
    # Not 100% sure about the function namings, location in the package, and the parameter API. I think that a single xr.Dataset should be returned so users can make modifications.

    ...  # work with field_data and ds_mesh
    ds = ...

    ds["grid"] = xr.DataArray(
        data=np.array([0]),
        attrs=sgrid.Grid2DMetadata(...).to_attrs(),
    )
    # ? does the "location" attribute need to be added to each field? (as per SGRID conventions)
    # ? would it be nice to use `sgrid.rename_dims()` to rename all dimensions to a common naming scheme used for Parcels from_* methods?
    # ? attach other cf-convention attributes? (standard_name etc.)
    # ? Add logger.info or logger.warning messages during ingestion for items the user needs to be aware of?

    return ds

# user code
ds_sgrid_compliant = from_nemo(...)

print(ds_sgrid_compliant)
# Looks good!
FieldSet.from_sgrid_conventions(ds_sgrid_compliant)
# OR
# Looks like the x variable is defined on the cell centers
ds_sgrid_compliant['x'].attrs['location'] = ...
# fixed
FieldSet.from_sgrid_conventions(ds_sgrid_compliant)
```

### DIFF
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -332,7 +332,7 @@ class FieldSet:
         if "time" in ds.dims and "time" in ds.coords:
             if "axis" not in ds["time"].attrs:
                 logger.debug(
-                    "Dataset contains 'time' dimension but no 'axis' attribute. Setting 'axis' attribute is set to 'T'."
+                    "Dataset contains 'time' dimension but no 'axis' attribute. Setting 'axis' attribute to 'T'."
                 )
                 ds["time"].attrs["axis"] = "T"
 


src/parcels/_core/fieldset.py
```diff
@@ -295,6 +296,92 @@ def from_fesom2(ds: ux.UxDataset):
...
+
+        See https://sgrid.github.io/ for more information on the SGRID conventions.
+        """
+        ds = ds.copy()
+
+        # Ensure time dimension has axis attribute if present
+        if "time" in ds.dims and "time" in ds.coords:
+            if "axis" not in ds["time"].attrs:
+                logger.debug(
+                    "Dataset contains 'time' dimension but no 'axis' attribute. Setting 'axis' attribute is set to 'T'."
```
erikvansebille:
```suggestion
                    "Dataset contains 'time' dimension but no 'axis' attribute. Setting 'axis' attribute to 'T'."
```

erikvansebille (commented):
Looks good!  A few comments/questions below

PR #2436, opened 2025-12-15
Nesting documentation guide

erikvansebille:
This PR creates a documentation guide for how to do nesting in Parcels v4, using a combination of unstructured (for the nest localisation) and structured (for the hydrodynamic data) Fields.

PR #2439, opened 2025-12-17
Creating a Nested Grid Tutorial

erikvansebille:
This PR creates a tutorial for how to do nesting in Parcels v4, using a combination of unstructured (for the nest localisation) and structured (for the hydrodynamic data) Fields.

It is different from #2436 in that it uses idealised flow fields, so can be run in CI


docs/examples/tutorial_nestedgrids.ipynb
erikvansebille: @fluidnumerics-joe, could you have a look at this notebook? There's an issue with the `ux.Grid()` function wrapping around the antemeridian, even if this is a flat mesh. Not sure how to prevent this. See the code at the end of the notebook (see screenshot)

Image

See also the rendered docs at https://parcels--2439.org.readthedocs.build/en/2439/user_guide/examples/tutorial_nestedgrids.html

docs/user_guide/examples/tutorial_nestedgrids.ipynb
reint-fischer: Very cool notebook, and so exciting to see this feature in v4! My comments are mostly considering the cases other people will use this feature for.

- When showing the velocity fields in rectangles with the Grid polygons in red, maybe add a note (admonition) discussing the case when the fields come from a nested model: the velocity fields would then be defined for the polygons only, and you would only need to pass the bounding coordinates?
- The triangulation function seems to be the core function that people will use (copy?) to set up their own nested fields. I was thinking about how to make this robust, maybe we can at least link to the documentation of the `triangle` function used: https://rufat.be/triangle/delaunay.html 
- The `AdvectionEE_NestedGrids` will also be used a lot I assume. The copying of the variables is not so bad, but maybe we should add the errorcode `TODO` as a box to be checked to #2143? Alternatively, is there not a way to pass `particles[mask]` as a fifth argument?

After writing this I also think it is okay to fix some of these challenges once people start implementing their own nested setups in v4 and come with feedback.

erikvansebille: I implemented the first two suggestions in https://github.com/Parcels-code/Parcels/pull/2439/commits/3d1be3e24fcf877d3127c2c7cf2354e150e7b58b. The third one will indeed be fixed/cleaned when we solve #2143

PR #1342, opened 2023-04-02
DOC: API reference restructure, contributing guidelines, and more...

VeckoTheGecko:
Part of #1324

- [x] API reference sectioning (making it easier to navigate)
  - Used the same modules that were covered in the prior documentation
- [x] Add contributing guidelines
- [x] Version switcher
  - Waiting on feedback from https://github.com/pydata/pydata-sphinx-theme/pull/1005#issuecomment-1493381984 about whether the version switcher from RTD can be used by default on desktop machines. Will see how to proceed from there.
    - Won't fix
- [x] Disable search engine indexing of previous versions
- [x] Automatic updating of copyright year (as per https://github.com/OceanParcels/parcels/pull/1341#issuecomment-1493318063)
- [x] Add support for markdown
- [x] Add backlink in navbar to OceanParcels website
- [x] Render grids in docstrings as raw text (rather than HTML table)
- [x] Rendering improvements (fixing numbered lists, inline code formatting)

[RTD deployment](https://veckoparcels.readthedocs.io/en/my_docs-new_layout/). Keen to hear what you think about these changes @erikvansebille

Issue #2376, opened 2025-11-13
Decide on English spelling rules (and add to docsguide)?


erikvansebille:
Perhaps we should standardise the English spelling across our documentation? English-UK or English-US? Or English-AU??

I'd vote for English-UK, but am open to other suggestions

And is there any tooling to check this spelling rule?

reint-fischer:
Sounds good, especially if we use tooling to check and fix it, since I think it is not such a large issue that we should spend checking things manually. I am happy to go either way, I mostly tend to write with US spelling myself.

PR #2442, opened 2025-12-19
add API reference links

reint-fischer:
This PR adds API reference links throughout the documentation, which resolves a TODO note.

Since we use sphinx autoapi (#2395) to build the API reference, and MyST to parse markdown (#1342), we should use the `{py:obj}` role to refer to parcels objects (see [executablebooks/discussions/q-a#1169](https://github.com/orgs/executablebooks/discussions/1169))

e.g.:
```
{py:obj}`parcels.FieldSet`
```

Unfortunately, I am only able to refer to classes, and not their methods, i.e. the following doesn’t create the link to the method:
```
{py:meth}`parcels.ParticleSet.execute()`
```

See also: https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-meth

Work done (contributing to #2449) :

- [x] quickstart tutorial
- [x] parcels.interpolators (done in #2441 )


VeckoTheGecko:
> ## Notebooks to add links
> * [x]  quickstart tutorial
> * [ ]  output tutorial
> ...

, I think this is something better stored in a tracking issue so that we can work on it gradually, rather than try for all in this PR

VeckoTheGecko:
Is there anything we're waiting on @erikvansebille before merging this?

I think it would be good to update the contributing guidelines at some point with

- #2376 
- Mentioning that you can do ``{py:meth}`parcels.FieldSet.from_copernicusmarine`` to show ``parcels.FieldSet.from_copernicusmarine`` OR you can shorten it with ``{py:meth}`~parcels.FieldSet.from_copernicusmarine`` to show ``from_copernicusmarine``
- Guidance on whether to crosslink all the time, or just on first mention (up to us to decide - though perhaps all the time would be nice so that we avoid broken docs down the line if things don't get updated)

I think this sort of guidance can be done as a mention in the current `docs/development/index.md`

PR #2444, opened 2025-12-23
Fixing bug in uxgrid index_search

erikvansebille:
This PR makes sure shapes stay consistent for single-face queries, and adds a small safeguard against zero norms


VeckoTheGecko (approved):
Looks good - just a clarifying comment below. I think it would be good to merge to enable https://github.com/Parcels-code/Parcels/pull/2443 if there's nothing more pending for this PR

PR #2451, opened 2026-01-06
Adding Field._land_value property

### LINKED ISSUE #2393: Land as np.nan or 0 in v4?
Maybe we should discuss whether land is `np.nan` or `0.0` in the internal model of v4?

_Originally posted by @erikvansebille in https://github.com/Parcels-code/Parcels/discussions/2387#discussioncomment-14991064_

In the new `XLinearInvdistLandTracer` interpolator, land values are assumed to be `np.nan`, instead of `0`, which is the current assumed value. I think this makes a lot of sense for many fields where no valid value exists on land and using `0` could result in ambiguous interpolation?

erikvansebille:
This PR fixes #2393 by giving users control over the value in their Fields that represents land points. This value is stored in `Field._land_value` (which can be modified by a `setter` on `Field.land_value`). 

Its value is taken from the `da._FillValue`, and set to `0.0` if there is no `_FillValue` present

The `Field._land_value` can then be used in Interpolators that require masking of land

### DIFF
--- a/tests/test_field.py
+++ b/tests/test_field.py
@@ -159,6 +159,20 @@ def test_field_invalid_interpolator():
         )
 
 
+@pytest.mark.parametrize("fill_value", [-999.99, 0.0, 42, None])
+def test_field_land_value(fill_value):
+    ds = datasets_structured["ds_2d_left"].copy()
+    if fill_value is not None:
+        ds["data_g"].attrs["_FillValue"] = fill_value
+    grid = XGrid.from_dataset(ds, mesh="flat")
+
+    field = Field(name="test_field", data=ds["data_g"], grid=grid, interp_method=XLinear)
+    if fill_value is None:
+        assert field.land_value == 0.0
+    else:
+        assert field.land_value == fill_value
+
+
 def test_vectorfield_invalid_interpolator():
     ds = datasets_structured["ds_2d_left"]
     grid = XGrid.from_dataset(ds, mesh="flat")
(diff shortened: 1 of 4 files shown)


tests/test_field.py
```diff
@@ -159,6 +159,20 @@ def invalid_interpolator_wrong_signature(particle_positions, grid_positions, inv
         )
 
 
+@pytest.mark.parametrize("fill_value", [-999.99, 0.0, 42, None])
+def test_field_land_value(fill_value):
+    ds = datasets_structured["ds_2d_left"].copy()
+    if fill_value is not None:
+        ds["data_g"].attrs["_FillValue"] = fill_value
```
VeckoTheGecko: I think here we're kind of overloading and conflicting with how xarray ingests this

https://docs.xarray.dev/en/stable/user-guide/io.html#scaling-and-type-conversions

i.e., the user merely opening a dataset with decoding enabled might cause the `_FillValue` to be interpretted as np.nan, and then removed from the xarray dataset metadata.

I think a more informative test would be to look at reading in a dataset from disk rather than using this example dataset

VeckoTheGecko (commented):
I think that there are other ways to implement this behaviour without having to work with `_FillValue` or assign a land value (which have drawbacks as mentioned in my other comments) . Perhaps a good place to start would be to create a dataset that has U V and something like salinity, and see from a user POV how they can get intended behaviour without modifying Parcels internals? (e.g., via [`xr.DataArray.fillna`](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.fillna.html))

erikvansebille:
Superseded by #2762

PR #2453, opened 2026-01-07
Using from_sgrid_conventions for simple_UV_dataset throughout

erikvansebille:
Now that we have a `FieldSet.from_sgrid_conventions()` method, we can clean up many of the tutorials that rely on the `simple_UV_dataset` by simplifying the fieldset creation. 

This requires the sgrid metadata to be attached to `simple_UV_dataset`, which this PR also implements

Another change is that `FieldSet.from_sgrid_conventions()` now always return a `FieldSet.UV` if `U` and `V` are present, irrespective of whether `W` is present (see https://github.com/Parcels-code/Parcels/pull/2453#discussion_r2675981249)

This PR also cleans AdvectEE_NestedGrids Kernel now that #2143 has been merged


VeckoTheGecko (approved):
Thanks or the clarifications - LGTM ! 🚀

PR #2458, opened 2026-01-09
Add Ruff rules for PEP naming

### LINKED ISSUE #2457: Enforcing PEP8 naming conventions in Parcels
I think we should discuss whether to enforce some of [PEP8](https://peps.python.org/pep-0008/) variable naming conventions in our codebase.

This can be done via the ['pep8-naming (N)' ruleset from Ruff](https://docs.astral.sh/ruff/rules/#pep8-naming-n).

Looking at our current codebase, enabling the N rule with have the following violations

```
{'N802': 193,
 'N803': 21,
 'N805': 4,
 'N806': 478,
 'N811': 2,
 'N816': 42,
 'N818': 2}
```
<details><summary>Repro instructions</summary>
<p>

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 86e1420b..4ce41eb6 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ select = [
   "ISC001", # single-line-implicit-string-concatenation
   "TID", # flake8-tidy-imports
   "T100", # Checks for the presence of debugger calls and imports
+  "N", # pep8 naming conventions
 ]
 
 ignore = [

```

then doing

```
ruff check . | grep -E '^N\d{3}' > pep-log.txt && python process-log.py
```
where
```python
# process-log.py
from pathlib import Path

FILE = 'pep-log.txt'

with open(FILE,'r') as f:
    errors = f.read().splitlines()

codes = {}
for error in errors:
    code = error.split(" ")[0]
    if code not in codes:
        codes[code] = 0
    codes[code] += 1

from pprint import pprint

pprint(codes)
```

</p>
</details> 

As with all Ruff rules, we should be selective about which ones to follow. Some conventions simply do not apply in certain domains (e.g., in `numpy`, `X = np.array(...)` is commonly used for multidimensional arrays, and so is `dX` - even though it violates PEP8) .

In Parcels, we have kernels following CamelCase naming. This is also a pep8 violation as functions aren't named so, but is really useful since its clear what's a kernel and I don't think we should change this convention.

I'll put together a PR that selects which of these rules to enforce and which to ignore - striking a balance that works best for Parcels.

VeckoTheGecko:
Picked up some inconsistencies in our code along the way which was nice!

See commit history for changes, and see updated `pyproject.toml` for updated rules.

Fixes #2457

Notable changes:
- renamed `applyConversion` to `apply_conversion` in Field and VectorField .eval() methods

### DIFF
--- a/docs/user_guide/examples/tutorial_nemo_curvilinear.ipynb
+++ b/docs/user_guide/examples/tutorial_nemo_curvilinear.ipynb
@@ -152,7 +152,7 @@
    "outputs": [],
    "source": [
     "u, v = fieldset.UV.eval(\n",
-    "    np.array([0]), np.array([0]), np.array([20]), np.array([50]), applyConversion=False\n",
+    "    np.array([0]), np.array([0]), np.array([20]), np.array([50]), apply_conversion=False\n",
     ")  # do not convert m/s to deg/s\n",
     "print(f\"(u, v) = ({u[0]:.3f}, {v[0]:.3f})\")\n",
     "assert np.isclose(u, 1.0, atol=1e-3)"
@@ -298,7 +298,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "test-notebooks",
+   "display_name": "default",
    "language": "python",
    "name": "python3"
   },
@@ -312,7 +312,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.0"
+   "version": "3.14.2"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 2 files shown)


VeckoTheGecko:
Also added a rule `PGH004` in  799db8d . 

This prevents the blanket use of `noqa` comments - this makes it specific which rule we want to ignore. So if there are multiple violations on a line, we only ignore the one we want.

For example,

```diff
- import parcels  # noqa
+ import parcels  # noqa: E402
```

Do you think that's ok @erikvansebille ? It would have the additional step of copying the code from the pre-commit output and pasting it next to `noqa: `
e.g, 
```
E402 Module level import not at top of file
  --> docs/conf.py:96:1
   |
94 | #
95 | # The short X.Y version.
96 | import parcels
   | ^^^^^^^^^^^^^^
97 |
98 | version = parcels.__version__
```

erikvansebille (approved):
Looks good to me!

PR #2464, opened 2026-01-12
Remove sphinx autoapi templates

### LINKED ISSUE #2448: DOC: Fix cross-referencing to API methods and functions
> Unfortunately, I am only able to refer to classes, and not their methods, i.e. the following doesn’t create the link to the method:

I dug a bit deeper into this, I also couldn't get it to link to methods either with the current setup - though I think that's due to our current config.

I tested the following

```
## works
{py:obj}`parcels.ParticleSet` 
{py:class}`parcels.ParticleSet`

## broken
{py:meth}`parcels.ParticleSet.execute` 
{py:meth}`parcels._core.particleset.ParticleSet.execute` 
{py:obj}`download_example_dataset`
```

On the https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#sphinx-autodoc2 page it has the following example

> `` - {py:obj}`myst_parser.sphinx_ext.main.setup_sphinx` ``
> ``- [](#myst_parser.sphinx_ext.main.setup_sphinx)``

which properly links to a function. I think the solution might be in the fact that they're using `autodoc2` instead of `autodoc` (maybe the way that generates references is different to what we have).

Should I make this into a separate issue to be pursued later @erikvansebille ? I think cross linking with the classes is suitable for now/a bit lower on the chopping block

_Originally posted by @VeckoTheGecko in https://github.com/Parcels-code/Parcels/issues/2442#issuecomment-3710169562_

VeckoTheGecko:
Fixes #2448

### DIFF
--- a/docs/getting_started/index.md
+++ b/docs/getting_started/index.md
@@ -9,7 +9,3 @@ Getting started with parcels is easy; here you will find:
 🎓 Output tutorial <tutorial_output.ipynb>
 📖 Conceptual workflow <explanation_concepts.md>
 ```
-
-```{note}
-TODO: Add links to Reference API in quickstart tutorial and concepts explanation
-```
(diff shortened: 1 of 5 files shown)


erikvansebille:
Yay, works now (after https://github.com/Parcels-code/Parcels/pull/2464/commits/39d6627f56feedb0216ff5f0cbde5fdf79f4d4c5)
Screenshot 2026-01-12 at 15 52 01

PR #2460, opened 2026-01-09
Implementing `convert.nemo_to_sgrid()`

VeckoTheGecko:
This PR supercedes https://github.com/Parcels-code/Parcels/pull/2447
implementing the following additional functionality:

- Refactor out the `FieldSet.from_nemo()` functionality into a `convert.py` module with a `nemo_to_sgrid()` function. Also added corresponding tests.
- Make the input parameters of the `nemo_to_sgrid` function more closely align with the input datasets
- Update `FieldSet.from_sgrid_conventions` to infer `mesh` from the metadata included in the dataset
- Add `get_value_by_id()` function to `Grid*DMetadata` classes allowing to easily refer to parts of the SGRID metadata model
- 

Future work:
- Improve logging and metadata conversion `convert.py` - its important that we refactor this module so that its clearer, and have logging so that changes to the input dataset are communicated to the user. To support this, we need
- Better testing against different NEMO datasets
  - Reach out to the [NEMO community](https://nemo-ocean.discourse.group/) asking for ncdump output so that better see the types of nemo models out there

I'll make an issue to track this future work.

PR #2466, opened 2026-01-13
Update sgrid renaming tooling to work with node_coordinates

VeckoTheGecko:
Before, the tooling in `sgrid` was used for the renaming of dimensions. Now that we're also parsing `node_coordinates` - which aren't dimensions - we need to make this tooling more flexible so that we can handle renaming of any dimension/coordinate name listed in the metadata.

This PR implements these changes, and updates all local variable names/error messages accordingly

Notable changes:
- Update `Grid*DMetadata.rename_dims` to `.rename`
- Update `sgrid.rename_dims` to `sgrid.rename` (along with internal xarray call)
- Add test

See commit history for full info

Done to help along #2460

### DIFF
--- a/src/parcels/_datasets/structured/generic.py
+++ b/src/parcels/_datasets/structured/generic.py
@@ -8,7 +8,7 @@ from parcels._core.utils.sgrid import (
     _attach_sgrid_metadata,
 )
 from parcels._core.utils.sgrid import (
-    rename as sgrid_rename_dims,
+    rename as sgrid_rename,
 )
 from parcels._datasets.utils import _attach_sgrid_metadata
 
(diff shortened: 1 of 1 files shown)


erikvansebille (approved):
Looks good to me!

src/parcels/_core/utils/sgrid.py
```diff
@@ -453,14 +453,6 @@ def get_unique_dim_names(grid: Grid2DMetadata | Grid3DMetadata) -> set[str]:
     return dims
 
 
-@overload
```
VeckoTheGecko: These were moved to be directly over the `_metadata_rename_dims` function (as this is how `@overloads` work).

Wasn't flagged before since we're not (yet) running mypy

PR #2470, opened 2026-01-14
Update Ruff rulesets

VeckoTheGecko:
Its been a while since we've fully reviewed the available linting rules in Ruff, and have updated our configs.

See `pyproject.toml` for the new rulesets that we're intentionally selecting and - of selected rulesets - which individual rules we're ignoring.

In the selected rules there are some new commented out rules. These will result in some larger code changes to Parcels. Let's discuss them here, and deliver them in a future PR.

No parent issue.

### DIFF
--- a/tests/test_spatialhash.py
+++ b/tests/test_spatialhash.py
@@ -15,7 +15,7 @@ def test_invalid_positions():
     ds = datasets["2d_left_rotated"]
     grid = XGrid.from_dataset(ds, mesh="flat")
 
-    j, i, _coords = grid.get_spatial_hash().query([np.nan, np.inf], [np.nan, np.inf])
+    j, i, _ = grid.get_spatial_hash().query([np.nan, np.inf], [np.nan, np.inf])
     assert np.all(j == -3)
     assert np.all(i == -3)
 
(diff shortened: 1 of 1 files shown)


src/parcels/_datasets/utils.py
```diff
@@ -4,26 +4,9 @@
...
-def _attach_sgrid_metadata(ds, grid: Grid2DMetadata | Grid3DMetadata):
-    """Copies the dataset and attaches the SGRID metadata in 'grid' variable. Modifies 'conventions' attribute."""
-    ds = ds.copy()
-    ds["grid"] = (
-        [],
-        0,
-        grid.to_attrs(),
-    )
-    ds.attrs["Conventions"] = "SGRID"
-    return ds
```
VeckoTheGecko: Looks like we had a double up of this function which was also defined in the `sgrid.py` module - thanks [F811](https://docs.astral.sh/ruff/rules/redefined-while-unused/) :)

tests/test_spatialhash.py
```diff
@@ -15,7 +15,7 @@ def test_invalid_positions():
     ds = datasets["2d_left_rotated"]
     grid = XGrid.from_dataset(ds, mesh="flat")
 
-    j, i, coords = grid.get_spatial_hash().query([np.nan, np.inf], [np.nan, np.inf])
+    j, i, _coords = grid.get_spatial_hash().query([np.nan, np.inf], [np.nan, np.inf])
```
erikvansebille: For my own learning: I assume this change is because we don't use `coords` in the rest of the function? 

Then why not simply
```suggestion
    j, i, _ = grid.get_spatial_hash().query([np.nan, np.inf], [np.nan, np.inf])
```

Or is that not ruff-compliant?

VeckoTheGecko: > I assume this change is because we don't use `coords` in the rest of the function?

Yes

> `j, i, _ = grid.get_s...`

Good point. Normally I would, but I was unfamiliar with the return of `.query()` and wasn't sure whether it was well documented in the definition. I see now that it is - happy to update this :)

PR #2473, opened 2026-01-15
Pin pixi version in CI

VeckoTheGecko:
Previously CI was pulling in the latest Pixi version and running based on that. This occasionally creates failing builds in (e.g.,) lock generation.

 It would be better to pin this so that we can keep local development in sync with CI and move together.

Changes:
- Pin CI for pixi versions to latest pixi release (0.63.0)
- Fix `pixi.toml` to be compatible with 0.63.0

@erikvansebille and @fluidnumerics-joe - you will need to update your pixi versions. If installed via brew: `brew upgrade pixi`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,14 @@ defaults:
   run:
     shell: bash -el {0}
 
+env:
+  PIXI_VERSION: "v0.63.0"
+
 jobs:
   cache-pixi-lock:
     uses: ./.github/workflows/cache-pixi-lock.yml
+    with:
+      pixi-version: "v0.62.2" # keep in sync with env var above
   unit-test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +16,14 @@ defaults:
...
     shell: bash -el {0}
 
+env:
+  PIXI_VERSION: "v0.63.0"
+
 jobs:
   cache-pixi-lock:
     uses: ./.github/workflows/cache-pixi-lock.yml
+    with:
+      pixi-version: "v0.62.2" # keep in sync with env var above
```
VeckoTheGecko: This wart in the code "keep in sync with above" is unfortunately necessary. Discussion about this and pinning of pixi versions in CI can be seen at https://github.com/pydata/xarray/pull/10888#discussion_r2538934409

PR #2479, opened 2026-01-15
Adding a tutorial on how to manipulate fields using xarray operations

erikvansebille:
During a discussion with @MeikeBos, I realised we don't yet show how to take advantage of xarray operations on fields. So this tutorial shows how to sum two fields (on the same grid) using `xarray` operations. It would be a good place to point users to when they ask about how to best adapt their fields

We could also add more complex `xarray` operations (e.g. vertical averaging of the horizontal flow-field), but then it would become very long


erikvansebille:
Good suggestion. Now done in https://github.com/Parcels-code/Parcels/pull/2479/commits/2801d59140731132e9e5d3169fe2860f79d186e1

PR #2480, opened 2026-01-15
Remove numpydoc validation from Sphinx build

VeckoTheGecko:
I thought the numpydoc validation provided during build was just looking at the public API, however this doesn't seem to be the case (it has behaviour closer to the CLI tool `numpydoc` - which doesn't suit our needs).

We'll need to rely fully on the `pixi run numpydoc-lint` task that we have.

Follow-on from #2474


erikvansebille (approved):
Looks good

PR #2485, opened 2026-01-19
Remove local `cache-pixi-lock` workflow

VeckoTheGecko:
This is now housed as an action in `Parcels-code/pixi-lock` to be moved into the official Pixi org down the line

https://github.com/Parcels-code/pixi-lock/pull/2


.github/workflows/ci.yml
```diff
@@ -46,9 +46,16 @@ jobs:
...
-    uses: ./.github/workflows/cache-pixi-lock.yml
-    with:
-      pixi-version: "v0.62.2" # keep in sync with env var above
+    runs-on: ubuntu-latest
+    outputs:
+      cache-key: ${{ steps.pixi-lock.outputs.cache-key }}
+      pixi-version: ${{ steps.pixi-lock.outputs.pixi-version }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Parcels-code/pixi-lock/create-and-cache@2b823a4a804631370fbde7e6088ee5db71a26c60 # TODO: Update to action in prefix-dev once available
```
VeckoTheGecko: Plans re. moving the repo to `prefix-dev`:

- Pending issues to be resolved/discussed on the `pixi-lock` repo
- (at some point) repo moves to prefix-dev
- We either:
  - delete our copy of the `pixi-lock` repo and update our reference here
  - archive our copy of the `pixi-lock` repo

Either way - this has minimal disruptions to us

PR #2510, opened 2026-02-12
DOCS: Add guide on sharing dataset metadata

VeckoTheGecko:
Added a documentation page with code showing how to share dataset metadata when posting issues.

The code cells in these are run during the doc build, hence testing them.

~~I also revamped our issue templates taking inspiration from Xarray's templates (also linking the new documentation page in these templates

Closes None

(thought it would be good to get this out there before OSM :))

### DIFF
--- a/docs/development/posting-issues.md
+++ b/docs/development/posting-issues.md
@@ -24,7 +24,7 @@ that only occur with specific datasets, having the dataset metadata is very valu
 
 This metadata could include information such as:
 
-- CF compliant metadata
+- the nature of the array variables (e.g., via CF compliant metadata)
 - descriptions about the origin of the dataset, or additional comments
 - the shapes and data types of the arrays
 
--- a/docs/getting_started/tutorial_output.ipynb
+++ b/docs/getting_started/tutorial_output.ipynb
@@ -560,7 +560,7 @@
  "metadata": {
   "celltoolbar": "Metagegevens bewerken",
   "kernelspec": {
-   "display_name": "docs",
+   "display_name": "test-notebooks",
    "language": "python",
    "name": "python3"
   },


.github/ISSUE_TEMPLATE/01_feature.yml
VeckoTheGecko: The motivation for adding more structure to the feature suggestion template is for when v4 is released. It will be important to trace which version of Parcels people are using (e.g., "I want to do xyz, by the way I'm on v3.1.2" then the answer might just be "use v4").

Perhaps this template is a bit verbose. Open to copy edits here or anywhere in this PR

docs/development/posting-issues.md
```diff
@@ -0,0 +1,70 @@
...
+## "Parcels doesn't work with my input dataset"
+
+Parcels is designed to work with a large range of input datasets.
+
+When extending support for various input datasets, or trying to debug problems
+that only occur with specific datasets, having the dataset metadata is very valuable.
+
+This metadata could include information such as:
+
+- CF compliant metadata
```
erikvansebille: Technically, the metadata doesn't have to be CF compliant; and often it won't be. So I would just use 'metadata' here?

docs/development/posting-issues.md
```diff
@@ -0,0 +1,70 @@
...
+As a user with access to your dataset, you would do:
+
+```{code-cell}
+import json
+
+import xarray as xr
+
+# defining an example dataset to illustrate
+# (you would use `xr.open_dataset(...)` instead)
+ds = xr.Dataset(attrs={"description": "my dataset"})
```
erikvansebille: Why not simply do 
```python
ds = xr.open_dataset(...)
```

Or is this code being run in CI?

erikvansebille (approved):
Looks good; only a few minor comments

docs/development/posting-issues.md
erikvansebille: Yep, good solution

PR #2541, opened 2026-03-10
Improve typechecking

VeckoTheGecko:
Supercedes #2422 

This PR adds a bunch of type annotations to the codebase, and centralises a lot of the types that are shared between files to `parcels._typing` (which in some places is imported as `ptyping` which i think is a nice alias).

### DIFF
--- a/src/parcels/_typing.py
+++ b/src/parcels/_typing.py
@@ -43,6 +43,7 @@ KernelFunction = Callable[..., None]
 
 XgridAxis = Literal["X", "Y", "Z"]
 XgcmAxisDirection = Literal["X", "Y", "Z", "T"]
+CfAxis = XgcmAxisDirection
 XgcmAxisPosition = Literal["center", "left", "right", "inner", "outer"]
 XgcmAxes = Mapping[XgcmAxisDirection, "xgcm.Axis"]
 


src/parcels/_reprs.py
```diff
@@ -47,16 +53,16 @@ def field_repr(field: Field, level: int = 0) -> str:
...
-        vector_type           : {fieldset.vector_type!r}
-    {field_repr(fieldset.U, level=1) if not from_fieldset_repr else ""}
-    {field_repr(fieldset.V, level=1) if not from_fieldset_repr else ""}
-    {field_repr(fieldset.W, level=1) if not from_fieldset_repr and fieldset.W else ""}"""
+        name                  : {vector_field.name!r}
+        vector_interp_method  : {vector_field.vector_interp_method!r}
+        vector_type           : {vector_field.vector_type!r}
+    {field_repr(vector_field.U, level=1) if not from_fieldset_repr else ""}
+    {field_repr(vector_field.V, level=1) if not from_fieldset_repr else ""}
+    {field_repr(vector_field.W, level=1) if not from_fieldset_repr and vector_field.W else ""}"""
```
VeckoTheGecko: quick win that was flagged by the typechecking

PR #2557, opened 2026-03-25
Rename sgrid.DimDimPadding to sgrid.FaceNodePadding

VeckoTheGecko:
After doing the diagrams in #2553  , I thought that this naming was more expressive.

Closes None


src/parcels/_core/utils/sgrid.py
```diff
@@ -316,39 +316,39 @@ def get_value_by_id(self, id: str) -> str:
...
+    In the context of a 2D grid, "face" corresponds with an edge.
+
+    Use the .to_diagram() method to visualize the representation.
     """
 
-    dim1: str
-    dim2: str
+    face: Dim
+    node: Dim
     padding: Padding
```
VeckoTheGecko: The test suite for the sgrid parsing etc. is very thorough, so the main thing to review is this block.

PR #2570, opened 2026-04-08
Sharing of representative datasets via compressed Zarr zips 

VeckoTheGecko:
I realised that there is both a better approach to https://github.com/Parcels-code/Parcels/pull/2510, as well as a way that we can pave the way for https://github.com/Parcels-code/parcels-benchmarks/pull/40 ("Future work: better testing").

This PR adds tooling and documentation that allows users to easily share datasets via the issue tracker. These datasets will be first trimmed of redundant array data, then shared as Zipped zarr files using `zarr.storage.ZipStore`.

This allows people to easily share their datasets - including the mesh itself - directly via GitHub (which has an upload limit of 25Mb for zips). Working with the MOI grid data in parcels-benchmarks the resulting dataset was only 26Mb (i.e., the `glamf` and `glamu`).

This PR also removes the old JSON approach as it is no longer needed.

- [x] xref https://github.com/Parcels-code/parcels-benchmarks/pull/40
  - This is a significant step in this direction
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Used it to update the docs (pointing it to the old docs, then to an example usecase) then made manual copy edits.

### DIFF
--- a/docs/development/posting-issues.md
+++ b/docs/development/posting-issues.md
@@ -67,7 +67,8 @@ ds = xr.open_dataset("my_dataset.nc")  # or xr.open_zarr(...), etc.
 # You can customise `except_for` to also retain actual values for specific variables:
 #   except_for='coords'         — keep coordinate arrays (useful for grid topology)
 #   except_for=['lon', 'lat']   — keep a specific list of variables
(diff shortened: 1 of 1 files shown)


erikvansebille (commented):
Looks very good, a few comments below

PR #2574, opened 2026-04-13
Add `parcels._datasets.remote`

### LINKED ISSUE #2571: Update tutorial dataset storage and loading

main

Currently I'm looking at building out tooling for #2570 so that we can add these items easily to the test suite. I will end up needing quite similar tooling to that already in

https://github.com/Parcels-code/Parcels/blob/1c6369438d8623d67722bc34441dde2dd9180041/src/parcels/_tutorial.py

except that in the `_tutorial.py` file it assumes that the underlying data is netcdf files, and there being some sort of implicit folder structure. This isn't really ideal:

- `download_example_dataset(...)` returns the folder in which the data is downloaded. From there users have to open it - which is cumbersome as it means they have to be familiar with the structure of the files, calling `open_dataset`/`open_mfdataset` etc appropriately with the right paths and globs
- users mainly just want to open an xarray dataset straight after downloading it (I don't see a usecase for users needing access to the files themselves)
- adding separate tooling similar to `_tutorial.py` in `tests/utils/_datasets.py` seems like unnecessary duplication

To simplify things, I wonder whether we should migrate all our example datasets to be these zipped zarr stores like in #2570

Changes:
- **make a new branch `v4` in `parcels-examples` repo**
  - unfortunately we can't just use main otherwise that would break people trying to use v3 code
- remove `download_example_dataset(...)` and replace with `open_example_dataset(...)` (the latter returning a dataset object)
  - optionally (but I think would be a good idea), maybe we should delineate between testing and tutorial datasets by doing `open_tutorial_dataset` and `open_testing_dataset` (similarly, `list_tutorial_datasets` and `list_testing_datasets`). This is solely to delineate stability (i.e., as devs, we can confidently add or remove testing datasets. Tutorial datasets can also be used in testing, but shouldn't be removed/changed in breaking ways as that would negatively affect users following tutorials. Testing datasets shouldn't be used in tutorials.

cc @erikvansebille @fluidnumerics-joe keen on your thoughts as this will significantly change how we approach datasets in tutorials and testing

Two separate files. This would be manageable in the short term, but I think that this issue is bets for the longterm maintainence of parcels

### Additional context

_No response_

VeckoTheGecko:
### Description

This PR adds tooling for ingesting remote datasets that are:
- For different purposes
  - testing or tutorials
- Backed by netcdf files or (the new approach) backed by a zipped Zarr store
  - This requires minimal changes to the original data repo (except adding a new folder `data-zarr` for when we add zipped zarr files

This PR also:
- Removes the `download_example_dataset` util (and renames the list_example_dataset util)
- Removes the data downloading utils from the root `parcels` namespace
- Creates a `parcels._datasets.remote` module as well as a thin wrapper module `parcels.tutorial`
- Adds a function `parcels._datasets.remote.open_dataset` which returns an xarray dataset
- So that CI passes, updates all references across the codebase (except references in v3 folders that are yet to be migrated)

### Checklist

- [x] Closes #2571
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Writing docstrings and updating examples to use new ingestion method


src/parcels/_datasets/remote.py
```diff
@@ -0,0 +1,273 @@
...
+    registry=_POOCH_REGISTRY,
+)
+
+
+class _ParcelsDataset(abc.ABC):
+    @abc.abstractmethod
+    def open_dataset(self) -> xr.Dataset: ...
+
+
+class _V3Dataset(_ParcelsDataset):
```
VeckoTheGecko, line 122: I thought that this approach was better than migrating the datasets to Zarr (i.e., working closer to the source netcdf files, without having to think about deprecation cycles and data storage etc.)

docs/user_guide/examples/tutorial_dt_integrators.ipynb
erikvansebille, line 1: But do we need all three fields for all tutorials? I reckon that for many tutorials, we only need the currents (`"CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-cur_anfc"`). For clarity, perhaps not load the other fields in these cases?

Do you want me to go through all the tutorials here and check which files should be loaded, and which don't need to?

docs/user_guide/examples/tutorial_nemo.ipynb
erikvansebille, line 1: Nice, this fieldset creation is much cleaner now!

src/parcels/_datasets/remote.py
```diff
@@ -0,0 +1,273 @@
...
+        self.pup = _ODIE
+        self.path_relative_to_root = path_relative_to_root
+
+    def open_dataset(self) -> xr.Dataset:
+        self.pup.fetch(self.path_relative_to_root)
+        return xr.open_zarr(Path(self.pup.path) / self.path_relative_to_root)
+
+
+def _preprocess_drop_time_from_mesh1(ds: xr.Dataset) -> xr.Dataset:
+    # For some reason on the mesh "NemoNorthSeaORCA025-N006_data/coordinates.nc" there are time dimensions. These dimension also has broken cf-time metadata
```
erikvansebille:
```suggestion
    # For some reason on the mesh "NemoNorthSeaORCA025-N006_data/coordinates.nc" there are two time dimensions. These dimension also has broken cf-time metadata
```

tests/test_index_search.py
```diff
@@ -56,13 +55,8 @@ def test_grid_indexing_fpoints(field_cone):
 
 
 def test_indexing_nemo_curvilinear():
-    data_folder = download_example_dataset("NemoCurvilinear_data")
-    ds = xr.open_mfdataset(
-        data_folder.glob("*.nc4"), combine="nested", data_vars="minimal", coords="minimal", compat="override"
```
erikvansebille, line 61: Maybe I missed it, but where are these xarray options (`combine="nested", data_vars="minimal", coords="minimal", compat="override"`) now set? These are quite important for performance. Without them, opening sets of very large files can take a very long time

erikvansebille (commented):
Looks good; a few small comments below

src/parcels/_datasets/remote.py
VeckoTheGecko: I was actually more confused that the grid had a time dimension at all? its of len 1

erikvansebille: Ah, then I misunderstood your comment.

```suggestion
    # For some reason on the mesh "NemoNorthSeaORCA025-N006_data/coordinates.nc" there are two time dimensions (of length 1). These dimension also has broken cf-time metadata
```

tests/test_index_search.py
VeckoTheGecko: I've now added these. the `combine='nested'` was giving me trouble though - looking at https://docs.xarray.dev/en/stable/generated/xarray.combine_nested.html and https://docs.xarray.dev/en/stable/generated/xarray.combine_by_coords.html#xarray.combine_by_coords it seems like the default combine by coord is what we want anyway?

Our tests are ~not~ now passing after removing `combine='nested'`

docs/user_guide/examples/tutorial_dt_integrators.ipynb
VeckoTheGecko: > But do we need all three fields for all tutorials?

No we don't

---

Actually, since this is all copernicus marine data it should be quite easy for us to just expose this as a single dataset right? There isn't a need to open 3 separate ones, and combine them since there's no grid staggering that affects the dimensions (like for NEMO)

```diff
diff --git a/src/parcels/_datasets/remote.py b/src/parcels/_datasets/remote.py
index 3b9d9c71..f956d750 100644
--- a/src/parcels/_datasets/remote.py
+++ b/src/parcels/_datasets/remote.py
@@ -201,9 +201,7 @@ _DATASET_KEYS_AND_CONFIGS: dict[str, tuple[_V3Dataset, _Purpose]] = dict([
     ("Peninsula_data/P", (_V3Dataset("data/Peninsula_data/peninsulaP.nc"), _Purpose.TUTORIAL)),
     ("Peninsula_data/T", (_V3Dataset("data/Peninsula_data/peninsulaT.nc"), _Purpose.TUTORIAL)),
     ("GlobCurrent_example_data/data", (_V3Dataset("data/GlobCurrent_example_data/*000000-GLOBCURRENT-L4-CUReul_hs-ALT_SUM-v02.0-fv01.0.nc", pre_decode_cf_callable=patch_dataset_v4_compat), _Purpose.TUTORIAL)),
-    ("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-cur_anfc", (_V3Dataset("data/CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-cur_anfc_0.083deg_P1D-m_uo-vo_31.00E-33.00E_33.00S-30.00S_0.49-2225.08m_2024-01-01-2024-02-01.nc"), _Purpose.TUTORIAL)),
-    ("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-so_anfc", (_V3Dataset("data/CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-so_anfc_0.083deg_P1D-m_so_31.00E-33.00E_33.00S-30.00S_0.49-2225.08m_2024-01-01-2024-02-01.nc"), _Purpose.TUTORIAL)),
-    ("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-thetao_anfc", (_V3Dataset("data/CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-thetao_anfc_0.083deg_P1D-m_thetao_31.00E-33.00E_33.00S-30.00S_0.49-2225.08m_2024-01-01-2024-02-01.nc"), _Purpose.TUTORIAL)),
+    ("CopernicusMarine_data_for_Argo_tutorial/data", (_V3Dataset("data/CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-*.nc"), _Purpose.TUTORIAL)),
     ("DecayingMovingEddy_data/U", (_V3Dataset("data/DecayingMovingEddy_data/decaying_moving_eddyU.nc"), _Purpose.TUTORIAL)),
     ("DecayingMovingEddy_data/V", (_V3Dataset("data/DecayingMovingEddy_data/decaying_moving_eddyV.nc"), _Purpose.TUTORIAL)),
     ("FESOM_periodic_channel/fesom_channel", (_V3Dataset("data/FESOM_periodic_channel/fesom_channel.nc"), _Purpose.TUTORIAL)),

```

```diff
diff --git a/docs/getting_started/tutorial_quickstart.md b/docs/getting_started/tutorial_quickstart.md
index 4c0bf02a..cf17e4bb 100644
--- a/docs/getting_started/tutorial_quickstart.md
+++ b/docs/getting_started/tutorial_quickstart.md
@@ -30,11 +30,8 @@ hydrodynamics fields in which the particles are tracked. Here we provide an exam
 [Global Ocean Physics Reanalysis](https://doi.org/10.48670/moi-00021) from the Copernicus Marine Service.
 
 ```{code-cell}
-ds_fields = xr.merge([
-    parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-cur_anfc"),
-    parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-so_anfc"),
-    parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/cmems_mod_glo_phy-thetao_anfc"),
-])
+ds_fields = parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/data")
+
 ds_fields
 ```

```

erikvansebille: Good point! Much simpler like this

erikvansebille (approved):

docs/user_guide/examples/tutorial_dt_integrators.ipynb
VeckoTheGecko: Done in feb67c1

PR #2575, opened 2026-04-13
Bugfix/2568 output fenceposting

### LINKED ISSUE #2568: Particle positions in output are lagged by one `dt` relative to time labels
### Parcels version

https://github.com/Parcels-code/Parcels/commit/afb58dae5652974580da10bfff4323f17ffd027a

### Description

When writing particle trajectories via `ParticleFile`, the position recorded at each output time corresponds to the particle state one `dt` earlier than the time label. For a uniform velocity field `u = U0` where the analytical trajectory is `x(t) = x0 + U0*t`, the final output at `t = T` contains the position at `t = T - dt`. The error is exactly proportional to `dt`.

## Reproducer

```python
import numpy as np
import parcels
import uxarray as ux
import xarray as xr

# --- Build a minimal Delaunay grid ---
NX = 10
lon, lat = np.meshgrid(
    np.linspace(0, 20, NX, dtype=np.float64),
    np.linspace(0, 20, NX, dtype=np.float64),
)
lon_flat, lat_flat = lon.ravel(), lat.ravel()
mask = (
    np.isclose(lon_flat, 0) | np.isclose(lon_flat, 20)
    | np.isclose(lat_flat, 0) | np.isclose(lat_flat, 20)
)
uxgrid = ux.Grid.from_points(
    (lon_flat, lat_flat),
    method="regional_delaunay",
    boundary_points=np.flatnonzero(mask),
)
uxgrid.attrs["Conventions"] = "UGRID-1.0"

# --- Uniform velocity field on face centers ---
U0 = 0.001  # degrees/s
V0 = 0.0
N_TIMESTEPS = 1
TIME = xr.date_range("2000-01-01", periods=N_TIMESTEPS, freq="1h")
zf = np.array([0.0, 1.0])
zc = np.array([0.5])

U = np.full((N_TIMESTEPS, 1, uxgrid.n_face), U0)
V = np.full((N_TIMESTEPS, 1, uxgrid.n_face), V0)
W = np.zeros((N_TIMESTEPS, 2, uxgrid.n_node))

ds = ux.UxDataset({
    "U": ux.UxDataArray(U, uxgrid=uxgrid, dims=["time", "zc", "n_face"],
             coords=dict(time=(["time"], TIME), zc=(["zc"], zc)),
             attrs=dict(location="face", mesh="delaunay", Conventions="UGRID-1.0")),
    "V": ux.UxDataArray(V, uxgrid=uxgrid, dims=["time", "zc", "n_face"],
             coords=dict(time=(["time"], TIME), zc=(["zc"], zc)),
             attrs=dict(location="face", mesh="delaunay", Conventions="UGRID-1.0")),
    "W": ux.UxDataArray(W, uxgrid=uxgrid, dims=["time", "zf", "n_node"],
             coords=dict(time=(["time"], TIME), nz=(["zf"], zf)),
             attrs=dict(location="node", mesh="delaunay", Conventions="UGRID-1.0")),
}, uxgrid=uxgrid)

# --- Run advection ---
T_TOTAL = np.timedelta64(3600, "s")
DT = np.timedelta64(300, "s")
dt_s = 300.0

fieldset = parcels.FieldSet.from_ugrid_conventions(ds, mesh="flat")
pset = parcels.ParticleSet(fieldset, lon=[5.0], lat=[5.0])
pfile = parcels.ParticleFile(store="test.zarr", outputdt=DT)

pset.execute(parcels.kernels.AdvectionRK4, runtime=T_TOTAL, dt=DT,
             output_file=pfile, verbose_progress=False)

# --- Check ---
lon_expected = 5.0 + U0 * 3600.0   # 8.6
lon_lagged   = 5.0 + U0 * 3300.0   # 8.3

print(f"pset.lon[0]        = {pset.lon[0]}")
print(f"analytical(T)      = {lon_expected}")
print(f"analytical(T - dt) = {lon_lagged}")
print(f"error vs T:        = {abs(pset.lon[0] - lon_expected):.6e}")
print(f"error vs T - dt:   = {abs(pset.lon[0] - lon_lagged):.6e}")
# Output:
#   error vs T      ≈ U0 * dt = 3e-1   (proportional to dt)
#   error vs T - dt ≈ 0                 (machine precision)
```

The issue reproduces with the AdvectionEE and  AdvectionRK4 time integrators (I have not tested AdvectionRK45, though I don't believe the integrator choice matters). I suspect there is an issue in the `particleset.execute()` and file io logic.

fluidnumericsJoe:
- [ X ] Closes #2568
- [ X ] Tests added
- [ X ] This PR targets the correct branch (`main` for normal development, `v3-support` for v3 support)

### AI Disclosure

- [ X ] This PR contains AI-generated content.
  - [ X ] I have tested any AI-generated content in my PR.
  - [ X ] I take responsibility for any AI-generated content in my PR.
  - Describe how you used it (e.g., by pasting your prompt): After making changes to dump output after each positionupdate and add initial condition IO, I've used Claude Code to review Issue #2568 with failing tests for context to suggest potential test failure resolution options.


src/parcels/_core/kernel.py
```diff
@@ -221,6 +220,12 @@ def execute(self, pset, endtime, dt):
                     f(pset[repeat_particles], self._fieldset)
                     repeat_particles = pset.state == StatusCode.Repeat
 
+            # apply position/time update only to particles still in a normal state
+            # (particles that signalled Stop*/Delete/errors should not have time/position advanced)
```
erikvansebille, line 218: So was this the fix to the bug in #2568? As simple as that?

src/parcels/_core/kernel.py
```diff
@@ -80,8 +80,7 @@ def __init__(
 
         self._kernels: list[Callable] = kernels
 
-        if pset._requires_prepended_positionupdate_kernel:
-            self.prepend_positionupdate_kernel()
+        self._position_update = self._make_position_update()
```
erikvansebille: Why not directly call `self._make_position_update()`? Why (effectively) rename the function here?

tests/test_advection.py
```diff
@@ -99,7 +98,7 @@ def test_advection_zonal_periodic():
     startlon = np.array([0.5, 0.4])
     pset = ParticleSet(fieldset, pclass=PeriodicParticle, lon=startlon, lat=[0.5, 0.5])
     pset.execute([AdvectionEE, periodicBC], runtime=np.timedelta64(40, "s"), dt=np.timedelta64(1, "s"))
-    np.testing.assert_allclose(pset.total_dlon, 4.1, atol=1e-5)
+    np.testing.assert_allclose(pset.total_dlon, 4.0, atol=1e-5)
```
erikvansebille, line 102: Nice to see that you also fixed all these 'bugs'!

tests/test_uxadvection.py
```diff
@@ -0,0 +1,25 @@
...
+    T = np.timedelta64(3600, "s")
+    dt = np.timedelta64(300, "s")
+    dt_s = 300.0
+
+    fieldset = parcels.FieldSet.from_ugrid_conventions(ds, mesh="flat")
+    pset = parcels.ParticleSet(fieldset, lon=[5.0], lat=[5.0])
+    pfile = parcels.ParticleFile(store="test.zarr", outputdt=dt)
+    pset.execute(integrator, runtime=T, dt=dt, output_file=pfile, verbose_progress=False)
+    expected_lon = 8.6
+    np.testing.assert_allclose(pset.lon, expected_lon, atol=1e-5)
```
erikvansebille: But the last locations in the output are not tested here. Also test that these are as expected?

erikvansebille (commented):
Nice improvement! But it seems that this doesn't (yet) fix the ParticleFile? That now raises some unit test errors. Do you want me to help?

fluidnumericsJoe:
> Nice improvement! But it seems that this doesn't (yet) fix the ParticleFile? That now raises some unit test errors. Do you want me to help?

It'd be great to get your input on those. It looks like I need to have some awareness for particles being injected at different points in time into the simulation for the explicit initial condition write in `particleset.execute()`

tests/test_advection.py
fluidnumericsJoe: With the extra update call that was there internally, there was a ton of these assertions that were right for the wrong reasons :)

src/parcels/_core/kernel.py
fluidnumericsJoe: The fix was really to have explicit initial condition IO and file output after each call to positionupdate. The previous logic, which pre-pended positionupdate after the first call to kernel.execute() inadvertently wrote the initial condition twice, once for the initial time and once for the initial time + dt; this introduced the "off-by-dt" error in the particlefile output.

src/parcels/_core/kernel.py
fluidnumericsJoe: That would be much simpler than what I concocted.

tests/test_particlesetview.py
```diff
@@ -178,5 +178,5 @@ def MoveLon(particles, fieldset):  # pragma: no cover
 
     # Should have deleted edge particles
     assert pset.size < initial_size
-    # Remaining particles should be in the middle range
-    assert np.all((pset.lon >= 0.2) & (pset.lon <= 0.8))
+    # Remaining particles should be in the middle range (with 0.02 of total displacement)
+    assert np.all((pset.lon >= 0.2) & (pset.lon <= 0.82))
```
VeckoTheGecko, line 182: I'm not really sure i understand this change in comment?

VeckoTheGecko (commented):
Should we have an additional test that isolates https://github.com/Parcels-code/Parcels/issues/2568 and tests it? I'm worried that at the moment this only surfaces in position assert statements within other tests

tests/test_uxadvection.py
fluidnumericsJoe: Good point. With the fix in this branch, I think we ought to make sure the output in the particle file matches the pset.lon value after a pset.execute - there should now be consistency here

tests/test_particlesetview.py
fluidnumericsJoe: The `MoveLon` function above explicitly sets `particles.dlon = 0.01` for each pset.execute call. The `pset.execute` call here runs two time steps (two `positionUpdates` so the total displacement is `0.02`) Given the initial particle longitudes are all between 0.2 and 0.8, the maximum longitude is 0.82 . I suppose minimum could be 0.22

src/parcels/_core/kernel.py
fluidnumericsJoe: Resolved in 90cebd3d

tests/test_uxadvection.py
fluidnumericsJoe: Resolved in 0e081238

fluidnumericsJoe:
> Should we have an additional test that isolates #2568 and tests it? I'm worried that at the moment this only surfaces in position assert statements within other tests

This is now covered in `tests/test_uxadvection.py` with commit 0e081238

fluidnumericsJoe:
Failed test looks like a 404 error unrelated to my changes....

PR #2581, opened 2026-04-16
Make `py-triangle` a tutorial dependency

VeckoTheGecko:
I realised that the PR with the tutorial for nested grid's accidentally added `py-triangle` as a core dependency for Parcels when (I think) it should be a tutorial dependency.

Thoughts @erikvansebille ? Do we want py-triangle to be part of the default Parcels installation?

- xref https://github.com/Parcels-code/Parcels/pull/2439

Additionally this PR wraps in a couple minor edits (to PR template and to .gitignore)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,7 @@
+### Description
+
+### Checklist
+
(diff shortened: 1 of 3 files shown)


erikvansebille (approved):
Good catch; py-triangle is indeed a package we only need for the nesting tutorial. Its not even generally needed for nesting; just this specific tutorial...

Issue #2582, opened 2026-04-17
Migrate particlefile output to tabular (via Parquet)


VeckoTheGecko:
We are looking to migrate out particle file output to a tabular output using Parquet. This will have the following benefits:

- simplifying the `particlefile.py` module as particle positions can now be trivially appended
- simplify analysis of trajectories (since trajectory output isn't fundamentally n-dimensional, storing it in Zarr can be quite confusing for those doing post-processing or analysis)
- easier integration into other trajectory analysis tools
- making it easier to implement other parcels features down the line (e.g., restarting simulations)

Tasks:
- [ ] Base implementation (no ability to call pset in a loop, no restarting simulations)
- [ ] Documentation
- [ ] Additional features
- [ ] Conversion tooling: Parquet to CF-trajectory compliant Zarr
  - As part of this migration, we can build tooling that converts this Parquet output to CF-trajectory compliant Zarr output (i.e., as a post-processing step) allowing for users to keep their existing scripts analysis scripts if they'd like, and for easier integration into packages like Trajan.

Additional things to consider:
- How does parquet scale up to large outputs? (potentially useful link: https://arrow.apache.org/docs/python/getstarted.html#working-with-large-data )
- How does integration of this output with other tools (e.g., https://github.com/oj-tooth/lt_toolbox ) in the ecosystem look like

Previous attempt: #1399 

FYI @oj-tooth

Keen to discuss further if people have ideas on what this should look like.

oj-tooth:
Great to see your progress on this @VeckoTheGecko.

Always happy to discuss / contribute to developing a schema and testing this down the line! I plan to refactor LT Toolbox this summer, so happy to start thinking ahead to user requirements.

VeckoTheGecko:
Sounds good @oj-tooth ! By schema, I assume that you mean for the Parquet file and which variables are being tracked for each particle? This for Parcels is heavily customisable by the user since they can add and record their own attributes (see https://github.com/Parcels-code/Parcels/blob/main/src/parcels/_core/particle.py#L125-L167 ). Perhaps we wouldn't be able to have a common schema, but keen to more generally discuss how to integrate Parcels output into LT Toolbox!

[...]

PR #2583, opened 2026-04-20
Base implementation of Parquet file writing

### LINKED ISSUE #2586: Consistent time handling
At the outset of Parcels v4 development, it was first thought that it would be possible to handle different datetime's on the fly - allowing users to use datetime objects (e.g., cftime, np.datetime) within their kernels where appropriate. With this in mind, the `TimeInterval` class was created to help determine bounds for simulations etc..

This approach - however - was discovered to have poor interoperability with numpy among other shortcomings.

In https://github.com/Parcels-code/Parcels/pull/2405 we reverted back to working with float datetimes. Upon further exploring #2583 , I realised that we need better support for how datetimes are handled in Parcels - how those are propogated into kernels - and how those are reflected in the ParticleFile output.

---

#### Flow data

NetCDF and Zarr data formats store datatimes as integers on disk ([netcdf Time Coordinate docs](file:///Users/Hodgs004/coding/repos/cf-conventions/build/cf-conventions.html#time-coordinate)) . These datetimes (which primarily are used for the coordinates in datasets, as opposed to the data variables) are optionally decoded from numeric to datetime objects when being ingested by Xarray into `xr.Dataset` objects ([Xarray docs](https://docs.xarray.dev/en/stable/internals/time-coding.html#cf-time-decoding)). These datetime objects are `np.datetime64` or `cftime` objects.

#### Kernels

Users working with time information will need to calculate timing themselves given the time origin, calendar, and numeric time of the simulation.

#### Trajectory output

When serialising out trajectory data, there are available date types (i.e., [`pyarrow.date64`](https://arrow.apache.org/docs/python/generated/pyarrow.date64.html) and [`pyarrow.timestamp`](https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html)) however no support for cftime objects. Pyarrow does however allow for the storing of dictionary metadata on the column level, allowing us to store CF-time metadata alongside the column (in a similar fashion to what is done for Zarr and NetCDF). This storing of CF-time metadata in the Parquet metadata is something that users would need to be aware of during analysis to fully contextualise their data (which they can do by using cftime alongside their favourite dataframe/tabular data processing library).

I see two options:

1. opt for float numeric output for time variables (irrespective of calendar), attaching appropriate metadata according to CF-time
2. opt for `pyarrow`'s datetime objects where appropriate (for select CFtime calendars) and opt for float numeric output for other cases

I think (1) is preffered for its simpler implementation and conistent behaviour across calendars.

---

Changes:
- Implement (1) from above
- See if we can simplify our internals (e.g., by enforcing that time dimensions in the datasets are numeric and aligned to the same time origin)

VeckoTheGecko:
### Description

This PR introduces Parquet file writing to Parcels.

I still need to work on:
- [x] How to work with cftime output in the Parquet (how does this work with our internal model of time in Parcels? How should it work?)
  - Scoped out in #2586 , and now implemented in this PR. Closes #2586. This PR also includes a function `parcels.read_particlefile(...)` to streamline ingestion into cftime objects.
- Reviewing the `test_particlefile.py` file - are there tests that are no longer needed? What would be the best testing approach here?
  - Will leave for a future PR
- Update documentation
  - Will leave for a future PR

Posting as draft for initial feedback

### Checklist

- [x] Contributes to #2582 
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Just to help with learning the PyArrow API (i.e., used it to create an example script - which I then used as an entry to exploring the docs on pyarrow)


VeckoTheGecko:
@erikvansebille let's table some of these questions for our meeting tomorrow (mainly around datetime serialization in the Parquet file)

VeckoTheGecko:
@erikvansebille I moved `read_particlefile` to the global scope (i.e.,`parcels.read_particlefile()`) as well as added a docstring.

VeckoTheGecko:
(we'll likely get doc failures on this PR - which is to be expected since they haven't been updated)

src/parcels/_core/particlefile.py
```diff
@@ -156,124 +137,37 @@ def write(self, pset: ParticleSet, time, indices=None):
...
-                        data = np.full(arrsize, _DATATYPES_TO_FILL_VALUES[var.dtype], dtype=var.dtype)
-                        data[ids, 0] = particle_data[var.name][indices_to_write]
-                        dims = ["trajectory", "obs"]
-                    ds[var.name] = xr.DataArray(data=data, dims=dims, attrs=attrs[var.name])
-                    ds[var.name].encoding["chunks"] = self.chunks[0] if var.to_write == "once" else self.chunks
-            ds.to_zarr(store, mode="w")
-            self._create_new_zarrfile = False
-        else:
-            Z = zarr.group(store=store, overwrite=False)
-            obs = particle_data["obs_written"][indices_to_write]
```
erikvansebille, line 229: I think `obs_written` is now obsolete and can be removed from the ParticleClass. Saves some memory ;-)

src/parcels/_core/particlefile.py
```diff
@@ -54,53 +60,51 @@ class ParticleFile:
...
 
         self._outputdt = outputdt
 
-        _assert_valid_chunks_tuple(chunks)
-        self._chunks = chunks
+        self._path = path  # TODO v4: Consider https://arrow.apache.org/docs/python/getstarted.html#working-with-large-data - though a significant question becomes how to partition, perhaps using a particle variable "partition"?
+        self._writer: pq.ParquetWriter | None = None
+        if path.exists():
+            # TODO: Add logic for recovering/appending to existing parquet file
+            raise ValueError(f"{path=!r} already exists. Either delete this file or use a path that doesn't exist.")
```
erikvansebille, line 90: This is a change from v3 behaviour, where the default would be to overwrite an existing file. Perhaps we can do that here too, and add an option to replace or append?

src/parcels/_core/particlefile.py
```diff
@@ -298,7 +192,7 @@ def _to_write_particles(particle_data, time):
     )[0]
```
erikvansebille, line 182: Is this section still necessary? If so, should we also fix the warning
```python
[/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:175](https://file+.vscode-resource.vscode-cdn.net/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:175): UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
  np.less_equal(
[/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:180](https://file+.vscode-resource.vscode-cdn.net/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:180): UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
  & np.greater_equal(
[/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:187](https://file+.vscode-resource.vscode-cdn.net/Users/erik/Codes/parcels/src/parcels/_core/particlefile.py:187): UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
  & np.equal(time, particle_data["time"], where=np.isfinite(particle_data["time"]))
```

I didn't worry too much about it before because thought we would refactor this out anyways, but if we keep it we should also address the warning

src/parcels/_core/particlefile.py
erikvansebille: should be "path"?

erikvansebille: This is not an argument to __init__?

src/parcels/_core/particlefile.py
```diff
@@ -54,53 +60,51 @@ class ParticleFile:
...
+        self._path = path  # TODO v4: Consider https://arrow.apache.org/docs/python/getstarted.html#working-with-large-data - though a significant question becomes how to partition, perhaps using a particle variable "partition"?
+        self._writer: pq.ParquetWriter | None = None
+        if path.exists():
+            # TODO: Add logic for recovering/appending to existing parquet file
+            raise ValueError(f"{path=!r} already exists. Either delete this file or use a path that doesn't exist.")
+        if not path.parent.exists():
+            raise ValueError(f"Folder location for {path=!r} does not exist. Create the folder location first.")
+
         self._maxids = 0
         self._pids_written = {}
```
erikvansebille: These `_maxids` and `_pids_written` are not used anymore?

src/parcels/_core/particlefile.py
```diff
@@ -54,53 +60,51 @@ class ParticleFile:
...
-            store = _get_store_from_pathlike(store)
-
-        self._store = store
-
-        # TODO v4: Enable once updating to zarr v3
-        # if store.read_only:
-        #     raise ValueError(f"Store {store} is read-only. Please provide a writable store.")
+        self.extra_metadata = {}
 
         # TODO v4: Add check that if create_new_zarrfile is False, the store already exists
```
erikvansebille: This comment can go? Irrelevant now?

src/parcels/_core/particlefile.py
```diff
@@ -54,53 +60,51 @@ class ParticleFile:
...
 
     def __repr__(self) -> str:
         return particlefile_repr(self)
 
     def set_metadata(self, parcels_grid_mesh: Literal["spherical", "flat"]):
-        self.metadata.update(
+        self.extra_metadata.update(
```
erikvansebille: What's the reason you changed from `metadata` to `extra_metadata`? That is a change in the API that does not appear super-important?

src/parcels/_core/particlefile.py
```diff
@@ -156,124 +137,37 @@ def write(self, pset: ParticleSet, time, indices=None):
...
-        #     )
-        #     return
         if isinstance(time, (np.timedelta64, np.datetime64)):
             time = timedelta_to_float(time - time_interval.left)
-        nparticles = len(particle_data["trajectory"])
         vars_to_write = _get_vars_to_write(pclass)
         if indices is None:
             indices_to_write = _to_write_particles(particle_data, time)
         else:
             indices_to_write = indices
```
erikvansebille, line 143: I did a quick look, but I don't think we call `write` with the `indices` keyword anymore, so this could maybe be simplified? Something to check?

src/parcels/_core/particlefile.py
```diff
@@ -156,124 +137,37 @@ def write(self, pset: ParticleSet, time, indices=None):
...
+            pa.table({v.name: pa.array(particle_data[v.name][indices_to_write]) for v in vars_to_write}),
+        )
 
-def _get_store_from_pathlike(path: Path | str) -> DirectoryStore:
-    path = str(Path(path))  # Ensure valid path, and convert to string
-    extension = os.path.splitext(path)[1]
-    if extension != ".zarr":
-        raise ValueError(f"ParticleFile name must end with '.zarr' extension. Got path {path!r}.")
+        # if len(indices_to_write) == 0: # TODO: Remove this?
+        #     return
```
erikvansebille: Probably good idea to remove if it's commented out already?

src/parcels/_core/particlefile.py
```diff
@@ -298,7 +192,7 @@ def _to_write_particles(particle_data, time):
     )[0]
 
 
-def _get_calendar_and_units(time_interval: TimeInterval) -> dict[str, str]:
+def _get_calendar_and_units(time_interval: TimeInterval) -> dict[str, str]:  # TODO: Remove?
```
erikvansebille: I don't think it's used anywhere in the codebase, so can indeed be removed

tests-v3/test_advection.py
erikvansebille, line 1: why change the tests in the v3-directory? We normally don't touch these, since they are broken already (also apply to other two files)

tests/test_particlefile.py
```diff
@@ -44,50 +47,36 @@ def fieldset() -> FieldSet:  # TODO v4: Move into a `conftest.py` file and remov
...
-    ds = xr.open_zarr(tmp_zarrfile)
-    assert ds.time.values[0, 1] == np.timedelta64(1, "s")
+    table = pq.read_table(tmp_parquet)
+    assert table.schema.field("time").metadata[b"units"] == b"seconds"
+    assert b"calendar" not in table.schema.field("time").metadata
+    assert table["time"].to_numpy()[1] == 1.0
 
 
-def test_pfile_array_remove_particles(fieldset, tmp_zarrfile):
+@pytest.mark.skip("Keep or remove? Introduced in 5d7dd6bba800baa0fe4bd38edfc17ca3e310062b ")
```
erikvansebille: I think ti's an important test to keep in. Or does it fail now?

tests/test_particlefile.py
```diff
@@ -334,19 +302,20 @@ def IncreaseAge(particles, fieldset):  # pragma: no cover
...
 
     pset.execute(IncreaseAge, runtime=np.timedelta64(npart * 2, "s"), dt=np.timedelta64(1, "s"), output_file=ofile)
 
-    ds = xr.open_zarr(tmp_zarrfile)
-    age = ds["age"][:].values.astype("timedelta64[s]")
-    ds_timediff = np.zeros_like(age)
-    for i in range(npart):
-        ds_timediff[i, :] = ds.time.values[i, :] - time[i]
-    np.testing.assert_equal(age, ds_timediff)
+    # df = pd.read_parquet(tmp_parquet)
```
erikvansebille: remove stray debug statement?

erikvansebille (commented):
Very nice work. And as I'm going through updating all the documentation, it seems to work smoothly too!

A few small comments below

tests/test_particlefile.py
```diff
@@ -187,7 +154,7 @@ def test_variable_written_once():
 @pytest.mark.skip(reason="Pending ParticleFile refactor; see issue #2386")
```
erikvansebille, line 150: Does this test work again? If not out of the box, perhaps we should discuss what we actually expect from output of a looped pset.execute...

src/parcels/_core/particle.py
erikvansebille, line 165: Now that we don't comply with the cf-conventions in the output anyways, should we consider to rename this to e.g. `particleid` or even `pid`, which is more internally logical to Parcels. I always found the name `trajectory` a bit strange...

src/parcels/_core/particlefile.py
VeckoTheGecko: Yeah true, I'll revert

tests-v3/test_advection.py
VeckoTheGecko: Yes - will fix. It was caught in the crossfire of my find-replace

src/parcels/_core/particle.py
VeckoTheGecko: tracking in https://github.com/Parcels-code/Parcels/issues/2592

src/parcels/_core/particlefile.py
VeckoTheGecko: tracking in https://github.com/Parcels-code/Parcels/issues/2593

src/parcels/_core/particlefile.py
VeckoTheGecko: Is `ParticleFile.write` public API? Wouldn't users want to use this?

src/parcels/_core/particlefile.py
```diff
@@ -298,33 +182,64 @@ def _to_write_particles(particle_data, time):
...
 
-    attrs = {"units": units}
-    if calendar is not None:
-        attrs["calendar"] = calendar
+    attrs = {k.decode(): v.decode() for k, v in time_field.metadata.items()}
 
-    return attrs
+    df = pd.read_parquet(path)
+    if not decode_times:
+        return df
```
erikvansebille, line 237: Would it not make sense to move this check to earlier in the function? If the user doesn't want to decode the times, most of the code above is not necessary?

src/parcels/_core/particlefile.py
erikvansebille: Yes, particiclefile.write() is public, but they can always call with a ParticleView in order to subset. Something like (not tested)

```python
pfile.write(pset[some_condition])
```

instead of
```python
pfile.write(pset, indices=np.where(pset==some_condition))
```

tests/test_particlefile.py
VeckoTheGecko: This test was introduced in 4b28ddb1a4a04c440031773255e1cb4086e3de0b . I'm ok cutting tests that are no longer valuable (/tested specific implementation bugs from the old codebase). I don't really understand this test - do you think its still valuable?

> perhaps we should discuss what we actually expect from output of a looped pset.execute...

Agreed. Should we open a tracking issue? Whatever we decide on that front we can add separate tests

erikvansebille: Yes, I think it is a valuable test (as it tests what happens in the particle file when `pset.execute()` is run in a for-loop). But perhaps better for a later PR

src/parcels/_core/particlefile.py
VeckoTheGecko: I would prefer to keep it here so that it checks that the metadata is valid as well, particlefile output should never contain corrupted time metadata anyway and this is useful for us

tests/test_particlefile.py
VeckoTheGecko: For context, this is the description of the commit:

```
commit 5d7dd6bba800baa0fe4bd38edfc17ca3e310062b
Author: Erik van Sebille <erikvansebille@gmail.com>
Date:   Fri Feb 17 09:23:56 2017 +0000

    Fixed a bug in JIT ParticleSet.remove
    
    Seems that particles need to be deleted before updating _cptr in JIT
    mode. Otherwise, ordering goes wrong. This wasn’t spotted before since
    we only deleted the last particle
    
    Now added a unit test to test_particle_sets where a particle somewhere
    in the middle of the ParticleFile is removed. Also use this test to
    check array-style writing of ParticleFile
```

I'll fix the test and enable it

tests/test_particlefile.py
VeckoTheGecko: And perhaps rename from `test_pset_repeated_release_delayed_adding_deleting` - since I find that quite confusing

src/parcels/_core/particlefile.py
VeckoTheGecko: The `pset` is not the same as the particle view - we don't support `pset[indices]`

erikvansebille: OK, good catch; I meant `pfile.write(particles[some_condition])` (in a kernel). But whether this works is to be tested

It's not a huge deal (and you're right it's a breaking change to the v3 API), but I think it would be cleaner to not have the extra option `indices` in `.write()` if we don't expect anyone to use it. This is the moment to update our API, if we want to

erikvansebille (approved):

PR #2595, opened 2026-04-24
Migrate back to mypy and enable typechecking in CI

VeckoTheGecko:
### Description

xref https://github.com/Parcels-code/Parcels/pull/2540 

After trying out `ty` some more, I think it's best if we go for mypy. It has better support in the ecosysytem, and ty was flagging a lot of 'errors' within xarray internals causing noise.

Changes:
- Switch from ty to mypy
- Enable `mypy` in CI (i.e., we will get red crosses if it fails)
- Curate the mypy settings taking inspiration from other open source projects
- Fix a bunch of typing errors
- Add an ignore list on files (as we improve typing across the codebase, we will reduce this ignore list)

### Checklist

- [x] Closes None


erikvansebille (approved):
Changes look good. What will the workflow be? Will we get errors when pushing PRs that fail type checking? Or will the process be more graceful?

VeckoTheGecko:
> Will we get errors when pushing PRs that fail type checking?

Yes, we will get errors that fail typechecking. Type checking will be like unit tests in this regard.

Devs will need to be familiar to some extent with how typechecking in Python works (how to fix and how to supress warnings), or maintainers will need to push edits to fix the typing should contributions mis-specify them

VeckoTheGecko:
> What will the workflow be?

You can do `pixi shell` then `mypy` to run the typechecking.

I realise I need to review the contributing instructions. Will do so before merging.

VeckoTheGecko:
> I realise I need to review the contributing instructions. Will do so before merging.

Just reviewed these. They're already up to date.

PR #2602, opened 2026-04-28
Refactor Parcels time strategies into tests.strategies submodule

VeckoTheGecko:
### Description

This makes it easier to share these strategies between tests.

Also moved `test_datasets.py` to `datasets/test_structured.py`. This is because we will want to have tests for creating and manipulating output data as well (as part of https://github.com/Parcels-code/Parcels/issues/2600 ) so this structure would be nice. At the end of last week I started work on https://github.com/Parcels-code/Parcels/issues/2600 , but since decided to stop in favour of other work. I think these changes are good to get into main regardless.

### Checklist

- [x] Closes None


tests/utils/test_time.py
```diff
@@ -6,62 +6,15 @@
 import pytest
 from cftime import datetime as cftime_datetime
 from hypothesis import given
-from hypothesis import strategies as st
 
+import tests.strategies as pst  # parcels strategies
```
VeckoTheGecko, line 10: This import alias is common convention for Hypothesis (and packages that provide hypothesis strategies).

```py
import hypothesis.strategies as st
import xarray.testing.strategies as xrst

# and here
import tests.strategies as pst # <-- parcels strategies
```

erikvansebille (approved):

Issue #2600, opened 2026-04-28
Parquet to CF-trajectory compliant Zarr conversion tooling


VeckoTheGecko:
As part of this migration, we can build tooling that converts this Parquet output to CF-trajectory compliant Zarr output (i.e., as a post-processing step) allowing for users to keep their existing scripts analysis scripts if they'd like, and for easier integration into packages like Trajan.

It's important that this tooling is able to handle large datasets without saturating working memory.

erikvansebille:
Perhaps have a look at https://oj-tooth.github.io/lt_toolbox/tutorials/tutorial_3_OceanParcels_output_with_TrajFrames.html#Converting-larger-than-memory-OceanParcels-output-.zarr-stores-to-multiple-file-.parquet-stores for inspiration on how @oj-tooth has implemented the inverse process (from Zarr to Parquet)

PR #2607, opened 2026-04-30
Update the documentation to use parquet output

erikvansebille:
### Description
This PR updates all the documentation and tutorial notebooks to parse the parquet output introduced in #2600, as tracked in #2582. It also updated the `parcels.read_particlefile()` to use polars, which scales better for large output files
### Checklist

- [x] Contributes to #2582 
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Help with how to use polars


docs/getting_started/tutorial_quickstart.md
```diff
@@ -123,11 +123,11 @@ Before starting the simulation, we must define where and how frequent we want to
 We can define this in a {py:obj}`parcels.ParticleFile` object:
 
 ```{code-cell}
-output_file = parcels.ParticleFile("output-quickstart.zarr", outputdt=np.timedelta64(1, "h"))
+output_file = parcels.ParticleFile("output-quickstart.parquet", outputdt=np.timedelta64(1, "h"))
 ```
 
-The output files are in `.zarr` [format](https://zarr.readthedocs.io/en/stable/), which can be read by `xarray`.
-See the [Parcels output tutorial](./tutorial_output.ipynb) for more information on the zarr format. We want to choose
+The output files are in `.parquet` [format](https://parquet.apache.org/), which can be read by `polars`.
```
VeckoTheGecko: Would this be a good place to link to Polars? 
```suggestion
The output files are in `.parquet` [format](https://parquet.apache.org/), which can be read by [Polars](https://pola.rs/).
```

I don't think we link to it yet in the docs here

docs/getting_started/tutorial_quickstart.md
```diff
@@ -13,12 +13,12 @@ read more, we have a [concepts overview](./explanation_concepts.md) discussing t
 
 ## Imports
 
-Parcels depends on `xarray`, expecting inputs in the form of [`xarray.Dataset`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html)
-and writing output files that can be read with xarray.
+Parcels depends on `xarray`, expecting inputs in the form of [`xarray.Dataset`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html). Output files can be read with `pandas`.
```
VeckoTheGecko:
```suggestion
Parcels depends on `xarray`, expecting inputs in the form of [`xarray.Dataset`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html). Output files can be read with `polars`.
```

docs/getting_started/tutorial_quickstart.md
```diff
@@ -155,23 +155,22 @@ pset.execute(
 To start analyzing the trajectories computed by **Parcels**, we can open the `ParticleFile` using `xarray`:
```
VeckoTheGecko: This needs to be updated from "xarray"

src/parcels/_core/particlefile.py
```diff
@@ -232,14 +233,18 @@ def read_particlefile(path: PathLike, decode_times: bool = True) -> pd.DataFrame
...
     var = xr.Variable(("time",), values, attrs)
     values = xr.coders.CFDatetimeCoder(time_unit="s").decode(var).values
-
-    df["time"] = values
+    if "since" in attrs["units"]:
+        values = values.astype("datetime64[ns]")
+        df = df.with_columns(pl.Series("time", values, dtype=pl.Datetime("ns")))
+    else:
+        values = values.astype("timedelta64[ns]") * 1e9
+        df = df.with_columns(pl.Series("time", values, dtype=pl.Duration("ns")))
```
VeckoTheGecko, line 248: I don't think this works properly with cf-time variables, and think that it will silently fail by providing incorrect times. I think its worth updating the docstring and also adding a check (if the calendar in the metadata isn't supported (ie. is CF specific) raise a `NotImplementedError`) since this is quite a user facing function.

tests/utils.py
```diff
@@ -160,7 +161,7 @@ def assert_cftime_like_particlefile(parquet_path: Path) -> None:
     df = parcels.read_particlefile(parquet_path, decode_times=True)
 
     # check first value (and hence rest of array) is what we expect
-    assert isinstance(df["time"].values[0], (cftime.datetime, np.datetime64)), (
+    assert isinstance(df["time"][0], (cftime.datetime, datetime)), (
         "CF-time values in Parquet did not get properly decoded. Are the attributes correct?"
     )
```
VeckoTheGecko, line 167: This assert should be updated pending the discussion from the other comment on the `read_particlefile` function.

docs/getting_started/tutorial_quickstart.md
VeckoTheGecko, line 223: This output is different to what is was before.

What are we highlighting with the red circles here?

(before on the left, this PR on the right)

Image

docs/getting_started/tutorial_output.ipynb
VeckoTheGecko, line 1: I've gone through this - really nice update! I think its quite clear

docs/user_guide/examples/tutorial_nemo.ipynb
VeckoTheGecko, line 1: FYI the colourbar is different here - seems to be better now (fits the range of values better)

Image

docs/user_guide/examples/tutorial_nestedgrids.ipynb
VeckoTheGecko, line 1: FYI The final plot here doesn't have black lines through it

Image

I'm happy either way

VeckoTheGecko (commented):
As part of review I've looked both at the code, as well as visually compared plots before and after.

I've gone through and pushed some edits which were quite straight-forward:

- 1b35bf9ff Fixing a notebook
- d977c88e7

Now the docs builds are passing


Other than that, I have some small comments - nothing major.

---

Given we're now using Polars in the docs, the tests, and in the `read_particlefile` function - I think its easiest just for us to add it as a core dependency to Parcels. We could make it an optional dependency, but we don't really have the tooling for that in Parcels (and I don't think its worth adding the tooling in this case).

If we add as a core dependency:
- Update `pyproject.toml` and `pixi.toml` (`run-dependencies` to `= ">=1.31.0"` and `feature.minimum.dependencies` to `= "1.31.*"`)
- Update `recipe.yaml`

I'm happy to make those updates.

docs/user_guide/examples/tutorial_nestedgrids.ipynb
erikvansebille: Good catch, was an issue with the `zorder`. Now fixed in afb4ce8fff5e97df0468d77c91a8fbd996e79fc0

docs/user_guide/examples/tutorial_nemo.ipynb
erikvansebille: Also a good catch, was an issue with the `vmin` and `vmax` not set explicitly, now that we use a for-loop for plotting the trajectories. Fixed in 12fb3b76ac9554e71ae1c1ece157054b08acfdc7

VeckoTheGecko: I was mainly pointing out there there was a difference. I think the rightmost plot is better no? I'm not sure why we should be setting vmin and vmax explicitly?

erikvansebille: No, setting vmin and max _is_ important. Because every trajectory is now plotted by itself, they will otherwise each use their 'own' vmin and max; but the colorbar is then only appropriate for the _last_ particle trajectory. This leads to wrong interpretations! So the fix in 12fb3b7 is important

VeckoTheGecko: Ahh, gotcha :)

docs/getting_started/tutorial_quickstart.md
erikvansebille: Fixed in ff12ea07a52acb89095c93ddc85013cb14531d11

tests/utils.py
erikvansebille: Added a TODO in 0241c28c06a7d1fd3b63b5b844ca46e341bf773f .

docs/getting_started/tutorial_quickstart.md
erikvansebille: Updated in 7cfad91e814c91b241cfbd2eb46916b4cd613f66

docs/getting_started/tutorial_quickstart.md
erikvansebille: Updated in 7cfad91e814c91b241cfbd2eb46916b4cd613f66

src/parcels/_core/particlefile.py
erikvansebille: Good point, but could I ask you to add such a `NotImplementedError`, @VeckoTheGecko? You are much better versed in cftime intricacies than I am

VeckoTheGecko: Let's postpone for another PR. Tracking in #2623

VeckoTheGecko:
> Given we're now using Polars in the docs, the tests, and in the `read_particlefile` function - I think its easiest just for us to add it as a core dependency to Parcels. We could make it an optional dependency, but we don't really have the tooling for that in Parcels (and I don't think its worth adding the tooling in this case).
> 
> If we add as a core dependency:
> 
> * Update `pyproject.toml` and `pixi.toml` (`run-dependencies` to `= ">=1.31.0"` and `feature.minimum.dependencies` to `= "1.31.*"`)
> * Update `recipe.yaml`
> 
> I'm happy to make those updates.

Do you agree with this @erikvansebille ?

erikvansebille:
Yes sorry, missed that comment about the polars dependency, and merged too fast. Can you make these updates in a new PR?

VeckoTheGecko:
All good, can do :)

PR #2609, opened 2026-04-30
Remove to_write="once" option

### LINKED ISSUE #2608: Removing support for to_write="once" option in ParticleFile

v4

Now that we moved to a tabular output format in #2582, I don't think it makes sense anymore to keep support for the `to_write="once"` option in `Variable`. For example, `particle_id` (formerly `trajectory`) is now set as `to_write="once"`, but this is actually _not_ desired in the output file because it is an important key to subset the output by, so would need to be stored on every row

Also for other variables stored only once, the effort of generating code to then compute these for every observation on a trajectory would probably be more complex than the little bit of storage saved

I suggest that, at least for the time-being, we remove the option `to_write="once"` in v4. We can always bring a similar function back if users desire

Keep it in; but then it needs to be fixed because currently it doesn't even work (i.e. Variables with `to_write="once"` are still written on all rows)

### Additional context

_No response_

erikvansebille:
### Checklist

- [x] Closes #2608

### AI Disclosure

### DIFF
--- a/docs/user_guide/examples/tutorial_interaction.ipynb
+++ b/docs/user_guide/examples/tutorial_interaction.ipynb
@@ -112,7 +112,7 @@
     "# Create custom particle class with extra variable that indicates\n",
     "# whether the interaction kernel should be executed on this particle.\n",
     "InteractingParticle = parcels.Particle.add_variable(\n",
-    "    parcels.Variable(\"attractor\", dtype=np.bool_, to_write=\"once\"),\n",
+    "    parcels.Variable(\"attractor\", dtype=np.bool_),\n",
     ")\n",
     "\n",
     "attractor = [\n",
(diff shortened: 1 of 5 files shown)


VeckoTheGecko (commented):
Looks good. Can we also update `docs/user_guide/examples/tutorial_sampling.ipynb`?

> In some simulations only the particles initial value within the field is of interest: the variable does not need to be known along the entire trajectory. To reduce computing we can specify the `to_write` argument to the temperature `Variable`. This argument can have three values: `True`, `False` or `'once'`. It determines whether to write the `Variable` to the output file. If we want to know only the initial value, we can enter `'once'` and only the first value will be written to the output file.

Happy to merge after that's been updated :)

erikvansebille:
Sure, I had removed that section as part of the #2607, but have now also done it here. This part was broken anyways, with the new kernel loop in #2575, which meant that initial values were NaN (will open another issue about this very soon)

VeckoTheGecko:
> I had removed that section as part of the #2607

Ah, didn't see that. Sorry about that

PR #2618, opened 2026-05-01
Add CI workflow for flaky runs

### LINKED ISSUE #2617: Separate CI run for flaky tests
A bunch of the tests in `test_remote.py` rely on connection to GitHub to check for data. These are regularly failing with 502 errors due to GitHub's reduced uptime https://www.githubstatus.com/ .

I think we should split out flaky tests into a separate CI run that isn't a "required check". In pytest this would just be an extra mark `@pytest.mark.flaky`.

This is the same approach used in xarray.

let me know what you think @erikvansebille

VeckoTheGecko:
### Changes

- Updates the `pixi run tests` task so that flaky tests aren't run by default
- Adds a CI workflow for flaky tests
- Marks `test_pooch_registry_url_reponse` as flaky

### Checklist

- [x] Closes #2617 
- [x] Tests added

### AI Disclosure

None used

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
           name: Unittest report ${{ matrix.os }}-${{ matrix.pixi-environment }}
           path: ${{ env.COVERAGE_REPORT }}
   flaky-unit-test:
-    name: "Falky unit tests: ${{ matrix.os }} | pixi run -e ${{ matrix.pixi-environment }} tests"
+    name: "Falky unit tests: ${{ matrix.os }} | pixi run -e ${{ matrix.pixi-environment }} tests -m 'flaky'"
     runs-on: ${{ matrix.os }}-latest
     needs: [cache-pixi-lock]
     permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -126,6 +126,66 @@ jobs:
...
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: .hypothesis/
+          key: cache-hypothesis-${{ runner.os }}-${{ github.run_id }}
+          restore-keys: |
+            cache-hypothesis-${{ runner.os }}-
+      - name: Unit test
+        id: unit-test
+        run: |
+          pixi run -e ${{ matrix.pixi-environment }} tests -m 'flaky' -v -s --cov=parcels --cov-report=xml --html="${COVERAGE_REPORT}" --self-contained-html
```
VeckoTheGecko, line 168: This is basically equivalent to `pytest -m 'not flaky' -m 'flaky'` (the second flag overrides the first one)

.github/workflows/ci.yml
```diff
@@ -126,6 +126,66 @@ jobs:
         with:
           name: Unittest report ${{ matrix.os }}-${{ matrix.pixi-environment }}
           path: ${{ env.COVERAGE_REPORT }}
+  flaky-unit-test:
```
VeckoTheGecko, line 129: This workflow is mostly the same as the unit test one. I kept in the coverage and the test reporting (in case we mark more tests as flaky in future, it would be handy to have coverage and this output I think )

.github/workflows/ci.yml
```diff
@@ -126,6 +126,66 @@ jobs:
...
+    name: "Falky unit tests: ${{ matrix.os }} | pixi run -e ${{ matrix.pixi-environment }} tests -m 'flaky'"
```
erikvansebille:
```suggestion
    name: "Flaky unit tests: ${{ matrix.os }} | pixi run -e ${{ matrix.pixi-environment }} tests -m 'flaky'"
```

erikvansebille (approved):
One small type-o. Looks good otherwise!

PR #2622, opened 2026-05-01
Check writing of initial values of sampled Variables

### LINKED ISSUE #2619: Initial values of sampled fields are not in output file
### Parcels version

v4-alpha

### Description

Since #2568 which moved the initial `pfile.write()` to _before_ the kernels are executed, we are back at the problem that particle Variables that are used to sample fields start with the initial value of these particles (which is generally not the field value); see also example below

Note that this was one of the main reasons to change the kernel loop in v3 (#1402) so that the writing to the particle file is only done _after_ the kernels are executed. But that created other problems that now seem fixed in #2568. 

In any case, we need to come up with a solution to the issue that initial values are not written. Options could be
1. again rethink the kernel loop and consider moving the writing until after the kernels have been executed
2. find a way to automatically execute all the sampling kernels before going into the for-loop in the `pset.execute()`
3. teach users that they first need to execute their sampling kernels before they run a pset.execute (this is [how it was like in v2](https://docs.oceanparcels.org/en/v2.4.2/examples/tutorial_sampling.html#Sampling-initial-values))

Option 3 is the easiest to implement, but will require most effort from users. I have no idea yet how to implement option 2.

### Code sample

```python
import numpy as np

import parcels
import parcels.tutorial

ds_fields = parcels.tutorial.open_dataset(
    "CopernicusMarine_data_for_Argo_tutorial/data"
)

# Convert to SGRID-compliant dataset and create FieldSet
fields = {"thetao": ds_fields["thetao"]}

SampleParticle = parcels.Particle.add_variable(
    parcels.Variable("temperature", dtype=np.float32, initial=np.nan)
)

def SampleT(particles, fieldset):
    particles.temperature = fieldset.thetao[
        particles.time, particles.z, particles.lat, particles.lon
    ]

pset = parcels.ParticleSet(
    fieldset=fieldset, pclass=SampleParticle, lon=[32, 33], lat=[-31.5, -30.5]
)

output_file = parcels.ParticleFile("sampletemp.parquet", outputdt=np.timedelta64(1, "h"))

pset.execute(
    SampleT,
    runtime=np.timedelta64(4, "h"),
    dt=np.timedelta64(30, "m"),
    output_file=output_file,
)

df = parcels.read_particlefile(output_file._path)
df
```

Output (note the NaNs on the first two rows):
Image

erikvansebille:
### Description
As discussed in #2619, there is a bug in that the initial values of sampled Variables are not in the particle file. This PR provides a (failing) unit test to show that

### Checklist

- [x] Closes #2619 
- [x] Tests added

### AI Disclosure

None


docs/user_guide/examples/tutorial_sampling.ipynb
```diff
@@ -11,8 +11,8 @@
     "In this tutorial we will go through how particles can sample `Fields`, using temperature as an example. Along the way we will get to know the parcels class `Variable` (see [here](https://parcels.readthedocs.io/en/latest/reference/particles.html#parcels.particle.Variable) for the documentation) and some of its methods. This tutorial covers several applications of a sampling setup:\n",
```
j-atkins: Curious about how to refer to **P**arcels in text... should it be capitalised or not?

I've seen it with capital P in other messaging e.g. warnings:

>/tmp/ipykernel_2767/2519569428.py:8: UserWarning: This is an alpha version of Parcels v4. The API is >not stable and may change without deprecation warnings.

docs/user_guide/examples/tutorial_sampling.ipynb
j-atkins: Maybe clarify that temperature is `thetao` in code blocks below, for avoidance of doubt? E.g. 

```
... We have a CopernicusMarine dataset with surface ocean velocities and the corresponding potential temperature (**"thetao"**) stored in netcdf files in the parcels example dataset repository...
```

docs/user_guide/examples/tutorial_sampling.ipynb
```diff
@@ -200,6 +200,78 @@
     "plt.show()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sampling initial Field values\n",
+    "\n",
+    "When creating a ParticleSet, the initial values of `Variables` are set as a constant (default `0`, but in the example above set to `np.nan`). Those will also be the values that appear as first entry in the ParticleFile."
```
j-atkins: Suggested change:

```
Those will also be the values that appear as first entry in the ParticleFile, **as seen when printing the initial values:**
```

docs/user_guide/examples/tutorial_sampling.ipynb
```diff
@@ -200,6 +200,78 @@
...
+    "        for traj in df.partition_by(\"particle_id\", maintain_order=True)\n",
+    "    ]\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you want to sample the initial value of a field, you can do so by calling the field with the ParticleSet as an argument. For example, in, the code above you would add one line (`pset.temperature = fieldset.thetao[pset]`) to sample the initial temperature values, beofe calling `pset.execute(...)`."
```
j-atkins: Typo here: "beofe" -> "before"

j-atkins (commented):
Looks good, a nice explanation! Just some tiny comments below

erikvansebille:
> Looks good, a nice explanation! Just some tiny comments below

Thanks! All addressed in 2c2dd989b0d959034ebc39d7a4ed61da123910bc

VeckoTheGecko (approved):

PR #2634, opened 2026-05-18
Add hypothesis strategy for SGRID dataset generation

VeckoTheGecko:
### Description

This PR:

- Adds a hypothesis strategy for SGRID _dataset_ generation under `parcels/_datasets/structured/strategies.py`
- Moves the strategies such that they are part of the `parcels` module (instead of the `tests` module) - `tests.strategies` -> `parcels._strategies`
  - Since the `_datasets` folder is in the Parcels package, I think this structure makes the most sense
  - Informatively errors out if someone tries to import strategies without having `hypothesis` installed

This in future will allow us to have more property based testing extending to full simulations - which would be really helpful for both improving the robustness and maintainability of the Parcels test suite.

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): For writing `tests/datasets/test_strategies.py`


src/parcels/_strategies/__init__.py
```diff
@@ -0,0 +1,13 @@
+# isort: skip_file
+
+try:
+    import hypothesis  # noqa: F401
+except ImportError as err:
+    err.add_note(
+        "To use strategies you must have hypothesis installed. Install it from PyPI, Conda, or using your preffered package manager."
+    )
+    raise err
```
erikvansebille, line 9: Why is this needed? Wouldn't `hypothesis` simply be part of our Pixi install?

src/parcels/_strategies/sgrid.py
erikvansebille, line 1: Perhaps this is just good practice, but I'm surprised that the strategies are moved out of the tests directory. Would they not more logically belong there?

erikvansebille (approved):
Looks good; just two questions below

src/parcels/_strategies/__init__.py
VeckoTheGecko: Its part of our Pixi install, but its not a Parcels run dependency (i.e., its not in `pixi.toml::run-dependencies` or the `recipe.yaml`/`pyproject.toml`. 

This is our first "optional dependency" for Parcels (i.e., a part of the codebase that needs a specific package in order to fulfill a function, but where it doesn't make sense to include it for everyone since most people wont use the specific function).

People doing `conda install parcels` then `import parcels._strategies` will encounter this more informative error message.

src/parcels/_strategies/sgrid.py
VeckoTheGecko: Yeah - good question.

Most of our dataset generation code we have shipping with Parcels via `parcels._datasets`. This means that users can have easy access to Parcels-compatible example datasets and dataset generation for posting issues etc. . When adding strategies for creating example datasets, I needed to use existing strategies that were in the `tests` module. Since I cant put code in `parcels` that depends on the `tests` module - the options I saw were:

1. Leave the strategies where they are in tests and carve out a part in `tests/...` for housing these dataset strategies. This would mean that some datasets are used via `parcels/_datasets/...` and some via `tests/...` which I felt was confusing for devs
2. Move the dataset generation to `tests` - foregoing the benefit of users having easy access to datasets
3. Move the strategies to `parcels`

I thought (3) was the best. It might seem weird to be putting "test" code in the parcels release, but its actually not that weird - some projects even put their whole test suites in the release itself.

src/parcels/_strategies/__init__.py
erikvansebille: OK,  I see. But then (in a next PR?) fix the type-o `preffered` to `preferred`?

PR #2638, opened 2026-05-20
Add SGRID Xarray accessor

VeckoTheGecko:
### Description

As I work to improve our performance, I would like to do significant work on our software architecture to make things easier to read, maintain, and refactor.

And to help with this re-organisation - I think it would better if we rely more on SGRID-based fieldset creation in our test suite. This will mean that refactoring changes down the line will require little to no updates to the test suite, will result in less setup code in the test suite, and improve readability.

This PR adds an SGRID Xarray dataset accessor to help inspect, and manipulate Xarray SGRID datasets.

Namely:

- `ds.sgrid.metadata` providing direct access to parsed SGRID metadata
- `ds.sgrid.rename` allowing the renaming of dataset dimensions alongside the renaming of associated metadata (replaces the old `parcels.sgrid.rename` function)
- `ds.sgrid.isel` selection of spatial dimensions (either by node or by face) which in conjunction selects the other dimension. Allows for the easy subsetting of regions and downsampling of data.
  - Subsetting, slicing with a step (aka. downsampling) etc. works out of the box for datasets that have padding as HIGH or LOW as there is a 1-1 correspondence between nodes and faces. Slicing with integer lists also works for this reason
  - With padding of BOTH (i.e., n_faces = n_nodes+1) and NONE (i.e., n_faces = n_nodes-1), it becomes ambiguous what slicing with an integer list or downsampling here means as its not clear where the extra node/face should come from. If padding is BOTH/NONE then only slice indexers with no step are supported (i.e., contiguous regions)

As we now have a bunch of xr.Dataset-SGRID logic - this PR also restructures the sgrid to a subpackage within parcels.

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Yes, only in the co-authored commits. Mapped out from a conceptual POV then asked claude to implement.


src/parcels/_sgrid/accessor.py
```diff
@@ -0,0 +1,230 @@
...
+                grid_da = self._ds[var_name]
+
+        if grid_da is None:
+            raise ValueError(
+                "No variable found in dataset with 'cf_role' attribute set to 'grid_topology'. This doesn't look to be an SGrid dataset - please make your dataset conforms to SGrid conventions https://sgrid.github.io/sgrid/"
+            )
+        return grid_da
+
+    def isel(self, indexers: Mapping[str, Any] | None = None, **indexers_kwargs):
+        """TODO: Docstring"""
```
erikvansebille: Something to still do now? Or in a next PR?

src/parcels/_sgrid/accessor.py
```diff
@@ -0,0 +1,230 @@
...
+            continue
+
+        expected_n_faces = get_n_faces(n_nodes, padding)
+        if expected_n_faces != n_faces:
+            raise SGridDatasetInconsistency(
+                f"Node dimension {node!r} has size {n_nodes}, and face dimension {face!r} has size of {n_faces}. "
+                f"Due to dataset padding of {padding!r}, expected face dimension {face} to actually be size {expected_n_faces}."
+            )
+
+    # TODO: Also check on coordinates
```
erikvansebille, line 116: For a next PR, I assume?

src/parcels/_sgrid/core.py
```diff
@@ -282,18 +280,18 @@ def from_attrs(cls, attrs):  # type: ignore[override]
...
                     tuple[FaceNodePadding, FaceNodePadding, FaceNodePadding], load_mappings(attrs["volume_dimensions"])
                 ),
-                node_coordinates=maybe_load_mappings(attrs.get("node_coordinates")),
+                node_coordinates=_maybe_load_mappings(attrs.get("node_coordinates")),
             )
         except Exception as e:
             raise SGridParsingException(f"Failed to parse Grid3DMetadata from {attrs=!r}") from e
 
     def to_attrs(self) -> dict[str, str | int]:
-        d = dict(
```
erikvansebille, line 291: Just for my own understanding, why has this changed from using `dict()` to `{}`? Why would the new approach be better?

src/parcels/_python.py
```diff
@@ -1,6 +1,10 @@
 # Generic Python helpers
 import inspect
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
+from typing import TypeVar
+
+K = TypeVar("K")
+V = TypeVar("V")
```
erikvansebille, line 7: What's happening here? What is this for?

erikvansebille (approved):
Looks good; just a few small comments and questions below

src/parcels/_python.py
VeckoTheGecko: These are generic type variables https://mypy.readthedocs.io/en/stable/generics.html

They allow you to describe (via typing) a relationship between the input and output types of a function, without  providing concrete types.

e.g.,
```python
T = TypeVar("T")
def make_list(obj: T) -> list[T]:
    # make a list of length 1 from provided object
    # works on any object type. Returns a list with objects of same type as input
    return [obj]
```

In our case here we define type variables K and V (for "key" and "value") for

```python
def invert_non_unique_mapping(d: Mapping[K, V]) -> Mapping[V, list[K]]:
    inv_map: dict[V, list[K]] = {}
    for k, v in d.items():
        inv_map[v] = inv_map.get(v, []) + [k]
    return inv_map
```

src/parcels/_sgrid/core.py
VeckoTheGecko: Actually, this change was unnecessary. I thought it would be needed for typing, but it works fine the other way too

src/parcels/_sgrid/accessor.py
VeckoTheGecko: yes :)

src/parcels/_sgrid/accessor.py
VeckoTheGecko: Thanks for the ping - I think its good to do now. Will implement before merge

PR #2640, opened 2026-05-20
Add fesom tutorial to documentation

fluidnumericsJoe:
### Description

This tutorial uses the provided fesom2 channel data for demonstrating the basics of getting started with FESOM2 (unstructured grid) in parcels

### Checklist

- [ X ] This PR targets the correct branch (`main` for normal development, `v3-support` for v3 support)


docs/user_guide/examples/tutorial_fesom.ipynb
erikvansebille, line 1: As I can't easily comment on specific lines, here two comments:

1. In other tutorials, we always use `ds = parcels.convert.fesom_to_ugrid(ds)` (instead of `ds = fesom_to_ugrid(ds)`) to make explicit that these convert functions are part of parcels. Do that here too?
2. Is there an easy/quick way to also plot the velocity field before the pset.execute()? In that way, users get a feel for what kind of flow to expect. And that's a nice help for them to visualise their own unstructured flow field

erikvansebille (commented):
Looks great; two small comments below

docs/user_guide/examples/tutorial_fesom.ipynb
fluidnumericsJoe: I've added particle trajectories on top of velocity field vectors + fluid speed (in color). I also made sure to use `ds = parcels.convert.fesom_to_ugrid(ds)`.

erikvansebille (approved):
Looks good!

PR #2643, opened 2026-05-21
Fix C/A grid dimensions on peninsula_dataset

VeckoTheGecko:
### Description

Discovered working on https://github.com/Parcels-code/Parcels/pull/2642 - splitting into a separate commit since its about correctness (while the other is about refactoring) and I want them to squash down to separate commits in main.

### AI Disclosure

None used


erikvansebille (approved):
Good catch!

PR #2645, opened 2026-05-27
Remove Field.{x,y,z}dim

VeckoTheGecko:
This already exists on the grid object - it doesnt have to be on the Field object.

I think this is a left over from v3.

### Checklist

- [x] Closes None
- [x] Tests added - None needed


erikvansebille (approved):
Good catch!

Issue #2629, opened 2026-05-07
`Field.eval` silently returns wrong values for out-of-domain queries


fluidnumericsJoe:
### Parcels version

sha: bc20889

## Summary
When `Field.eval` is called directly (without a `ParticleSet`), the search-error sentinels returned by the index-search routines are passed straight through to the interpolator's corner-lookup, where they are used as raw indices. `numpy` negative-indexing then wraps them to the opposite edge of the grid, and the interpolator silently returns a weighted blend of corners pulled from unrelated parts of the field — no exception, no NaN, no warning.

This was discovered while investigating #2521. The PIC bug there has been fixed; this sentinel-leak path is an independent issue.

## Affected sentinels

All three search-error sentinels share the same defect:

| Sentinel | Source | Triggered by |
|---|---|---|
| `LEFT_OUT_OF_BOUNDS  = -2` | `_search_1d_array` (rectilinear) | `query < arr[0]` |
| `RIGHT_OUT_OF_BOUNDS = -1` | `_search_1d_array` (rectilinear) | `query > arr[-1]` |
| `GRID_SEARCH_ERROR   = -3` | `_search_indices_curvilinear_2d` (curvilinear) | spatial hash + PIC find no containing cell |

## Suggested acceptance criteria for a fix

- `Field.eval(...)` on an out-of-domain query without a `ParticleSet` raises an explicit error or returns `NaN` deterministically (not via index wrap).
- Behavior is uniform across all three sentinels (`LEFT_OUT_OF_BOUNDS`, `RIGHT_OUT_OF_BOUNDS`, `GRID_SEARCH_ERROR`) and across rectilinear and curvilinear paths.
- A regression test covers the direct-call path on a query that lies outside the mesh.

### Code sample

```python
import warnings

import numpy as np

from parcels import FieldSet
from parcels._datasets.structured.generated import simple_UV_dataset

warnings.filterwarnings("ignore")

# Tiny flat-mesh dataset so the corner indices are easy to read.
# Defaults: dims=(time=360, depth=2, ydim=30, xdim=4); we'll keep ydim, xdim small
# but informative.
ydim, xdim = 6, 5
ds = simple_UV_dataset(dims=(2, 1, ydim, xdim), mesh="flat")

# Functional U so that U[t, z, j, i] = j + 0.001 * i.
# The integer part identifies the j index, the fractional part the i index,
# so the value returned by Field.eval reveals which cells were sampled.
[...]

PR #2646, opened 2026-05-27
Model-centric refactoring to reduce dataset creation

VeckoTheGecko:
# Description

## Overview

The central change is the introduction of a new `Model` abstraction layer between raw xarray/uxarray data and the `Field`/`FieldSet` objects. Previously, `Field` owned its data and grid directly. Now, `Field` is a thin view over a `Model`, and `FieldSet` is a container of `Model` objects rather than `Field` objects.

---

### `Model` (abstract base class)

Abstract class with three required attributes:
- `data: Any` — the underlying dataset
- `grid: BaseGrid` — the grid object
- `field_to_interpolator: dict[str, ScalarInterpolator | VectorInterpolator]` — maps field names to interpolator instances

Abstract methods:
- `construct_fields() -> list[Field | VectorField]` — build field objects from this model
- `scalar_field_names -> list[str]` — names of scalar fields in the data
- `assert_valid_field_data(field_data)` — validate a single field's data

Concrete methods on `Model`:
- `assert_valid_model_data()` — iterates `scalar_field_names` and calls `assert_valid_field_data` on each
- `time_interval -> TimeInterval | None` — computed from `self.data`

### `StructuredModel(Model)`

For structured (SGRID) grid data backed by `xr.Dataset`.

Constructor: `StructuredModel(data: xr.Dataset, mesh: Mesh)`
- Calls `preprocess_sgrid_model_data(data)` to transpose fields to `(t, z, y, x)` order
- Creates an `XGrid(data, mesh)` grid
- Initializes `field_to_interpolator = {}`
- Calls `assert_valid_model_data()` on construction

`from_sgrid_conventions(cls, ds, mesh=None)` classmethod:
- Copied/moved from `FieldSet.from_sgrid_conventions` — handles time axis renaming, mesh type inference
- Sets default interpolator `XLinear()` on all scalar fields after construction
- Returns a `StructuredModel` instance

`construct_fields()`:
- Creates `Field("U", self)`, `Field("V", self)` etc., then wraps them in `VectorField("UV", ...)` if U+V present
- Uses `XLinear_Velocity()` for A-grids, `CGrid_Velocity()` for C-grids

### `UnstructuredModel(Model)`

For unstructured (UGRID) grid data backed by `ux.UxDataset`.

Constructor: `UnstructuredModel(data: ux.UxDataset, grid: UxGrid)`

`from_ugrid_conventions(cls, ds, mesh="spherical")` classmethod:
- Validates required dimensions (`time`, `zf`, `zc`)
- Creates `UxGrid`, calls `_discover_ux_U_and_V`, returns instance

`construct_fields()`:
- Uses `_select_uxinterpolator(da)` to pick the appropriate interpolator per field
- Note: interpolator is passed as 3rd arg to `Field(name, model, interp)` — see Field changes below

### Helper functions moved from `fieldset.py` to `model.py`

- `_discover_ux_U_and_V(ds)` — unchanged logic
- `_select_uxinterpolator(da)` — unchanged logic
- `_get_mesh_type_from_sgrid_dataset(ds)` — unchanged logic
- `_is_coordinate_in_degrees(da)` — unchanged logic
- `_get_time_interval(data)` — logic adjusted: checks `"time" not in data or data["time"].size == 1` (previously checked `data.shape[0] == 1`)
- `_assert_valid_uxdataarray(data)` — unchanged logic
- `_assert_has_time_coordinate(da)` — new helper extracted from old `Field.__init__`

### New helper in `model.py`

- `preprocess_sgrid_model_data(ds)` — transposes all non-grid-topology data vars to `(t, z, y, x)` using `_transpose_xfield_data_to_tzyx`

---

### `Field.__init__` signature change

**Before:**
```python
Field(name: str, data: xr.DataArray | ux.UxDataArray, grid: UxGrid | XGrid, interp_method: Callable)
```

**After:**
```python
Field(name: str, model: Model)
```

- `data`, `grid`, and `interp_method` are no longer constructor arguments
- The constructor only sets `self.name`, `self.model`, and `self.igrid = -1`
- Validation (data type checks, axis checks, time interval extraction) removed from `__init__`

### `Field` properties (delegating to model)

Three new properties proxy into the model:
```python
@property
def data(self):
    return self.model.data[self.name]

@property
def grid(self):
    return self.model.grid

@property
def time_interval(self):
    return self.model.time_interval
```

These preserve backward compatibility for code that reads `field.data`, `field.grid`, `field.time_interval`.

### `Field.interp_method` property/setter

**Before:** stored as `self._interp_method`; validated via `assert_same_function_signature` against `ZeroInterpolator`

**After:** stored in `self.model.field_to_interpolator[self.name]`
- Getter raises `AttributeError` (not `KeyError`) if no interpolator is set for this field
- Setter validates `isinstance(value, ScalarInterpolator)` instead of checking function signature

### Interpolator call convention change

**Before:** `self._interp_method(particle_positions, grid_positions, self)`

**After:** `self.interp_method.interp(particle_positions, grid_positions, self)`

Interpolators are now objects with an `.interp(...)` method, not plain callables.

### `VectorField` changes

- `interp_method` parameter type annotation changed from `Callable | None` to `VectorInterpolator | None`
- Validation changed from `assert_same_function_signature(...)` to `isinstance(interp_method, VectorInterpolator)`
- Setter similarly validates `isinstance(method, VectorInterpolator)`
- Call site: `self._interp_method.interp(...)` instead of `self._interp_method(...)`

### Removed from `field.py`

- `_assert_valid_uxdataarray` — moved to `model.py`
- `_assert_compatible_combination` — removed (validation now handled per-model)
- `_get_time_interval` — moved to `model.py`
- Imports: `uxarray`, `xarray`, `Callable`, `TimeInterval`, `ZeroInterpolator`, `ZeroInterpolator_Vector`, `assert_same_function_signature`, `_transpose_xfield_data_to_tzyx`, `assert_all_field_dims_have_axis`

---

### `FieldSet.__init__` signature change

**Before:** `FieldSet(fields: list[Field | VectorField])`

**After:** `FieldSet(models: list[Model])`

- Now stores `self.models: list[Model]`
- Calls `self.reconstruct_fields()` on init to build `self._fields`
- `assert_compatible_calendars(fields)` call commented out (TODO)

### New `FieldSet.fields` property

`_fields` is now the backing store; `fields` is a lazy property that calls `reconstruct_fields()` if `_fields` is `None`.

### New `FieldSet.reconstruct_fields()` method

Iterates `self.models`, calls `model.construct_fields()` on each, flattens into `self._fields` dict.

### New `FieldSet.__add__` operator

```python
def __add__(self, other: FieldSet) -> FieldSet:
    assert_compatible_fieldsets(self, other)
    combined = FieldSet(self.models + other.models)
    combined.constants = {**self.constants, **other.constants}
    return combined
```

### `from_ugrid_conventions` simplified

**Before:** ~15 lines building grid, discovering U/V, creating Field objects, returning `cls(list(fields.values()))`

**After:**
```python
model = UnstructuredModel.from_ugrid_conventions(ds, mesh)
return cls([model])
```

### `from_sgrid_conventions` simplified

**Before:** ~50 lines handling time axis, xgcm grid creation, field creation

**After:**
```python
model = StructuredModel.from_sgrid_conventions(ds, mesh)
return cls([model])
```

### `add_field` constant field creation updated

The inline `xgcm.Grid(...)` call when adding a constant scalar field is replaced with constructing `XGrid(ds, mesh=mesh)` directly (after attaching SGRID metadata via `sgrid._attach_sgrid_metadata`).

### New module-level function: `assert_compatible_fieldsets`

```python
def assert_compatible_fieldsets(left: FieldSet, right: FieldSet) -> None
```

Raises `ValueError` if the two fieldsets share any field names or constant names.

### Removed from `fieldset.py`

- `xgcm` import
- `UxGrid` import
- `_DEFAULT_XGCM_KWARGS` import
- `logger` import
- `_ds_rename_using_standard_names` import
- Most interpolator imports (only `XConstantField` remains)
- `_discover_ux_U_and_V` — moved to `model.py`
- `_select_uxinterpolator` — moved to `model.py`
- `_get_mesh_type_from_sgrid_dataset` — moved to `model.py`
- `_is_coordinate_in_degrees` — moved to `model.py`

---

## Summary of architectural intent

| Concern | Before | After |
|---|---|---|
| Data ownership | `Field` (held `self.data`, `self.grid`) | `Model` (holds `self.data`, `self.grid`) |
| Interpolator storage | `Field._interp_method` (per-field callable) | `Model.field_to_interpolator` (dict of objects) |
| Interpolator type | Any callable matching `ZeroInterpolator` signature | Instance of `ScalarInterpolator` / `VectorInterpolator` |
| Interpolator invocation | `interp_method(positions, grid_positions, field)` | `interp_method.interp(positions, grid_positions, field)` |
| `FieldSet` contents | `list[Field \| VectorField]` | `list[Model]` |
| Field construction | Done in `FieldSet.from_*` classmethods | Delegated to `Model.construct_fields()` |
| `FieldSet` combination | Not supported | `fieldset_a + fieldset_b` via `__add__` |

## Open questions

- Is `Model` the best name for this level of abstraction? (this question doesn't _have_ to be answered now - talking with @erikvansebille it seemed confusing, but there wasn't a clear better alternative)

# Checklist

- [x] Closes None
- [x] Tests added

# AI Disclosure

- [x] This PR contains AI-generated content - None(yet)
  - Describe how you used it (e.g., by pasting your prompt): AI was used in the refactoring of tests (mainly for simple updates, and for categorisation of failures - before coming back and updating by hand). This PR description was also written by AI based on the diff.


VeckoTheGecko:
OK - now I've mapped out most of the architectural changes here. I'm going to now profile performance and memory consumption to see where we stand, and how to optimise that under the new structure.

Once i have results for that, I will populate this PR with them as well as a full description over the proposed refactoring.

Once we decide on structure and performance differences and agree, we can go through the test suite and update.

VeckoTheGecko:
Still need to fix some small merge conflicts with main - but keen for review in the meantime!

docs/superpowers/plans/2026-06-18-update-tests-for-model-refactor.md
erikvansebille, line 1: What is this file? Why does it need to be in the repo? What does it have to do with this PR?

docs/getting_started/installation.md
erikvansebille, line 1: Not really part of this PR, is it? Would have been better to make a separate PR?

src/parcels/_core/field.py
```diff
@@ -86,69 +83,51 @@ class Field:
     def __init__(
         self,
         name: str,
-        data: xr.DataArray | ux.UxDataArray,
-        grid: UxGrid | XGrid,
-        interp_method: Callable,
+        model: Model,
```
erikvansebille: see general reflection on the naming of `Model` above

src/parcels/_core/field.py
```diff
@@ -86,69 +83,51 @@ class Field:
...
-        try:
-            if isinstance(data, ux.UxDataArray):
-                _assert_valid_uxdataarray(data)
-                # TODO: For unstructured grids, validate that `data.uxgrid` is the same as `grid`
-            else:
-                pass  # TODO v4: Add validation for xr.DataArray objects
-        except Exception as e:
-            e.add_note(f"Error validating field {name!r}.")
-            raise e
+        self.igrid = -1  # Default the grid index to -1
```
erikvansebille, line 99: Why this choice of -1? Don't we run the same risk of accidentally wrapping negative indices as in #2629?

src/parcels/_core/field.py
```diff
@@ -86,69 +83,51 @@ class Field:
...
+                f"{type(self).__name__} doesn't have an interp_method defined for it. Use `.interp_method = ...`"
+            ) from e
 
     @interp_method.setter
-    def interp_method(self, method: Callable):
-        assert_same_function_signature(method, ref=ZeroInterpolator, context="Interpolation")
-        self._interp_method = method
+    def interp_method(self, value):
+        # Setting the interpolation method dynamically
+        if not isinstance(value, ScalarInterpolator):
```
erikvansebille, line 128: I like this distinction between ScalarInterpolators and VectorInterpolators!

src/parcels/_core/field.py
```diff
@@ -86,69 +83,51 @@ class Field:
...
+        try:
+            return self.model.field_to_interpolator[self.name]
+        except KeyError as e:
+            raise AttributeError(
+                f"{type(self).__name__} doesn't have an interp_method defined for it. Use `.interp_method = ...`"
+            ) from e
 
     @interp_method.setter
-    def interp_method(self, method: Callable):
-        assert_same_function_signature(method, ref=ZeroInterpolator, context="Interpolation")
```
erikvansebille, line 150: Should we now also remove the ZeroInterpolators? It's not necessary (I wouldn't know any situation where a field interpolation should _always_ return zero; and can only lead to confusion

src/parcels/_core/fieldset.py
```diff
@@ -69,34 +55,57 @@ class FieldSet:
...
-    def __init__(self, fields: list[Field | VectorField]):
-        for field in fields:
-            if not isinstance(field, (Field, VectorField)):
-                raise ValueError(f"Expected `field` to be a Field or VectorField object. Got {field}")
-        assert_compatible_calendars(fields)
+    def __init__(self, models: list[Model]):
+        for model in models:
+            if not isinstance(model, Model):
+                raise ValueError(f"Expected `model` to be a Model object. Got {model}")
+        # assert_compatible_calendars(fields)
```
erikvansebille, line 62: Debug comment-out? Is this check needed?

src/parcels/_core/fieldset.py
```diff
@@ -69,34 +55,57 @@ class FieldSet:
...
+        self.models = list(models)
+        self._fields: dict[str, Field | VectorField] | None = None
+        self.reconstruct_fields()
         self.constants: dict[str, float] = {}
 
+    @property
+    def fields(self):
+        if self._fields is None:
+            self.reconstruct_fields()
+        assert self._fields is not None
```
erikvansebille, line 82: Should this assert raise an understandable error? Or will this never happen? Can we imagine empty FieldSets?

src/parcels/_core/fieldset.py
```diff
@@ -69,34 +55,57 @@ class FieldSet:
...
+        return self._fields
+
+    def reconstruct_fields(self):
+        fields = []
+        for model in self.models:
+            fields += model.construct_fields()
```
erikvansebille, line 88: What's the difference between constructing and _re_constructing?

src/parcels/_core/fieldset.py
```diff
@@ -69,34 +55,57 @@ class FieldSet:
...
             raise AttributeError(f"FieldSet has no attribute '{name}'")
 
-    def __repr__(self):
-        return fieldset_repr(self)
+    def __add__(self, other: FieldSet) -> FieldSet:
+        if not isinstance(other, FieldSet):
+            return NotImplemented
+        assert_compatible_fieldsets(self, other)
+        combined = FieldSet(self.models + other.models)
+        combined.constants = {**self.constants, **other.constants}
```
erikvansebille: Should be `contexts` since #2678?

src/parcels/_core/fieldset.py
```diff
@@ -69,34 +55,57 @@ class FieldSet:
...
+        return combined
+
+    # def __repr__(self):
+    #     return fieldset_repr(self)
```
erikvansebille, line 109: Needs to be removed or reimplemented?

src/parcels/_core/fieldset.py
```diff
@@ -143,15 +152,19 @@ def add_constant_field(self, name: str, value, mesh: Mesh = "spherical"):
...
-            ds, coords={"X": {"left": "lon"}, "Y": {"left": "lat"}}, autoparse_metadata=False, **_DEFAULT_XGCM_KWARGS
-        )
-        grid = XGrid(xgrid, mesh=mesh)
-        self.add_field(Field(name, ds[name], grid, interp_method=XConstantField))
+        try:
+            model = constant_field_models[mesh]
+        except KeyError as e:
+            raise ValueError(f"mesh must be one of ['flat', 'spherical']. Got {mesh!r}.") from e
+
+        model.data[name] = (["lat", "lon", "depth", "time"], np.full((1, 1, 1, 1), value))
```
erikvansebille: Make this `time`, `depth`, `lat`, `lon` order(following #2663)

src/parcels/_core/fieldset.py
```diff
@@ -258,70 +248,30 @@ def from_sgrid_conventions(
...
+        The two FieldSets to check.
+
+    Raises
+    ------
+    ValueError
+        If the FieldSets share field names or constant names.
+    """
+    overlapping_fields = set(left.fields) & set(right.fields)
+    if overlapping_fields:
+        raise ValueError(f"Cannot add FieldSets with overlapping field names: {sorted(overlapping_fields)}")
```
erikvansebille: "overlapping" may be a confusing word. Why not simply "that contain fields with the same name"

src/parcels/_core/fieldset.py
```diff
@@ -258,70 +248,30 @@ def from_sgrid_conventions(
...
+
+    overlapping_constants = set(left.constants) & set(right.constants)
+    if overlapping_constants:
+        raise ValueError(f"Cannot add FieldSets with overlapping constant names: {sorted(overlapping_constants)}")
```
erikvansebille: Same comment as above about the term overlapping

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+
+import parcels._sgrid as sgrid
+from parcels._core.basegrid import BaseGrid
+from parcels._core.field import Field, VectorField
+from parcels._core.utils.time import TimeInterval
+from parcels._core.uxgrid import UxGrid
+from parcels._core.xgrid import (
+    XGrid,
+    _transpose_xfield_data_to_tzyx,
+    assert_all_field_dims_have_axis,  # noqa: F401, leave import for now until decision is made
```
erikvansebille: add a TODO here, so we know to remove it later?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+            )
+            raise e
+        return time_interval
+
+
+def preprocess_sgrid_model_data(ds: xr.Dataset) -> xr.Dataset:
+    metadata: sgrid.SGrid2DMetadata = ds.sgrid.metadata
+
+    for field_name in set(ds.data_vars) - {ds.sgrid._get_grid_topology().name}:
+        ds[field_name] = _transpose_xfield_data_to_tzyx(ds[field_name], metadata)
```
erikvansebille, line 77: This would not work for raw zarr? Because we would not use dask then?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+        grid = XGrid(data, mesh)
+
+        self.data = data
+        self.grid = grid
+        self.field_to_interpolator = {}
+        self._fields: list[Field | VectorField] | None = None
+        self.assert_valid_model_data()
+
+    def assert_valid_field_data(self, field_data: xr.DataArray) -> None:
+        # assert_all_field_dims_have_axis(field_data, self.grid.xgcm_grid) #! These checks should be revisited
```
erikvansebille: Add a TODO or fix in this PR?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+            if self.data[var].attrs.get("cf_role") == "grid_topology":
+                skip_vars.add(var)
+        return list(set(self.data.data_vars) - skip_vars)
+
+    def construct_fields(self) -> list[Field | VectorField]:
+        single_fields: dict[str, Field] = {}
+        vector_fields: dict[str, VectorField] = {}
+        scalar_field_names = self.scalar_field_names
+        if "U" in scalar_field_names and "V" in scalar_field_names:
+            vector_interp_method = XLinear_Velocity() if _is_agrid(self.data) else CGrid_Velocity()
```
erikvansebille: Should this just be interp_method, instead of vector_interp_method?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+                raise ValueError("Multiple time coordinates found in dataset. This is not supported by Parcels.")
+            (time_dim,) = time_dims
+            if time_dim != "time":
+                logger.debug(f"Renaming time axis coordinate from {time_dim} to 'time'.")
+                ds = ds.rename({time_dim: "time"})
+
+        # if "lon" not in ds.coords or "lat" not in ds.coords:
+        #     node_dimensions = sgrid.load_mappings(ds.grid.node_dimensions)
+        #     ds["lon"] = ds[node_dimensions[0]]
+        #     ds["lat"] = ds[node_dimensions[1]]
```
erikvansebille, line 161: Is this a debug comment?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+
+        model = cls(ds, mesh=mesh)
+        model._fields = model.construct_fields()
+        for f in model._fields:
+            if isinstance(f, Field):
+                f.interp_method = XLinear()
+        return model
+
+
+constant_field_models = {
```
erikvansebille: Is this the most logical place/file for this script? Should it not go to dataset generation?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+        model._fields = model.construct_fields()
+        for f in model._fields:
+            if isinstance(f, Field):
+                interp_cls = _select_uxinterpolator(model.data[f.name])
+                if interp_cls is not None:
+                    f.interp_method = interp_cls()
+        return model
+
+
+# TODO: Refactor later into something like `parcels._metadata.discover(dataset)` helper that can be used to discover important metadata like this. I think this whole metadata handling should be refactored into its own module.
```
erikvansebille, line 261: Agree, but not for this PR I assume?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+    # Map to supported interpolators
+    if vdim and ldim:
+        key = f"{vdim},{ldim}"
+        if key in supported_uxinterp_mapping.keys():
+            return supported_uxinterp_mapping[key]
+
+    return None
+
+
+def _is_agrid(ds: xr.Dataset) -> bool:
```
erikvansebille, line 369: Hmm, this would also return True for a B-grid (where U and V are on the same node, but W and tracers are on different nodes). So the naming here is not correct. Perhaps change to _is_cgrid and then return the negate?

src/parcels/_core/model.py
```diff
@@ -0,0 +1,406 @@
...
+    # check if U and V are defined on the same dimensions
+    # if yes, interpret as A grid
+    return set(ds["U"].dims) == set(ds["V"].dims)
+
+
+def _get_time_interval(data: xr.DataArray | ux.UxDataArray) -> TimeInterval | None:
+    if "time" not in data or data["time"].size == 1:
+        return None
+
+    return TimeInterval(data.time.values[0], data.time.values[-1])
```
erikvansebille, line 379: Ok, so now we assume that all Fields in a Model have the same time (or no time). Is this a change from before in v4-dev, when Fields could share a spatial grid, but have a different time dimension? or had we changed that already?

erikvansebille (commented):
_I started reviewing this PR, but it was difficult to do the full PR because it wasn't rebased on main, so many of the changes were already merged in main (and hence reviewed). Since I didn't want to do double work, I stopped reviewing after `model.py` (which is the most important change anyways)._

Just to make sure I understand it: a `Model` is the set of `Fields` that share a `Grid`; right?

I'm munching on the term `Model` though. Most oceanographers would probably assume that a model is e.g. NEMO, MITgcm, etc. What we here refer to as a model, they would call a dataset, simulation or modelrun; the _output_ of a model on a specific grid.

But `Dataset` would be too confusing with the `xarray` terminology, and `Simulation` would be confusing with the output of Parcels itself (since Parcels _simulates_). Not sure if `Modelrun` is better than `Model`

How about renaming model to `SharedGrid`, or something like that? 
Or `ParcelsDataset`, to highlight that these are "parcels-flavoured" xarray Datasets?

VeckoTheGecko:
Ah, didn't realise that this not being synced with main was messing with the diff

Also didn't realise those temporary files were still committed.

I'll respond to the feedback, fix those files and sync when I'm back in the office next week

VeckoTheGecko:
As term, I do like ParcelsDataset

docs/superpowers/plans/2026-06-18-update-tests-for-model-refactor.md
VeckoTheGecko: Resynced with main - this is no longer applicable

docs/getting_started/installation.md
VeckoTheGecko: I think this was a rendering issue since we weren't synced with main

src/parcels/_core/field.py
VeckoTheGecko: This was already in `main`.

AFAICT the caching of ei isn't even working  (searching for `.igrid =` gives no meaningful results). I already suspected this was the case, hence why I was mentioning before that we should have explicit tests for this as part of our release roadmap.

For another issue/pr

src/parcels/_core/fieldset.py
VeckoTheGecko: This will never happen and is not user facing.

Re. assertions and checking values: if its user facing code, we should not really be using Assert statements. We should be using

```python
if condition:
    raise SomeError(...) # e.g., ValueError
```

If its internal facing things, like errors that should never happen (and as far as we know, will never happen) we use asserts.

This is common Python convention. And the asserts (such as here) are actually used for type narrowing within mypy. This tells mypy that at that point in execution `._fields` is `list[Field | VectorField]` not `list[Field | VectorField] | None`

In the case of an empty fieldset, its `[]` which satisfies `list[Field | VectorField]`

src/parcels/_core/fieldset.py
VeckoTheGecko: The way I was thinking about it is that `construct_fields` returns the fields, while `reconstruct_fields` sets the value. Also, from the FieldSet POV, having `reconstruct_fields` indicates that the fields need to be _re_constructed when the underlying models change.

I think this wording is good to keep.

src/parcels/_core/fieldset.py
VeckoTheGecko: resynced with main - this should be fixed now

src/parcels/_core/fieldset.py
VeckoTheGecko: This is ahead of #2683

> @VeckoTheGecko
> [Disable reprs](https://github.com/Parcels-code/Parcels/pull/2646/commits/0b2b1470bb48ec9aae72b05cca0185b82def09a1) 
> [0b2b147](https://github.com/Parcels-code/Parcels/pull/2646/commits/0b2b1470bb48ec9aae72b05cca0185b82def09a1)
> Field, VectorField, FieldSet, ParticleSet, XGrid
> 
> Ahead of https://github.com/Parcels-code/Parcels/issues/2683 and so that we don't have to refactor too much in this PR

I would suggest we keep as is

src/parcels/_core/model.py
VeckoTheGecko: AFAICT I think zarr would still work. I think the Zarr arrays still support a range of operations (e.g., reshaping, transposing, expanding dims) since they're really simply operations that don't actually change the data - just the way its represented/traversed.

The fact we were getting those performance improvements despite this already being in main shows that.

But yes, operations like this should be on our mind when we want to support workflows like loading from raw Zarr :)

src/parcels/_core/model.py
VeckoTheGecko: now fixed

src/parcels/_core/model.py
VeckoTheGecko: added a TODO

src/parcels/_core/model.py
VeckoTheGecko: Added a TODO

src/parcels/_core/fieldset.py
VeckoTheGecko: Is this good?
> f"Cannot add FieldSets that have field names in common. Duplicate field names are: {sorted(common_fields)}"

863328e7e5bbb0eabd16fc65d12dabf17e302118

src/parcels/_core/fieldset.py
VeckoTheGecko: Done now

src/parcels/_core/field.py
VeckoTheGecko: Yes - this is a good idea.

Done now in 9e4e34cd7fed3d501a97a99d9c0af7efa679f7e6

src/parcels/_core/model.py
VeckoTheGecko: Yes - this comment was already in `main`

src/parcels/_core/model.py
VeckoTheGecko: Could you copy paste/point out the code snippet that you're refering to here?

src/parcels/_core/model.py
VeckoTheGecko: Sounds good. Quickly making a separate issue for this to avoid scope creep in this PR :)

VeckoTheGecko: https://github.com/Parcels-code/Parcels/issues/2692

src/parcels/_core/model.py
VeckoTheGecko: > Is this a change from before in v4-dev, when Fields could share a spatial grid, but have a different time dimension?

Yes, you're right that this is a change - one that can result in more of a memory footprint in `ei`'s should people run with the same spatial grids but differing temporal dimensions.

Given xarrays data model doesn't support multiple time dimensions in the same dataset - I think that this is unavoidable without significantly reconsidering how we ingest data.

VeckoTheGecko:
> Just to make sure I understand it: a `Model` is the set of `Fields` that share a `Grid`; right?

Yes, that is correct

VeckoTheGecko:
Will do the Model renamings after lunch and re-request a review

src/parcels/_core/field.py
fluidnumericsJoe: Hmm.. I'll look into this grid index caching issue today.

VeckoTheGecko: Awesome - thanks for picking this up @fluidnumericsJoe !

VeckoTheGecko:
Hmmm.

> I'm munching on the term `Model` though. Most oceanographers would probably assume that a model is e.g. NEMO, MITgcm, etc. What we here refer to as a model, they would call a dataset, simulation or modelrun; the _output_ of a model on a specific grid.
> 
> But `Dataset` would be too confusing with the `xarray` terminology, and `Simulation` would be confusing with the output of Parcels itself (since Parcels _simulates_). Not sure if `Modelrun` is better than `Model`
> 
> How about renaming model to `SharedGrid`, or something like that? Or `ParcelsDataset`, to highlight that these are "parcels-flavoured" xarray Datasets?

Thinking more about this, I think the renaming to `ParcelsDataset` would be too confusing to us as devs since it would warrant replacing all mentions with `models` to `dataset` which I think makes it very easy to conflate with xarray's meaning of datasets.

I agree that `Model` by itself might be a bit confusing to users - but at the same time `Model` and mentions to it is not public API (in fact, I don't think that it will show up anywhere in public API - perhaps only error messaging in the file paths). `Model` here is mainly only used for internal bookkeeping and reasoning - and can be said as such in our docs.

I think having the term `model` in parcels is quite clear for developers and how the abstraction fits within parcels (i.e,. a model encapsulates how the spatial domain relates to the underlying data representation by defining indexing, and by performing transformations on the input data).

I think there is a degree to which we need to use language that works best for development (particularly for internal API) even if there is alternative usage in other domains (i.e., oceanography or atmosphere) - there's just not enough words.

Perhaps a middleground would be to use `ModelData` for the base class name, `StructuredModelData` and `UnstructuredModelData` and keep everything else the same? (i.e., keep `FieldSet.models`)

VeckoTheGecko:
I was looking at the docs build, and saw an example that adds a second velocity field

```python
import numpy as np
import xarray as xr

import parcels
import parcels.tutorial

ds_fields = parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/data")

# Create an idealised wind field and add it to the dataset
tdim, ydim, xdim = (len(ds_fields.time),len(ds_fields.latitude), len(ds_fields.longitude))
ds_fields["UWind"] = xr.DataArray(
    data=np.ones((tdim, ydim, xdim)) * np.sin(ds_fields.latitude.values)[None, :, None],
    coords=[ds_fields.time, ds_fields.latitude, ds_fields.longitude])

ds_fields["VWind"] = xr.DataArray(
    data=np.zeros((tdim, ydim, xdim)),
    coords=[ds_fields.time, ds_fields.latitude, ds_fields.longitude])

fields = {
    "U": ds_fields["uo"],
    "V": ds_fields["vo"],
    "UWind": ds_fields["UWind"],
    "VWind": ds_fields["VWind"],
}

# Create a vecorfield for the wind
windvector = parcels.VectorField(
    "Wind",
    fieldset.UWind,
    fieldset.VWind,
    interp_method=parcels.interpolators.XLinear_Velocity
)
fieldset.add_field(windvector)

```

We no longer have `add_field` - I think we need to map out exactly how we should be defining velocity fields under this new refactoring... I think that can be done in a future PR

VeckoTheGecko:
Just chatted f2f with Erik, and decided on vvvvv

> Perhaps a middleground would be to use `ModelData` for the base class name, `StructuredModelData` and `UnstructuredModelData` and keep everything else the same? (i.e., keep `FieldSet.models`)

src/parcels/_core/fieldset.py
erikvansebille: Hmm, I feel that this difference between constructing and reconstructing is too subtle to be intuitive to new developers. Perhaps use terms that are more clearly distinct (`construct_fields` and `set_field_values`?)

src/parcels/_core/model.py
```diff
@@ -0,0 +1,404 @@
...
+                raise e
+        return
+
+    @property
+    def time_interval(self) -> TimeInterval | None:
+        try:
+            time_interval = _get_time_interval(self.data)
+        except ValueError as e:
+            e.add_note(
+                f"Error getting time interval for model with data {self.data!r}. Are you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects?"
```
erikvansebille: So here the term 'model' is exposed to users/oceanographers. Can we not rephrase this to stay away from the term model, as we indicated it could be confusing?

src/parcels/_core/model.py
erikvansebille: I mean the creation of the `constant_field_models` dictionary (lines 171-196 here)

```python
constant_field_models = {
    mesh: StructuredModelData.from_sgrid_conventions(
        xr.Dataset(
            {},
            coords={
                "lat": (["lat"], [0], {"axis": "Y"}),
                "lon": (["lon"], [0], {"axis": "X"}),
                "depth": (["depth"], [0], {"axis": "Z"}),
                "time": (["time"], [0], {"axis": "T"}),
            },
        ).pipe(
            sgrid._attach_sgrid_metadata,
            sgrid.SGrid2DMetadata(
                cf_role="grid_topology",
                topology_dimension=2,
                node_dimensions=("lon", "lat"),
                face_dimensions=(
                    sgrid.FaceNodePadding("XC", "lon", sgrid.Padding.LOW),
                    sgrid.FaceNodePadding("YC", "lat", sgrid.Padding.LOW),
                ),
            ),
        ),
        mesh=mesh,  # type:ignore
    )
    for mesh in ["flat", "spherical"]
}
```

src/parcels/_core/model.py
erikvansebille: OK, I think this is a good idea, to move `time` into our internal concept of one of the four dimensions of a `Grid`

src/parcels/_core/xgrid.py
```diff
@@ -68,37 +71,86 @@ def assert_all_field_dims_have_axis(da: xr.DataArray, xgcm_grid: xgcm.Grid) -> N
...
+@dataclass
+class XgcmLikeAxis:
+    coords: dict[ptyping.XgcmAxisPosition, str]
+
+
+class XgcmLikeGrid:
+    """Adapter class to circumvent XGCM as a dep.
+
+
+    This is only used as a temporary class for the moment. Down the line we should refactor to remove XGCM entirely and work with SGRID metadata (especially since COMODO metadata isn't standard, and nor is the xgcm data model).
```
erikvansebille: Add a TODO tag here, so we don't lose track of it? Or (perhaps even better) create an issue (with a post-v4 tag)?

tests/test_field.py
```diff
@@ -118,36 +49,20 @@ def test_field_init_fail_on_float_time_dim():
         ds["time"].attrs,
     )
 
-    data = ds["data_g"]
-    grid = XGrid.from_dataset(ds, mesh="flat")
+    model = StructuredModelData.from_sgrid_conventions(ds, mesh="flat")
     with pytest.raises(
         ValueError,
-        match=r"Error getting time interval.*. Are you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects\?",
+        match=r"Are you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects\?",
```
erikvansebille, line 55: Should this not be singular ?

```suggestion
        match=r"Are you sure that the time dimension on the xarray dataset is stored as a timedelta, datetime or cftime datetime object\?",
```

tests/test_fieldset.py
```diff
@@ -291,3 +206,62 @@ def test_fieldset_from_sgrid_conventions(ds_name):
...
+
+def test_fieldset_add():
+    """Test that two FieldSets can be combined with + (fset1 + fset2)."""
+    ds1 = datasets_structured["ds_2d_left"][["U_A_grid", "grid"]].rename({"U_A_grid": "U1"})
+    ds2 = datasets_structured["ds_2d_left"][["V_A_grid", "grid"]].rename({"V_A_grid": "V2"})
+
+    fset1 = FieldSet.from_sgrid_conventions(ds1, mesh="flat")
+    fset2 = FieldSet.from_sgrid_conventions(ds2, mesh="flat")
+
+    fset = fset1 + fset2
```
erikvansebille, line 219: Have we also implemented `fset1 += fset2`?

tests/test_index_search.py
```diff
@@ -48,6 +48,8 @@ def test_grid_indexing_fpoints(field_cone):
             assert y > np.min(cell_lat) and y < np.max(cell_lat)
 
 
+# XGrid no longer accepts xgcm.Grid objects; now requires xr.Dataset with SGRID metadata; NEMO curvilinear workflow via xgcm is not the construction path
+@pytest.mark.skip("Uses now removed API. TODO: What is the goal of this test?")
```
erikvansebille: I think this unit test can go now. We test curvilinear indexing in the test_advection tests and tutorials for NEMO and MITgcm already; so this one is redundant. Was ported from v3?

erikvansebille (commented):
Looks good, just a few comments/questions below

tests/test_field.py
VeckoTheGecko: The time dimension is an array of objects, so I would say its plural

src/parcels/_core/model.py
VeckoTheGecko:
```diff
-"Error getting time interval for model with data {self.data!r}. Are you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects?"
+"Error getting time interval for:\n {self.data!r}\n\nAre you sure that the time dimension on the xarray dataset is stored as timedelta, datetime or cftime datetime objects?"

```

VeckoTheGecko:
I've gone and disabled the kernelloop doc page from running in CI 7c46987 (see #2695)

VeckoTheGecko:
Hmmm - docs are failing since quite a few use the old `add_field` API. Should I resolve all the docs errors in this PR @erikvansebille  or should we merge this in so that I can work on other items?

I think you have more of a picture on the documentation/what it's like to _use_ parcels, so you might be quicker resolving those.

VeckoTheGecko:
🥳🥳🥳🥳🥳🥳

PR #2648, opened 2026-05-28
Remove XGrid.from_dataset

VeckoTheGecko:
### Description

This code path isn't used by `from_sgrid_conventions` and was mainly used before we had SGRID support by piggy-backing off xgcm functionality.

This PR removes this classmethod and updates the test suite. Some parts of the test suite haven't been updated and have instead been flagged (as they're likely to be removed in a future PR).

This PR is blocked by https://github.com/Parcels-code/Parcels/pull/2647 (once this PR is merged this will be ready for review)

### Checklist

- [x] Closes None (related to https://github.com/Parcels-code/Parcels/pull/2647 )
- [ ] Tests added - N/A

### AI Disclosure
None used

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,8 @@
 import pytest
 
+from parcels import FieldSet
+from parcels._datasets.structured.generic import datasets as datasets_structured
+
 SKIP_BY_DEFAULT = {"validation", "flaky"}
 
 
@@ -17,3 +20,16 @@ def pytest_collection_modifyitems(config, items):
 @pytest.fixture
 def tmp_parquet(tmp_path):
     return tmp_path / "tmp.parquet"
+
+
+@pytest.fixture
+def fieldset() -> FieldSet:
+    """FieldSet with U and V"""
+    ds = datasets_structured["ds_2d_left"].copy()
+    ds = ds[["U_A_grid", "V_A_grid", "grid"]].rename(
+        {
+            "U_A_grid": "U",
+            "V_A_grid": "V",
+        }
+    )
+    return FieldSet.from_sgrid_conventions(ds, mesh="flat")
--- a/src/parcels/_core/xgrid.py
+++ b/src/parcels/_core/xgrid.py
@@ -133,18 +133,6 @@ class XGrid(BaseGrid):
         ptyping.assert_valid_mesh(mesh)
         self._ds = ds
 
-    @classmethod
-    def from_dataset(cls, ds: xr.Dataset, mesh, xgcm_kwargs=None):
-        """WARNING: unstable API, subject to change in future versions."""  # TODO v4: make private or remove warning on v4 release
-        if xgcm_kwargs is None:
-            xgcm_kwargs = {}
-
-        xgcm_kwargs = {**_DEFAULT_XGCM_KWARGS, **xgcm_kwargs}
-
-        ds = _drop_field_data(ds)
-        grid = xgcm.Grid(ds, **xgcm_kwargs)
-        return cls(grid, mesh=mesh)
-
     def __repr__(self):
         return xgrid_repr(self)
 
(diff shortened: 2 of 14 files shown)


VeckoTheGecko:
Hmmmmm, not sure why the conda package build process is failing

```
 │ │
 │ │ ╭─ Installing test environment
 │ │ │ ✔ Successfully updated the test environment
 │ │ │
 │ │ ╰─────────────────── (took 14 seconds)
 │ │ Matplotlib is building the font cache; this may take a moment.
 │ │ $SRC_DIR/conda_build_script.py:1: UserWarning: This is an alpha version of Parcels v4. The API is not stable and may change without deprecation warnings.
 │ │   import parcels
 │ │ ✔ python imports test passed!
 │ │ panel 1.9.2 requires panel-material-ui, which is not installed.
 │ │ × error Script failed with status 1
 │ │ × error 
 │ │ × error Script execution failed.
 │ │ × error 
 │ │ × error   Work directory: /var/folders/j5/x0w7f5sn7gg25__6pg48j82c0000gn/T/.tmpVuU6ZK/parcels-0.1.0dev-pyh4616a5c_0
 │ │ × error   Prefix: /Users/Hodgs004/coding/repos/parcels/output/test/test_parcelsOYpBf8/test_env
 │ │ × error   Build prefix: None
 │ │ × error 
 │ │ × error To run the script manually, use the following command:
 │ │ × error 
 │ │ × error   cd "/var/folders/j5/x0w7f5sn7gg25__6pg48j82c0000gn/T/.tmpVuU6ZK/parcels-0.1.0dev-pyh4616a5c_0" && ./conda_build.sh
 │ │ × error 
 │ │ × error To run commands interactively in the build environment:
 │ │ × error 
 │ │ × error   cd "/var/folders/j5/x0w7f5sn7gg25__6pg48j82c0000gn/T/.tmpVuU6ZK/parcels-0.1.0dev-pyh4616a5c_0" && source build_env.sh
 │ │
 │ ╰─────────────────── (took 74 seconds)
 │
 ╰─────────────────── (took 74 seconds)
```

This is happening on `main` as well - likely a problem with a dependency in the build environment.... Not a blocker for this PR

tests/conftest.py
```diff
@@ -17,3 +20,16 @@ def pytest_collection_modifyitems(config, items):
...
+@pytest.fixture
+def fieldset() -> FieldSet:
+    """FieldSet with U and V"""
+    ds = datasets_structured["ds_2d_left"].copy()
+    ds = ds[["U_A_grid", "V_A_grid", "grid"]].rename(
+        {
+            "U_A_grid": "U",
+            "V_A_grid": "V",
+        }
+    )
```
erikvansebille, line 34: Do we need to worry about the values in the arrays of these datasets? For example, if they are by default zero then some advection tests may pass irrespective of how good the advection scheme is. Or is it simply not intended that the values in these arrays are used?

tests/conftest.py
```diff
@@ -17,3 +20,16 @@ def pytest_collection_modifyitems(config, items):
 @pytest.fixture
 def tmp_parquet(tmp_path):
     return tmp_path / "tmp.parquet"
+
+
+@pytest.fixture
+def fieldset() -> FieldSet:
```
erikvansebille, line 26: It's a bit 'magical' that the fieldset() is defined here, but can be used in all the test files without an explicit import in these test files. New devs may get confused?

erikvansebille (approved):
Looks good; two minor questions below

tests/conftest.py
VeckoTheGecko: Discussed in person - these datasets have random data at the moment, where other datasets can be used for more flow specific tests.

tests/conftest.py
VeckoTheGecko: Discussed in person - this is just [pytest magic](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files) :)

PR #2655, opened 2026-06-04
Adding overwrite option to ParticleFile API

erikvansebille:
### Description

This PR adds an option to `ParticleFile.__init__` to control when the file already exists: either raise an error (default), or overwrite the exisiting file. 

### Checklist

- [x] Closes #2583
- [x] Tests added

### AI Disclosure

I used Claude code to help with the implementation of the append option in particlefile.write()

### DIFF
--- a/src/parcels/_core/particlefile.py
+++ b/src/parcels/_core/particlefile.py
@@ -101,7 +101,6 @@ class ParticleFile:
 
         self._path = path  # TODO v4: Consider https://arrow.apache.org/docs/python/getstarted.html#working-with-large-data - though a significant question becomes how to partition, perhaps using a particle variable "partition"?
         self._writer: pq.ParquetWriter | None = None
-        self._tmp_path: Path | None = None
 
         if self._mode not in {None, "w"}:
             raise ValueError(f"Invalid mode value {self._mode!r}. Expected one of None or 'w'.")
@@ -182,9 +181,6 @@ class ParticleFile:
         if self._writer is not None:
             self._writer.close()
             self._writer = None
-            if self._tmp_path is not None:
-                self._tmp_path.replace(self.path)
-                self._tmp_path = None
 
 
 def _get_vars_to_write(particle: ParticleClass) -> list[Variable]:
--- a/src/parcels/_core/particlefile.py
+++ b/src/parcels/_core/particlefile.py
@@ -97,18 +97,17 @@ class ParticleFile:
             raise ValueError(f"outputdt must be positive/non-zero. Got {outputdt=!r}")
 
         self._outputdt = outputdt
-        self._mode = mode
 
         self._path = path  # TODO v4: Consider https://arrow.apache.org/docs/python/getstarted.html#working-with-large-data - though a significant question becomes how to partition, perhaps using a particle variable "partition"?
         self._writer: pq.ParquetWriter | None = None
 
-        if self._mode not in {None, "w"}:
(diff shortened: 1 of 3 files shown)


src/parcels/_core/particlefile.py
```diff
@@ -61,6 +61,11 @@ class ParticleFile:
         It is either a numpy.timedelta64, a datimetime.timedelta object or a positive float (in seconds).
     compression : {"zstd", "gzip", "snappy", "brotli", None}, optional
         Compression algorithm to use for the Parquet file. Default is "zstd".
+    if_exists : {"error", "overwrite", "append"}, optional
+        Behavior when the output file already exists.
+        - "error" (default): raise a ValueError.
+        - "overwrite": remove the existing file before writing.
+        - "append": preserve existing rows and append new rows.
```
VeckoTheGecko: I think we should stick closer to convention here

```suggestion
    mode : {"w", "a", None}, optional
        Writing behaviour.
        - None (default): Write dataset, and raise an error if it already exists.
        - "w": Write dataset, overwriting it.
        - "a": Append to dataset.
```

also rename `._if_exists` to `._mode`

src/parcels/_core/particlefile.py
```diff
@@ -122,6 +143,42 @@ def outputdt(self):
...
+            # Parquet can't directly append, so we need to rewrite the existing data along with the new data.
+            for batch in existing_file.iter_batches():
+                self._writer.write_table(pa.Table.from_batches([batch], schema=existing_schema))
+        else:
+            assert not self.path.exists(), "If the file exists, the writer should already be set"
+            self._writer = pq.ParquetWriter(
+                self.path,
+                schema,
+                compression=self._compression,
+            )
```
VeckoTheGecko: Just taking a step back here - why do we need an append mode for the ParticleFile? Could users easily just create multiple particlefiles and join them into one after the fact?

VeckoTheGecko: I think especially since the file format of Parquet doesn't support this, and calling "append" would require rewriting the current data that we have, is maybe an indication that we either shouldn't have append or should consider something else.

src/parcels/_core/particlefile.py
erikvansebille: Good call, address in aafde6a0354acc6fed24c0d50763c1efc1103194

src/parcels/_core/particlefile.py
erikvansebille: You're right; it doesn't make sense to add an "append" option if we are copying the data internally anyways. I now removed it in aafde6a0354acc6fed24c0d50763c1efc1103194

The only thing users will need to be aware of, is that the last record of the first file is the same as the first second of the second file when concatenating, so they will probably need to filter these duplicate records out. Is that something we should add to the tutorial?

src/parcels/_core/particlefile.py
```diff
@@ -89,12 +97,20 @@ def __init__(
             raise ValueError(f"outputdt must be positive/non-zero. Got {outputdt=!r}")
 
         self._outputdt = outputdt
+        self._mode = mode
```
erikvansebille: Now that we don't use `mode` outside of this function anymore, do we need to make it an object attribute? What is best practice in such cases?

erikvansebille: Fixed in 27cc62fa80a580e3886f98504040ddf7967a2a4f

VeckoTheGecko (approved):
Looks good to me!

Issue #2374, opened 2025-11-13
Consistent dimensions order


reint-fischer:
> Very minor comment, but should we try to keep the order `time, z, lat, lon` wherever we can in Parcels?
> 
> 
> ```suggestion
> pset = parcels.ParticleSet(
>     fieldset=fieldset, pclass=parcels.Particle, time=time, z=z, lat=lat, lon=lon
> )
> ``` 

 _Originally posted by @erikvansebille in [#2362](https://github.com/Parcels-code/Parcels/pull/2362/files#r2521978659)_

I like this consistency and think it will be a quick edit

PR #2663, opened 2026-06-15
Issue: Consistent dimensions order #2374

PeterWolfram:
### Changed dimension order to time,z,lat,lon or time,z,y,x for consistency

As requested in issue https://github.com/Parcels-code/Parcels/issues/2374 changed order of dimensions to be consistent throughout the code.


src/parcels/_core/particleset.py
```diff
@@ -59,20 +59,20 @@ def __init__(
...
     ):
         self._data = None
         self._kernel = None
 
         self.fieldset = fieldset
-        lon = np.empty(shape=0) if lon is None else np.array(lon).flatten()
-        lat = np.empty(shape=0) if lat is None else np.array(lat).flatten()
         time = np.empty(shape=0) if time is None else np.array(time).flatten()
+        lat = np.empty(shape=0) if lat is None else np.array(lat).flatten()
+        lon = np.empty(shape=0) if lon is None else np.array(lon).flatten()
```
erikvansebille, line 75: Why si there no `z` here in this list? Should it be added?

src/parcels/_core/particleset.py
```diff
@@ -252,7 +252,7 @@ def populate_indices(self):
                     "X": grid_positions["X"]["index"],
                     "Y": grid_positions["Y"]["index"],
                     "Z": grid_positions["Z"]["index"],
-                }
+                }  # TODO check if reordering to Z,Y,X for dimension order consistency is possible
```
erikvansebille: So still needs to be done?

erikvansebille (commented):
Looks good, two small comments

src/parcels/_core/particleset.py
PeterWolfram: z is handled explicitly just below

src/parcels/_core/particleset.py
PeterWolfram: yes, but no change needed and comment is removed.

erikvansebille (approved):
Looks good; thanks for your contribution, @PeterWolfram

PR #2664, opened 2026-06-15
Moving RK2 and RK4 methods to separate integrators.py functions

### LINKED ISSUE #2338: Writing kernels and integration schemes in a more 'function'-like way
In a `parcels` kernel, currently (in v3 and v4), you must write your own integration scheme for timestepping your problem. While we provide RK2 and RK4 advection schemes out of the box, writing an RK2/4 kernel for a different process can be incredibly cumbersome (e.g. for Stokes drift, or, especially, biofouling).

A fantastic new enhancement of v4 is the ability to call functions inside a kernel (which was probably possible in v3, but only in slower Scipy mode), which will make coding a kernel far easier, and more modular. However, in the current implementation, a user will still need to write their own integration scheme for their own problem.

An extra enhancement for `v4` that would be nice is to have 'integrators'. Users can write their problem in a typical ODE-solver way, e.g. provide a `state` (e.g. lon, lat, depth from `particles`), and `derivatives` (in the case of advection, a function that samples `fieldset.UVW`), and they can simply call an `RK2_integrator`, or `RK4_integrator` (and so on), that will compute the `new_state` (or, probably better, the `dlon`, `dlat`, and so forth).

If of interest, I can probably come up with a working example using advection as an example.

erikvansebille:
### Description
This PR implements an idea by @michaeldenes (#2338) to provide users with the option to call an integrator (`RK2` or `RK4`) when they themselves specify the right-hand-side of a tendency equation.

It also shows how to use this, in a new section at the end of the dt integrators tutorial

Things to discuss/explore
1. Is this a good idea in the first place?
2. How/where should we put the RK2 and RK4 functions of _integrators.py in the public API? Directly under kernels? or under integrators?
3. Any other suggestions/improvements?

### Checklist

- [x] Closes #2338
- [x] Tests added

### AI Disclosure


michaeldenes:
As a preliminary thought, a way to generalise this even further might be to do something like this:

```python
# Functions defining the RHS for advection in 2D and 3D
def advection_rhs_2D(fieldset, time, particles, positions):
    lon, lat = positions #evaluate the derivative at the provided positions/intermediate values
    return fieldset.UV[time,  particles.depth, lon, lat, particles]


def advection_rhs_3D(fieldset, time, particles, positions):
    depth, lon, lat = positions #evaluate the derivative at the provided positions/intermediate values
    return fieldset.UVW[time,  depth, lon, lat, particles]


# A single RK2 function that is generalised to any dimension
def RK2(fieldset, particles, rhs, variables):
    intermediateVals = np.zeros_like(variables)
    fields = rhs(fieldset, particles.time, particles, [particles[variable] for variable in variables])
    for i, variable in enumerate(variables):
        intermediateVals[i] = particles[variable] + fields[i]*0.5*particles.dt
    
    t = particles.time + 0.5 * particles.dt
    return rhs(fieldset, t, particles, intermediateVals)

# Advection kernels
def AdvectionRK2(particles, fieldset):  # pragma: no cover
    """Advection of particles using second-order Runge-Kutta integration."""
    dt = _constrain_dt_to_within_time_interval(fieldset.time_interval, particles.time, particles.dt)

    u, v = RK2(fieldset, particles, advection_rhs_2D, ['U', 'V'])
    particles.dlon += u * dt
    particles.dlat += v * dt

def AdvectionRK2_3D(particles, fieldset):  # pragma: no cover
    """Advection of particles using second-order Runge-Kutta integration."""
    dt = _constrain_dt_to_within_time_interval(fieldset.time_interval, particles.time, particles.dt)

    u, v, w = RK2(fieldset, particles, advection_rhs_3D, ['U', 'V', 'W'])
    particles.dlon += u * dt
    particles.dlat += v * dt
    particles.ddepth += w * dt

```

In this way, you can apply RK2, RK4, etc. to any variable. (Haven't tested this, need @VeckoTheGecko help for the pixi install stuff ;)

BTW, I think a cleaner way is to just define the RHS inside the kernel, and then apply the integrator against it (like in an earlier commit).

EDIT: Added a line that would create `fields` which was missing

erikvansebille:
Nice idea, @michaeldenes. We'd need to double-check that this also works if particles is an array of depths, lats and lons (kernels are now vectored!) and think what the memory overhead is of the intermediateVals. But that will come when you try this out

Also, instead of `RK2(fieldset, ... fields)`, would it not be more intuitive to call as `RK2([fieldset.U, fieldset.V, fieldset.W], ...)?

PR #2668, opened 2026-06-16
Add `open_raw_zarr` helper

VeckoTheGecko:
### Description

This helper ingests a Zarr dataset into an Xarray dataset by bypassing Dask.

xref https://github.com/pydata/xarray/discussions/11236#discussioncomment-17266708

PR is (somewhat) blocked by #2646 - at least #2646 greatly simplifies the number of Xarray objects we have flying around.

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Just got it to right the test, and fix the code based on that

### DIFF
--- a/src/parcels/_xarray.py
+++ b/src/parcels/_xarray.py
@@ -14,26 +14,21 @@ def open_raw_zarr(store: Store):
     """Open a Zarr dataset in an Xarray dataset, bypassing Dask."""
     with xr.open_zarr(store) as ds:
         var_to_dims = {name: var.dims for name, var in ds.variables.items()}
-        coord_names = list(ds.coords)
+        var_to_attrs = {name: var.attrs for name, var in ds.variables.items()}
+        coords = {name: ds[name].variable.load() for name in ds.coords}
+        ds_attrs = ds.attrs
 
     group = zarr.open(store, mode="r")
     assert isinstance(group, zarr.Group)
 
     data_vars = {}
-    coords = {}
     for name, array in group.members():
         if not isinstance(array, zarr.Array):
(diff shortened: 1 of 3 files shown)


src/parcels/_xarray.py
erikvansebille, line 1: Is it common practice to name a file after another package? What does this signal?

tests/test_xarray.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.parametrize("ds", [pytest.param(v, id=k) for k, v in datasets.items()])
+def test_open_raw_zarr_roundtrip(ds, tmp_path):
+    path = tmp_path / "ds.zarr"
+    ds.to_zarr(path)
+
+    result = open_raw_zarr(path)
+
+    xr.testing.assert_identical(result.load(), ds)
```
erikvansebille, line 20: Do we also need to check that result is not accidentally coerced to a numpy array?

erikvansebille (commented):
Looks good; one comment and one question for my own learning

src/parcels/_xarray.py
VeckoTheGecko: > Is it common practice to name a file after another package?

I wouldn't say its wrong to do so, but its not super common either.

> What does this signal?

I'm using it to signal this is for Xarray-only utilities (ie. things that have largely nothing to do with Parcels and only to do with Xarray)

erikvansebille (approved):

tests/test_xarray.py
VeckoTheGecko: good call - done in 8b19af6d3af1

Note that we aren't doing `.data` as that _does_ get coerced. Hence we need to test private API here

PR #2672, opened 2026-06-16
Rename `FieldSet.add_constant` to `.add_context`

### LINKED ISSUE #2644: Rename FieldSet.add_constant to FieldSet.add_context
> Erik and I discussed this yesterday on a call.
> ...
> We ended up just deciding to update the naming to better reflect v4 of Parcels. We can rename:
> 
> - `constants` -> `context`
> - `add_constant` -> `add_context`
> 
> One difficulty to be aware of is that `fieldset.my_constant = 2` does not update the `constants`/`context` dict (instead defaulting to Python's built-in behaviour of accepting this since "everything is an object"). I think we should disallow setting values via `fieldset.my_constant = 2` in favour of `fieldset['my_constant']=2` or `fieldset.context['my_constant']=2` (the same way and for the same reasons that Xarray does this).
> 
> Closing this and opening a new issue for this renaming. 

 _Originally posted by @VeckoTheGecko in [#2337](https://github.com/Parcels-code/Parcels/issues/2337#issuecomment-4552021462)_

PeterWolfram:
As described in issue https://github.com/Parcels-code/Parcels/issues/2644 

1. renamed FieldSet.add_constant -> FieldSet.add_context
2. renamed FieldSet.constants -> FieldSet.context
3. added __setattr__ to raise an error if trying to attribute access a variable in FieldSet.context to prevent silent unexpected behavior 

Additionally deleted empty "_core/constants.py" file


docs/user_guide/examples_v3/documentation_stuck_particles.ipynb
erikvansebille, line 1: This is a v3 file; so shouldn't be changed. Can you revert this change? The same is true for the other files in docs/user_guide/examples_v3

src/parcels/_core/constants.py
erikvansebille, line 1: what changed here? Whitespace only?

tests-v3/test_fieldset.py
erikvansebille, line 1: This is an old file from tests-v3; these don't need to be changed

erikvansebille (commented):
Looks good; just a few small comments below!

docs/user_guide/examples_v3/documentation_stuck_particles.ipynb
VeckoTheGecko: git tip for Peter: you can quickly revert by doing `git checkout main -- docs/user_guid/examples_v3`. After the `-- ` specifies files, which is basically like "copy the files from the main branch"

src/parcels/_core/constants.py
VeckoTheGecko: Removed file :)

VeckoTheGecko:
Also, small note - would you mind following our PR template? It nice to keep things consistent, and also the `Closes #xxx` syntax means that the upstream issue gets auto-closed when this is merged

src/parcels/_core/fieldset.py
```diff
@@ -153,24 +162,24 @@ def add_constant_field(self, name: str, value, mesh: Mesh = "spherical"):
...
 
         """
         _assert_str_and_python_varname(name)
 
-        if name in self.constants:
-            raise ValueError(f"FieldSet already has a constant with name '{name}'")
+        if name in self.context:
+            raise ValueError(f"FieldSet already has a context with name '{name}'")
         if not isinstance(value, (float, np.floating, int, np.integer)):
-            raise ValueError(f"FieldSet constants have to be of type float or int, got a {type(value)}")
```
VeckoTheGecko, line 172: Just confirming @erikvansebille that this limitation of constants being numeric no longer applies? We can deal with this in a future PR

erikvansebille: Indeed, these can now be anything. @PeterWolfram, would you mind removing this ValueError - and also adding a unit test to check whether context can now be e.g. a string or a numpy array?

PR #2665, opened 2026-06-16
Fixing OutOfBounds errors on field.eval

erikvansebille:
### Description

In the previous implementation, the `GRID_SEARCH_ERROR`, `LEFT_OUT_OF_BOUNDS` and `RIGHT_OUT_OF_BOUNDS` were negative values, but this could create confusion when a user would call a `field.eval()` directly. Instead, these constants are now set to very large values

### Checklist

- [x] Closes #2629
- [x] Tests added
- [x] This PR targets the correct branch (`main` for normal development, `v3-support` for v3 support)

### AI Disclosure

- [ ] This PR contains AI-generated content.
  - [ ] I have tested any AI-generated content in my PR.
  - [ ] I take responsibility for any AI-generated content in my PR.
  - Describe how you used it (e.g., by pasting your prompt):

PR #2681, opened 2026-06-17
Fixing OutOfBounds errors on field.eval (attempt 2) 

### LINKED ISSUE #2629: `Field.eval` silently returns wrong values for out-of-domain queries

erikvansebille:
### Description
This PR supersedes #2665 by going for another approach to fixing OutOfBound errors. Instead of setting the `GRID_SEARCH_ERROR`, `LEFT_OUT_OF_BOUNDS` and `RIGHT_OUT_OF_BOUNDS` variables to very large numbers (which leads to IndexErrors in e.g. the RK4 kernel), this PR masks the returned values of particles that are out of bound to zero

In doing so, I also found that a few of the default interpolators didn't return numpy arrays (see also 2114), so that had to be fixed for all the tests to pass too

### Checklist

- [x] Closes #2629 (attempt 2)
- [x] Tests added

### AI Disclosure


src/parcels/interpolators/_uxinterpolators.py
```diff
@@ -158,5 +158,5 @@ def Ux_Velocity(
     if "3D" in vectorfield.vector_type:
         w = vectorfield.W._interp_method(particle_positions, grid_positions, vectorfield.W)
     else:
-        w = 0.0
+        w = np.zeros_like(u)
```
fluidnumericsJoe: good catch here!

src/parcels/_core/field.py
```diff
@@ -367,6 +369,15 @@ def _update_particle_states_position(particles, grid_positions: dict):
             )
 
 
+def _mask_outofbounds_values(grid_positions: dict, value):
+    mask = np.zeros(value.shape, dtype=bool)
+    for dim in ["X", "Y", "Z"]:
+        if dim in grid_positions:
+            mask[grid_positions[dim]["index"] < 0] = True
+    if np.any(mask):
+        value[mask] = 0
```
fluidnumericsJoe: A bit picky, but be explicit about this being a float, e.g.
```
    if np.any(mask):
        value[mask] = 0.0
```

src/parcels/_core/field.py
```diff
@@ -367,6 +369,15 @@ def _update_particle_states_position(particles, grid_positions: dict):
             )
 
 
+def _mask_outofbounds_values(grid_positions: dict, value):
+    mask = np.zeros(value.shape, dtype=bool)
+    for dim in ["X", "Y", "Z"]:
```
fluidnumericsJoe: This will definitely break for out of bounds access on unstructured grids

(uxgrid.py (Line 123)](https://github.com/erikvansebille/Parcels/blob/07f6f72f2330e1a38d2b35acf56ddcdd65df141e/src/parcels/_core/uxgrid.py#L123) shows the `grid_positions` dimensions are `Z` and `FACE`

```
        return {"Z": {"index": zi, "bcoord": zeta}, "FACE": {"index": fi, "bcoord": coords}}
```

tests/test_field.py
```diff
@@ -270,6 +271,26 @@ def test_field_constant_in_time():
...
+
+    # eval outside of bounds should return 0.0
+    np.testing.assert_allclose(fieldset.U.eval(0.0, 0.0, 0.0, 5e6), 0.0)
+    np.testing.assert_allclose(fieldset.UV.eval(0.0, 0.0, 0.0, 5e6), [[0.0], [0.0]])
+    np.testing.assert_allclose(fieldset.U.eval(0.0, 0.0, 5e6, 0.0), 0.0)
+    np.testing.assert_allclose(fieldset.UV.eval(0.0, 0.0, 5e6, 0.0), [[0.0], [0.0]])
+    np.testing.assert_allclose(fieldset.U.eval(0.0, 5e6, 0.0, 0.0), 0.0)
+    np.testing.assert_allclose(fieldset.UV.eval(0.0, 5e6, 0.0, 0.0), [[0.0], [0.0]])
+
+
```
fluidnumericsJoe, line 236: It'd be good to add an unstructured test here. You could make use of the `ux_constant_flow_face_centered_2D` field in https://github.com/erikvansebille/Parcels/blob/07f6f72f2330e1a38d2b35acf56ddcdd65df141e/src/parcels/_datasets/unstructured/generic.py#L408

that domain is [0,20]^2 and you could add a similar check as above.

fluidnumericsJoe (changes requested):
This PR is a straightforward fix that applies a mask to the interpolated values for particles that are aligned with negative indices. The masking function, in its current form, assumes `grid_position` dimensions are `X`, `Y` and `Z`; this is only valid for structured grids - while this resolves the associated issue, it would introduce a breaking scenario for unstructured grids. 

I've provided comments on the code changes that suggest a path forward .

tests/test_field.py
erikvansebille: Added in c93133b05237c0eb28ab12c6b787981f7d48e197

src/parcels/_core/field.py
erikvansebille: Added "FACE" to this list in c93133b05237c0eb28ab12c6b787981f7d48e197

src/parcels/_core/field.py
erikvansebille: Implemented in c93133b05237c0eb28ab12c6b787981f7d48e197

fluidnumericsJoe (approved):

tests/test_field.py
```diff
@@ -270,6 +273,54 @@ def test_field_constant_in_time():
...
+def test_field_eval_out_of_bounds_structured():
+    """Test that Field.eval returns IndexError when queried outside the grid boundaries."""
+    ds = simple_UV_dataset(mesh="flat")
+    fieldset = FieldSet.from_sgrid_conventions(ds, mesh="flat")
+    fieldset.U.data[:] = 1.0
+    fieldset.V.data[:] = 2.0
+
+    # eval inside bounds should work
+    np.testing.assert_allclose(fieldset.U.eval(0.0, 0.0, 0.0, 0.0), 1.0)
+    np.testing.assert_allclose(fieldset.UV.eval(0.0, 0.0, 0.0, 0.0), [[1.0], [2.0]])
```
VeckoTheGecko: I see that in future we would (by default) have the test suite fail if any unexpected warnings are printed. This includes FieldEval warnings like this.

As such - I think this "eval inside bounds should work" test should be removed as the whole rest of the test suite already tests it.

tests/test_field.py
```diff
@@ -270,6 +273,54 @@ def test_field_constant_in_time():
...
+
+    # eval outside of bounds should return 0.0
+    with pytest.warns(
+        FieldEvalWarning,
+        match="Some interpolated values are out-of-bounds. These values are set to 0. Treat carefully.",
+    ):
```
VeckoTheGecko, line 208: This block only tests that the warning is raised at some point during the evaluations below - not that it's being raised each time.

I think we can refactor the following test to something like the following

```python

@pytest.fixture(scope="session")
def fieldset_1_2():
    ds = simple_UV_dataset(mesh="flat")
    fieldset = FieldSet.from_sgrid_conventions(ds, mesh="flat")
    fieldset.U.data[:] = 1.0
    fieldset.V.data[:] = 2.0
    return fieldset


@pytest.mark.parametrize(
    "field_name, location, expected",
    [
        ("U", (0.0, 0.0, 0.0, 5e6), 0.0),
        ("UV", (0.0, 0.0, 0.0, 5e6), [[0.0], [0.0]]),
        ("U", (0.0, 0.0, 5e6, 0.0), 0.0),
        ("UV", (0.0, 0.0, 5e6, 0.0), [[0.0], [0.0]]),
        ("U", (0.0, 5e6, 0.0, 0.0), 0.0),
        ("UV", (0.0, 5e6, 0.0, 0.0), [[0.0], [0.0]]),
    ],
)
def test_field_eval_out_of_bounds_structured(fieldset_1_2, field_name, location, expected):
    """Test that Field.eval returns IndexError when queried outside the grid boundaries."""
    # eval outside of bounds should return 0.0
    field = getattr(fieldset_1_2, field_name)
    with pytest.warns(
        FieldEvalWarning,
        match="Some interpolated values are out-of-bounds. These values are set to 0. Treat carefully.",
    ):
        np.testing.assert_allclose(field.eval(*location), expected)

```

tests/test_field.py
```diff
@@ -270,6 +273,54 @@ def test_field_constant_in_time():
...
+        np.testing.assert_allclose(fieldset.U.eval(0.0, 0.0, 0.0, 5e6), 0.0)
+        np.testing.assert_allclose(fieldset.UV.eval(0.0, 0.0, 0.0, 5e6), [[0.0], [0.0]])
+        np.testing.assert_allclose(fieldset.U.eval(0.0, 0.0, 5e6, 0.0), 0.0)
+        np.testing.assert_allclose(fieldset.UV.eval(0.0, 0.0, 5e6, 0.0), [[0.0], [0.0]])
+        np.testing.assert_allclose(fieldset.U.eval(0.0, 5e6, 0.0, 0.0), 0.0)
+        np.testing.assert_allclose(fieldset.UV.eval(0.0, 5e6, 0.0, 0.0), [[0.0], [0.0]])
+
+
+def test_field_eval_out_of_bounds_unstructured():
```
VeckoTheGecko: Same as above

src/parcels/__init__.py
```diff
@@ -61,6 +61,7 @@
     "OutsideTimeInterval",
     "StatusCode",
     # Warnings
+    "FieldEvalWarning",
     "FieldSetWarning",
     "FileWarning",
     "KernelWarning",
```
VeckoTheGecko, line 67: nit: I wonder whether we should have a `parcels.warnings` namespace.

I'm not strongly for or against - just a convention some packages adopt.

Similarly for `parcels.errors`.

VeckoTheGecko (commented):
Code changes look good! Some comments about the tests though.

Do any changes to docs/migration guide need to be made?

Also, could we wait to merge this until #2646 is merged? (I think it would be easier to do that than have to deal with conflicts introduced in this PR on the other end)

tests/test_field.py
erikvansebille: OK, fixed in 4f015e725e8e340670f256c2fface5deecdd9524

tests/test_field.py
erikvansebille: OK, fixed in 4f015e725e8e340670f256c2fface5deecdd9524

tests/test_field.py
erikvansebille: OK, fixed in 4f015e725e8e340670f256c2fface5deecdd9524

src/parcels/__init__.py
erikvansebille: Something for a separate PR? Do you want to turn this into an Issue?

erikvansebille:
> Do any changes to docs/migration guide need to be made?

No, I don't think so. The normal Parcels workflow hasn't changed; the only change is that users now see a Warning if they direct call `Field.eval()`. I'd say not impactful enough for the migration guide

> Also, could we wait to merge this until https://github.com/Parcels-code/Parcels/pull/2646 is merged? (I think it would be easier to do that than have to deal with conflicts introduced in this PR on the other end)

Sure, let's first merge #2646

PR #2688, opened 2026-06-19
Add FACE to dims to check in _update_particle_states_position

### LINKED ISSUE #2687: Least square interpolators and piece-wise constant face interpolators particles out of bounds
### Parcels version

3.1.3.dev2079

### Description

https://github.com/danliba/elphe-hackathon/blob/interpolation-comparison-fesom/interpolation_comparison_DBG_fesom.ipynb

### Code sample

```python
# Paste your code within this block
```

```text
# Paste your error message within this block
```

fluidnumericsJoe:
Resolves the TODO comment and issue #2687

### Checklist

- [x] Closes #2687 
- [x] Tests added

### DIFF
--- a/tests/test_particleset_execute.py
+++ b/tests/test_particleset_execute.py
@@ -18,6 +18,7 @@ from parcels import (
     Variable,
     VectorField,
 )
+from parcels._core.statuscodes import GridSearchingError
 from parcels._core.utils.time import timedelta_to_float
 from parcels._datasets.structured.generated import simple_UV_dataset
 from parcels._datasets.structured.generic import datasets as datasets_structured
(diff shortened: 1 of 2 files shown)


tests/test_particleset_execute.py
```diff
@@ -592,3 +593,26 @@ def test_uxstommelgyre_pset_execute_output():
...
+    UV = VectorField(name="UV", U=U, V=V, interp_method=Ux_Velocity)
+    fieldset = FieldSet([UV, UV.U, UV.V])
+
+    # Uniform eastward flow (U0 = 0.001 deg/s); release 0.1 deg inside the eastern
+    # outflow boundary (domain spans lon, lat in [0, 20]). The particle crosses
+    # lon=20 after 100 s of travel.
+    lon_max = float(ds.uxgrid.node_lon.max())
+    pset = ParticleSet(fieldset, lon=[lon_max - 0.1], lat=[10.0], z=[0.5], pclass=Particle)
+
+    with pytest.raises(GridSearchingError):
```
erikvansebille, line 617: Why does this not throw an ErrorOutOfBounds? Would that not be more similar to how it's done in Sgrid?

erikvansebille (commented):
Good catch and solution; only one question below

tests/test_particleset_execute.py
fluidnumericsJoe: The grid search returns a value of [`GRID_SEARCH_ERROR`](https://github.com/Parcels-code/Parcels/blob/6d0a417b68c7a35596fbb03ca07926739c31eba3/src/parcels/_core/index_search.py#L15) on the face index when it's not found in the grid; this is done in `spatialhash.query()`

fluidnumericsJoe: [ErrorOutOfBounds is thrown when the search index is `-1`](https://github.com/Parcels-code/Parcels/blob/6d0a417b68c7a35596fbb03ca07926739c31eba3/src/parcels/_core/field.py#L343)

`GRID_SEARCH_ERROR=-3` which [triggers this line](https://github.com/Parcels-code/Parcels/blob/6d0a417b68c7a35596fbb03ca07926739c31eba3/src/parcels/_core/field.py#L347)

fluidnumericsJoe: Why does the Sgrid return a value of grid index -1, corresponding to [`RIGHT_OUT_OF_BOUNDS`](https://github.com/Parcels-code/Parcels/blob/6d0a417b68c7a35596fbb03ca07926739c31eba3/src/parcels/_core/index_search.py#L17) ?

```
GRID_SEARCH_ERROR = -3
LEFT_OUT_OF_BOUNDS = -2
RIGHT_OUT_OF_BOUNDS = -1
```

erikvansebille: Ah I see; I was confused

We use these LEFT and RIGHT only in the 1d_search:
https://github.com/Parcels-code/Parcels/blob/481decc20c661da6093b1b4d622f73e5b20a7fa7/src/parcels/_core/index_search.py#L59-L60

This is especially useful for z/depth, where users may want to distinguish between a particle moving into the air or into the ground. And it just so happens we can also use it on rectilinear grids in the horizontal, because these use 1d_search

On CurviLinear grids, Xgrid would also not be able to disinguish between LEFT and RIGHT. So your test is fine as is

erikvansebille (approved):

PR #2694, opened 2026-06-23
Clean out `tests/test_data`

VeckoTheGecko:
### Description

Pre-v4 of Parcels there were a bunch of datasets that were committed to the repo for testing and tutorials. This PR removes a bunch of these datasets, only keeping those that are currently used in the v4 test suite in the `test_interp_regression_v3` test (i.e., datasets `test_interpolation_*`).

I imagine those datasets will also be removed at some point (likely post-v4 release, or after we have other validation outputs).

Do you think this PR makes sense to do now @erikvansebille ?

### Checklist

- [x] xref https://github.com/Parcels-code/Parcels/issues/1950 (not "fixes" since we haven't removed all the datasets)

### AI Disclosure

None used


erikvansebille (approved):
I see you removed the tests already; fine to merge it now

PR #2697, opened 2026-06-24
MAINT: Add `docs-clean` Pixi task 

VeckoTheGecko:
### Description

It would be great to clean out artifacts (e.g., generated Parquet files) from the docs folder during building.

This PR does the following:
- Adds a `docs-clean` Pixi task which first cleans the `docs` folder then builds the docs
- Adds a `docs-only-clean` which just cleans the docs folder
- Updates the Pixi tasks to use the tasks used in the makefile definition
  - It would be great if we could just remove the makefile entirely, and leverage Pixi's task running for everything, however unfortunately the Pixi task runner [isn't as flexible as we need it to be.](https://github.com/prefix-dev/pixi/issues/4208#issuecomment-4789425517)

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

None

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,6 +14,6 @@ build:
     build:
       html:
         - pixi run -e docs python ci/scripts/download-all-tutorial-datasets.py
-        - pixi run -e docs sphinx-build -T -b html docs $READTHEDOCS_OUTPUT/html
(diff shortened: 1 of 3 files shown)


erikvansebille (approved):
`pixi run docs-clean` is a great idea!

VeckoTheGecko:
Will merge once CI is fixed and this is rebased

PR #2703, opened 2026-06-25
Add tutorial_write_in_kernel to the docs page

### LINKED ISSUE #2599: Explore and expose use of writing particles in a Kernel

erikvansebille:
### Description

While #2620 implemented the functionality to write particle sets to output inside a Kernel and provided a tutorial for it, I now realise that that tutorial was not added to the docs page, so was not easily findable there. This PR adds it

### Checklist

- [x] Closes #2599 

### AI Disclosure

### DIFF
--- a/docs/user_guide/index.md
+++ b/docs/user_guide/index.md
@@ -51,6 +51,7 @@ examples/tutorial_delaystart.ipynb
 examples/explanation_kernelloop.md
 examples/tutorial_sampling.ipynb
 examples/tutorial_statuscodes.ipynb
+examples/tutorial_write_in_kernel.ipynb
 ```
 
 ```{toctree}


erikvansebille:
Oops, I now see that I had already addressed this in #2657, so closing here

PR #2715, opened 2026-07-01
Add `vector_fields` parameter to FieldSet creation methods

### LINKED ISSUE #2695: Decide on API for defining VectorFields
With #2646 - we need to decide on an API for adding VectorFields.

---

I was looking at the docs build, and saw an example that adds a second velocity field

```python
import numpy as np
import xarray as xr

import parcels
import parcels.tutorial

ds_fields = parcels.tutorial.open_dataset("CopernicusMarine_data_for_Argo_tutorial/data")

# Create an idealised wind field and add it to the dataset
tdim, ydim, xdim = (len(ds_fields.time),len(ds_fields.latitude), len(ds_fields.longitude))
ds_fields["UWind"] = xr.DataArray(
    data=np.ones((tdim, ydim, xdim)) * np.sin(ds_fields.latitude.values)[None, :, None],
    coords=[ds_fields.time, ds_fields.latitude, ds_fields.longitude])

ds_fields["VWind"] = xr.DataArray(
    data=np.zeros((tdim, ydim, xdim)),
    coords=[ds_fields.time, ds_fields.latitude, ds_fields.longitude])

fields = {
    "U": ds_fields["uo"],
    "V": ds_fields["vo"],
    "UWind": ds_fields["UWind"],
    "VWind": ds_fields["VWind"],
}

# Create a vecorfield for the wind
windvector = parcels.VectorField(
    "Wind",
    fieldset.UWind,
    fieldset.VWind,
    interp_method=parcels.interpolators.XLinear_Velocity
)
fieldset.add_field(windvector)

```

We no longer have `add_field` - I think we need to map out exactly how we should be defining velocity fields under this new refactoring... I think that can be done in a future PR

_Originally posted by @VeckoTheGecko in https://github.com/Parcels-code/Parcels/issues/2646#issuecomment-4778812419_

VeckoTheGecko:
### Description

This PR adds a `vector_fields` parameter which allows users to control vector fields that are created.

Looking at the following tests you can see how they're used

```python

def test_fieldset_structured_vectorfield_default():
    ds = datasets_structured["ds_2d_left"][["U_A_grid", "V_A_grid", "grid"]].rename({"U_A_grid": "U", "V_A_grid": "V"})

    fset = FieldSet.from_sgrid_conventions(ds, mesh="flat")

    assert "U" in fset.fields
    assert "V" in fset.fields
    assert "UV" in fset.fields


def test_fieldset_structured_vectorfield_custom():
    ds = datasets_structured["ds_2d_left"][["U_A_grid", "V_A_grid", "grid"]].rename({"U_A_grid": "U", "V_A_grid": "V"})
    ds = ds.rename({"U": "U_wind", "V": "V_wind"})

    fset = FieldSet.from_sgrid_conventions(ds, mesh="flat", vector_fields={"UV_wind": ("U_wind", "V_wind")})

    assert "U_wind" in fset.fields
    assert "V_wind" in fset.fields
    assert "UV_wind" in fset.fields


def test_fieldset_structured_vectorfield_none():
    ds = datasets_structured["ds_2d_left"][["U_A_grid", "V_A_grid", "grid"]].rename({"U_A_grid": "U", "V_A_grid": "V"})

    fset = FieldSet.from_sgrid_conventions(ds, mesh="flat", vector_fields=None)

    assert "U" in fset.fields
    assert "V" in fset.fields
    assert "UV" not in fset.fields
```

Now users can:
- specify the names of vectorfields and their components
- explicitly set that no vector fields are created (by doing `vector_fields = {}` or `vector_fields = None`)

If `vector_fields` is not provided it uses the default behaviour of constructing UV and UVW if the component fields are present.

These changes have also been added to the unstructured grid side as well.

### Checklist

- [x] Closes #2695 
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): only for updating the docstring

### DIFF
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -211,7 +211,7 @@ class FieldSet:
         cls,
         ds: ux.UxDataset,
         mesh: str = "spherical",
-        vector_fields: ptyping.VectorFields | None | NotSetType = NOTSET,
+        vector_fields: ptyping.VectorFields | NotSetType = NOTSET,
     ):
         """Create a FieldSet from a Parcels compliant uxarray.UxDataset.
 
@@ -229,8 +229,7 @@ class FieldSet:
         vector_fields : Mapping[str, tuple[str, ...]] or None, optional
             Mapping of vector field names to tuples of component variable names in the dataset.
             For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
-            If ``None``, no vector fields are constructed. If omitted (default), vector fields
-            are auto-discovered from standard variable names (``U``/``V``/``W``).
+            If omitted (default), vector fields are auto-discovered from standard variable names (``U``/``V``/``W``).
 
         Returns
         -------
@@ -249,7 +248,7 @@ class FieldSet:
         cls,
         ds: xr.Dataset,
         mesh: ptyping.Mesh | None = None,
-        vector_fields: ptyping.VectorFields | None | NotSetType = NOTSET,
+        vector_fields: ptyping.VectorFields | NotSetType = NOTSET,
     ):  # TODO: Update mesh to be discovered from the dataset metadata
         """Create a FieldSet from a dataset using SGRID convention metadata.
 
@@ -267,8 +266,7 @@ class FieldSet:
         vector_fields : Mapping[str, tuple[str, ...]] or None, optional
             Mapping of vector field names to tuples of component variable names in the dataset.
             For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
-            If ``None``, no vector fields are constructed. If omitted (default), vector fields
-            are auto-discovered from standard variable names (``U``/``V``/``W``).
+            If omitted (default), vector fields are auto-discovered from standard variable names (``U``/``V``/``W``).
 
         Returns
         -------
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -263,7 +263,7 @@ class FieldSet:
         mesh : str
             String indicating the type of mesh coordinates used during
             velocity interpolation. Options are "spherical" or "flat".
-        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
+        vector_fields : Mapping[str, tuple[str, ...]], optional
             Mapping of vector field names to tuples of component variable names in the dataset.
             For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
             If omitted (default), vector fields are auto-discovered from standard variable names (``U``/``V``/``W``).
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -14,7 +14,7 @@ from parcels._core.model import (
     CONSTANT_FIELD_MODELS,
     ModelData,
     StructuredModelData,
-    TVectorFieldMapping,
+    TVectorField,
     UnstructuredModelData,
 )
 from parcels._core.utils.string import _assert_str_and_python_varname
@@ -212,7 +212,7 @@ class FieldSet:
         cls,
         ds: ux.UxDataset,
         mesh: str = "spherical",
-        vector_fields: TVectorFieldMapping | None | NotSetType = NOTSET,
+        vector_fields: TVectorField | None | NotSetType = NOTSET,
     ):
         """Create a FieldSet from a Parcels compliant uxarray.UxDataset.
 
@@ -250,7 +250,7 @@ class FieldSet:
         cls,
         ds: xr.Dataset,
         mesh: Mesh | None = None,
-        vector_fields: TVectorFieldMapping | None | NotSetType = NOTSET,
+        vector_fields: TVectorField | None | NotSetType = NOTSET,
     ):  # TODO: Update mesh to be discovered from the dataset metadata
         """Create a FieldSet from a dataset using SGRID convention metadata.
 
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -226,7 +226,7 @@ class FieldSet:
         ----------
         ds : uxarray.UxDataset
             uxarray.UxDataset as obtained from the uxarray package but with appropriate named vertical dimensions
-        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
+        vector_fields : Mapping[str, tuple[str, ...]], optional
             Mapping of vector field names to tuples of component variable names in the dataset.
             For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
             If omitted (default), vector fields are auto-discovered from standard variable names (``U``/``V``/``W``).
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -20,7 +20,7 @@ from parcels._core.model import (
 from parcels._core.utils.string import _assert_str_and_python_varname
 from parcels._core.utils.time import get_datetime_type_calendar
 from parcels._core.utils.time import is_compatible as datetime_is_compatible
-from parcels._python import _MISSING, _MissingType
+from parcels._python import NOTSET, NotSetType
 from parcels._typing import Mesh
 from parcels.interpolators import (
     XConstantField,
@@ -212,7 +212,7 @@ class FieldSet:
         cls,
         ds: ux.UxDataset,
         mesh: str = "spherical",
-        vector_fields: TVectorFieldMapping | None | _MissingType = _MISSING,
+        vector_fields: TVectorFieldMapping | None | NotSetType = NOTSET,
     ):
         """Create a FieldSet from a Parcels compliant uxarray.UxDataset.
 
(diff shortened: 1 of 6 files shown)


src/parcels/_core/fieldset.py
```diff
@@ -201,7 +208,12 @@ def gridset(self) -> list[BaseGrid]:
         return grids
 
     @classmethod
-    def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: str = "spherical"):
+    def from_ugrid_conventions(
+        cls,
+        ds: ux.UxDataset,
+        mesh: str = "spherical",
+        vector_fields: TVectorFieldMapping | None | NotSetType = NOTSET,
```
VeckoTheGecko: `NOTSET` here is an example of a "sentinel value". 

They're handy in the case of wanting to distinguish between `None` and "the default behaviour"

in our case here we want `vector_fields=None` to mean no vector fields, and the default behaviour to autodiscover based on if there are U and V fields.

src/parcels/_core/fieldset.py
```diff
@@ -201,7 +208,12 @@ def gridset(self) -> list[BaseGrid]:
         return grids
 
     @classmethod
-    def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: str = "spherical"):
+    def from_ugrid_conventions(
+        cls,
+        ds: ux.UxDataset,
+        mesh: str = "spherical",
+        vector_fields: TVectorField | None | NotSetType = NOTSET,
```
erikvansebille: I don't understand why we would allow both `vector_fields={}` and `vector_fields=None` - which then end up with the same result. Is that not confusing to users? Can we not choose one of those options?

src/parcels/_core/model.py
```diff
@@ -32,11 +34,14 @@
 )
 from parcels.interpolators._base import ScalarInterpolator, VectorInterpolator
 
+TVectorField = dict[str, tuple[str, str] | tuple[str, str, str]]
```
erikvansebille: Maybe it's me, but I find this difficult to read. I keep seeing "Television" in the first two letters. Does the `T` stand for Type? If so, why not spell it out (either prepended or appended)?

src/parcels/_core/fieldset.py
erikvansebille: Just discussed in person, we will not support `vector_fields = None`

src/parcels/_core/fieldset.py
```diff
@@ -215,6 +227,11 @@ def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: str = "spherical"):
         ----------
         ds : uxarray.UxDataset
             uxarray.UxDataset as obtained from the uxarray package but with appropriate named vertical dimensions
+        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
+            Mapping of vector field names to tuples of component variable names in the dataset.
+            For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
+            If ``None``, no vector fields are constructed. If omitted (default), vector fields
```
erikvansebille: Also update doctoring here

src/parcels/_core/fieldset.py
```diff
@@ -245,6 +265,11 @@ def from_sgrid_conventions(
         mesh : str
             String indicating the type of mesh coordinates used during
             velocity interpolation. Options are "spherical" or "flat".
+        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
+            Mapping of vector field names to tuples of component variable names in the dataset.
+            For example, ``{"UV": ("U", "V"), "UVW": ("U", "V", "W")}``.
+            If ``None``, no vector fields are constructed. If omitted (default), vector fields
+            are auto-discovered from standard variable names (``U``/``V``/``W``).
```
erikvansebille: Update this doctoring to remove `None` option

src/parcels/_core/model.py
```diff
@@ -110,30 +116,25 @@ def construct_fields(self) -> list[Field | VectorField]:
...
+            component_fields = [single_fields[name] for name in components]
+            vector_fields[vfield_name] = VectorField(vfield_name, *component_fields, interp_method=interp_method)  # type:ignore[misc,arg-type]
+
+        fields: dict[str, Field | VectorField] = {**single_fields, **vector_fields}
         return list(fields.values())
 
     @classmethod
-    def from_sgrid_conventions(cls, ds: xr.Dataset, mesh: Mesh | None = None) -> Self:
+    def from_sgrid_conventions(
+        cls, ds: xr.Dataset, mesh: Mesh | None, vector_fields: TVectorField | None | NotSetType
```
erikvansebille: Also remove None option here

src/parcels/_core/model.py
```diff
@@ -110,30 +116,25 @@ def construct_fields(self) -> list[Field | VectorField]:
...
-        fields: dict[str, Field | VectorField] = {**single_fields, **vector_fields}
-        for varname in set(scalar_field_names) - set(fields.keys()):
-            fields[varname] = Field(str(varname), self)
+        for varname in set(scalar_field_names):
+            single_fields[varname] = Field(str(varname), self)
 
+        for vfield_name, components in self.vector_field_components.items():
+            interp_method = (
+                XLinear_Velocity() if _is_agrid(self.data, u=components[0], v=components[1]) else CGrid_Velocity()
+            )
```
erikvansebille, line 125: How would this work for `ugrids`? Should we not test for them too? Or is this function not used by from_ugrid?

src/parcels/_core/model.py
```diff
@@ -239,7 +259,7 @@ def scalar_field_names(self) -> list[str]:
         return list(self.data.data_vars)
 
     @classmethod
-    def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: str = "spherical"):
+    def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: Mesh, vector_fields: TVectorField | None | NotSetType):
```
erikvansebille: And again, remove the None here?

src/parcels/_python.py
```diff
@@ -1,11 +1,15 @@
...
+import enum
 import inspect
 from collections.abc import Callable, Mapping
 from typing import TypeVar
 
 K = TypeVar("K")
 V = TypeVar("V")
 
+NotSetType = enum.Enum("NotSetType", "VALUE")
+NOTSET = NotSetType.VALUE
```
erikvansebille, line 11: Not for this PR, but would there be a benefit in also using this `NOTSET` when postponing setting the particle.time?
https://github.com/Parcels-code/Parcels/blob/32a7f2a2a7737eefa7cb3db2af079183bd5b091b/src/parcels/_core/particleset.py#L89-L91

tests/test_fieldset.py
erikvansebille, line 1: I don't see that we also test creating a fieldset with two vector_fields. Would that not be a good test to have?

erikvansebille (commented):
Thanks for this quick PR! See below some comments

src/parcels/_core/model.py
VeckoTheGecko: Prefixing with a T like this (as well as the camelcasing) is common naming convention in Python code for types.

Alternatively we could move this to parcels._typing and use `pt.VectorFields`?

Thinking about it now, i think i prefer the latter

erikvansebille: If it's common practise then I'm ok with it. Although I also like the `pt.VectorFields` approach. I leave it up to you

src/parcels/_core/model.py
VeckoTheGecko: > Or is this function not used by from_ugrid?

Yes.

The unstructured ModelData class has a separate `construct_fields` implementation

```python
for vfield_name, components in self.vector_field_components.items():
            interp_method = Ux_Velocity()

            component_fields = [single_fields[name] for name in components]
            vector_fields[vfield_name] = VectorField(vfield_name, *component_fields, interp_method=interp_method)  # type:ignore[misc, arg-type]
```

tests/test_fieldset.py
VeckoTheGecko: Done now in 61c8c6ad5548 by improving `test_fieldset_add`

src/parcels/_python.py
VeckoTheGecko: I'm not quite sure what you mean by this. Do you mean to use `NOTSET` instead of `None` for the time parameter in the init of the particleset?

VeckoTheGecko:
All feedback has been addressed - ready for re-review

src/parcels/_core/fieldset.py
```diff
@@ -215,6 +226,10 @@ def from_ugrid_conventions(cls, ds: ux.UxDataset, mesh: str = "spherical"):
         ----------
         ds : uxarray.UxDataset
             uxarray.UxDataset as obtained from the uxarray package but with appropriate named vertical dimensions
+        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
```
erikvansebille:
```suggestion
        vector_fields : Mapping[str, tuple[str, ...]], optional
```

src/parcels/_core/fieldset.py
```diff
@@ -245,6 +263,10 @@ def from_sgrid_conventions(
         mesh : str
             String indicating the type of mesh coordinates used during
             velocity interpolation. Options are "spherical" or "flat".
+        vector_fields : Mapping[str, tuple[str, ...]] or None, optional
```
erikvansebille:
```suggestion
        vector_fields : Mapping[str, tuple[str, ...]], optional
```

erikvansebille (approved):
Looks good. Two suggested changes for the docstring below

PR #2719, opened 2026-07-03
Rename particle position attributes from "lon" and "lat" to "x" and "y"

### LINKED ISSUE #2667: Rename particle position attributes from (lon, lat, depth) to (x,y,z)

v4

Parent models don't always define positions on globe/spherical coordinates. Hard-setting the names of the particle positions to `.lon`, `.lat`, and `.depth` implies that this is indeed the norm. Many high resolution regional simulation use cartesian coordinates and users of these kinds of models often just use `x` `y` and `z` . 

I would propose that `particles` use attributes `x` `y` and `z` as the names for the particle position attributes (rather than `lon`, `lat`, and `depth`). This would also mean that the increments in position be recorded as `dx`, `dy`, `dz`.

The interpretation of the meaning of the generic `x` `y` and `z` is in the hands (minds?) of the parcels user and in the units aligned with each dimension (degrees longitude/latitude for spherical coordinates or meters for cartesian).

_No response_

### Additional context

_No response_

VeckoTheGecko:
### Description

Kernel variables `lon` and `lat` have been renamed to `x` and `y`, and `dlon` and `dlat` have been renamed to `dx` and `dy`. These changes are also reflected on the ParticleSet as well as the particlefile output.

Changes:

- Update particle definition and all related code
  - particle.py particle variables
  - ParticleSet args
  - all internal mentions of .lon and .lat
  - test suite code
- See commit history for other changes

This now means that users will need to use `x` and `y` when refering to particle positions. This includes in kernels, when defining the particlesets themselves, and when working with output after the face.

Note that I haven't gone about updating local variable names in tests. I don't think it provides any benefit to do so (and for those tests using "lon" might be more informative than using "x" as it hints at the mesh type used).

### Checklist

- [x] Closes #2667
- [x] Tests added

### AI Disclosure

None used.


VeckoTheGecko:
Also note that I haven't updated the docs as these are currently broken due to the last refactoring. We should fix them in a separate PR.

docs/user_guide/examples_v3/example_brownian.py
erikvansebille, line 1: These are in the examples_v3 directory; I thought we wouldn't change these? Keep them in v3 format?

docs/user_guide/v4-migration.md
```diff
@@ -18,6 +18,7 @@ Version 4 of Parcels is unreleased at the moment. The information in this migrat
 - Users need to explicitly use `convert_z_to_sigma_croco` in sampling kernels (such as the `AdvectionRK4_3D_CROCO` or `SampleOMegaCroco` kernels) when working with CROCO data, as the automatic conversion from depth to sigma grids under the hood has been removed.
 - We added a new AdvectionRK2 Kernel. The AdvectionRK4 kernel is still available, but RK2 is now the recommended default advection scheme as it is faster while the accuracy is comparable for most applications. See also the Choosing an integration method tutorial.
 - Functions shouldn't be converted to Kernels before adding to a pset.execute() call. Instead, simply pass the function(s) as a list to pset.execute().
+- Kernel variables `lon` and `lat` have been renamed to `x` and `y`, and `dlon` and `dlat` have been renamed to `dx` and `dy`. These changes are also reflected on the ParticleSet as well as the particlefile output.
```
erikvansebille: As we try to consequently use the time, depth, lat, Lon order, I would swap these here too


```suggestion
- Kernel variables `lat` and `lon` have been renamed to `y` and `x`, and `dlat` and `dlon` have been renamed to `dy` and `dx`. These changes are also reflected on the ParticleSet as well as the ParticleFile output.
```

src/parcels/_core/particle.py
```diff
@@ -141,18 +141,26 @@ def get_default_particle(spatial_dtype: type[np.float32] | type[np.float64]) ->
...
             Variable(
-                "lat",
+                "y",
                 dtype=spatial_dtype,
-                attrs={"standard_name": "latitude", "units": "degrees_north", "axis": "Y"},
+                attrs={
+                    "standard_name": "latitude",
+                    "units": "degrees_north",
+                    "axis": "Y",
+                },  # TODO v4: Update to ensure that units come from the model
```
erikvansebille: Make this an Issue? Or link to an Issue if it already exists?

docs/user_guide/v4-migration.md
erikvansebille: Should we also update `time` to `t`? 

Because a user would now sample by `fieldset.field[particles.time, particles.z, particles.y, particles.x]`. This could be more 'symmetric' with
```python
fieldset.field[particles.t, particles.z, particles.y, particles.x]
```

tests-v3/tools/test_warnings.py
erikvansebille, line 1: Again, a few changes to files in tests-v3; not necessary?

erikvansebille (commented):
Code changes look good; see below a few minor comments

Two more important ones
1. Should we also change `particles.time` to `particles.t`, for symmetry with the other one-letter attributes?
2. Would it be helpful to users to provide some informative error message if they create a ParticleSet or Kernel with `lon`, informing them that this has changed in v4? Or do we expect them to just read the migration guide? I guess that this will be one of the most impactful changes for users wanting to port over their code (along with the vectorization of the particleset).

VeckoTheGecko:
> 2\. Would it be helpful to users to provide some informative error message if they create a ParticleSet or Kernel with `lon`, informing them that this has changed in v4? Or do we expect them to just read the migration guide? I guess that this will be one of the most impactful changes for users wanting to port over their code (along with the vectorization of the particleset).

That's a good idea. Added to https://github.com/Parcels-code/Parcels/issues/2109

erikvansebille (approved):

VeckoTheGecko:
> 1. Should we also change `particles.time` to `particles.t`, for symmetry with the other one-letter attributes?

OK, let's do that. Tracking in a separate issue to be done in another PR.

VeckoTheGecko:
@erikvansebille merged now!

PR #2722, opened 2026-07-06
Pin xgcm version

VeckoTheGecko:
### Description

https://github.com/xgcm/xgcm/pull/743 resulted in the dropping of Axis from the public API hence we can't use it for typechecking. We'll very likely drop xgcm entirely before release, so just pinning for now.

### Checklist

- [x] Closes None

### AI Disclosure

None

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -30,7 +30,7 @@ holoviews = ">=1.22.0" # https://github.com/prefix-dev/rattler-build/issues/2326
 uxarray = ">=2026.04.1"
 dask = ">=2024.5.1"
 zarr = ">=3"
-xgcm = { git = "https://github.com/xgcm/xgcm", rev = "master" } # TODO: Switch to release version after release is cut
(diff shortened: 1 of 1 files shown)


hdrake:
@VeckoTheGecko , it is removed from the top-level public API but can't you still use it like this?
```
from xgcm.axis import Axis
```

hdrake:
Also, can you point me towards an Issue/Discussion/PR that discusses dropping xgcm entirely? What is the reason for this? Helpful for me to know why `xgcm` is or is not serving community needs!

VeckoTheGecko:
Hi @hdrake . We only used xgcm for the parsing of the COMODO/SGRID metadata conventions. Due to a bug in xgcm's parser (https://github.com/xgcm/xgcm/issues/686 ) we now we have our own SGRID parsing ( https://github.com/Parcels-code/Parcels/pull/2418 ) which is more reliable and maintainable than the logic in xgcm. After some refactoring, xgcm will become an unused dependency.

Deepak and Julius flagged interest in https://github.com/xgcm/xgcm/issues/686 about adding SGRID to cf_xarray (or an `xarray-sgrid` package) so its more interoperable. Our sgrid module is intentionally quite orthogonal, and can be easily lifted into a different package at some point for those interested in doing so.

PR #2729, opened 2026-07-06
FEAT: Add `fieldset.describe()`

### LINKED ISSUE #2683: Add `FieldSet.describe()` and remove all heavy, verbose reprs
In v3, the reprs were the main way to introspect the Parcels objects (also since the fields were quite hidden under the hood). Now that we work from datasets, this is less of a priority.

I propose that we remove all verbose reprs from Parcels, and instead have a `.describe()` method on FieldSet for users to request pretty summaries. 

This is nicer since reprs are used everywhere, and often users aren't interested in this prettified output

VeckoTheGecko:
### Description

Add `fieldset.descibe()`. Also adds interpolator reprs of `self.__class__.__name__ + '(...)'`

Example output:

```
| Type          | Dataset origin   | Name     | Interp method         |
|:--------------|:-----------------|:---------|:----------------------|
| Context value | -                | my_list  | [1, 2, 'hello']       |
| Context value | -                | my_value | 2.0                   |
| Field         | 0                | U        | XLinear(...)          |
| Field         | 0                | V        | XLinear(...)          |
| VectorField   | 0                | UV       | XLinear_Velocity(...) |
| Field         | 1                | U_wind   | XLinear(...)          |
| Field         | 1                | V_wind   | XLinear(...)          |
| VectorField   | 1                | UV_wind  | XLinear_Velocity(...) |
```

As part of this PR we add the `tabulate` dependency (optional dependency of Pandas).

I think that's all the information needed to be surfaced by the `.describe()` method, and I don't think any of the other classes need a method like this.

### Checklist

- [x] Closes #2683
- [x] Tests added

### AI Disclosure

None

### DIFF
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -393,17 +393,17 @@ def test_fieldset_add_context_values():
 def test_fieldset_describe(fieldset_two_models: FieldSet):
     fieldset = fieldset_two_models
     expected = """\
-| Name           | Type        | Dataset origin   | Interp method / value   |
-|:---------------|:------------|:-----------------|:------------------------|
-| my_list        | Context     | -                | [1, 2, 'hello']         |
-| my_value       | Context     | -                | 2.0                     |
-| U              | Field       | 0                | XLinear(...)            |
-| V              | Field       | 0                | XLinear(...)            |
-| UV             | VectorField | 0                | XLinear_Velocity(...)   |
-| U_wind         | Field       | 1                | XLinear(...)            |
-| V_wind         | Field       | 1                | XLinear(...)            |
-| UV_wind        | VectorField | 1                | XLinear_Velocity(...)   |
-| constant_field | Field       | 2                | XConstantField(...)     |
+| Name           | Type        | Grid number   | Interp method / value   |
+|:---------------|:------------|:--------------|:------------------------|
+| my_list        | Context     | -             | [1, 2, 'hello']         |
+| my_value       | Context     | -             | 2.0                     |
+| U              | Field       | 0             | XLinear(...)            |
+| V              | Field       | 0             | XLinear(...)            |
+| UV             | VectorField | 0             | XLinear_Velocity(...)   |
+| U_wind         | Field       | 1             | XLinear(...)            |
+| V_wind         | Field       | 1             | XLinear(...)            |
+| UV_wind        | VectorField | 1             | XLinear_Velocity(...)   |
+| constant_field | Field       | 2             | XConstantField(...)     |
 
 mesh: flat
 time interval: (np.datetime64('2000-01-01T00:00:00.000000000'), np.datetime64('2001-01-01T00:00:00.000000000'))"""
--- a/tests/test_fieldset.py
+++ b/tests/test_fieldset.py
@@ -393,16 +393,16 @@ def test_fieldset_add_context_values():
 def test_fieldset_describe(fieldset_two_models):
     fieldset = fieldset_two_models
     expected = """\
-| Type          | Dataset origin   | Name           | Interp method / value   |
-|:--------------|:-----------------|:---------------|:------------------------|
-| Context value | -                | my_list        | [1, 2, 'hello']         |
-| Context value | -                | my_value       | 2.0                     |
-| Field         | 0                | U              | XLinear(...)            |
-| Field         | 0                | V              | XLinear(...)            |
-| VectorField   | 0                | UV             | XLinear_Velocity(...)   |
-| Field         | 1                | U_wind         | XLinear(...)            |
-| Field         | 1                | V_wind         | XLinear(...)            |
-| VectorField   | 1                | UV_wind        | XLinear_Velocity(...)   |
-| Field         | 2                | constant_field | XConstantField(...)     |"""
+| Type        | Dataset origin   | Name           | Interp method / value   |
+|:------------|:-----------------|:---------------|:------------------------|
+| Context     | -                | my_list        | [1, 2, 'hello']         |
+| Context     | -                | my_value       | 2.0                     |
+| Field       | 0                | U              | XLinear(...)            |
+| Field       | 0                | V              | XLinear(...)            |
+| VectorField | 0                | UV             | XLinear_Velocity(...)   |
+| Field       | 1                | U_wind         | XLinear(...)            |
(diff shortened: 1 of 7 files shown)


VeckoTheGecko:
Just realised that the header `Interp method` doesn't make sense for the Context values. Should I update to `Interp method / value`?

VeckoTheGecko:
Updated table:

```
| Type          | Dataset origin   | Name           | Interp method / value   |
|:--------------|:-----------------|:---------------|:------------------------|
| Context value | -                | my_list        | [1, 2, 'hello']         |
| Context value | -                | my_value       | 2.0                     |
| Field         | 0                | U              | XLinear(...)            |
| Field         | 0                | V              | XLinear(...)            |
| VectorField   | 0                | UV             | XLinear_Velocity(...)   |
| Field         | 1                | U_wind         | XLinear(...)            |
| Field         | 1                | V_wind         | XLinear(...)            |
| VectorField   | 1                | UV_wind        | XLinear_Velocity(...)   |
| Field         | 2                | constant_field | XConstantField(...)     |
```

tests/test_fieldset.py
```diff
@@ -373,3 +388,21 @@ def test_fieldset_add_context_values():
 
     assert fset.context["c1"] == 1.0
     assert fset.context["c2"] == 2.0
+
+
+def test_fieldset_describe(fieldset_two_models):
+    fieldset = fieldset_two_models
+    expected = """\
+| Type          | Dataset origin   | Name           | Interp method / value   |
```
erikvansebille: Also add `_mesh` information to the table?

erikvansebille: Make `Name` the first column?

erikvansebille: I'm not sure I understand what the integer in the Dataset origin column refers to

tests/test_fieldset.py
```diff
@@ -373,3 +388,21 @@ def test_fieldset_add_context_values():
...
+|:--------------|:-----------------|:---------------|:------------------------|
+| Context value | -                | my_list        | [1, 2, 'hello']         |
+| Context value | -                | my_value       | 2.0                     |
```
erikvansebille: Can we change the order to
1. Fields
2. VectorFields
3. Context

tests/test_fieldset.py
```diff
@@ -373,3 +388,21 @@ def test_fieldset_add_context_values():
...
     assert fset.context["c1"] == 1.0
     assert fset.context["c2"] == 2.0
+
+
+def test_fieldset_describe(fieldset_two_models):
+    fieldset = fieldset_two_models
+    expected = """\
+| Type          | Dataset origin   | Name           | Interp method / value   |
+|:--------------|:-----------------|:---------------|:------------------------|
+| Context value | -                | my_list        | [1, 2, 'hello']         |
```
erikvansebille: Should this not be `Context` instead of `Context value`?

erikvansebille (commented):
Nice addition! See below some comments on the layout of the table output

tests/test_fieldset.py
VeckoTheGecko: Yeah, this is which "ModelData" the (vector)field comes from. I thought to avoid ModelData as a name to avoid confusion. I'm not sure how important this column is - perhaps this is exposing unneeded lowlevel info. Open to ideas!

tests/test_fieldset.py
VeckoTheGecko: At the moment it's ordered by the "Dataset origin" column. Can do if we drop that column https://github.com/Parcels-code/Parcels/pull/2729#discussion_r3528973553

tests/test_fieldset.py
VeckoTheGecko: Yes, we can go for that

tests/test_fieldset.py
VeckoTheGecko: The mesh will be the same for the whole fieldset. I can print it at the end?

erikvansebille: Yep, good idea to print it at the end. And then perhaps also the `time_interval`?

tests/test_fieldset.py
VeckoTheGecko: Done now

tests/test_fieldset.py
VeckoTheGecko: Done now

VeckoTheGecko: Done now

tests/test_fieldset.py
VeckoTheGecko: Would be great to know what you think about this @erikvansebille . Perhaps we can rename from `Dataset origin` to `Grid number`? They're very much related and maybe that's clearer for users

erikvansebille: Yes, I think Grid number is better

VeckoTheGecko: Repr looks like this now:

```
| Name           | Type        | Grid number   | Interp method / value   |
|:---------------|:------------|:--------------|:------------------------|
| my_list        | Context     | -             | [1, 2, 'hello']         |
| my_value       | Context     | -             | 2.0                     |
| U              | Field       | 0             | XLinear(...)            |
| V              | Field       | 0             | XLinear(...)            |
| UV             | VectorField | 0             | XLinear_Velocity(...)   |
| U_wind         | Field       | 1             | XLinear(...)            |
| V_wind         | Field       | 1             | XLinear(...)            |
| UV_wind        | VectorField | 1             | XLinear_Velocity(...)   |
| constant_field | Field       | 2             | XConstantField(...)     |

mesh: flat
time interval: (np.datetime64('2000-01-01T00:00:00.000000000'), np.datetime64('2001-01-01T00:00:00.000000000'))
```
I think its fine like this? I like the ordering of the grid numbers since that's also the logical separation of the underlying data.

We can move the `Context` values to the bottom, but its a bit non-trivial in the Pandas API and think that can come in a future PR if we want it.

VeckoTheGecko:
CI is failing due to test pollution occuring between `test_fieldKh_Brownian` and this new test. Basically at the moment `add_constant_field` uses models which are global state, which results in subsequent fieldset constructions re-using this model.

This can be fixed in a separate PR - opening an issue to track.

erikvansebille (approved):

tests/test_fieldset.py
erikvansebille: Looks good like this!

Issue #2702, opened 2026-06-24
Updating parcels website and documentation for v4 preview release


erikvansebille:
This is a Tracking Issue to list all the changes that need to be made on the parcels-code.org and docs.parcels-code.org pages before we can release v4-preview.

- [x] Update banner on docs.parcels-code.org (#2835)
- [x] Publish a release blog post for https://parcels-code.org/blog (https://github.com/Parcels-code/parcels-code.org/pull/291)
- [x] Clean up the migration guide, and put in a nicer format (https://github.com/Parcels-code/Parcels/pull/2742)
- [x] Write a page about Grids (https://github.com/Parcels-code/Parcels/pull/2786)
- [x] Remove the example_script section (#2749)
- [x] Remove the explicit `ds.load()` calls in the examples, once raw zarr and/or WindowedArray are implemented (#2748)
- [x] Remove the unstable API warning in `__init__` (#2835)
- [x] Update conda feedstock recipe

VeckoTheGecko:
there are a bunch of unchecked items in this issue. re-opening for now

erikvansebille:
v4 is released, so closing this tracking issue!

PR #2737, opened 2026-07-08
[DOCS] Cleaning up the User Guide

erikvansebille:
### Description

This PR cleans up the user guide on the docs.parcels-code.org page. in particular, it moves the Getting Started section into the User Guide itself, so that everything is in one place. It also adds some clarifying wording, removes the (duplicate) installation-instructions on the v4 page and adds anchor links to the headings on the User Guide

### Checklist

- [x] Supports #2702 

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Creating custom css for adding anchors to the headers that don't show up in the side-bar


docs/_static/custom.css
VeckoTheGecko, line 1: Could you elaborate on what this does? Which anchor tags were you having problems with? I think it would be better to work within the pydata-sphinx theme we're using rather than adding custom CSS (as its more difficult to maintain)

VeckoTheGecko (commented):
Changes look good. Just a question about the custom css

docs/_static/custom.css
erikvansebille: Yes, I added the custom CSS because I wanted both headers on the main page (that could be anchored so we could directly link to a section) as well as a table-of-content in the left side-bar. See screenshot
Screenshot 2026-07-09 at 12 43 03


Without this custom css, the headers would appear twice in the main body, see orange circles in the screenshot below
Screenshot 2026-07-09 at 12 52 36

I asked an LLM how to do this and after a lot of back-and-forth this was the solution it came up with. If you have a better idea, then I'd be curious to hear!

VeckoTheGecko: Yep - just remove the captions

```diff
diff --git a/docs/user_guide/index.md b/docs/user_guide/index.md
index 6f2138de1..0d7d83429 100644
--- a/docs/user_guide/index.md
+++ b/docs/user_guide/index.md
@@ -9,7 +9,6 @@ For a description of the specific classes and functions, check out the [API refe
 ## Installation
 
 ```{toctree}
-:caption: Installation instructions
 :name: installation
 :titlesonly:
 getting_started/installation.md
@@ -22,7 +21,6 @@ If you have code that uses Parcels v3, you can migrate to Parcels v4 using [this
 ## Getting started
 
 ```{toctree}
-:caption: Getting Started
 :name: getting-started
 :titlesonly:
 getting_started/tutorial_quickstart.md
```

image

VeckoTheGecko: I misunderstood your original message here. I don't know if there's a way to do this without custom css

I'll leave it up to you, but just wanted to flag dabbling in custom css does add an additional maintenance burden

PR #2738, opened 2026-07-09
FEAT: Update FieldSet.describe() to print output and update docs

VeckoTheGecko:
### Description

Turns out that Jupyter notebooks format the output of `__repr__`'s differently from other functions (really, I should have expected that).

This PR updates the describe function to print to terminal by default instead (this implementation is the same as what `xr.Dataset.info()` does.

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

None used


erikvansebille (approved):
Looks good, implemented these also in the notebooks. Happy to be merged!

PR #2739, opened 2026-07-09
Add support for defining planetary radius

### LINKED ISSUE #2731: Make planetary radius configurable

v4

I'm a planetary scientist and have been using Parcels (v3.1.4) for some simulations of trajectories in the Venus atmosphere. Venus has a very similar radius to Earth, so degrees are a similar size in meters.

However, I'd like to use the package for Mars and potentially Jupiter in the future. The hardcoded degree to meter conversions (1852 * 60) will be way off for these planets. 

Make radius configurable by adding it as an attribute to the grid (defaulting to Earth radius). Then calculate deg2m = radius * PI/180. Replace instances of 1852 * 60 with the new deg2m constant.

_No response_

### Additional context

I'm happy to fork/implement this myself and try it in v4. I could allow deg2rad to revert to 1852 * 60 whenever Earth's radius or no radius is specified so it's fully back compatible.

maureenjcohen:
### Checklist

- [x] Closes #2731 
- [x] Tests added (5 in test_mesh.py)

### AI Disclosure

I used Claude to explore the repo, find the locations with hardcoded 1852 * 60 conversion factor, and suggest code snippets. I wrote the code manually (even if suggested by Claude) and made the design choices. I ran the full test suite with and without a custom radius, plus a short Mars simulation with a Mars radius.

Note: I decided to unpack the SphericalMesh object into a string to maintain the mesh == "flat"/"spherical" string comparisons throughout. This minimizes the impact of the change on the existing code. Another option would be for all mesh types to be objects with a string/flat attribute that could be checked instead, which might be useful for adding further mesh-specific properties.


src/parcels/_core/mesh.py
```diff
@@ -0,0 +1,28 @@
...
+    """Spherical mesh object with configurable planetary radius.
+
+    Pass to FieldSet object as ``mesh=SphericalMesh(radius=...)``.
+    radius is in meters; None reverts degree to meter conversion
+    to 1852 * 60 .
+    """
+
+    def __init__(self, radius: float | None = None):
+        if radius is not None and not isinstance(radius, (int, float, np.number)):
+            raise TypeError(f"radius must be a number of None, got {type(radius).__name__}")
```
erikvansebille:
```suggestion
            raise TypeError(f"radius must be a number or None, got {type(radius).__name__}")
```

src/parcels/_core/mesh.py
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+
+
+class SphericalMesh:
+    """Spherical mesh object with configurable planetary radius.
+
+    Pass to FieldSet object as ``mesh=SphericalMesh(radius=...)``.
+    radius is in meters; None reverts degree to meter conversion
+    to 1852 * 60 .
```
erikvansebille:
```suggestion
    Pass to FieldSet object as ``mesh=SphericalMesh(radius=...)``.
    radius is in meters; None reverts to default for Earth, where 
    arcdegree to meter conversion is defined as 1852 * 60
    (1852 meters per arcminute * 60 arcminutes per arcdegree).
```

src/parcels/_core/uxgrid.py
```diff
@@ -73,6 +79,14 @@ def get_axis_dim(self, axis: _UXGRID_AXES) -> int:
         elif axis == "FACE":
             return self.uxgrid.n_face
 
+    @property
+    def deg2m(self) -> float:
+        """Metres per degree of arc for this grid's mesh."""
```
erikvansebille:
```suggestion
        """Metres per arcdegree for this grid's mesh."""
```

src/parcels/_core/uxgrid.py
```diff
@@ -73,6 +79,14 @@ def get_axis_dim(self, axis: _UXGRID_AXES) -> int:
...
+        if self._radius is None:
+            return 1852 * 60.0
+        else:
+            return self._radius * np.pi / 180.0
```
erikvansebille: This may be a bit overcomplicated. Why not simply use Earth radius as default (`6366707.019493707`?), so that its `deg2m` ends up being 1852 * 60? (same for xgrid below)

erikvansebille: An advantage would then also be that a `radius is None` could indicate a flat Mesh, so that we don't need to check for the string `mesh == flat` but instead `radius is None`? But that may be another PR?

src/parcels/_typing.py
```diff
@@ -73,4 +75,6 @@ def _validate_against_pure_literal(value, typing_literal):
 
 # Assertion functions to clean user input
 def assert_valid_mesh(value: Any):
+    if isinstance(value, SphericalMesh):
+        return
```
erikvansebille: @VeckoTheGecko, can you check if this is correct behaviour for typing?

tests/test_mesh.py
erikvansebille, line 1: Very nice tests!

erikvansebille (commented):
Great work, @maureenjcohen! Just a few comments and suggestions below

We should also make sure to highlight this new feature in the documentation; perhaps in the Grids tutorial that @michaeldenes is working on? Can you do that, @michaeldenes?

src/parcels/_core/uxgrid.py
maureenjcohen: I wasn't sure if the exact 1852 * 60 was needed, or if using Earth radius might lead to small inconsistencies when switching between v3 and v4. I think your suggestion is good.

maureenjcohen: But if the user doesn't specify a radius, should radius default to None (flat mesh) or to Earth's? Most people are working on Earth and it might be more convenient for users not to have to put in a radius every time.

erikvansebille: I think we should keep the default as it is now (spherical Earth), but the idea is that the correct mesh is discovered when creating the FieldSet, see below
https://github.com/Parcels-code/Parcels/blob/a6bb153a031872d12f9b47738f400e621c1759e5/src/parcels/_core/model.py#L320

maureenjcohen: I've updated SphericalMesh to have Earth's radius as default. However, I was getting crashes from the advection tests for flat mesh cases because the advection still calls grid.deg2m, which has radius = None for flat meshes. I kept an if/else for xgrid and uxgrid which sets deg2m = 1.0 if there's no radius (i.e. the mesh is flat).

src/parcels/_core/xgrid.py
```diff
@@ -169,7 +170,12 @@ def __init__(self, model_data: xr.Dataset, mesh):
         self._ds = model_data
         grid = XgcmLikeGrid(self.sgrid_metadata, model_data)
         self.xgcm_grid = grid
-        self._mesh = mesh
+        if isinstance(mesh, SphericalMesh):
+            self._mesh = "spherical"
+            self._radius = mesh.radius
```
erikvansebille: PR looks good to me! The only question I have for @VeckoTheGecko is if this combination of `grid._mesh` as a string and the `SphericalMesh` class is not too obfuscating/confusing. Would it be worth to streamline this and e.g. use the class throughout? Or shall we do that in a next PR?

erikvansebille (approved):

VeckoTheGecko:
I've gone and done some refactorings in 93a364b which I think make things more readable and address the review comments.

src/parcels/_core/uxgrid.py
```diff
@@ -82,9 +76,9 @@ def get_axis_dim(self, axis: _UXGRID_AXES) -> int:
     @property
     def deg2m(self) -> float:
         """Metres per arcdegree for this grid's mesh."""
-        if self._radius is None:  # flat mesh; None causes crash in advection
+        if not self._mesh.is_spherical():
             return 1.0
-        return self._radius * np.pi / 180.0
+        return self._mesh.deg2m
```
erikvansebille: Would this not be easier to parse in the reverse? This avoids the negation (`not`)

```python
if self._mesh.is_spherical():
    return self._mesh.deg2m
else:
    return 1.0
```

src/parcels/_core/xgrid.py
```diff
@@ -258,9 +252,9 @@ def time(self):
     @property
     def deg2m(self) -> float:
         """Metres per degree of arc for this grid's mesh."""
-        if self._radius is None:  # flat mesh; None causes crash in advection
+        if not self._mesh.is_spherical():
             return 1.0
-        return self._radius * np.pi / 180.0
+        return self._mesh.deg2m
```
erikvansebille: Same comment as above about avoiding the negation

erikvansebille (commented):
Thanks for the refactoring, @VeckoTheGecko. Two comments below, looks good otherwise. Happy for you to merge afterwards

src/parcels/_core/uxgrid.py
VeckoTheGecko: I don't mind, I find `if not spherical` and `if spherical` to be equally readable. Would you like me to make the change?

erikvansebille: Yes, please make the change. I'd prefer to avoid if-statements with "not" where we can

VeckoTheGecko: Done in c82a69ea66467c63d56bfec00ecfb1e3c7ecc5c1

PR #2743, opened 2026-07-10
fix: correct misplaced parentheses in _Spatialslip zeta corrections

steps-re:
### What

In the 3D branch of `_Spatialslip` (`src/parcels/interpolators/_xinterpolators.py`), the four vertical (`zeta`) slip corrections have a misplaced closing parenthesis:

```python
f_u = np.where(
    is_land(0, 0, 0, 0) & is_land(0, 0, 0, 1) & is_land(0, 0, 1, 0 & is_land(0, 0, 1, 1) & (zeta > 0)),
    f_u * (a + b * zeta) / zeta,
    f_u,
)
```

The `)` after `is_land(0, 0, 1, 0` is missing, so `0 & is_land(0, 0, 1, 1) & (zeta > 0)` is passed as `is_land`'s 4th positional argument (the x-index) instead of `0`. The `& is_land(...) & (zeta > 0)` mask that should gate `np.where` is swallowed into that argument, so both the land test and the `zeta` position gate are wrong. This breaks the vertical free-/partial-slip boundary correction for 3D C-grid velocities.

### Fix

Move the parenthesis so each line matches the analogous, correctly-formed `eta`/`xsi` corrections in the same function:

```python
is_land(0, 0, 0, 0) & is_land(0, 0, 0, 1) & is_land(0, 0, 1, 0) & is_land(0, 0, 1, 1) & (zeta > 0)
```

Same correction applied to all four affected lines (the `zeta > 0` and `zeta < 1` gates for `f_u` and `f_v`).

### Testing

`pytest tests/test_interpolation.py tests/test_convert.py` passes (29 passed, 1 xfailed) with the change. Note the `cgrid_velocity` case in `test_interpolation.py` is currently commented out, which is likely why this went unnoticed.


VeckoTheGecko:
Hi @steps-re , thanks for the PR! I think @erikvansebille will be able to provide a review on this.

In future could you please make sure to follow our PR template? That makes things easier on our end :)

erikvansebille (approved):
Yep, this change is good; thanks for catching this, @steps-re!

PR #2744, opened 2026-07-10
Spatialhash regional spherical improvements

fluidnumericsJoe:
### Description
This PR changes the spatialhash initialization for `spherical` mesh types. Previously the bounding box was set to the unit cube for spherical mesh types. However, spherical mesh types might be desirable for regional simulations (e.g. for improved accuracy in barycentric coordinates, dealing with regions close to the poles, or for leveraging built in unit conversions in Parcels velocity field interpolators).

The main problem with the unit cube for regional configurations is that the hash grid resolution is often far too coarse for a regional domain. This results in a high number of mesh cells being aligned with a single morton key. Calling `spatialhash.query()` then results in an abnormally high number of particle in cell checks.

This is remedied here by choosing the extents of the hash grid as the bounding cube (x,y,z) of the regional domain. 
### Checklist

- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Claude Code was used to identify and implement documentation changes.


fluidnumericsJoe:
I still need to test this out on the example you shared @erikvansebille, but wanted to get this in and make sure that existing tests all pass

fluidnumericsJoe:
looks like there's a few tests to clean up here.

erikvansebille:
Thanks for this PR, @fluidnumericsJoe! Seems like there are some segmentation errors in the tests. Do you want to try fix them; or can I help?

fluidnumericsJoe:
> Thanks for this PR, @fluidnumericsJoe! Seems like there are some segmentation errors in the tests. Do you want to try fix them; or can I help?

I'll get to it today

fluidnumericsJoe:
So, I've narrowed down what's going on here. To recap what we primarily changed here - we've changed the spatialhash initialization to use a more narrow bounding box, based on the actual x,y,z extents for spherical mesh types. effectively, for regional domains this makes the hash grid size smaller. This happens because the number of hash grid cells is fixed, but we're now using a smaller domain extent for the hash grid.

The upshot of this change is that we have fewer faces in the "parent" mesh that we need to check, when a particle is found in a given hash cell. This reduces the number of calls to the particle in-cell checks. 

The issue here is in the construction of the hash table that maps a hash cell index (the morton code) to a list of faces. During construction we have to use the bounding box of a face in the parent grid to find the range of hash cells that overlap. This information is used to pre-allocate a lookup table that maps from parent grid face to hash grid cells; the hash table is contructed as the inverse of that table.  If the hash cells are significantly smaller than the parent model face sizes, we end up with a fairly large memory footprint (high number of hash cells per face).

I'm looking at tidying up that initializer here to reduce the memory footprint..

fluidnumericsJoe:
hashgrid_overview
This figure shows the antimeridian spanning grid that the CI is failing on with OOM errors. The grid locations are plotted in cartesian (x,y,z) space after mapping the latitude, longitude through a spherical coordinate transformation. A box shows a region that we'll zoom in on in the next figure


hashgrid_zoom
This zoomed in region shows the hash grid (light gray lines) and the parent grid edges (blue lines). The bouding cube for the cell shaded light blue has 55x59x46 hash cells. Summing the number of hash cells per face across all faces (to get the size of the work arrays for constructing the hash table), each work array is roughly 193 million entries (too big).

In addition to tidying up in the initializer, I'll probably look at putting in a lever to manipulate the hash cell size - with the morton cells, this is controlled by the bit width . A higher bitwidth gives more hash cells per direction. At the moment, we default to 1024.

Another thing we could possibly do is be a bit smarter about tossing out hash cells that are clearly outside of elements (bounding box overlaps are kind of wasteful). In both plane views in the zoomed in plot, there's a large number of hash cells that are completely outside the element.

fluidnumericsJoe:
I pushed up a couple of commits created with claude that focus on memory footprint reduction in the `spatialhash._initialize_hash_table` method. 

The [first commit ](https://github.com/Parcels-code/Parcels/pull/2744/commits/34af90324c24c157d6462748c14d2bf814d7d706)removes a conditional path that is never traversed; that branch would only be executed if a face is found that doesn't overlap a single hash cell. Since the hash grid bounds are the bounding box of the whole grid, it's safe to assume each face will overlap at least 1 hash cell. Relying on this assumption allows us to trim back on a number of intermediate arrays.

The [second commit](https://github.com/Parcels-code/Parcels/pull/2744/commits/35dd7d9cec0c93cb4d596ea274deb2f732b511cd) clears previously used intermediate arrays from memory after all the morton codes are calculated. Then since an entry in the hash table lookup consists of the morton code with a corresponding face id and each is a `uint32`, we can store the combination is a single `uint64`. This allows us to use a single sort call, rather than an argsort followed by a sort (as we previously did). This single sort gives us a list of integers sorted by the morton code first then face id.

Using the `fesom2_square_delaunay_uniform_z_coordinate ` example dataset, I tracked the max memory usage of a benchmark that just loads the dataset and constructs the spatialhash object.

| | main | https://github.com/Parcels-code/Parcels/pull/2744/commits/34af90324c24c157d6462748c14d2bf814d7d706 | https://github.com/Parcels-code/Parcels/pull/2744/commits/35dd7d9cec0c93cb4d596ea274deb2f732b511cd |
|---|---|---|---|
| peak build RSS (101M-entry UxGrid) | 17.2 GB | 11.3 GB | **6.35 GB** |

Even with these improvements, I still think we might need to provide controls to coarsen the hash grid.

erikvansebille:
Thanks for this further digging, @fluidnumericsJoe! I must say I don't entirely understand what's going on, but taking a step back I think this is the problem:
1. We had a working morton/hashing procedure for full-sphere (global) grids before this PR
2. For grids with small, regional domains, this procedure is suboptimal because many of the global hash cells (those outside the grid) remain empty - and other hash cells have lots of entries
3. The obvious solution is to bound the hashtable by the grid extent; which is what this PR implements

    _(and the next step is where I'm unclear)_
4. However, this creates problems with grids that cross the antemeridian (right?)

    _(If so, is the following approach suitable?)_
6. We keep the original full-sphere/global hashtable for any grid spanning the antemeridian - and use a bounding box hashtable for regional grids
7. If users do want to improve their hash performance on regional grids across the antemeridian, they could(?) do a coordinate transformation from longitude [180, 180] to [0, 360]. Since these are per definition regional grids, the grid would then not span the antemeridian anymore?
8. The only remaining issue is polar caps (so spanning all longitudes but only a limited range of latitudes). For that, this wouldn't be a solution

fluidnumericsJoe:
> Thanks for this further digging, @fluidnumericsJoe! I must say I don't entirely understand what's going on, but taking a step back I think this is the problem:
> 1. We had a working morton/hashing procedure for full-sphere (global) grids before this PR

It was working. Reducing the extent of the hash grid ballooned the number of hash cells per face. This happens because reducing the hash grid extent while keeping the number of hash cells the same causes the hash cells to become smaller. With smaller hash cells, more of them are withing each parent mesh face.  The number of hash cells per face determines the size of work arrays during hash table construction. For some grids, this led to oom errors during construction. Now the game is to reduce the memory footprint of the hash table construction.

> 2. For grids with small, regional domains, this procedure is suboptimal because many of the global hash cells (those outside the grid) remain empty - and other hash cells have lots of entries

Ultimately, once the hash table is constructed, since we use a compressed sparse row matrix format, hash entries that have no face overlaps don't get stored in the lookup table.

> 3. The obvious solution is to bound the hashtable by the grid extent; which is what this PR implements

Yes. This optimizes query, but balloons the memory footprint for the hash table creation

> 
>     _(and the next step is where I'm unclear)_
> 4. However, this creates problems with grids that cross the antemeridian (right?)
It just so happened that I caught it on the anti meridian grid. Nothing about the anti meridian crossing was part of the problem. The faces are quite coarse resolution relative to the hash cells after reducing the hash grid extent. Again,the number of hash cells per face became quite large causing work arrays used in hash table construction to balloon in size. I found other grids where the memory footprint went up as well during construction, but just happened to be below the allotted memory in the GitHub runners.

> 
>     _(If so, is the following approach suitable?)_
> 6. We keep the original full-sphere/global hashtable for any grid spanning the antemeridian - and use a bounding box hashtable for regional grids
No. That's not necessary. I've reduced the memory footprint of the hash table creation by removing unnecessary work arrays. This also speeds up hash table construction. The other route is to provide an option to coarsen the hash grid; this could be implemented by using a lower bit width. 

> 7. If users do want to improve their hash performance on regional grids across the antemeridian, they could(?) do a coordinate transformation from longitude [180, 180] to [0, 360]. Since these are per definition regional grids, the grid would then not span the antemeridian anymore?

Antimeridian is not the issue. It's a ratio of hash grid cell size to parent mesh face/cell size.

> 8. The only remaining issue is polar caps (so spanning all longitudes but only a limited range of latitudes). For that, this wouldn't be a solution

This is not an issue..for polar caps we have to use spherical mesh types. Flat mesh types would end up with colinear cells . The changes made in this pr will improve construction time of the hash grid for these cases and help reduce the number of particle in cell checks.
> 
>

fluidnumericsJoe:
I put some effort into having the hash grid cell spacing change according to a budget number of hash cells per parent model face. Rather than passing the choice to users to set the `bitwidth` option (which admittedly might be off-putting for users) the `init` method now sets the bitwidth to ensure we have no more than 16 hash cells per face. This value was chosen because it still results in a low number of particle in cell checks during query, but cuts way down on memory requirements during construction

Here's the full progression on the fesom2_square_delaunay_uniform_z_coordinate spherical benchmark (721 faces), each commit measured in its own process:

| commit | change | build time | peak RSS (build) | entries | stored table |
|---|---|---|---|---|---|
| ac22b349 | tight bounds (baseline) | 39.9 s | 17.22 GB | 101.6M | 2,827 MB |
| 34af9032 | trim 1: dead code + no scatter | 32.5 s | 11.27 GB | 101.6M | 2,827 MB |
| 35dd7d9c | trim 2: packed in-place sort | 31.7 s | 6.35 GB | 101.6M | 2,827 MB |
| badcf7d3 | trim 3: diff-based CSR | 28.3 s | 6.35 GB | 101.6M | 2,827 MB |
| 9a9b4f2f | trim 4: flat face id | 25.7 s | 6.35 GB | 101.6M | 1,607 MB |
| e54fca50 | entry-budget cap | **0.81 s** | **0.26 GB** | **4.18M** | **64 MB** |

That last change, with the adjustable hash cell size, makes a huge difference in peak memory consumption and runtime of the hash table construction.

In terms of the number of particle in cell checks, below are some stats on the number of faces per hash table key; this is directly related to the number of particle in cell checks that are run for finding particles.

| mesh | bitwidth | keys | mean | median | p99 | max | 1 face | 2 | 3 | 4 | ≥5 |
|---|---|---|---|---|---|---|---|---|---|---|---|
| UxGrid uniform_z | 340 | 2.36M | 1.77 | 2 | 4 | 8 | 35.8% | 53.8% | 8.0% | 2.3% | 0.1% |
| UxGrid antimeridian | 273 | 2.28M | 1.84 | 2 | 4 | 8 | 30.4% | 58.5% | 7.9% | 3.1% | 0.1% |
| XGrid 2d_left_rotated | 551 | 3.46M | 1.21 | 1 | 2 | 4 | 80.0% | 19.1% | 0.8% | 0.0% | — |

fluidnumericsJoe:
@erikvansebille - I'd be interested to know how this branch works on the example you shared. I won't be able to give that a go until monday, but if you have the time to try it out before me, I'd be interested to know what your experience is like

erikvansebille:
> @erikvansebille - I'd be interested to know how this branch works on the example you shared. I won't be able to give that a go until monday, but if you have the time to try it out before me, I'd be interested to know what your experience is like

Yep, it works like a charm on https://github.com/Parcels-code/Drifter_campaign_prediction/blob/main/Matroos/run_dcsm7.py

A simulation with the old code (coercing to a flat mesh) takes approximately 6 minutes, while running with this new PR and a spherical mesh takes only 3 minutes. So a nice speed-up!

src/parcels/_core/spatialhash.py
```diff
@@ -11,6 +11,17 @@
...
+# Budget on the total number of (face, hash cell) pairs in the hash table:
+# max(_HASH_ENTRIES_PER_FACE * nfaces, _HASH_ENTRY_BUDGET_MIN).
+# When the hash grid cell size, set by the bitwidth, would result in too many
+# hash entries per face, the hash grid is coarsened by lowering the bitwidth
+# The target value for the number of hash entries per face is chosen to keep
+# the number of particle in cell checks as small as possible, while also
+# minimizing the hash table construction memory footprint.
+_HASH_ENTRIES_PER_FACE = 16
+_HASH_ENTRY_BUDGET_MIN = 2**22
+_HASH_MAX_BITWIDTH = 1023
```
erikvansebille, line 23: So how would users then change these values? By hardcoding? They are not optional parameters of the functions, are they?

src/parcels/_core/spatialhash.py
```diff
@@ -41,21 +51,23 @@ def __init__(
             raise ValueError("Expected `grid` to be a parcels.XGrid or parcels.UxGrid")
 
         self._source_grid = grid
-        self._bitwidth = bitwidth  # Max integer to use per coordinate in quantization (10 bits = 0..1023)
+        self._bitwidth = _HASH_MAX_BITWIDTH  # Max integer to use per coordinate in quantization (10 bits = 0..1023)
```
erikvansebille, line 54: Would it then make more sense to define the number of bits needed (so `10` instead of `1023`)?

erikvansebille (approved):
Looks very good; and as I mentioned earlier it is also very performant! A few small questions below

src/parcels/_core/spatialhash.py
fluidnumericsJoe: TBH, users shouldn't change these values. I didn't want to have arbitrary numbers showing up in a few spots further down in the spatialhash module - figured it'd be best to define these kinds of module-scoped constants in one spot here for developers. The main thing this controls is the auto-resizing of the hash grid so that users don't have to think about this.

src/parcels/_core/spatialhash.py
fluidnumericsJoe: Not really. If we change this to the number of bits, then further down when we recast any of `x`, `y`, or `z` we'd have to calculate the range of an integer that fits entirely within the number of bits, e.g. convert `10` to `1023`. This value just starts the bisection method that is used to calculate the "optimal" bitwidth while keeping the number of hash cells per face below the `_HASH_ENTRIES_PER_FACE`

PR #2763, opened 2026-07-22
Moving relevant v3 tutorials to v4 examples folder

erikvansebille:
### Description

This PR updates the old tutorials in examples-v3 folder; removing some that are not relevant anymore (e.g. on zarr output) and updating the others to v4 API

Note that, with this PR, we'll likely use more than the 15 minutes of ReadTheDocs build time on their free plan, so we need to think about how to build the documentation...

### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

None


docs/user_guide/index.md
```diff
@@ -121,12 +103,14 @@ examples/tutorial_gsw_density.ipynb
 examples/tutorial_Argofloats.ipynb
 examples/tutorial_diffusion.ipynb
 examples/tutorial_interaction.ipynb
+examples/documentation_homepage_animation.md
```
VeckoTheGecko: Wouldn't this be better in the "Other" section below?

VeckoTheGecko:
Should we make a distinction anymore between `documentation_...ipynb` and `tutorial_...ipynb`? From what I understand we intend to run them all in CI?

In a future PR/at the end of this one maybe we should do renamings

erikvansebille:
> Should we make a distinction anymore between `documentation_...ipynb` and `tutorial_...ipynb`? From what I understand we intend to run them all in CI?
> 
> In a future PR maybe we should do renamings

Good point; the distinction between `documentation_...ipynb` and `tutorial_...ipynb` is a relic from v3, when only the tutorial notebooks were run in CI. But now indeed we will run all of them. 

The only 'notebook' that shouldn't run on CI is the homepage_animation one - because it takes approximately 30 minutes just to compute the full set of trajectories. But that's why I now made it a markdown file with code cells that aren't triggered on CI

Why shall I not do the renaming to tutorials-* in this PR?

VeckoTheGecko:
> Why shall I not do the renaming to tutorials-* in this PR?

Doing so in a separate PR or at the end of this one would be good. I'm mid-review at the moment, so doing the renaming right now would mean that I have to resubmit my other review comments.

src/parcels/_datasets/structured/generated.py
```diff
@@ -333,23 +333,23 @@ def stommel_gyre_dataset(xdim=200, ydim=200, grid_type="A"):
                 U[j, i] = -(1 - math.exp(-xi / es) - xi) * math.pi**2 * np.cos(math.pi * yi) * scalefac
                 V[j, i] = (math.exp(-xi / es) / es - 1) * math.pi * np.sin(math.pi * yi) * scalefac
     if grid_type == "C":
-        V[:, 1:] = (P[:, 1:] - P[:, 0:-1]) / dx * a
         U[1:, :] = -(P[1:, :] - P[0:-1, :]) / dy * b
-        Udims = ["YC", "XG"]
-        Vdims = ["YG", "XC"]
+        V[:, 1:] = (P[:, 1:] - P[:, 0:-1]) / dx * a
+        Udims = ["YG", "XC"]
+        Vdims = ["YC", "XG"]
```
VeckoTheGecko, line 339: I think that this change is the wrong way around and should be `U=["YC", "XG"]`?

- YC -> Y-axis centers 
- YG -> Y-axis edges
- XC -> X-axis centers 
- XG -> X-axis edges

I don't think its urgent to fix in this PR - perhaps just add a TODO. I'm wondering what the error was that led to this change?

VeckoTheGecko (commented):
a couple clarifications - happy to merge after those are discussed

docs/user_guide/index.md
erikvansebille: OK, fixed in c4797750d8efa45cb848e686a22c8b82b6182610

erikvansebille:
> > Why shall I not do the renaming to tutorials-* in this PR?
> 
> Doing so in a separate PR or at the end of this one would be good. I'm mid-review at the moment, so doing the renaming right now would mean that I have to resubmit my other review comments.

Now done in 3dfb6f59a5919b3c49027faf3add9569c2e53c34

src/parcels/_datasets/structured/generated.py
erikvansebille: No, the U-velocities are defined on the center of the right y-edges (which are located at XC). See also the plot below
image

Furthermore, with this definition the results agree with the v3 version. I updated this function when moving the tutorial_peninsula_AvsCgrid over to v4.

PR #2766, opened 2026-07-23
Removing old v3 unit tests

erikvansebille:
### Description

This PR moves some of the last relevant tests from v3 over to the new v4 test suite, and removes the other irrelevant or already-copied tests. It also updates the interpolation_regression test

The only file that remains is [tests-v3/test_reprs.py](https://github.com/erikvansebille/Parcels/blob/removing-v3-tests/tests-v3/test_reprs.py); @VeckoTheGecko, can you confirm if this test file can also go? 

### Checklist

- [x] Tests added

### AI Disclosure

None


tests-v3/tools/test_helpers.py
VeckoTheGecko, line 1: I think this test file is worth keeping.

- `_format_list_items_multiline` exists in `src/parcels/_reprs.py`
- `deprecated` and `deprecated_made_private` exist in `src/parcels/_decorators.py`

Perhaps two separate files `test_decorators.py` and `test_repr_utils.py` (and maybe a rename of `src/parcels/_reprs.py` to `src/parcels/_repr_utils.py`)

tests/test_kernel.py
```diff
@@ -145,3 +162,81 @@ def kernel_with_forced_kwarg(particles, *, fieldset=0):
...
+        add_lon = 0.1
+        particles.dx += add_lon
+
+    def Kernel2(particles, fieldset):  # pragma: no cover
+        add_lon = -0.3
+        particles.dx += add_lon
+
+    pset = ParticleSet(fieldset, x=[0.5], y=[0.5])
+    pset.execute([Kernel1, Kernel2], runtime=1.0, dt=1.0)
+    np.testing.assert_allclose(pset.x, 0.3, rtol=1e-5)
```
VeckoTheGecko: I'm not quite sure what this is testing - we don't have the code-generator anymore nor merge kernels internally (this test was introduced in cd6feea068b494e0b2a911ea8ba5d542c1178d18 ).

Perhaps just remove?

VeckoTheGecko (commented):
some prelim feedback

tests/test_kernel.py
erikvansebille: Removed in 08eb89af518a821681023c98ce1d6cf7b8d195d6

tests-v3/tools/test_helpers.py
erikvansebille: Tests moved back in de73712bf03a642ff0c2962a6fd72f316b1f3611 and renamed `_reprs.py` to `_repr_utils.py` in 356d83f9139000b81df637131b8b3fbfe6e4632b

tests/test_interpolation.py
```diff
@@ -193,22 +194,8 @@ def test_interpolation_mesh_type(mesh, npart=10):
...
-
-
-@pytest.mark.xfail(reason="ParticleFile not implemented yet")
-@pytest.mark.parametrize(
-    "interp_name",
-    [
-        "linear",
-        # "freeslip",
-        # "nearest",
-        # "cgrid_velocity",
```
VeckoTheGecko, line 289: Mainly out of curiousity: Do we have these other interpolators in v4?(/are they tested?)

tests/test_particle.py
```diff
@@ -117,6 +117,13 @@ def test_particleclass_add_variable_in_loop():
         assert var1.to_write == var2.to_write
 
 
+def test_variable_special_names(fieldset):
+    """Test that checks errors thrown for special names."""
+    for vars in ["z", "x"]:
+        with pytest.raises(ValueError):
+            Particle.add_variable(Variable(vars, dtype=np.float32, initial=10.0))
+
+
```
VeckoTheGecko: This test should be removed.

Although it passes, its now testing something different than why it was introduced ( 2d15dd231789ed43ecebce8b577b3d92ae311eb9 ). Now its doubling up `test_particleclass_add_variable_collision`

VeckoTheGecko (changes requested):
Looks good for the most part. Some small edits (including getting CI working again - some of the changed tests raise Parcels warnings which are now interpretted as failures)

tests/test_particle.py
erikvansebille: Removed in a9baaa960b676d473aa316e3786ffd4a11d92d04

tests/test_interpolation.py
erikvansebille: Good catch! These indeed should also be tested. I worked on it this afternoon; but it's a bit more involved than I expected. Freeslip worked out of the box, but the other two gave errors, possibly to do with the issue that @willirath found in #2801 

I propose merging this PR now - and then I'll open a new issue to work on the fixes after my holidays

PR #2778, opened 2026-07-28
Spatial hash docs update

fluidnumericsJoe:
## Description
* Update unstructured grid search docs to reflect variable bitwidth in construction 
* Documentation on use of the gnomonic projection, rather than planar normal projection ( PIC check algorithm adjustment )
* Physical element - hash grid cell bounds for spherical mesh types


erikvansebille:
WHta's the status on this PR, @fluidnumericsJoe? Should we get this merged soon? It won't be in the official release, but would be good to get it on the website asap. 

But before we merge: is it useful also adding some info on #2883 in this documentation?

fluidnumericsJoe:
Thanks for the update here @wyatt-fluidnumerics - I'm going to run another pass through to make sure we're all nice and consistent with everything implemented between spatialhash and index search. On a first glance, I'm wanting to fill in a bit more detail on curvilinear coordinate mapping (bilinear interpolation). Seems we could also use local links to other sections in a few spots to make navigation easier

Issue #2693, opened 2026-06-23
Grid search (ei/igrid) not caching between successive grid search calls


fluidnumericsJoe:
There are reports of ei/igrid not caching between grid search's. See, e.g.  https://github.com/Parcels-code/Parcels/pull/2646#discussion_r3458637178

fluidnumericsJoe:
I'll need to look into this one further.

Cursory review from Claude
> The premise is wrong, and I verified this empirically back in June. particle.ei is cached across Field.eval calls and within pset.execute: 36 particles, RK4, 40 steps → 113 hash queries vs 648 if uncached (82.6% reduction). Reproducers are in verification/ei_cache_test.py and verification/ei_cache_execute_test.py.

> There is one genuine bug hiding in it, still present today: the batch gates at index_search.py:287 (if np.any(xi)) and uxgrid.py:117 (if np.any(ei)). These are batch-wide, so when every particle's cached guess is index 0 — all particles at init, or a lone particle genuinely in cell 0 — the cache is discarded and a full re-hash runs. Retitle #2693 to that narrow bug and drop the "not caching" framing, otherwise whoever picks it up will chase a phantom.

Issue #2669, opened 2026-06-16
Check that there are unit tests that check for correct particles._ei setting and updating


erikvansebille:
> Kind of related - are we sure that we're saving the ei's properly? I think that we should explicitly test that (+grid de-duplication). Is there an existing issue for that? 

 _Originally posted by @VeckoTheGecko in [#2144](https://github.com/Parcels-code/Parcels/issues/2144#issuecomment-4718149952)_

PR #2779, opened 2026-07-29
Grid explanations user guide

michaeldenes:
### Description
This PR includes a grid explanations user guide, to help users understand how their underlying model data is defined and interpreted.

### Checklist

- [x] Contributes to #2702 
- [ ] Tests added

### AI Disclosure


docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+Under the hood, every `Field` in a `FieldSet` has a `grid` attribute. This `grid` stores the spatial and temporal information of the Field coordinates. The number of Grids in a FieldSet is thus always smaller or equal to the number of Field objects; and this is what the "grid number" column in `FieldSet.describe()` refers to.
+
+## Structured grids
+
+A structured grid is composed of quadrilateral elements, that are indexed using logical 2D or 3D indices, like $(i,j,k)$. In `xarray` terminology, these indices correspond to `dimensions`, for example [...]. In structured grids, grid cell neighbours are easily found by decrementing or incrementing these dimensions.
```
erikvansebille: The [...] here needs to be replaced by something? What did you have in mind?

erikvansebille: The sentence "In structured grids, grid cell neighbours are easily found by decrementing or incrementing these dimensions" is true, but not relevant for Parcels v4. We don't search for neighbours anymore. Instead: 
 
- In grids with 1-dimensional coordinates (= rectilinear grids), we do a quick search of the coordinate to find the right value: 
https://github.com/Parcels-code/Parcels/blob/e19411f7f77b88e2f2ff42759db8a06a3e67b705/src/parcels/_core/index_search.py#L47-L51
- In grids with 2-dimensional coordinates (= curvilinear grids; but _could_ also be rectilinear), we use the hash-table approach that Joe documented extensively in https://docs.oceanparcels.org/en/main/development/unstructured_grid_search.html. Note that there is no fundamental difference between how we treat unstructured grids and structured grids with 2D coordinates. The only implementation difference is the point-in-cell check
- Before going into the hash-table query, I think(?) that we first check if the particle is stil in the old cell (through its element index `ei`). But whether this works is to be explored (see #2693 and #2669)

So I would rephrase this sentence to better highlight that we don't use neighbour search

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+There are two styles of structured grids, rectilinear and curvilinear, as shown in Figure 1. Rectilinear grids are nearly always aligned with the coordinate axes (that is, there is a one-to-one mapping between the dimensions and the physical coordinate space), making them simple and computationally efficient to query. However, they're limited by the fact that it is difficult to resolve complex coastlines without high resolution, and they suffer from singularities in the flow fields at the poles where the grid lines converge. Curvilinear grids allow for curved grid lines, which allow for better representation of coastlines. They often move the poles onto land to avoid singularities occuring in the ocean. However, they become trickier to work with, as the spatial positions of their vertices are stored in separate 2D arrays, and moving to an eastward neighbour is not as simple as incrementing the $i$-th coordinate by 1.
```
erikvansebille: Why 'nearly' always? In Parcels, I think they have to be always. The index_search works with either 1D coordinates or 2D coordinates

erikvansebille:
```suggestion
There are two styles of structured grids, rectilinear and curvilinear, as shown in Figure 1. 

1. Rectilinear grids are nearly always aligned with the coordinate axes (that is, there is a one-to-one mapping between the dimensions and the physical coordinate space), making them simple and computationally efficient to query. However, they're limited by the fact that it is difficult to resolve complex coastlines without high resolution, and they suffer from singularities in the flow fields at the poles where the grid lines converge. 

2. Curvilinear grids allow for curved grid lines, which allow for better representation of coastlines. They often move the poles onto land to avoid singularities occurring in the ocean. However, they become trickier to work with, as the spatial positions of their vertices (their ["coordinates" in xarray-parlance](https://docs.xarray.dev/en/latest/user-guide/terminology.html#term-Coordinate)) are stored in separate 2D coordinate arrays, and moving to an eastward neighbour is not as simple as incrementing the $i$-th dimension by 1.
```

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+![Figure 1 from parcels v2 paper](image.png)
```
erikvansebille: Add a caption to this figure?

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+TODO: For Joe.
+
+## How your data may be defined
+
+Regardless of your grid type, how your data is defined on your grid is equally important. Here, we will describe how you can interpret your data at a very general level, and the concept applies for any $n$-gon ($n$-sided polygon with $n \ge 3$).
+
+### Nodes, edges, and faces
+
+Let's assume we have a simple 2D rectlinear grid. The grid is composed of a number of nodes (or vertices), connected by edges, which together construct grid cells and grid cell faces. In Figure 2, we draw a simple grid cell. Here, your data may be defined on the corners of the grid cell (the nodes/vertices), at the centre of the cell face, or across a cell edge.
```
erikvansebille:
```suggestion
Let's assume we have a simple 2D rectilinear grid. The grid is composed of a number of nodes (or vertices), connected by edges, which together construct grid cells and grid cell faces. In Figure 2, we draw a simple grid cell. Here, your data may be defined on the corners of the grid cell (the nodes/vertices), at the centre of the cell face, or across a cell edge.
```

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+![Figure 2 - A simple grid cell. Blue circles denote nodes (or vertices) of the grid cell. The red circle denotes the centre of the cell face.](image-1.png)
```
erikvansebille, line 31: Why not provide the code to generate this (and following) plot too? That will for sure be useful if someone later wants to adapt the plot - and it also aligns with the rest of the tutorials/documentation by making the file as self-contained as possible. I think you can even collapse/hide the input cell if it's too long and distracting

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+You will need to make several assumptions about your data. Your data may represent a point-wise "sample" of some field. For example, velocity data may be defined at the nodes of your grid, and a fairly safe assumption to make is that you can bi-linearly interpolate these data points to your particle positions. In such an example, your velocity field may look like Figure 3.
```
erikvansebille: Why "fairly safe"? I don't understand what you mean. Under what conditions is it fairly safe to make this assumption?

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+![Figure 3 - Bi-linear interpolation of point-wise data at nodes](image-2.png)
+
+Alternatively, your data may represent an "average value" across a cell face. For example, your temperature and salinity data may be defined at the cell centre, and represent an average value for the entire grid cell. A fairly safe assumption to make is that you can nearest-neighbour interpolate these data points to your particle positions. In such a case, your temperature field may look like figure 4.
```
erikvansebille: Can we be a bit more specific than "may be defined" here? Again, under what conditions would this be typical? How could a user find out?

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+![Figure 4 - Nearest-neighbour interpolation of "grid cell averaged" data at cell centres](image-4.png)
+
+Lastly, your data may represent a value across a cell edge. For example, in (2D) Arakawa C-grid datasets, velocities are defined across an edge as they represent a "flux" across that cell edge. [for some reason an accepted assumption is] to perform a 1D linear interpolation of the zonal velocity in the zonal direction, and similarly a 1D linear interpolation of the meridional velocity in the meridional direction. In such a case, your velocity field may look like figure 5.
```
erikvansebille:
```suggestion
Lastly, your data may represent a value across a cell edge. For example, in (2D) Arakawa C-grid datasets, velocities are defined across an edge as they represent a "flux" across that cell edge. 

For structured grids, [Blanke and Raynaud](https://doi.org/10.1175/1520-0485(1997)027%3C1038:KOTPEU%3E2.0.CO;2) proposed in 1997 to perform a 1D linear interpolation of the zonal velocity in the zonal direction, and similarly a 1D linear interpolation of the meridional velocity in the meridional direction. In such a case, your velocity field may look like figure 5.

This (uni)linear velocity interpolation is now often referred to as the Analytical interpolation scheme, and is what the [Ariane](https://ariane-code.cnrs.fr) and [TRACMASS](https://www.tracmass.org/index.html) Lagrangian codes also use. In Parcels, the time-stepping version of this interpolation is provided in the `CGrid_Velocity` Interpolator function
```

docs/user_guide/examples/explanation_grids.md
```diff
@@ -1,10 +1,39 @@
...
+
+![Figure 5 - C-grid 1D interpolation of velocity data defined across a cell edge](image-5.png)
```
erikvansebille, line 178: Anything to say about vertical grids? Or even just put a placeholder?

erikvansebille (changes requested):
Thanks @michaeldenes. See below for some suggestions/questions and comments

docs/user_guide/examples/explanation_grids.md
michaeldenes: I think I've seen models with a rotated rectilinear grid (say, along the coast of the Netherlands), which is not aligned East/North, hence the careful choice of wording. Happy to change to "typically" rather than "nearly always".

docs/user_guide/examples/explanation_grids.md
michaeldenes: Again, being safe with word choices. I'll use "a typical assumption to make"

docs/user_guide/examples/explanation_grids.md
michaeldenes: This figure is hand drawn. If you'd like, I can turn it into a python script.

docs/user_guide/examples/explanation_grids.md
michaeldenes: Resolved

docs/user_guide/examples/explanation_grids.md
michaeldenes: Caption (and updated figure) now added.

michaeldenes:
Whacky stuff going on with commits. Have decided to redo this in another branch and PR, using an ipynb instead of markdown file.

PR #2781, opened 2026-07-29
Resolved bug in the selection of particle's default z value when it is not provided by the user

### LINKED ISSUE #2772: ParticleSet default initial z not set correctly (does not respect direction of depth axis)
### Parcels version

v4

### Description

The current code in main/particleset.py for calculating the initial `z` of the ParticleSet does not work as expected:
https://github.com/Parcels-code/Parcels/blob/df9a313c17de4638ce950585f10a117a8decc91c/src/parcels/_core/particleset.py#L80-L85

This means that if the `Grid.depth` is e.g. `[1, 5, 10]` m, the particles will be initiated at `z=0` m, which immediately throws an out of bound error. 

Better to initialise `zmin= np.inf`; but then this still wouldn't work for depths that are negative (e.g. `[-1, -5, -10]` m). So a proper fix would be to make this code block aware of the direction of the grid. 

An alternative (simpler) solution would be to force users to set the initial `z` explicitly; lik ewe already do for `x` and `y`. But this would be an API-change from v3, where `depth` would be inferred from he grid when users don't specify it.

wyatt-fluidnumerics:
### Description
Previously the logic to initialize the depth level of particles when no depth was provided did not work as intended, and lead to the initialization of particles out of bounds. This fixes that by selected the depth level from the fields that is closest to zero.
### Checklist

- [x] Closes #2772 

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Once I had written the new logic, I had Claude Code generate a few test cases.


erikvansebille (approved):
Code looks good! I pushed an update to the docstring. I'll merge now

PR #2796, opened 2026-08-03
Add spatial hash describe function

wyatt-fluidnumerics:
### Description
Currently, it is difficult to get a sense for many of the dimensions of the spatial hash table that gets built over unstructured, and many structured grids. This makes understanding the performance speed up due to spatial hashing, or potential bugs in the spatial hashing such as that addressed by PR #2780, difficult. This PR adds functionality for `.describe()` to the `SpatialHash` class. The function is used exclusively for debugging or logging and provides useful statistics regarding the size of the hash grid and the ratio of hash grid cells to mesh grid faces. 
### Checklist

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): I used Claude Code to generate an initial implementation of the `spatialhash_describe` function. I then made significant modifications to the resulting code to improve the utility of the reported statistics.


src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+    n_entries = int(hash_table["faces"].size)
+    n_occupied_cells = int(hash_table["keys"].size)
+    n_total_cells = (spatialhash._bitwidth + 1) ** 3
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
+        "Total Mesh Faces": str(n_faces),
+        "Bitwidth (current / max)": f"{spatialhash._bitwidth} / 1023  (higher = finer resolution hash grid)",
+        "Total hash cells": str(n_total_cells),
```
erikvansebille: These can be large numbers, so print this with a comma for each 10^3 separator (`:,d`)? 

(Perhaps for all numbers that get printed?)

src/parcels/_core/spatialhash.py
erikvansebille, line 1: Also add a unit test (like we do for `fieldset.describe`)? 
https://github.com/Parcels-code/Parcels/blob/8ee28489da583f8a27ae9af1535b7581a8ee9799/tests/test_fieldset.py#L460-L529

erikvansebille, line 1: Make sure to also add an explanation somewhere (e.g. in https://github.com/Parcels-code/Parcels/blob/main/docs/development/unstructured_grid_search.md) showcasing this function? Otherwise, users are unlikely to find/use it

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+        "Occupied hash cells": str(n_occupied_cells),
+        "Total (cell --> face) entries": str(n_entries),
+        "Entries per occupied cell (avg)": f"{n_entries / n_occupied_cells:.2f}" if n_occupied_cells else "-",
+        "Entries per face (avg)": f"{n_entries / n_faces:.2f}" if n_faces else "-",
+        "Faces per occupied cell (min / mean / max)": (
```
erikvansebille:
```suggestion
        "Faces per occupied hash cell (min / mean / max)": (
```

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
+        "Total Mesh Faces": str(n_faces),
+        "Bitwidth (current / max)": f"{spatialhash._bitwidth} / 1023  (higher = finer resolution hash grid)",
+        "Total hash cells": str(n_total_cells),
+        "Occupied hash cells": str(n_occupied_cells),
+        "Total (cell --> face) entries": str(n_entries),
+        "Entries per occupied cell (avg)": f"{n_entries / n_occupied_cells:.2f}" if n_occupied_cells else "-",
```
erikvansebille:
```suggestion
        "Entries per occupied hash cell (avg)": f"{n_entries / n_occupied_cells:.2f}" if n_occupied_cells else "-",
```

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+    n_total_cells = (spatialhash._bitwidth + 1) ** 3
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
+        "Total Mesh Faces": str(n_faces),
+        "Bitwidth (current / max)": f"{spatialhash._bitwidth} / 1023  (higher = finer resolution hash grid)",
+        "Total hash cells": str(n_total_cells),
+        "Occupied hash cells": str(n_occupied_cells),
+        "Total (cell --> face) entries": str(n_entries),
```
erikvansebille:
```suggestion
        "Total (hash cell --> grid face) entries": str(n_entries),
```

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+    n_occupied_cells = int(hash_table["keys"].size)
+    n_total_cells = (spatialhash._bitwidth + 1) ** 3
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
+        "Total Mesh Faces": str(n_faces),
+        "Bitwidth (current / max)": f"{spatialhash._bitwidth} / 1023  (higher = finer resolution hash grid)",
+        "Total hash cells": str(n_total_cells),
+        "Occupied hash cells": str(n_occupied_cells),
```
erikvansebille: Also add percentage occupied here?

```suggestion
        "Occupied hash cells": f"{n_occupied_cells}, ({n_occupied_cells/n_total_cells*100}%)",
```

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+
+    n_faces = int(np.size(spatialhash._xlow))
+    n_entries = int(hash_table["faces"].size)
+    n_occupied_cells = int(hash_table["keys"].size)
+    n_total_cells = (spatialhash._bitwidth + 1) ** 3
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
+        "Total Mesh Faces": str(n_faces),
```
erikvansebille:
```suggestion
        "Total mesh faces": str(n_faces),
```

src/parcels/_reprs.py
```diff
@@ -280,6 +281,36 @@ def fieldset_describe(fieldset: FieldSet) -> str:
...
+    counts = hash_table["counts"]
+
+    n_faces = int(np.size(spatialhash._xlow))
+    n_entries = int(hash_table["faces"].size)
+    n_occupied_cells = int(hash_table["keys"].size)
+    n_total_cells = (spatialhash._bitwidth + 1) ** 3
+
+    rows = {
+        "Grid type": type(grid).__name__,
+        "Mesh": "spherical" if grid._mesh.is_spherical() else "flat",
```
erikvansebille: Why not print the actual object here, instead of a string? That could(?) also show the radius (for spherical)

erikvansebille (commented):
Thanks for this PR, @wyatt-fluidnumerics. I like it a lot!

See below a screenshot for one of my Delft3D curvilinear meshes
Image

Very useful information; as I now wonder why the max faces per occupied cell is more than 175k(!)

See below some suggestions/comments on this PR itself

erikvansebille:
> Very useful information; as I now wonder why the max faces per occupied cell is more than 175k(!)

I dug a bit deeper, and it turns out that `counts.max()` is for those hash cells where `keys == 0`. 

I think/assume(?) that these are the hash cells that fall outside the model grid (i.e. no mesh faces occupy that hash cell), which makes sense because my grid is highly curved (see screenshot below)

So the question now is; should we only consider the counts where `keys > 0` in this describe method? 

If I do `counts = hash_table["counts"][hash_table["keys"] > 0]` on the grid below I get a max count of **44**, which is _much_ more reasonable than the 175k hash cells with `keys==0`! 

image

wyatt-fluidnumerics:
Interesting, 44 certainly seems a lot more reasonable, let me dig into this some myself as well. My understanding was that keys only stores hash cells that contain faces from the model grid's mesh, but this could be wrong.

erikvansebille:
If `hash_table["keys"] == 0` indeed represents all hash cells that don't contain any mesh face, then we could add that number as a separate line in the `describe()`. For example as "Hash cells without mesh faces"

And note that the computation of `n_occupied_cells = int(hash_table["keys"].size)` should then probably also ignore `keys == 0`

wyatt-fluidnumerics:
I am pretty sure that every entry in keys corresponds to an occupied hash cell. 
https://github.com/Parcels-code/Parcels/blob/main/src/parcels/_core/spatialhash.py#L339-L366
The `keys` for the hash table are constructed by indexing into an array of sorted morton codes at the location where that hash cell's entries in `faces` begin.

erikvansebille:
OK, but if I do `np.isnan(fieldset_wind.models[0].grid._spatialhash._xlow).sum()` I get **175698**, which is exactly the same number as the `counts[0]`...

So, could these be all the grid cells with NaN lon/lat locations from Delft3D?

wyatt-fluidnumerics:
This would make sense, the actual meaning of `key == 0` is the cell occupies the (0, 0, 0) cell on the unit cube the hash grid creates. So if NaN lon/lat get masked to 0 then I believe that would be expected. I don't totally see how this leads to an enormous number of faces.

wyatt-fluidnumerics:
I think I found where the problem occurs. The code snippet below prints 0. So if you have a NaN value then it gets clipped to (0, 0, 0) coord equivalent int in `quantize_coordinates`
```
  import numpy as np
  with np.errstate(invalid="ignore"):
      xq = np.clip(np.array([np.nan]) * 1023, 0, 1023).astype(np.uint32)

  print(xq)
```
Normally I think this would produce a RuntimeWarning but those are silenced for the block of code.

erikvansebille (approved):
PR looks great; thanks!

Issue #2307, opened 2025-10-07
Changing Field.depth to Field.z throughout v4 code


erikvansebille:
Following on from #2257 (and its implementation in #2303), it would be good to also change the vertical coordinate of our internal `Field` model from `depth` to `z`

VeckoTheGecko:
actually, this issue is different to what I thought it was

PR #2799, opened 2026-08-04
Removing depth from the tutorials

erikvansebille:
This PR removes references to `field.depth` or `grid.depth` from the tutorials, so that users don't immediately see it. This is useful in preparation for considering #2307

Note that many of the explicit setting of `z` below (e.g. `z = np.repeat(0.5, npart)`) can be removed altogether when #2781 gets merged

### Checklist

- [x] Contributes to #2307 

### AI Disclosure

None


VeckoTheGecko (commented):
Are you wanting to completely remove setting the depth? I thought you'd pull the depth from the original Xarray dataset?

erikvansebille:
Now that we have a good default setting (thanks to @wyatt-fluidnumerics's fix in #2781), I think the best way to show how users can set depth is either with an explicit value that they control; or by letting the pset logic do it based on the xarray vertical axis

VeckoTheGecko:
Sounds good

VeckoTheGecko (approved):

PR #2801, opened 2026-08-04
Fix corner index ordering

willirath:
### Description

Fix corner index ordering in the A-grid and C-grid interpolators: the flattened gather currently in main silently produced wrong corners when particless time or depth indices were not aligned, so a particle could pick up a batch-mate's `ti`/`zi`.
Both interpolators now go through a single and tested `_gather_corners` helper that derives index arrays and output shape from the same layout.

Adds a regression test that a trajectory does not depend on its batch-mate release times, plus interpolation tests for mixed per-particle indices.

Also updates the expected z values in `test_nemo_3D_curvilinear_fieldset[RK4_3D]`, which had recorded the buggy mixing.

Reproducer with U uniform in space, varying in time, so each particle's displacement depends only on its own clock:

```python
# fieldset: flat A-grid, U = cos(2*pi*t/86400) m/s (uniform in space), V = 0, 3-hourly
t0 = np.timedelta64(0, "s")

def run(release_times):
    npart = len(release_times)
    pset = ParticleSet(
        fieldset, pclass=Particle,
        t=np.array(release_times),
        z=np.zeros(npart), y=np.zeros(npart), x=np.zeros(npart),
    )
    pset.execute(AdvectionRK4, dt=np.timedelta64(1, "h"), endtime=np.timedelta64(48, "h"))
    return pset.x[0]

alone     = run([t0])
staggered = run([t0] + [t0 + np.timedelta64(3, "h")] * 3)

assert staggered == pytest.approx(alone)  # fails before the fix
```

Particle 0 should be identical in both runs as only its batch-mates differ. Before the fix, `staggered` moves particle 0 onto the other particle time index and the assert fails. The depth analogue is the same shape: two particles at different `z` in one set get each other's `zi`.

### Checklist

- [x] Tests added: Integration test and test for the new corner gather logic.
- [x] This PR targets the correct branch: Only relevant for v4. So targeted at `main`.

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Bug was found by Claude Opus when developing a downstream package. Diagnostics and implementation of the fix were done with Claude Opus. I guided the session and reviewed the code.

~~Draft PR because it's late and I want to have a fresh pair of eyes (my own or someone else's) tomorrow.~~


erikvansebille:
Great that you spotted this bug @willirath, and thanks for the fix! I had a quick skim and I think your problem diagnosis and solution are sound. Let me know when you want me to do a full/thorough review

willirath:
I think this is ready for review.

tests/test_advection.py
```diff
@@ -466,9 +466,13 @@ def test_nemo_3D_curvilinear_fieldset(kernel):
         np.testing.assert_allclose([p.z for p in pset], z_initial)
     elif kernel == AdvectionRK4_3D:
         # TODO check why decimals needs to be so low in RK4_3D (compare to v3)
+        # These particles sit at different depth levels, so the C-grid gather used to
+        # mix their zi indices; the previous expectations recorded that. Each value
+        # below is what the same particle gets when advected on its own.
```
erikvansebille: Great! So this comment can go?


```suggestion
        # TODO check why decimals needs to be so low in RK4_3D (compare to v3)
```

(edit Willi: The genrated commit would have dropped the pre-existing TODO comment as well.)

src/parcels/interpolators/_xinterpolators.py
```diff
@@ -17,6 +17,61 @@
...
+    data : np.ndarray or xr.DataArray
+        Field data, with dimensions ordered ``(time, Z, Y, X)``.
+    axis_dim : dict
+        Maps ``"X"``, ``"Y"`` and ``"Z"`` to dimension names. The ``"T"``
+        dimension is always named ``"time"``.
+    levels : dict
+        Maps an axis to the per-particle index arrays of its corner levels, e.g.
+        ``{"T": (ti, ti + 1), "Y": (yi, yi + 1)}``. An axis absent from
+        ``levels`` contributes a single level. An axis the field has no
+        dimension for still shapes the result, but is not indexed, so its
```
erikvansebille: I don't understand the "for still shapes the result". Type-o?

src/parcels/interpolators/_xinterpolators.py
```diff
@@ -17,6 +17,61 @@
...
+    ----------
+    data : np.ndarray or xr.DataArray
+        Field data, with dimensions ordered ``(time, Z, Y, X)``.
+    axis_dim : dict
+        Maps ``"X"``, ``"Y"`` and ``"Z"`` to dimension names. The ``"T"``
+        dimension is always named ``"time"``.
+    levels : dict
+        Maps an axis to the per-particle index arrays of its corner levels, e.g.
+        ``{"T": (ti, ti + 1), "Y": (yi, yi + 1)}``. An axis absent from
+        ``levels`` contributes a single level. An axis the field has no
```
erikvansebille: Would it not be simpler to then give only one value for axes that contribute to a single level - instead of not including it in the dict? E.g. `"T": (ti)`?

tests/test_interpolation.py
```diff
@@ -201,6 +202,78 @@ def test_interpolation_mesh_type(mesh, npart=10):
...
+
+    for p in range(npart):
+        single = _agrid_corners(
+            corner_gather_data, ti[p : p + 1], zi[p : p + 1], yi[p : p + 1], xi[p : p + 1], lenT, lenZ, 1
+        )
+        np.testing.assert_array_equal(batch[..., p], single[..., 0])
+
+
+def test_corner_gather_axes_are_ordered_t_z_y_x(corner_gather_data):
+    """The returned axes must mean (T, Z, Y, X, particle), as the interpolators assume."""
```
erikvansebille:
```suggestion
    """The returned axes must be (T, Z, Y, X, particle), as the interpolators assume."""
```

erikvansebille (commented):
Thanks for this fix @willirath. Looks very good! The new way of creating the selection dict in the xinterpolation functions is much cleaner now. 

A few small comments below.

src/parcels/interpolators/_xinterpolators.py
willirath: Broken grammar. Will reformulate.

src/parcels/interpolators/_xinterpolators.py
willirath: Yes, and this also allows for dropping another conditional.

tests/test_advection.py
```diff
@@ -468,7 +468,8 @@ def test_nemo_3D_curvilinear_fieldset(kernel):
         # TODO check why decimals needs to be so low in RK4_3D (compare to v3)
```
erikvansebille: Also remove this TODO now?

willirath: Ah, yes. I thought is was unrelated TODO but it hinted at the problem already. Will remove

willirath:
096192e addresss all comments.

willirath:
CI errors are pull errors from Nick's XGCM fork. Unrelated to this PR and more likely from in progress work on the env?

tests/test_advection.py
erikvansebille: This PR indeed fixed that TODO. I reran the code in cell [1] of https://docs.parcels-code.org/en/v3.1.4/examples/tutorial_nemo_3D.html (using parcels v3) and updated the expected values for the test in cc98f98aceb5aed3ab7724dde73cdfed1feffb11. Agreement is now much better!

erikvansebille (approved):

PR #2802, opened 2026-08-04
Exclude faces that touch NaN nodes from the SpatialHash table

wyatt-fluidnumerics:
### Description
Previously, any faces that touched NaN nodes in curvilinear or unstructured grids were treated as a normal face. This meant they were quantized and added to the SpatialHash table. The resulting behavior was a pileup of faces being assigned to (0, 0, 0). For more details see the discussion on #2796. This PR resolves this issue by manually setting values in `num_hash_per_face` to 0 for faces which touch a node that is NaN. This prevents these faces from ever entering the SpatialHash table. This PR also adds testing to confirm that faces that touch NaN nodes are never in the SpatialHash table, and that this new behavior does not effect the query resolution of any other mesh cells in the grid.
### Checklist

- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Used Claude code in plan mode to discuss the best approaches for adding the NaN filter and testing. Claude found and fixed a bug in the new test that arose from not deep copying the lat/lon grid before injecting NaN values.

### DIFF
--- a/src/parcels/_core/spatialhash.py
+++ b/src/parcels/_core/spatialhash.py
@@ -788,12 +788,7 @@ def _generate_valid_mask(xlow, xhigh, ylow, yhigh, zlow, zhigh):
         False where it contains a NaN.
     """
     invalid_face = (
-        np.isnan(xlow)
-        | np.isnan(xhigh)
-        | np.isnan(ylow)
-        | np.isnan(yhigh)
-        | np.isnan(zlow)
-        | np.isnan(zhigh)
+        np.isnan(xlow) | np.isnan(xhigh) | np.isnan(ylow) | np.isnan(yhigh) | np.isnan(zlow) | np.isnan(zhigh)
     )
 
     return ~invalid_face
--- a/src/parcels/_core/spatialhash.py
+++ b/src/parcels/_core/spatialhash.py
@@ -260,15 +260,8 @@ class SpatialHash:
         nz = zqhigh.astype(np.int64) - zqlow + 1
         # NaN values are not allowed in the SpatialHash table, so faces with a NaN
         # bounding box do not contribute to the entry count
-        invalid_face = (
-            np.isnan(self._xlow)
-            | np.isnan(self._xhigh)
-            | np.isnan(self._ylow)
-            | np.isnan(self._yhigh)
-            | np.isnan(self._zlow)
-            | np.isnan(self._zhigh)
-        )
-        return int(np.where(invalid_face, 0, nx * ny * nz).sum())
+        valid_face = _generate_valid_mask(self._xlow, self._xhigh, self._ylow, self._yhigh, self._zlow, self._zhigh)
+        return int(np.where(valid_face, nx * ny * nz, 0).sum())
 
     def _initialize_hash_table(self):
         """Create a mapping that relates unstructured grid faces to hash indices by determining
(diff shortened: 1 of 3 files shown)


wyatt-fluidnumerics:
@erikvansebille, curious to hear your thoughts on this solution! To me it seemed like this approach would take the least amount of refactoring as it doesn't mess with the 2d array reconfiguration issues you had mentioned earlier today.

erikvansebille:
This works beautifully! My example that I showed earlier in https://github.com/Parcels-code/Parcels/pull/2796#pullrequestreview-4850996851 now looks much better
Screenshot 2026-08-05 at 08 39 35

src/parcels/_core/spatialhash.py
erikvansebille, line 1: This solution is very elegant! Nice work @wyatt-fluidnumerics 

The only thing I wonder about is if it's worth to make a helper function to construct the `invalid_face` array. Because the same code is now used twice. Or will that make going through the code only more confusing?

src/parcels/_core/spatialhash.py
```diff
@@ -261,7 +261,17 @@ def _total_hash_entries(self, bitwidth):
         nx = xqhigh.astype(np.int64) - xqlow + 1
         ny = yqhigh.astype(np.int64) - yqlow + 1
         nz = zqhigh.astype(np.int64) - zqlow + 1
-        return int((nx * ny * nz).sum())
+        # NaN values are not allowed in the SpatialHash table, so faces with a NaN
+        # bounding box do not contribute to the entry count
+        invalid_face = (
```
erikvansebille: We tend to think of masks here, so perhaps rename to `valid_mask`?

src/parcels/_reprs.py
```diff
@@ -298,7 +298,7 @@ def spatialhash_describe(spatialhash: SpatialHash) -> str:
         "Bitwidth (current / max)": f"{spatialhash._bitwidth} / 1023  (higher = finer resolution hash grid)",
         "Total hash cells": f"{n_total_cells:,d}",
         "Occupied hash cells": f"{n_occupied_cells:,d}, {n_occupied_cells / n_total_cells * 100:.4f}%",
-        "Total (hash cell --> gird face) entries": f"{n_entries:,d}",
+        "Total (hash cell --> grid face) entries": f"{n_entries:,d}",
```
erikvansebille, line 303: Left-over typo from #2796 that I only spotted now...

tests/test_spatialhash.py
erikvansebille, line 1: Nice test, but should we also add a check that the number of unique mesh faces is less than the number of original grid cells - because the NaNs are filtered out?

I think this should be `len(np.unique(spatialhash._hash_table["faces"]))`, but double-check. Perhaps that's also what the "Total mesh faces" of hashtable.describe() should show?

src/parcels/_core/spatialhash.py
wyatt-fluidnumerics: I think a helper here makes sense, just added.

src/parcels/_core/spatialhash.py
wyatt-fluidnumerics: Agreed, I modified the logic so that the True corresponds to valid non-NaN cells.

tests/test_spatialhash.py
wyatt-fluidnumerics: Just added this check, and changed hashtable.describe() to display "Total (non-NaN) mesh faces" as a separate metric. I could imagine it may be useful to see this in addition to the raw total

src/parcels/_core/spatialhash.py
```diff
@@ -302,7 +305,13 @@ def _initialize_hash_table(self):
...
         ny = (yqhigh - yqlow + 1).astype(np.int32, copy=False)
         nz = (zqhigh - zqlow + 1).astype(np.int32, copy=False)
-        num_hash_per_face = (nx * ny * nz).astype(
+
+        # prevent NaN values from entering the SpatialHash table by setting their
+        # num_hash_per_face equal to 0
+        valid_face = _generate_valid_mask(
+            self._xlow, self._xhigh, self._ylow, self._yhigh, self._zlow, self._zhigh
+        ).ravel()
+        num_hash_per_face = np.where(valid_face, nx * ny * nz, 0).astype(
```
erikvansebille: Put the .ravel on the np.where line?

```suggestion
        valid_face = _generate_valid_mask(self._xlow, self._xhigh, self._ylow, self._yhigh, self._zlow, self._zhigh)
        num_hash_per_face = np.where(valid_face.ravel(), nx * ny * nz, 0).astype(
```

erikvansebille (commented):
Nice! One more small suggestion

erikvansebille (approved):
Very good! Now it's custom that you, as the initiator of this PR, also merge it. DO you have the rights to do that?

wyatt-fluidnumerics:
Just initiated it, I believe it should go through when the checks complete?

PR #2812, opened 2026-08-12
Regression tests: Fixing interpolation to match v3 

### LINKED ISSUE #2807: Fix v3 regression tests for all interpolators
### Parcels version

v4

### Description

#2766 only implemented the interpolation regression test for the XLinear interpolator. The others will need to be checked too. From some first poking around, it seems that at least Nearest has the same issue as solved in #2801

erikvansebille:
### Description

This PR makes some fixes to v4 so that the regression tests of all four interpolation schemes tested in v3 pass. Note that I had to add a "NearestNeighbour Velocity" interpolation; but since I don't see any use-cases for it I put it in the tests directory instead of _xinterpolators

### Checklist

- [x] Closes #2807
- [x] Tests added

### AI Disclosure

None


tests/test_interpolation.py
```diff
@@ -333,53 +358,21 @@ def DeleteParticle(particles, fieldset):
...
-        # Skip if all NaN
-        if np.all(np.isnan(v3_lon)) or np.all(np.isnan(v4_lon)):
-            continue
-
-        tol = 1e-6
-        np.testing.assert_allclose(v3_lon, v4_lon, atol=tol)
-        np.testing.assert_allclose(v3_lat, v4_lat, atol=tol)
-        np.testing.assert_allclose(v3_z, v4_z, atol=tol)
+    # v3 zarr is not sorted by particle_id, so we sort it here to match the v4 output
+    ds_v3 = ds_v3.sortby("trajectory")
```
VeckoTheGecko, line 373: I'm curious about this - wasn't the zarr output in order of ascending particle_id?

VeckoTheGecko (approved):
Looks good to me! One question

tests/test_interpolation.py
erikvansebille: Yes, I was surprised about this too. The v3 output files indeed weren't sorted...

PR #2816, opened 2026-08-12
Add .load() call on UxGrid.uxgrid.face_node_connectivity

### LINKED ISSUE #2815: Optimization for unstrucutred grid point-in-cell checking and interpolation

v4

When running the fesom2 component of the parcels benchmarking repository on the most up to date version of parcels v4, and running a Cprofile to capture function calls, it can be seen that the code makes 375 total calls to dask compute. These calls account for a substantial component of the total simulation length:

Image

These calls can be traced by wrapping `dask.compute` in a function that stores the stack frames preceding the `dask.compute call`, before calling `dask.compute` itself and returning that result to maintain the functionality of the existing code. 

Image

Running the fesom2 benchmark after reassigning `dask.compute` to this function, and then printing the most common stack frames leads to the following:

Image

The two most responsible call sites are:
https://github.com/Parcels-code/Parcels/blob/c9ccd66d0e7eca4d3938512f6e5ef04eba39d6b0/src/parcels/interpolators/_uxinterpolators.py#L101

and 

https://github.com/Parcels-code/Parcels/blob/c9ccd66d0e7eca4d3938512f6e5ef04eba39d6b0/src/parcels/_core/index_search.py#L328

which loads a dask backed array linking face_ids to nodes.

The fix to this is trivial, the dask backed face_id to node connectivity array is not large. In the UxGrid constructor a `.load()` can be called directly on `self.uxgrid.face_node_connectivity` since the array will never be larger than `(n_faces, face_dim)` (where `face_dim=3` for all parcels supported unstructured meshes), this should not ever cause significant memory overhead. 

Applying this fix results in significant performance increases. The fesom2 benchmark with 10000 particles goes from 6.28s to 2.23s, without altering the peak memory consumption. As particle counts increase the impact of this optimization decreases:

Image

_No response_

### Additional context

_No response_

wyatt-fluidnumerics:
### Description
See #2815 for major details, this PR serves to decrease the number of `dask.compute` calls during particle-in-cell checks and during interpolation for unstructured grids by backing the `uxgrid.face_node_connectivity` array with numpy rather than leaving it lazy as a dask array. 
### Checklist

- [ ] Closes #2815 
- [ ] This PR targets the correct branch (`main` for normal development, `v3-support` for v3 support)

### DIFF
--- a/src/parcels/_core/uxgrid.py
+++ b/src/parcels/_core/uxgrid.py
@@ -45,6 +45,7 @@ class UxGrid(BaseGrid):
         self.z = z
         self._mesh = get_mesh(mesh)
         self._spatialhash = None
+        self.uxgrid.face_node_connectivity.load()
 
     @property
     def depth(self):
--- a/src/parcels/_core/uxgrid.py
+++ b/src/parcels/_core/uxgrid.py
@@ -4,6 +4,7 @@ from typing import Literal
 
 import numpy as np
 import uxarray as ux
+from dask import is_dask_collection
 
 from parcels._core.basegrid import BaseGrid
 from parcels._core.index_search import GRID_SEARCH_ERROR, _search_1d_array, uxgrid_point_in_cell
@@ -45,7 +46,8 @@ class UxGrid(BaseGrid):
         self.z = z
         self._mesh = get_mesh(mesh)
         self._spatialhash = None
-        self.uxgrid.face_node_connectivity.load()
+        if is_dask_collection(self.uxgrid.face_node_connectivity.data):
+            self.uxgrid.face_node_connectivity.load()
 
     @property
     def depth(self):


erikvansebille (changes requested):
Wow, this is a great and simple performance boost; well done on finding this @wyatt-fluidnumerics!

It's conceptually similar to the performance gain of loading the `lon` and `lat` arrays in xgrid
https://github.com/Parcels-code/Parcels/blame/c9ccd66d0e7eca4d3938512f6e5ef04eba39d6b0/src/parcels/_core/xgrid.py#L159-L161

Perhaps here also add some safeguards that the .load is only called if the array is a dask collection?

But these kinds of low-hanging fruit performance boosts are exactly what we should be looking for now; great job!

wyatt-fluidnumerics:
Great! Just added the dask collection check.

erikvansebille (approved):
Looks good now!

PR #2823, opened 2026-08-14
Fix silent fail on pset sampling when time is NaN

### LINKED ISSUE #2822: Field Sampling fails when particleset has no time set yet
### Parcels version

v4

### Description

While #2622 implemented sampling of a Field directly with a particleset (`var = fieldset.field[pset]`) I now realise that this does not work if the particle times are not set explicitly. 

In that case, (because time is set to NaN) the interpolation returns zero - without warning!

### Code sample

```python
fieldset.U.data[:] = 2.
pset = ParticleSet(fieldset, x=0., y=0.)
pset.sample, _ = fieldset.UV[pset]  # Sample initial value
np.testing.assert_allclose(pset.sample, 2.0, rtol=1e-12)
```

```text
AssertionError: 
E       Not equal to tolerance rtol=1e-12, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference among violations: 2.
E       Max relative difference among violations: 1.
E        ACTUAL: array(0)
E        DESIRED: array(2.)
```

Note that the test above passes if we change
```diff
- pset = ParticleSet(fieldset, x=0., y=0.)
+ pset = ParticleSet(fieldset, x=0., y=0., t=np.timedelta64(0, "s")

```

erikvansebille:
### Description

This PR fixes #2822, where field interpolation would silently fail if the `particles.t` included NaN values. Now, this raises a ValueError

### Checklist

- [x] Closes #2822 
- [x] Tests added

### AI Disclosure

None


src/parcels/_core/field.py
```diff
@@ -393,6 +393,9 @@ def _assert_same_time_interval(fields: Sequence[Field]) -> None:
 
 def _get_positions(field: Field, t, z, y, x, particles, _ei) -> tuple[dict, dict]:
     """Initialize and populate particle_positions and grid_positions dictionaries"""
+    if np.any(np.isnan(t)):
+        nan_indices = np.where(np.isnan(t))[0]
+        raise ValueError(f"Time values for particles with indices {nan_indices} cannot be NaN.")
```
VeckoTheGecko, line 398: Wondering whether this should be somewhere in the execute loop?

i.e., make sure that the time for each particle is set before we go into the main loop

Is there a usecase of `t` being `nan` during execution?

erikvansebille: No, the point is that time is never NaN in a pset.execute (see [here](https://github.com/Parcels-code/Parcels/blob/a21221a1bd0169f809c0670c0940b1042d8257b6/src/parcels/_core/particleset.py#L414-L415)), but it can be _before_ the execute loop; e.g. during an initial sampling (`particles.var = fieldset.var[particles]`). That is why we need this error here and not in the execute loop

VeckoTheGecko: Ah, I see now

VeckoTheGecko (approved):

PR #2832, opened 2026-08-17
API: Decide on keyword only API

### LINKED ISSUE #2716: Enforcing keyword-only arguments in certain API
Working on https://github.com/Parcels-code/Parcels/pull/2715 I realised that there might be parts of the codebase where we want to force keyword only API.

Here are some code samples to show what keyword-only arguments are:

```python
# current code
def from_sgrid_conventions(
    ds, mesh, vector_fields
):
    pass

ds = ...
mapping = ...
from_sgrid_conventions(ds, "flat", mapping)
# passes
```
```python
def from_sgrid_conventions(
    ds, *, mesh, vector_fields # the `*` forces all subsequent arguments to be called by keyword=
):...

ds = ...
mapping = ...
from_sgrid_conventions(ds, "flat", mapping)
# Traceback (most recent call last):
#   File "/Users/Hodgs004/coding/repos/parcels/t.py", line 22, in <module>
#    from_sgrid_conventions(ds, "flat", mapping)
#    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
#TypeError: from_sgrid_conventions() takes 1 positional argument but 3 were given

from_sgrid_conventions(ds, mesh="flat", vector_fields=mapping)
# passes
```
.

---

In the case of the `Fieldset.from_...` constructor methods I find that using keyword arguments is much more readable and less error prone. Keyword only arguments also has the very significant benefit of allowing us to seamlessly make API changes. If we want to remove `mesh` as an argument down the line we can just see if the user uses it and raise a warning. This becomes quite difficult if they provide this as a positional argument.

I wonder as well if other parts of the codebase can benefit from keyword only arguments - something to discuss more @erikvansebille  before 4.x release

VeckoTheGecko:
### Description

I was looking through the tutorial to see how we're using our functions, and to decide on which API should be keyword only. These are my thoughts

- `parcels.FieldSet.from_sgrid_conventions`
  - make `mesh`, `vector_fields`, `skip_field_data_validation` keyword only
- `parcels.FieldSet.from_ugrid_conventions`
  - make `mesh`, `vector_fields` keyword only
- `parcels.ParticleSet.__init__`
  - make particle initial values keyword only
- `ParticleFile.__init__`
  - leave as is (minimal chance of confusion)
- the rest - keep as is

This is a bit subjective. Happy to discuss if you have other thoughts @erikvansebille 

### Checklist

- [x] Closes #2716
- [x] Tests added

### AI Disclosure

None used

### DIFF
--- a/src/parcels/_core/fieldset.py
+++ b/src/parcels/_core/fieldset.py
@@ -263,7 +263,7 @@ class FieldSet:
         -----
         See https://ugrid-conventions.github.io/ugrid-conventions/ for more information on the UGRID conventions.
         """
-        model = UnstructuredModelData.from_ugrid_conventions(ds, mesh, vector_fields)
+        model = UnstructuredModelData.from_ugrid_conventions(ds, mesh=mesh, vector_fields=vector_fields)
         return cls([model])
 
     @classmethod
@@ -308,7 +308,7 @@ class FieldSet:
         See https://sgrid.github.io/sgrid/ for more information on the SGRID conventions.
         """
         model = StructuredModelData.from_sgrid_conventions(
-            ds, mesh, vector_fields, skip_field_data_validation=skip_field_data_validation
+            ds, mesh=mesh, vector_fields=vector_fields, skip_field_data_validation=skip_field_data_validation
         )
         return cls([model])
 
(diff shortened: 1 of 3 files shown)


VeckoTheGecko:
(I still need to update the test suite and the docs, but that is quite straightforward)

EDIT: Done

erikvansebille (approved):
Yep I agree with this list. PR looks good

PR #2841, opened 2026-08-21
MAINT: Remove `--install-types` from typing Pixi command

VeckoTheGecko:
### Checklist

- [x] Closes None
- [x] Tests added

### AI Disclosure

None used


VeckoTheGecko:
actually, nevermind

> ✨ Pixi task (typing in default): mypy src/parcels --non-interactive --html-report mypy-report: (Run static type checking with mypy.)
> error: --non-interactive is only supported with --install-types

PR #2846, opened 2026-08-24
Refactor windowed arrays to use a numpy backed DataArray cache

### LINKED ISSUE #2843: Windowed Array Improvements/Optimization

v4

Currently, windowed arrays are the encouraged option for particle simulations that use large grids. Nonetheless, the size of the grid severely limits the speed at which the computation occurs. A significant portion of this slowdown occurs during interpolation, during which the `WindowedArray.isel()` method is called to gather the model data surrounding particles. In the current implementation of `WindowedArray` there is a costly `np.stack` call on already loaded data from the cache:
https://github.com/Parcels-code/Parcels/blob/febb39e46440e1056948ec3da39764ce0a89a7ec/src/parcels/_core/_windowed_array.py#L93
This is costly because `np.stack` allocates the size of the stacked array, and then copies in the relevant data. For large datasets such as moi curvillinear in the parcels-benchmarks repository, this results in a roughly 10GB RAM copy on every call to `WindowedArray.isel()`. For advanced particle kernels such as RK2 or RK4 which call `.isel` multiple times per particle timestep, this is pretty detrimental. The damage is compounded by the fact that the newly stacked array is almost immediately dropped after a very small subset of it is indexed into:
https://github.com/Parcels-code/Parcels/blob/febb39e46440e1056948ec3da39764ce0a89a7ec/src/parcels/_core/_windowed_array.py#L94-L97

There appear to be two different solutions to this problem:

1. Remove the reliance on a `.isel` call in L97 by making an index into the numpy arrays that are stored in the cache. This avoids the need for stacking them and the construction of the `nda` xarray. 
2. Change the cache data structure from a dictionary to a numpy backed `xr.DataArray`. This approach allows you to retain nearly all of the same `.isel` behavior on L97 as the current implementation, without the need for the `np.stack`. The cache is a `xr.DataArray` in which the current cache's keys are stored in the `t_dim` of the DataArray. 

Both approaches have the same benefit of avoiding the large `np.stack` copy during `.isel`. However, I prefer the second because it reduces the complexity of the required code by remaining reliant on xarray's `.isel` for the indexing operation. The tradeoff is re-implementing `WindowedArray._esure()` to be compatible with the modification to the cache's data structure. 

Implementing the second of these approaches results in significant performance benefits as the size of the grid in a fieldset increases. For the moi curvilinear benchmark, I am seeing performance increases of ~26x for a 100 timestep RK2 integration of 10000 particles on the moi curvilinear dataset. 

Image

I am currently putting the finishing touches on a PR with the windowed array cache refactor.

_No response_

### Additional context

_No response_

wyatt-fluidnumerics:
### Description
This improvement is largely outlined in #2843. The goal is to remove the current codes dependence on a `np.stack` call in `WindowedArray.isel()` which causes a large in memory copy. This implementation replaces the existing dictionary cache with a numpy backed DataArray, which allows for direct indexing and removes the need to reconstruct a DataArray from the dictionary cache on every call to `.isel()`. This results in significant performance improvement for large datasets (as much 20x or more depending on size and advection scheme).

Additionally, these changes mean that in theory `WindowedArray` now has all the needed functionality to support non-synchronous clocks. In practice however, this is likely not a very practical use case as for large datasets holding even a few time levels in the cache can lead to OOM errors on many machines. 
### Checklist

- [x] Closes #2843 

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): Claude code was used to write code to benchmark this change against the previous version and create the plot in #2843. I thoroughly reviewed the benchmarking code. It was also used to write parts of the new implementation, though significant changes and rewrites to those components were made by myself.


src/parcels/_core/_windowed_array.py
```diff
@@ -54,47 +62,88 @@ def _read_level(self, lvl: int) -> np.ndarray:
...
+            # If isel requests more levels to be loaded than self._max, then the
+            # request must be granted. Otherwise an indexing error will occur when
+            # isel attempts to index into the cache. This can cause large memory
+            # overhead, potentially beyond the cap set by self._max.
+            # The most likely reason for this to occur is non-synchronous particle clocks.
+            warnings.warn(
+                f"The windowed array cache is attempting to hold {levels.size} time levels "
+                f"which exceeds max_level={self._max}; the cache will hold {levels.size} to maintain "
+                f"simulation accuracy. This may cause significant memory usage or an OOM error. "
+                f"This most likely occured due to non-synchronous particle clockes. Raise max "
```
fluidnumericsJoe: change "clockes" to "clocks" -- typo

erikvansebille (approved):
Looks good!

Issue #2849, opened 2026-08-25
MAINT: Fix all deprecationwarnings


VeckoTheGecko:
We're starting to have failing CI as some of these deprecations complete their cycles.

# Numpy

> NumPy 2.5.0 released
> 21 Jun, 2026 – Numpy 2.5.0 is a transitional release. It drops support for Python 3.11, marking the end of distutils, and expires a large number of deprecations made in the 2.0.x release. It also improves free threading and brings sorting into compliance with the array-api standard with the addition of descending sorts. Highlights are:
> 
> Distutils has been removed,
> Many expired deprecations, see below,
> Many new deprecations, see below,
> Many static typing improvements,
> Improved support for free threading,
> Support for descending sorts.
> This release supports Python versions 3.12-3.14

Might be worth cross xrefing with:
- [PRs from `sujeito-operator`](https://github.com/Parcels-code/Parcels/issues?q=sort%3Aupdated-desc+is%3Apr+author%3Asujeito-operator+) - though looking through them now the fixes are needlessly overcomplicated (e.g., replacing `sum` with `functools.reduce(operator.add, dts)` instead of `np.sum`)

# ...

PR #2850, opened 2026-08-25
MAINT: Avoid deprecated generic timedelta

VeckoTheGecko:
### Description

Per https://numpy.org/doc/2.5/release/2.5.0-notes.html 

> Using the generic unit in numpy.timedelta64 is deprecated since this can lead to unexpected behavior such as non-transitive comparison, see [gh-28287](https://github.com/numpy/numpy/issues/28287) for details. As an alternative, specify an explicit unit such as 's' (seconds) or 'D' (days) when constructing numpy.timedelta64. Due to this change, operations that implicitly rely on the generic unit are also deprecated. For example:
> ```
> arr = np.array([1, 2, 3], dtype="m8[s]")
> 
> # `1` is implicitly converted to generic timedelta64
> arr + 1
> ```

The implementation of Python's `sum` is `0 + element_1 + ...`, which now fails for these timedelta objects.

### Checklist

- [x] contributes to #2849
- [x] Tests added

### AI Disclosure

None used

### DIFF
--- a/src/parcels/_core/utils/time.py
+++ b/src/parcels/_core/utils/time.py
@@ -182,7 +182,7 @@ def maybe_convert_python_timedelta_to_numpy(dt: timedelta | np.timedelta64) -> n
                 dts.append(np.timedelta64(value, np_unit))
 
         if dts:
-            return sum(dts)
+            return np.sum(dts)
         else:
             return np.timedelta64(0, "s")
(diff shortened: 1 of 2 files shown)


VeckoTheGecko:
CI was failing on main because of this (since numpy updated). Merging on green

Issue #2854, opened 2026-08-29
New Parcels backend - `ChunkCachedArray`


VeckoTheGecko:
#2656 introduced time windowing to Parcels. This greatly improved performance when working with larger-than-memory datasets as it eliminate's Dask overhead on the vectorized indexing level. However, there are still improvements that can be made:

- to reduce the memory overhead: at the moment we're loading whole timeslices which has significant memory use when the particles may only be in a small portion of the domain
- to improve integration with Xarray: currently we're working completely in Parcels - at a level of abstraction above Xarray. This is a somewhat messy abstraction level to work at, resulting in maintainer burden, and I think it can be put within the underlying arrays within Xarray instead.

Something like 

<img width="1247" height="703" alt="Image" src="https://github.com/user-attachments/assets/22ed1c86-6030-4512-bc9c-ac267cb3764c" />

I've made significant progress on an implementation, which (from prelim testing on my laptop) has significantly better performance than windowed array up to 1M particles. Just wanting to do some further work to make sure the abstractions work well with Xarray (and that it's properly tested), before making a PR.

Checklist:
- [ ] Testing against nightly Xarray
  - Since this change would rely on potentially unstable Xarray behaviour, it would be good to have a nightly testing workflow so that any regressions are picked up before Xarray cuts a release.
- [ ] CachedChunkArray implementation
- [ ] (low prio) Cleanup of methods
  - Should we go for `FieldSet.use_backend("windowed_arrays")`, `FieldSet.use_backend("chunked_arrays")` or similar for the API? I think that would be a clearer API than having a bunch of methods.

VeckoTheGecko:
Note that `CachedChunkArray` only implements a small subset of the array API. Hence my mention about its reliance on upstream Xarray (in case they change indexing such that it requires more than what we've implemented).

erikvansebille:
Implemented in #2860

PR #2860, opened 2026-09-01
Implement `.to_chunk_cached_arrays()` with `ChunkCachedArray`

VeckoTheGecko:
### Description

Provides an opt-in optimization that wraps dask-backed field data in
chunk-level LRU caches. 
vectorized .isel() calls hit an in-memory cache instead of recomputing
dask task graphs, giving large speedups for particle simulations.

This PR adds `ChunkCachedArray` according to the approach detailed in #2854 , providing an opt-in optimization wrapping dash-backed field data with (dask)chunk-level LRU caches on the array level. This PR:

- Adds a subpackage `parcels._chunk_cached_array` with the implementation of the Array and the needed LRU cache. Note the cache is on the array level with configurable memory limits per array.
- Exposes a function `parcels._chunk_cached_array.wrap_dataset` providing the main entrypoint for Parcels for wrapping Xarray dataset objects
- Adds `.to_chunk_cached_arrays` to `FieldSet` and `ModelData` classes.
- Adds equivalence testing (`test_backends.py`) on a NEMO dataset, asserting that the same results are received regardless of backend.

Still needed:
- (separate PR) Upstream testing
- [ ] (separate PR?) Backend equivalency testing on Unstructured grids?
- [x] Benchmarking with realworld simulations
- [x] (pre-merge) Remove temporary testing files (`data-generation.py`,`benchmark_chunk_cache.py`,`benchmark_chunk_cache.png`)

Future work:
- `max_cache_bytes` on `FieldSet.to_chunk_cached_arrays` operates _per variable_ (i.e., per `Field`). This requires the user to think about how many Fields are in their simulation in order to tune their performance. Down the line, we could tune this to allocate a total cache of (e.g.,) 50% of the system memory (which would allow for other memory overhead of coordinates (not stored in the cache), the Python process, or other user processes). This would require significant refactoring not only on how the chunk cache is handled, but also backends in general.

(cc @erikvansebille, @wyatt-fluidnumerics )

### Performance

I have some some preliminary profiling with a 20Gb idealised dataset, and found the following performance profile which looks like a promising first step.

image

I've included `data-generation.py` and `benchmark_chunk_cache.py` here for testing in case it helps.

Its not clear to me the performance with real-world simulations and how that scales (I only have access to my laptop, and found working/debugging on Lorenz quite frustrating with my high latency from travels). Let me know if there's anything I can do to help here @erikvansebille .

### Checklist

- [x] Contributes to #2854
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): I guided an LLM to explore the Xarray codebase, and to help work on the implementation of the Array class. I also used an LLM to generate and iterate on the profiling script, and write docstrings.


erikvansebille:
> Its not clear to me the performance with real-world simulations and how that scales (I only have access to my laptop, and found working/debugging on Lorenz quite frustrating with my high latency from travels). Let me know if there's anything I can do to help here @erikvansebille .

Thanks for this exciting PR, @VeckoTheGecko! I plan to do some real-world performance testing (also including/comparing #2846 and v3) later this week. Will report back here when I know more!

src/parcels/_chunk_cached_array/lru_cache.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def get(self, key: Hashable) -> np.ndarray | None:
+        if key not in self._cache:
+            return None
+        self._cache.move_to_end(key)
+        return self._cache[key]
+
+    def put(self, key: Hashable, value: np.ndarray) -> None:
+        nbytes = value.nbytes
+        if nbytes > self._max_bytes:
```
wyatt-fluidnumerics, line 34: I wonder if this should be accompanied by a warning. My understanding is the chunk size is set by dask/xarray and if it is greater than the cache size then this would silently skip use of the cache altogether.

src/parcels/_chunk_cached_array/core.py
```diff
@@ -0,0 +1,156 @@
...
+            local_indices[d] = indices[d] - self._boundaries[d][cid]
+
+        # Step 2: Group points by chunk using a structured array for vectorized grouping.
+        # Encode each point's chunk coords as a single int for fast grouping.
+        # Use np.ravel_multi_index on chunk_ids to get a flat chunk key per point.
+        numblocks = np.array(self.array.numblocks, dtype=np.intp)
+        flat_keys = np.ravel_multi_index(chunk_ids, numblocks)
+
+        # Sort points by flat chunk key to group them.
+        sort_order = np.argsort(flat_keys, kind="mergesort")
```
wyatt-fluidnumerics: From the testing I've done it looks like for randomly distributed particles quicksort may be faster here?

wyatt-fluidnumerics:
This is really cool! The potential it has to lower peak memory usage for comparable speed to the current windowed array implementation is very exciting. I left a couple of comments and will also plan on doing some performance testing through the week!

VeckoTheGecko:
> I left a couple of comments

I can't see them on my end, maybe you still need to click 'submit' on the review?

src/parcels/_chunk_cached_array/lru_cache.py
VeckoTheGecko: I think that's reasonable to raise a warning. Any thoughts about how we can only raise it once? (since raising on every put would flood the output)

src/parcels/_chunk_cached_array/core.py
VeckoTheGecko: Awesome! I'll update

src/parcels/_chunk_cached_array/core.py
erikvansebille, line 1: Note that the `fieldset.describe()` also needs to be updated. That currently shows 'NumPy`

VeckoTheGecko: Good catch - fixed in latest updates

src/parcels/_chunk_cached_array/core.py
VeckoTheGecko: Done

VeckoTheGecko:
Wondering what the order of the should be for the performance - ideally we want our most recommended to be nearer the top. Currently:

1. load the full FieldSet into memory
1. use (cached) zarr files
1. use Windowed Arrays
1. use Dask
1. use ChunkCachedArrays

Thoughts @erikvansebille ?

docs/user_guide/examples/explanation_performance.md
```diff
@@ -110,3 +110,26 @@ The long-term plan for Parcels development is to make this Option 4 work well fo
 If you have ideas for how to make Parcels faster, we'd love to hear from you!
 Feel free to [open an issue](https://github.com/Parcels-code/Parcels/issues) or reach out to us on [Zulip](https://clam-community.github.io).
 ```
+
+## Option 5: use ChunkCachedArrays
+
+_Uses Parcels Backend: ChunkCachedArray_
+
+**Best for: large Datasets (more than a few GB) and particles distributed over the entire domain**
```
erikvansebille, line 118: Should this be particles _not_ distributed over the entire domain?

erikvansebille:
> Wondering what the order of the should be for the performance - ideally we want our most recommended to be nearer the top. Currently:

Yep, I was also thinking about this. I would say we do it in the following order
 
1. load the full FieldSet into memory
2. use ChunkCachedArrays
3. use Windowed Arrays
4. use (cached) zarr files
5. use Dask

docs/user_guide/examples/explanation_performance.md
```diff
@@ -110,3 +110,26 @@ The long-term plan for Parcels development is to make this Option 4 work well fo
 If you have ideas for how to make Parcels faster, we'd love to hear from you!
 Feel free to [open an issue](https://github.com/Parcels-code/Parcels/issues) or reach out to us on [Zulip](https://clam-community.github.io).
 ```
+
+## Option 5: use ChunkCachedArrays
```
erikvansebille, line 114: Update order based on https://github.com/Parcels-code/Parcels/pull/2860#issuecomment-5566941290?

src/parcels/_chunk_cached_array/core.py
```diff
@@ -0,0 +1,156 @@
...
+    xr.Dataset
+        Copy with dask arrays wrapped in ChunkCachedArray.
+    """
+    from dask.base import is_dask_collection
+
+    ds = ds.copy()
+    # Load coordinates eagerly — they are small 1D arrays and keeping them
+    # as dask arrays causes expensive task-graph construction on every .isel().
+    for name in list(ds.coords):
+        ds[name].load()
```
erikvansebille, line 43: Note that we should also load `lon` and `lat` here. They are not coordinates in a curvilinear grid - but are called _very_ often. See also #2752 

https://github.com/Parcels-code/Parcels/blob/9dc6bb4fce5374be7a5bb0d712ba518fda348272/src/parcels/_core/model.py#L113-L116

src/parcels/_core/model.py
```diff
@@ -119,6 +120,29 @@ def to_windowed_arrays(self, *, max_levels: int | None = None) -> Self:
...
 
+    def to_chunk_cached_arrays(self, *, max_cache_bytes: int = 600_000_000) -> Self:
+        """Wrap dask-backed field data in chunk-level LRU caches.
+
+        Opt-in optimization that replaces each dask-backed data variable's
+        internal storage with a :class:`~parcels._chunk_cached_array.ChunkCachedArray`.
+        Repeated vectorized ``.isel()`` calls then hit an in-memory LRU cache
+        keyed by chunk coordinates instead of recomputing dask task graphs.
+
+        Coordinate variables are loaded eagerly into memory (they are small 1D
```
erikvansebille, line 131: Or "non-time-varying"? As per #2752?

benchmark_chunk_cache.py
erikvansebille, line 1: Should this file be part of the repo? Should it not go to the benchmarks repository, if anywhere?

benchmark_chunk_cache.png
erikvansebille, line 1: Should this file be on the main repo?

data-generation.py
erikvansebille, line 1: Should this file be on the main repo?

erikvansebille (commented):
Looks good. A few comments/questions below

src/parcels/_chunk_cached_array/core.py
VeckoTheGecko: Why aren't they set as coordinates on a curvilinear grid? I feel they should be set as such

benchmark_chunk_cache.png
VeckoTheGecko: > Still needed:
> ...
>  - [ ] (pre-merge) Remove temporary testing files (data-generation.py,benchmark_chunk_cache.py,benchmark_chunk_cache.png)

src/parcels/_core/model.py
VeckoTheGecko: I read through  #2752 , and I don't understand it

src/parcels/_chunk_cached_array/core.py
erikvansebille: Oh you are right - they are coordinates. So your code does the same as #2752. Perhaps we should also change  `for dim in ["lon", "lat", "depth"]` on line 113 of model.py to `for name in list(ds.coords):`?

src/parcels/_core/model.py
erikvansebille: Never mind this comment

src/parcels/_chunk_cached_array/core.py
VeckoTheGecko: Yeah, I think so too. Though I think that should be done in a separate PR (since it would also be good to have some checks tests in place - as the cost of these not being coordinates is many many dask computes)

erikvansebille (approved):
Happy to be merged after the benchmark files are removed

VeckoTheGecko:
> the benchmark files are removed

Done

PR #2866, opened 2026-09-01
Add upstream CI testing workflow

VeckoTheGecko:
### Description

Adds a workflow to test against `main` for our various core dependencies, allowing us to quickly flag incompatibilities with changes upstream.

This workflow runs on PRs labelled with `run-upstream`, on commits to `main`, as well as on a schedule (midnight on Monday and Wednesday).

For schedule-triggered runs, on failure it will create an issue with the sumarized results and post an issue.

This setup is based on the tooling in Xarray.

### Checklist

- [x] Contributes to #2854
- [x] Tests added - N/A

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): an LLM was used to help translate the workflow from Xarray, which was then manually adjusted


ci/nightly/pixi.toml
VeckoTheGecko, line 1: This exists as a separate TOML since the nightly environment in particular takes a while to solve - and for reasons discussed a couple times before, we have to go through the solving process on `pixi shell` reasonably often.

Having it separate means that we don't have to worry about long solves for the main manifest (which would negatively effect not only our local developer experience, but also negatively effect RTD build times)

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
   "cftime >=1.6.3",
   "numpy >=2.1.0",
   "dask >=2024.5.1",
-  "netCDF4 >=1.7.2",
+  "netCDF4 >=1.7.2", # TODO: should we use h5netcdf here instead of netCDF4?
```
VeckoTheGecko, line 26: I saw this during review of the code, hence adding the TODO comment.

I think we should consider which NetCDF dependency/ies we ship  by default with Parcels. Opening an issue to discuss further...

VeckoTheGecko:
I can test the issue creation feature in a future couple PRs (this needs to be in main first before that can be tested).

I'll work on that this week - logging off for now

.github/workflows/upstream-dev-ci.yaml
```diff
@@ -0,0 +1,161 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize, labeled]
+  schedule:
+    - cron: "0 0 * * 1,4" # At 00:00 UTC on Monday and Wednesday.
+  workflow_dispatch: # allows you to trigger the workflow run manually
```
erikvansebille, line 12: I don't understand this comment. _How_ does it allow me to trigger the workflow manually?

erikvansebille (approved):

.github/workflows/upstream-dev-ci.yaml
VeckoTheGecko: This only takes effect once merged into main . See [github docs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)

VeckoTheGecko:
After a bunch of PRs, the issue creation is working properly https://github.com/Parcels-code/Parcels/issues/2872

PR #2880, opened 2026-09-03
Remove 'mesh' arg from add_constant_field

### LINKED ISSUE #2844: Error out when combining fieldsets of different mesh types

VeckoTheGecko:
### Description

Since we have `fieldset.mesh` in #2848, we no longer need this argument.

### Checklist

- [x] Closes #2844
- [x] Tests added

### AI Disclosure

None used


src/parcels/_core/fieldset.py
```diff
@@ -201,17 +201,9 @@ def add_constant_field(self, name: str, value, mesh: ptyping.TMesh = "spherical"
...
-
-            1. spherical (default): Lat and lon in degree, with a
-               correction for zonal velocity U near the poles.
-            2. flat: No conversion, lat/lon are assumed to be in m.
         """
-        if mesh not in ("flat", "spherical"):
-            raise ValueError(f"mesh must be one of ['flat', 'spherical']. Got {mesh!r}.")
-
         if self.constant_model is None:
+            mesh = "flat" if isinstance(self.mesh, FlatMesh) else self.mesh
```
erikvansebille: Why can't mesh be `FlatMesh` here? Why does it need to be a string only for flat meshes, and an object otherwise?

VeckoTheGecko: Since within `create_empty_constant_field_model`,  `from_sgrid_conventions` expects `mesh: Literal['flat', spherical'] | SphericalMesh` (`FlatMesh` is not part out our public API, and is instead created internally from the `"flat"` keyword).

I have moved this cast statement to be in `create_empty_constant_field_model` now so its closer to the part that needs it

erikvansebille (approved):

PR #2883, opened 2026-09-07
Account for spherical face bulging during hashgrid construction and point in cell checks

### LINKED ISSUE #2878: Fix `tutorial_nested_grids.ipynb` - grid searching
OK - running into a separate (unrelated) problem surfaced by this change which I think runs deeper...

`docs/user_guide/examples/tutorial_nestedgrids.ipynb` combines different datasets together - but it was specifying `"flat"` for the gridIDs and `"spherical"` for the data itself (which although somehow worked - doesn't make sense to me, and doesn't align with the work we've been doing).

This is in lon/lat space, so I updated to use `"spherical"` and got grid searching errors.

```
GridSearchingError                        Traceback (most recent call last)
Cell In[10], [line 18](vscode-notebook-cell:?execution_count=10&line=18)
     14 def SampleGridID(particles, fieldset):
     15     particles.gridID = fieldset.GridID[particles]
     16 
     17 
---> [18](vscode-notebook-cell:?execution_count=10&line=18) pset.execute(
     19     SampleGridID,
     20     runtime=np.timedelta64(1, "s"),
     21     dt=np.timedelta64(1, "s"),

File ~/coding/repos/parcels/src/parcels/_core/particleset.py:453, in ParticleSet.execute(self, kernels, dt, endtime, runtime, output_file, verbose_progress)
    450 else:
    451     next_time = end_time
--> [453](https://file+.vscode-resource.vscode-cdn.net/Users/Hodgs004/coding/repos/parcels/docs/user_guide/examples/~/coding/repos/parcels/src/parcels/_core/particleset.py:453) self._kernel.execute(self, endtime=next_time, dt=dt)
    455 if next_output is not None:
    456     if np.abs(next_time - next_output) < 0.001:

File ~/coding/repos/parcels/src/parcels/_core/kernel.py:244, in Kernel.execute(self, pset, endtime, dt)
    242                 error_func(pset[inds].t)
    243             else:
--> [244](https://file+.vscode-resource.vscode-cdn.net/Users/Hodgs004/coding/repos/parcels/docs/user_guide/examples/~/coding/repos/parcels/src/parcels/_core/kernel.py:244)                 error_func(pset[inds].z, pset[inds].y, pset[inds].x)
    246 return pset

File ~/coding/repos/parcels/src/parcels/_core/statuscodes.py:85, in _raise_grid_searching_error(z, y, x)
     84 def _raise_grid_searching_error(z, y, x):
---> [85](https://file+.vscode-resource.vscode-cdn.net/Users/Hodgs004/coding/repos/parcels/docs/user_guide/examples/~/coding/repos/parcels/src/parcels/_core/statuscodes.py:85)     raise GridSearchingError(f"Grid searching failed at (z={z}, y={y}, x={x})")

GridSearchingError: Grid searching failed at (z=array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0.], dtype=float32), y=array([-18.        , -18.        , -18.        , -18.        ,
       -18.        , -15.052631  , -15.052631  , -15.052631  ,
       -15.052631  , -12.105263  , -12.105263  ,  -9.157895  ,
        -9.157895  ,  -9.157895  ,  -6.2105265 ,  -3.2631578 ,
        -3.2631578 ,  -3.2631578 ,  -3.2631578 ,  -3.2631578 ,
        -3.2631578 ,  -0.31578946,  -0.31578946,   2.631579  ,
         5.5789475 ,   8.526316  ,   8.526316  ,  11.473684  ,
        11.473684  ,  11.473684  ,  14.421053  ,  14.421053  ,
        14.421053  ,  17.368422  ,  17.368422  ,  17.368422  ,
        17.368422  ,  17.368422  ,  20.31579   ,  20.31579   ,
        23.263159  ,  26.210526  ,  29.157894  ,  29.157894  ,
        29.157894  ,  29.157894  ,  32.105263  ,  35.05263   ,
        35.05263   ,  38.        ,  38.        ,  38.        ],
      dtype=float32), x=array([ 0.25,  3.  ,  5.75,  8.5 , 11.25,  8.5 , 11.25, 14.  , 52.5 ,
       16.75, 49.75, -2.5 , 44.25, 47.  , 41.5 , -8.  ,  3.  , 16.75,
       19.5 , 30.5 , 38.75, -8.  , 33.25, -8.  , -8.  , -8.  , -5.25,
       -8.  , -5.25, 25.  , -8.  , -5.25, 25.  , -8.  , -5.25, -2.5 ,
       25.  , 44.25, 25.  , 47.  , 47.  , 49.75, 16.75, 22.25, 49.75,
       52.5 , 52.5 , 49.75, 55.25, 22.25, 55.25, 58.  ], dtype=float32))
```

Adapting the plot I was able to find the grid search failure locations

image

Using an LLM it was talking about there being a bug in the point-in-cell check, but to be honest I don't know how accurate that is.

<details><summary>LLM output</summary>
<p>

# Spherical UxGrid search fails for large triangles

When using `mesh="spherical"` with `UxGrid`, grid searching fails for points inside large triangular faces. Two independent bugs:

## Bug 1: SpatialHash Cartesian bounding box too tight

`SpatialHash.__init__` (`spatialhash.py:182-188`) computes each face's bounding box from the Cartesian (x,y,z) coordinates of its **vertices only**. On a sphere, great-circle edges curve outward relative to the chord, so for large faces a query point on the sphere surface can lie inside the face but outside the vertex-only Cartesian bounding box. The hash never considers the face as a candidate.

**Suggested fix**: Include great-circle edge midpoints (average vertex Cartesian coords, renormalize to unit sphere) in the bounding box computation.

## Bug 2: `uxgrid_point_in_cell` barycentric check too strict

`uxgrid_point_in_cell` (`index_search.py:298-372`) projects the query onto the face plane and computes barycentric coordinates using unsigned 3D triangle areas. For large spherical triangles, projection distortion causes `sum(bcoords)` to exceed 1 (e.g. 1.023 for a ~70°-wide face), failing the `np.isclose(..., 1.0, rtol=1e-3)` check at line 370.

## Minimal reproducer

```python
import numpy as np
import uxarray as ux
import xarray as xr
from shapely.geometry import Point, Polygon
from triangle import triangulate

from parcels._core.index_search import uxgrid_point_in_cell
from parcels._core.uxgrid import UxGrid

# Build a multi-face constrained Delaunay triangulation from three nested polygons
polygons = [
    np.array([(10, 15), (25, 10), (25, 25), (17, 36), (10, 32)]),
    np.array([(0, -5), (35, 0), (35, 25), (0, 20)]),
    np.array([(-10, -20), (60, -20), (60, 40), (-10, 40)]),
]
verts, segments = [], []
offset = 0
for poly in polygons:
    Ni = len(poly)
    verts.extend(poly.tolist())
    segments.extend([[offset + j, offset + ((j + 1) % Ni)] for j in range(Ni)])
    offset += Ni
B = triangulate(
    {"vertices": np.asarray(verts, dtype=float), "segments": np.asarray(segments, dtype=int)},
    "p",
)
points, face_tris = B["vertices"], B["triangles"].astype(int)

ds = xr.Dataset(
    {
        "node_lon": ("n_node", points[:, 0]),
        "node_lat": ("n_node", points[:, 1]),
        "face_node_connectivity": (("n_face", "n_max_face_nodes"), face_tris),
    },
    coords={
        "n_node": np.arange(points.shape[0]),
        "n_face": np.arange(face_tris.shape[0]),
    },
    attrs={"Conventions": "UGRID-1.0"},
)
uxgrid = ux.Grid(ds)
z = ux.UxDataArray(np.array([0.0, 1.0]), dims="z", uxgrid=uxgrid)
grid = UxGrid(uxgrid, z, mesh="spherical")

# --- Bug 1: SpatialHash misses points inside a large face ---
sh = grid.get_spatial_hash()
for lon, lat in [(0.25, -18.0), (3.0, -18.0), (5.75, -18.0)]:
    _, face_ids, _ = sh.query(
        np.array([lat], dtype=np.float32),
        np.array([lon], dtype=np.float32),
    )
    print(f"Bug 1 — ({lon}, {lat}): face_id={face_ids[0]} (expected >=0, got GRID_SEARCH_ERROR)")

# --- Bug 2: point-in-cell rejects a point near the edge of a large face ---
# Find which face contains (52.5, -15) by brute force — none pass the check
for fi in range(face_tris.shape[0]):
    is_in, coords = uxgrid_point_in_cell(
        grid,
        np.array([-15.0], dtype=np.float32),
        np.array([52.5], dtype=np.float32),
        np.array([fi]),
        np.array([fi]),
    )
    if abs(coords[0].sum() - 1.0) < 0.05:  # close but fails the rtol=1e-3 check
        print(
            f"Bug 2 — (52.5, -15) face {fi}: is_in={is_in[0]}, "
            f"bcoords sum={coords[0].sum():.4f} (fails isclose rtol=1e-3)"
        )
```

Output:

```
Bug 1 — (0.25, -18.0): face_id=-3 (expected >=0, got GRID_SEARCH_ERROR)
Bug 1 — (3.0, -18.0): face_id=-3 (expected >=0, got GRID_SEARCH_ERROR)
Bug 1 — (5.75, -18.0): face_id=-3 (expected >=0, got GRID_SEARCH_ERROR)
Bug 2 — (52.5, -15) face 12: is_in=0, bcoords sum=1.0234 (fails isclose rtol=1e-3)
Bug 2 — (52.5, -15) face 13: is_in=0, bcoords sum=1.0492 (fails isclose rtol=1e-3)
```

Bug 1 is the primary blocker — it prevents `tutorial_nestedgrids.ipynb` from running with `mesh="spherical"` on the triangulation FieldSet.

</p>
</details> 

@erikvansebille did you run into something similar when authoring the notebook? cc @fluidnumericsJoe

_Originally posted by @VeckoTheGecko in https://github.com/Parcels-code/Parcels/issues/2848#issuecomment-5507062450_

---

cc @erikvansebille, @wyatt-fluidnumerics, @fluidnumericsJoe - nicer to continue here than in the original PR :)

wyatt-fluidnumerics:
### Description
Issue #2878 raised awareness of two separate bugs:

1. During point in cell checking on spherical unstructured or cuvilinear grids the old code did a projection of the particle position by calculating the direction normal to the face and moving the particle in that direction. This meant that for particles close to the edge of a face the normal projection could project them outside of the cell face, which was incorrect. This could result in a GridSearchingError. 
2. The spatialhash grid bounds were previously calculated only by using the positions of a cell's corners. This does not account for the bulged component of the cell that the use a spherical mesh creates. As a result, the spatialhash grid does not fully enclose spherical grids. 

Both of these issues are magnified for grids that contain large cells, which is why for most of the small celled grids that Parcels is tested on, they did not show up earlier.

The fixes are the following:

1. Change from a projection normal to the face to a gnomonic projection which moves the particle on a ray that extends through the face and the center of the unit sphere. This way particles very close to the edge of their cell remain in the cell after projection.
2. Add a much more extensive bounding box calculation for the spatialhash grid. For any give cell their are three possibilities for its minima/maxima. The first is that is on one of the nodes (this is all that was used for the calculation previously). The second is that the face overlaps a global minama/maxima of the unit sphere, think a face that extends over the North Pole, in this case the minima/maxima can be anywhere on the inside of the face, for this reason we introduce and explicit check to see whether a face contains one of the 6 global minima/maxima of the unit sphere, and if it does set the respective xyz bound accordingly. Third, the minima maxima could sits on the face edge. In this case picture a face that has a bottom edge extending from lon -5 to 5 at a lat of 10, the final node forming the face could sit at lon 0, lat 20. In this case the point lon 0, lat 10 serves as an extrema for the face due to its bulge between longitudes of -5 and 5. However this extrema lies only on an edge rather than on a node, or on a global extrema. These types of points can be calculated analytically by parameterizing the curve between to vertices as a cos function and then determining whether that function has a local extrema along the curve. If it does, then that point is one of the faces extrema, if it doesn't, then the node at the end of the curve is the extrema. I am happy to help put this calculation into the docs that @fluidnumericsJoe is writing up. 

The PR in this state covers the case for unstrucutred grids. Structured curvilinear grids still suffer the same bug and I have yet to implement these fixes there. I'll try and do that tomorrow so this is resolved before the v4 release!

### Checklist

- [x] Closes #2878 
- [x] Tests added

### AI Disclosure

  - Describe how you used it (e.g., by pasting your prompt): I guided claude code to generate parts of the new _spherical_triangle_bounds function. I made significant modifications to the output for clarity, structure, and scope. I also guided claude code to modify and improve the new testing associated with the PR. I have thoroughly reviewed and understand all code associated with the PR.

Currently a draft PR until the spherical curvilinear fixes are in.

### DIFF
--- a/docs/user_guide/examples/tutorial_nestedgrids.ipynb
+++ b/docs/user_guide/examples/tutorial_nestedgrids.ipynb
@@ -334,7 +334,7 @@
     ")\n",
     "uxda = ux.UxDataArray(ds_tri[\"GridID\"], uxgrid=ux.Grid(ds_tri)).to_dataset()\n",
     "uxda = uxda.assign_coords({\"zf\": ds_tri.coords[\"zf\"]})\n",
-    "fieldset = parcels.FieldSet.from_ugrid_conventions(uxda, mesh=\"flat\")"
+    "fieldset = parcels.FieldSet.from_ugrid_conventions(uxda, mesh=\"spherical\")"
    ]
   },
   {
--- a/src/parcels/_core/index_search.py
+++ b/src/parcels/_core/index_search.py
@@ -335,20 +335,31 @@ def uxgrid_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray, xi:
             axis=-1,
         )
 
-        # Get projection points onto element plane. Keep the leading
-        # dimension even for single-face queries so shapes remain (M,3).
-        r1 = face_vertices[:, 1, :] - face_vertices[:, 0, :]
-        r2 = face_vertices[:, 2, :] - face_vertices[:, 0, :]
-        nhat = np.cross(r1, r2)
-        norm = np.linalg.norm(nhat, axis=-1)
-        # Avoid division by zero for degenerate faces
-        norm = np.where(norm == 0.0, 1.0, norm)
-        nhat = nhat / norm[:, None]
-        # Calculate the component of the points in the direction of nhat
-        ptilde = points - face_vertices[:, 0, :]
-        pdotnhat = np.sum(ptilde * nhat, axis=-1)
-        # Reconstruct points with normal component removed.
-        points = ptilde - pdotnhat[:, None] * nhat + face_vertices[:, 0, :]
+        # Barycentric coordinates of each point solved directly: coords = w0,w1,w2
+        # such that w0*v0 + w1*v1 + w2*v2 = point, normalized to sum to 1.
+        # This is the gnomonic projection such that each point is projected
+        # onto the face a ray that goes through the center of the unit sphere.
+        face_matrix = np.stack(
+            (face_vertices[:, 0, :], face_vertices[:, 1, :], face_vertices[:, 2, :]),
+            axis=-1,
+        )  # (M, 3, 3), columns v0, v1, v2
+
+        # The trailing single axis on `points` tells np.linalg.solve to treat
+        # the leading M as a batch dimension (one 3x3 solve per point), not as
+        # the matrix size.
+        weights = np.linalg.solve(face_matrix, points[..., None])  # (M, 3, 1)
+        # drop the trailing axis
+        weights = weights[..., 0]  # (M, 3)
+        weight_sum = weights.sum(axis=-1, keepdims=True)
+
+        # A point is inside only if it's a non-negative combination of the 3
+        # vertices.
+        is_in_cell = np.where(np.all(weights >= -1e-9, axis=1) & (weight_sum[:, 0] > 0), 1, 0)
+
+        # weight_sum can be exactly 0 for a candidate face the point isn't in,
+        # is_in_cell already excludes these, this avoids a spurious divide-by-zero warning only.
+        safe_weight_sum = np.where(weight_sum == 0.0, 1.0, weight_sum)
+        coords = weights / safe_weight_sum
 
     else:
         nids = grid.uxgrid.face_node_connectivity[xi, :].values
@@ -361,13 +372,10 @@ def uxgrid_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray, xi:
         )
         points = np.stack((x, y), axis=-1)
 
-    M = len(points)
+        coords = _barycentric_coordinates(face_vertices, points)
 
-    coords = _barycentric_coordinates(face_vertices, points)
-
-    is_in_cell = np.zeros(M, dtype=np.int32)
-    is_in_cell = np.where(np.all((coords >= -1e-6), axis=1), 1, 0)
-    is_in_cell &= np.isclose(np.sum(coords, axis=1), 1.0, rtol=1e-3, atol=1e-6)
+        is_in_cell = np.where(np.all((coords >= -1e-6), axis=1), 1, 0)
+        is_in_cell &= np.isclose(np.sum(coords, axis=1), 1.0, rtol=1e-3, atol=1e-6)
 
     return is_in_cell, coords
 
--- a/src/parcels/_core/index_search.py
+++ b/src/parcels/_core/index_search.py
@@ -335,16 +335,15 @@ def uxgrid_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray, xi:
             axis=-1,
         )
 
-        # Barycentric coordinates of each point solved directly: coords = w0,w1,w2
+        # Construcuct a matrix of face positions to solve for the
+        # byarycentric coordinates of each point, and then convert to coords = w0,w1,w2
         # such that w0*v0 + w1*v1 + w2*v2 = point, normalized to sum to 1.
-        # This is the gnomonic projection such that each point is projected
-        # onto the face a ray that goes through the center of the unit sphere.
         face_matrix = np.stack(
             (face_vertices[:, 0, :], face_vertices[:, 1, :], face_vertices[:, 2, :]),
             axis=-1,
         )  # (M, 3, 3), columns v0, v1, v2
 
-        # The trailing single axis on `points` tells np.linalg.solve to treat
+        # The trailing single axis on points tells np.linalg.solve to treat
         # the leading M as a batch dimension (one 3x3 solve per point), not as
         # the matrix size.
         weights = np.linalg.solve(face_matrix, points[..., None])  # (M, 3, 1)
--- a/src/parcels/_core/index_search.py
+++ b/src/parcels/_core/index_search.py
@@ -94,11 +94,11 @@ def _search_time_index(field: Field, time: np.ndarray):
 def curvilinear_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray, xi: np.ndarray):
     """Locate points within 2D curvilinear grid cells.
 
-    For spherical grids, the bilinear inverse runs in a tangent plane on a sphere
-    through each cell's centroid — robust across the antimeridian and the
-    polar cap. For flat grids it runs directly in (lon, lat). In both
-    cases the returned (xsi, eta) are bilinear weights in the space the inverse
-    was solved in; downstream interpolators must agree on that same space.
+    For spherical grids, the bilinear inverse runs in a plane tangent to the sphere at
+    each cell's centroid, which is robust across the antimeridian and the polar cap. For
+    flat grids it runs directly in (lon, lat). In both cases the returned (xsi, eta) are
+    bilinear weights in the space the inverse was solved in; downstream interpolators
+    must agree on that same space.
     """
     clon = np.asarray(
         [grid.lon[yi, xi], grid.lon[yi, xi + 1], grid.lon[yi + 1, xi + 1], grid.lon[yi + 1, xi]],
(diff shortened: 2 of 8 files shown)


src/parcels/_core/index_search.py
```diff
@@ -335,20 +335,30 @@ def uxgrid_point_in_cell(grid, y: np.ndarray, x: np.ndarray, yi: np.ndarray, xi:
...
-        nhat = nhat / norm[:, None]
-        # Calculate the component of the points in the direction of nhat
-        ptilde = points - face_vertices[:, 0, :]
-        pdotnhat = np.sum(ptilde * nhat, axis=-1)
-        # Reconstruct points with normal component removed.
-        points = ptilde - pdotnhat[:, None] * nhat + face_vertices[:, 0, :]
+        # Construcuct a matrix of face positions to solve for the
+        # byarycentric coordinates of each point, and then convert to coords = w0,w1,w2
+        # such that w0*v0 + w1*v1 + w2*v2 = point, normalized to sum to 1.
+        face_matrix = np.stack(
```
erikvansebille, line 359: Is this the most efficient implementation? Does using np.stack here (and later in this PR) not lead to a similar performance loss as #2846? Or are the arrays so small here that that doesn't matter?

docs/user_guide/examples/tutorial_nestedgrids.ipynb
erikvansebille, line 1: I updated the nested grids tutorial so that gets run again - hope that's ok?

erikvansebille (commented):
Looks good; one comment and one question below

src/parcels/_core/index_search.py
fluidnumericsJoe: Hmm, yeah, maybe preallocating `face_matrix` and concatenating into this array would be better? I wonder if concatenating in Python concats by reference or by value

fluidnumericsJoe: Looks like we'd have to use the `out=` argument with concatenate (or stack) to avoid additional allocation costs, see https://numpy.org/devdocs/reference/generated/numpy.concatenate.html and https://numpy.org/doc/stable/reference/generated/numpy.stack.html

Source code for `stack` shows it just calls `concatenate` under the hood .

Neither avoids the copy cost here though. Buffers are copied . Only numpy only route I see possible here is to use an array `view` to avoid copies.

erikvansebille: Thanks for digging into this, @fluidnumericsJoe. I don't think we need to go too deep into optimisation of this statement specifically before we know whether it is a bottle-neck or not. So I'd avoid creating our own views. I just raised the question in case there was a low-hanging-fruit solution to make this faster

docs/user_guide/examples/tutorial_nestedgrids.ipynb
wyatt-fluidnumerics: That's great, thank you!

src/parcels/_core/index_search.py
fluidnumericsJoe: Pre-allocating and using `np.stack` with the `out` parameter could be helpful - but that pre-allocation might need to have broader scope than just a single PIC call to be of any value now that I think of it. IMO, after thinking on this a bit, I wouldn't worry about this too much here since its a single call per vectorized PIC check.

wyatt-fluidnumerics:
@erikvansebille @fluidnumericsJoe I've added the changes for spherical curvilinear grids. I generalized the _spherical_face_bounds function to work for both cases, and then adjusted _spherical_project_cell_and_query to use a gnomonic projection instead of orthonormal. I also changed some of the new tests to use xgrid/uxgrid parameterizations. I think things should be good to go now, but would be great to get another review on the changes!

erikvansebille (approved):
Looks good to me, @wyatt-fluidnumerics!

erikvansebille:
As we want to release Parcels v4 today, I've gone ahead and merged this PR - hope that's ok @wyatt-fluidnumerics

wyatt-fluidnumerics:
That's great! Glad to get the bugfix in before the release

fluidnumericsJoe:
huzzah! nice work @wyatt-fluidnumerics

PR #2889, opened 2026-09-09
Update README to v4

erikvansebille:
### Description

This PR updates the links in README.md to v4. it also aligns some of the text with that on https://docs.parcels-code.org/en/stable/index.html

### Checklist

### AI Disclosure

None


README.md
```diff
@@ -14,14 +14,11 @@
 [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/Parcels-code/parcels/main?labpath=docs%2Fexamples%2Fparcels_tutorial.ipynb)
 [![LinkedIn](https://custom-icon-badges.demolab.com/badge/LinkedIn-0A66C2?logo=linkedin-white&logoColor=fff)](https://www.linkedin.com/company/parcelscode/)
 
-> [!WARNING]
-> This branch contains version 4 of Parcels which is in active development. See `v3-support` to see v3 or the tags to browse stable versions of Parcels.
-
```
VeckoTheGecko, line 19: Are we still maintaining the `v3-support` branch?

I think it would be a good idea at least until we get more user feedback.

README.md
```diff
@@ -39,13 +36,13 @@ _Kehl, C, PD Nooteboom, MLA Kaandorp and E van Sebille (2023) Efficiently simula
 
 ### Further information
 
-See [parcels-code.org](http://parcels-code.org/) for further information about [installing](https://docs.parcels-code.org/en/latest/installation.html) and [running](https://docs.parcels-code.org/en/latest/documentation.html) the Parcels code, as well as extended [documentation](https://docs.parcels-code.org/en/latest/reference.html) of the methods and classes.
+See [parcels-code.org](http://parcels-code.org/) for further information about [installing](https://docs.parcels-code.org/en/stable/user_guide/getting_started/installation.html) and [running](https://docs.parcels-code.org/en/stable/user_guide/index.html) the Parcels code, as well as extended [API reference](https://docs.parcels-code.org/en/stable/reference/parcels/index.html) of the methods and classes.
```
VeckoTheGecko: I have now finished reconfiguring RTD - so we have the `latest` tag again.

- `latest` - builds off `main`
- `stable` - builds off the latest release

I think we should point everything in README's and in code to the `latest` tag, which is now also the default version in the docs

README.md
erikvansebille: Yes, I think we do keep the v4-support branch. But perhaps not as a warning - I'll change it to an INFO message

README.md
erikvansebille: OK, but if the (default) documentation builds off main, while most users will install the latest release - don't we run the risk that the documentation doesn't (entirely) align with the code users have.

Or is that the reason why we "freeze" the API, so that from a user perspective nothing changes until the next release?

README.md
erikvansebille: Done in d363cedc1ca10f780581173dfd5d2fa425c8bf24

README.md
erikvansebille: Done in cdc58610dd78030c1f36f4169e866c2bdffb2f95

README.md
```diff
@@ -14,14 +14,13 @@
 [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/Parcels-code/parcels/main?labpath=docs%2Fexamples%2Fparcels_tutorial.ipynb)
 [![LinkedIn](https://custom-icon-badges.demolab.com/badge/LinkedIn-0A66C2?logo=linkedin-white&logoColor=fff)](https://www.linkedin.com/company/parcelscode/)
 
-> [!WARNING]
-> This branch contains version 4 of Parcels which is in active development. See `v3-support` to see v3 or the tags to browse stable versions of Parcels.
+> This branch contains version 4 of Parcels. Change to `v3-support` to see v3 or the tags to browse stable versions of Parcels.
```
VeckoTheGecko:
```suggestion
> This branch contains version 4 of Parcels. Change to `v3-support` to see v3 (which is currently in maintainence mode) or the tags to browse stable versions of Parcels.
```

VeckoTheGecko (approved):
small edit - otherwise looks good

